Issue #31, opened 2021-01-20
Custom Date Range Buttons Broke


Rperry2174:
![image](https://user-images.githubusercontent.com/23323466/105132598-0af49380-5aa0-11eb-8b7f-22d943db4630.png)

- [ ] Fix the date range buttons so they work
- [ ] (optional) Write a test to make sure that they don't break again in the future

LouisInFlow:
This is already fixed. It seems that adding "resolve #issue#" is not auto closing issues

Rperry2174:
closed in #36

Issue #27, opened 2021-01-18
Cleanup JSX lint errors and warnings


LouisInFlow:
To package.json add

```
    "lint": "eslint ./ --ext .js, .jsx --cache --fix"
```

Fix all the issues

gwuah:
I wanted to pick this up but it seems it's been fixed. guess this issue can be closed now.

petethepig:
@gwuah You're right, I'm closing the issue. There's a similar one for CSS / SCSS linting https://github.com/pyroscope-io/pyroscope/issues/1 if you're interested.

PR #36, opened 2021-01-22
Refactor and Modernize Webapp

LouisInFlow:
* Enable JSX Lint
* Convert class components to function components
* Will refactor FlameGraphRenderer in follow-up

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -16,7 +16,10 @@
     "jsx-a11y/heading-has-content": ["warn"],
     "jsx-a11y/control-has-associated-label": ["warn"],
     "no-undef": ["warn"],
-    "react/jsx-one-expression-per-line": ["warn"]
+    "react/jsx-one-expression-per-line": ["warn"],
+    "jsx-a11y/mouse-events-have-key-events": ["warn"],
+    "jsx-a11y/click-events-have-key-events": ["warn"],
+    "jsx-a11y/no-static-element-interactions": ["warn"]
   },
   "env": {
     "browser": true,
--- a/.eslintrc
+++ b/.eslintrc
@@ -13,7 +13,8 @@
     "no-restricted-globals": ["warn"],
     "react/button-has-type": ["warn"],
     "react/prop-types": ["warn"],
-    "jsx-a11y/heading-has-content": ["warn"]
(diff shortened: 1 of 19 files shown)


LouisInFlow:
Resolve #27

Rperry2174:
Thanks @LouisInFlow can you please explain some of the benefits of switching from class components to functional components?

LouisInFlow:
Function components are a solution the React core team developed to solve problems discovered through the many years of seeing React out in the wild, it's where React is going and preferred by the core team. Class components maintained for backwards compatibility only. 

The main things are: 

1.  Lower complexity of component (no this, no constructor, [more intuitive life-cycle control](https://reactjs.org/docs/hooks-effect.html))
2. [React Hooks](https://reactjs.org/docs/hooks-intro.html ) - share stateful logic easier/cleaner, [this is how it works with redux](https://react-redux.js.org/api/hooks)
3. Better performance and fits JS model of programming better (JS doesn't have native classes, ES6 classes are sugar on top of prototype inheritance, JS/React is more declarative and functional programming.)
4. Easier to test (props are arguments and the output is jsx)
5. It's not either/or, codebase can be some class/some function components. 
6. New major features will only work with function components, like the [currently experimental concurrent mode (better user experience when fetching data)
](https://reactjs.org/docs/concurrent-mode-intro.html)

Rperry2174 (changes requested):
Hey @LouisInFlow, I noticed an error with the dateRangePicker component:

This is partially not working in the current version (#31), but somehow was made worse in this version:
When you start from the `dateRangePicker`:
Screen Shot 2021-01-22 at 8 00 32 PM

And then you "delete" the "1" from "now - 1h", then it breaks the page. This is probably a result of our initial code being more forgiving of the bug, but either way this version is worse because before it was a no-op and now it breaks the page. Could you either:

- Maintain the current functionality of just having them be no-ops and well fix in a follow-up
- Fix the button so that they work as intended

Also the refresh button no longer works. When I click it i get this error in the console:
image

LouisInFlow:
resolve #31

Rperry2174 (approved):
LGTM 🔥  Thanks a ton @LouisInFlow

Issue #49, opened 2021-01-28
Lint Go Code


petethepig:
### Community Call To Action

Feel free to open PRs, even if it's one or two small fixes. This kind of work is important but often overlooked. This is a great way to learn more about the internals of the project and we'd appreciate any help on this.

### Summary

We use [revive](https://github.com/mgechev/revive) as the main linter.

### How to tackle these

Run the linter. It will print a bunch of issues in the code
```
make lint
```
Sample output:
```
pkg/util/serialization/metadata.go
  (17, 2)  https://revive.run/r#unhandled-error  Unhandled error in call to function w.Write  

pkg/storage/key.go
  (104, 4)  https://revive.run/r#unhandled-error       Unhandled error in call to function sb.WriteString               
  (36, 1)   https://revive.run/r#cyclomatic            function ParseKey has cyclomatic complexity 12                   
  (36, 1)   https://revive.run/r#cognitive-complexity  function ParseKey has cognitive complexity 12 (> max enabled 7)  

pkg/server/labels.go
  (19, 2)  https://revive.run/r#unhandled-error   Unhandled error in call to function w.Write                              
  (34, 2)  https://revive.run/r#unhandled-error   Unhandled error in call to function w.Write                              
  (8, 62)  https://revive.run/r#unused-parameter  parameter 'r' seems to be unused, consider removing or renaming it as _  

pkg/config/config.go
  (19, 18)  https://revive.run/r#struct-tag          malformed tag                                                         
  (80, 39)  https://revive.run/r#struct-tag          malformed tag                                                                       
  (85, 0)   https://revive.run/r#line-length-limit   line is 161 characters, out of limit 120   
...
```

Pick any of these and try to fix it. Some of them are trivial (e.g line length ones), and some require deeper understanding of the underlying code. If you have any questions, or if something doesn't work, reach out to me on [our slack](https://pyroscope.io/slack) or open issues.

We also have instruction on how to set up VSCode to do the linting inside the editor: https://pyroscope.io/docs/developer-guide#go

### Progress

At the time of writing (Jan 28th) we're at `224 problems (51 errors) (173 warnings)`

Jan 31st — `202 problems (51 errors) (151 warnings)` Thanks to @Pranay0302!

Jun 9th — `139 problems (30 errors) (109 warnings)` Thanks to @AdrK !

Pranay0302:
hey @petethepig, I sent a PR #52 regarding this issue. 
please review the code and tell me your opinion on it

[...]

PR #52, opened 2021-01-30
cleanup: lint go code

Pranay0302:
at present ```202 problems (51 errors) (151 warnings) ```, reduced the problem count by 22.
deals with issue #49


petethepig (approved):
Looks great! I changed a couple of things:
* brought back `work` function code. It might seem redundant, because it's basically a while true loop, but it's actually used.
* renamed `errFoo` to `errClosing` so that the name makes more sense

Thank you so much for contributing @Pranay0302!

Issue #55, opened 2021-01-31
DateRange breaks page if dates are updated in the wrong order


Rperry2174:
Bug introduced in #53 

Steps to reproduce:
- Open the page
- Select a new date from the second `DateRange` component

![image](https://user-images.githubusercontent.com/23323466/106396900-9e429880-63bf-11eb-9ad8-f4416622b347.png)

It's worth noting that this happens when you change the second `DateRange` component _before_ you change the first one. 

/cc @abaali

abaali:
this happens because the value of from is still a string ('now-1h') which makes moment.js return a non-valid date 
I'm working on this.

petethepig:
I think the best way of fixing it would be to copy the time parsing logic from the backend to frontend:

* this is where it is on the go side: https://github.com/pyroscope-io/pyroscope/blob/master/pkg/util/attime/attime.go#L18
* that one is based on the original logic in graphite project: https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/attime.py#L30

and so if we had an `attime` parser on the frontend we could parse these `now-1h` strings and turn them into time

abaali:
good suggestion @petethepig  but I already worked on a different one but it needs some improvements.
check it in my new [PR](https://github.com/pyroscope-io/pyroscope/pull/56)

Rperry2174:
closed in #56

PR #56, opened 2021-02-01
split the code into reusable pieces, fixed issue#55

abaali:
these changes fix the issue [#55](https://github.com/pyroscope-io/pyroscope/issues/55) but I'm not satisfied with the code 
it needs a lot of improvement like:

- [ ] change functions names to make more sense
- [ ] write better comments
- [ ] add comment about type/form of arguments and type/form of functions return
- [ ] add tests

if you spot anything else that must be changed like the code logic,  repetition pls let me know

### DIFF
--- a/webapp/javascript/util/formatDate.js
+++ b/webapp/javascript/util/formatDate.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-underscore-dangle */
 import moment from "moment";
 
 const multiplierMapping = {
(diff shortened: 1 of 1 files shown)


webapp/javascript/util/formatDate.js
```diff
@@ -10,32 +11,65 @@ const multiplierMapping = {
...
-    const m = from.match(/^now-(?<number>\d+)(?<multiplier>\D+)$/);
-    if (m && multiplierMapping[m.groups.multiplier]) {
-      let multiplier = multiplierMapping[m.groups.multiplier];
-      if (m.groups.number > 1) {
-        multiplier += "s";
-      }
-      const readableDateForm = {
-        range: `Last ${m.groups.number} ${multiplier}`,
-        from: moment().add(-m.groups.number, multiplier).toDate(),
-        until: moment().toDate(),
```
LouisInFlow, line 24: Would be nice to add comment about type/form of arguments and type/form of functions return. This will make it easier for someone to understand exactly what this functions does.

webapp/javascript/util/formatDate.js
```diff
@@ -10,32 +11,65 @@ const multiplierMapping = {
...
+    _multiplier += "s";
+  }
+  // convert preset to a Date object
+  const fromObjectForm = moment().add(-number, _multiplier).toDate();
+
+  return { fromObjectForm, number, _multiplier };
+}
+
+function formateDate(from, until, preset = true) {
+  const { fromObjectForm, number, _multiplier } = preset
```
LouisInFlow, line 28: How does function handle case where the de-structured variables are null?

webapp/javascript/util/formatDate.js
```diff
@@ -10,32 +11,65 @@ const multiplierMapping = {
...
+
+  let readableDateForm = {
+    range: `Last ${number} ${_multiplier}`,
+    from: _from,
+    until: _until,
+  };
+
+  if (until === "now") {
+    // check if from is not a preset
+    if (!/^now-/.test(from)) {
```
LouisInFlow, line 48: Can probably do ```if(until === "now" && !/^now-...."

webapp/javascript/util/formatDate.js
```diff
@@ -10,32 +11,65 @@ const multiplierMapping = {
...
+  readableDateForm = {
+    ...readableDateForm,
+    range: `${moment(_from).format("lll")} - ${moment(_until).format("llll")}`,
+    until: _until,
   };
 
   return readableDateForm;
+}
+
+export default function humanReadableRange(until, from) {
```
LouisInFlow, line 68: Can this logic just be a part of formatDate? Seems like needless abstraction

webapp/javascript/util/formatDate.js
abaali: can't do that or the range won't be the same as the default presets label
example: "Last 5 minutes"
instead, it will be date - date
example: "Jan 25, 2021 6:12 PM - Tue, Feb 2, 2021 6:12 PM"

LouisInFlow:
Wouldn't hurt to add tests to confirm functionality

webapp/javascript/util/formatDate.js
abaali: in that case, it gets replaced by another value but that's only for `fromObjectForm` but the other two variables stay unhandled.
is there a better way to handle it?

webapp/javascript/util/formatDate.js
abaali: sure, that's better thank u for the review @LouisInFlow

webapp/javascript/util/formatDate.js
LouisInFlow: Think about from perspective of user. What experience would user have currently if those values were null, what would I as a user expect if something wasn't working correctly

abaali: `number` and `_multiplier` are used only in the default range value when `/^now-/.test(from` is true and `until === "now" `. otherwise the variables won't be used in any part of the code.
so I think there is no need to handle it.

LouisInFlow: What if I decided to use this function in a different context/different use case?

Rperry2174 (approved):
Thanks @abaali for the quick fix we really like the new data pickers! Also thanks @LouisInFlow for the extra review. We're gonna go ahead and merge this so `main` is fixed and then create a separate issue for adding tests

Issue #60, opened 2021-02-02
Date range selectors send too many requests


Rperry2174:
Currently when using the date range selectors, they send an API request every time they're updated. Instead there should be an "Apply" button that only sends one request once a user is done updating the date range.

Rperry2174:
Fixed in #64

PR #64, opened 2021-02-02
improved the date range selector and fixed #54 #60

### LINKED ISSUE #54: Alert user with text if startDate > endDate
We have a new DatePicker dropdown, but now its easier to accidentally choose an `endDate` that is smaller than `startDate`. When `state.startDate > state.endDate` then we should add some red text to the bottom to alert the user that they have an invalid date range. 

This way they will not accidentally wait for the data to update for an invalid date range.

Screen Shot 2021-01-30 at 4 54 22 PM

abaali:
this PR fixes #54 and #60 
any remarks about the code or the comments let me know
I tested the functionality before this pr but if something isn't working open an issue or mention it here so I can fix it

### DIFF
--- a/webapp/javascript/components/CustomDatePicker.jsx
+++ b/webapp/javascript/components/CustomDatePicker.jsx
@@ -50,7 +50,6 @@ function CustomDatePicker({ setRange, dispatch, setDateRange }) {
           selectsStart
           showTimeSelect
           startDate={selectedDate.from}
-          endDate={selectedDate.until}
           dateFormat="yyyy-MM-dd hh:mm aa"
         />
       </div>
--- a/webapp/javascript/components/CustomDatePicker.jsx
+++ b/webapp/javascript/components/CustomDatePicker.jsx
@@ -51,7 +51,7 @@ function CustomDatePicker({ setRange, dispatch, setDateRange }) {
           showTimeSelect
           startDate={selectedDate.from}
           endDate={selectedDate.until}
-          dateFormat="yyyy-dd-MM hh:mm aa"
+          dateFormat="yyyy-MM-dd hh:mm aa"
         />
       </div>
       <div className="until">
@@ -66,7 +66,7 @@ function CustomDatePicker({ setRange, dispatch, setDateRange }) {
           startDate={selectedDate.from}
           endDate={selectedDate.until}
           minDate={selectedDate.from}
-          dateFormat="yyyy-dd-MM hh:mm aa"
+          dateFormat="yyyy-MM-dd hh:mm aa"
         />
       </div>
       {warning && <p style={{ color: "red" }}>Warning: invalid date Range</p>}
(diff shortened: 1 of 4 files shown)


Rperry2174:
Hey @abaali I like this solution a lot! There are a couple of UI/UX things that I think would make it better, however I'm not sure if they are all possible. 

#### Improvement 1 
We should change the date formate to be YYYY-mm-dd HH:MM
image

2021-19-01 04:30 PM --> 2021-01-19 04:30 PM
2021-10-02 04:55 PM --> 2021-02-10 04:55 PM

#### Improvement 2
This one is a little trickier, but I think if we use these date pickers that select a range we have to have it affect both dates. 

Is it possible to make just one daterange selector affect _both_ start date and end date? Right now its confusing because when you move the mouse around while selecting an end date, it seems like you are selecting a full range because a full range gets highlighted.

Here's a gif where I'm only adjusting the start date, but when I change the start date from the Jan 15 to Jan 1, it looks as if I'm changing _the full range_ to be:
from: January 1, 2020 to January 15, 2020. 

Do you think we could get ours to function more like [this one](https://k8s.devstats.cncf.io/d/12/dashboards?orgId=1&refresh=15m)?

abaali:
@LouisInFlow I added some small changes 
-  changed the date format
-  from selector doesn't look like you are changing the whole range anymore

the selector can be changed to look like the one you mentioned but it will need a little work.

Rperry2174:
Hey @abaali the thing stopping us from merging this version is that the UI/UX is actually more confusing than the version before. The reason why, is because in this version there are two colors, Grey and Blue... The problem is that these colors don't  _consistently_ mean something.

 In the [example I linked](https://k8s.devstats.cncf.io/d/12/dashboards?orgId=1&refresh=15m) the meaning of the colors is very straightforward: 
 - If a date is "blue" then that date is part of the selected date range state. The color is a function of what date range is selected
 
 In the version from this pr:
 - If a date is "blue" then that actually doesn't necessarily mean anything. It actually flickers depending on wherever your mouse is hovering. It is therefore a function of where your mouse is hovering (which is more confusing than useful) 
 - if a date is "grey" then that means that it is part of the selected range UNLESS you move your mouse around and then even the selected range will sometimes turn blue depending on where your mouse is
 


Here is what I propose in order to merge this PR:

- [ ] Make the current selected range be **blue** instead of the light grey that it currently is
- [ ] Remove the light grey color totally

So this:
- removes the "on hover" coloring so that the "flickering" stops
- Only colors a date when someone clicks **and** that date is a part of the date range

What do you think?

abaali:
I worked on a quick solution check it out

Rperry2174:
Hey @abaali yeah this is great! If you can resolve the conflicts then we can merge this

abaali:
conflicts resolved!  :)

Rperry2174:
Thanks for this @abaali Pyroscope's whole flow is much smoother thanks to your last couple PRs 🙌🏾

abaali:
You're welcome 😄

PR #65, opened 2021-02-03
Crosshair implementation

cmonez:
-Implemented crosshair functionality when hovering over timeline graph (issue [51](https://github.com/pyroscope-io/pyroscope/issues/51))
-Used blue to make crosshair pop when overlayed the gold bars
crosshair

### DIFF
--- a/webapp/javascript/components/PyroscopeApp.jsx
+++ b/webapp/javascript/components/PyroscopeApp.jsx
@@ -26,7 +26,7 @@ const flotOptions = {
   },
   crosshair: {
     mode: "x",
-    color: "blue",
+    color: "#C3170D",
     lineWidth: "1",
   },
   grid: {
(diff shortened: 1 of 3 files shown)


petethepig:
Looks awesome!

It would be cool if the cursor also changed to crosshair, I know there's a CSS option for that:

IMG_5700 copy

Rperry2174:
I agree thanks @cmonez !!! I also think that we should change the color to red instead of blue so that it stands out a little bit more. Can you change it to this red: `#C3170D`?

cmonez:
will change that to #C3170D , @petethepig so you want the cursor to disappear, and just be represented by the crosshair instead, correct? will work on this rn.

Rperry2174:
Yeah just to be clear so instead of your typical cursor it will be a _cross_ instead:

So the red line should still show, but instead of this:
image

when you move your mouse it will look like this:
image

cmonez:
kk
crosshair-cross

Rperry2174 (approved):
Looks 🔥!!! Thanks so much. I think this will help usability a lot and get people to utilize the timeline more with these indicators that they can interact with it.

PR #71, opened 2021-02-12
Implement previous time range functionality

cmonez:
Implementing back button for the timechart for issue [38](https://github.com/pyroscope-io/pyroscope/issues/38)
- Create a Redux state object and action 
- Keep track of the previous called time by switching between previous and new date ranges whenever you try to fetch new data

PreviousTime


Rperry2174:
Thanks for the PR @cmonez this is much needed. We will test this out over the weekend and merge beginning of next week. The one thing that we will likely change is that instead of the red button we will probably use something more like this:

image

But let me test this out first and I'll update here after

cmonez:
> Thanks for the PR @cmonez this is much needed. We will test this out over the weekend and merge beginning of next week. The one thing that we will likely change is that instead of the red button we will probably use something more like this:
> 
> image
> 
> But let me test this out first and I'll update here after

Sounds good. I'll look out for your comments, i'll work on change once I get your feedback.

abaali:
@cmonez @Rperry2174 
i suggest this icon:
https://fontawesome.com/icons/history
image

Rperry2174:
this is really interesting... I really like this idea as I use the functionality of this with chrome back / forward buttons a lot: 
image

1. The icon is much better because its showing "going back in time" vs "un-zooming" which "going back in time" is both clearer and more intuitive
2. I like the idea of having two icons so you can go backwards and forwards if you want

From a logistics standpoint we should maybe change the implementation of this to allow for that in the future, which means we may need to use an actual library for this since it stores history in an array as opposed to one at a time. 

For example React Router keeps a copy of the history in an array so that you can navigate to / from particular entries in the history:

https://github.com/ReactTraining/react-router
https://github.com/Treora/redux-query-sync/issues/13#issuecomment-327361957

What do you think about this as an alternative @abaali @cmonez ? I'm going to do a little more digging the next couple of days and see what the best next course of action is (i.e. merging this in the meantime or going directly with another implementation) but very interested to get your thoughts as well

cmonez:
Yeah I think that icon is great @abaali , also love the idea of keeping the history of all previously selected time ranges. It would make it easier to keep going closer and closer within a given time window and then going back out. I'm gonna look at those links you sent @Rperry2174 while you decide what the next course of action is.

Rperry2174:
hey @cmonez I'm going to close this as I think we should reapproach this from a different route now that we are using react router and there is a `history` object that allows to move back in forth in the list of routes. I'll update the issue in case you're still interested in tackling this issue from this other angle but no worries if not!

Issue #16, opened 2021-01-15
Add view to compare Flamegraphs from two time periods


Rperry2174:
While this has a lot of potential to add value, it's definitely a much trickier issue to tackle. Posting anyway as we may move towards trying to make this happen depending on if people feel this is necessary.

Currently a common use case for Pyroscope is:

**A user has determined that part of their application or service is running slow or consuming too much CPU resources.** So, they take the following steps:

1. They look at Pyroscope "problem period" to pinpoint the file and function most responsible 
2. They fix / improve the speed of that function 
3. They look at a period post-fix to determine if their fix improved their CPU problem

Here is a picture that demonstrates this situation. 

![500x500_flame_graph_table](https://user-images.githubusercontent.com/23323466/104780914-65a38d80-5736-11eb-9fe5-96ca13472dd4.jpg)

- Before the fix, a function in `B` was taking up 75% of CPU resources while a function in `C` was taking up 25% 
- User fixed function in `B`
- After the fix, `B` now uses 25% of CPU resources while `C` now takes up 75%

The most important part about this picture is the sum of the total CPU time at the bottom. **_With context_** i.e. if you are the dev who wrote these functions, fixed them, and are now reviewing the results, then you would know that your overall CPU resource consumption decrease (congrats!)

However, even without the context that `B` is the only place where changes were made, looking at the aggregate CPU time of the table on the left and comparing it with that of the table on the right

Rperry2174:
Make some mechanism by which there can be two separate timelines associated with each side of the screen. Either by passing their data into the `FlamegraphRenderer` instead of `FlamegraphRenderer` using state or make the state able to support two different timelines / sets of flamegraph data.

End goal of this would be something that looks like this:
![small_ai_preview](https://user-images.githubusercontent.com/23323466/107100527-8311c780-67c9-11eb-9f9e-aeb2bc9765e2.gif)

Rperry2174:
I've started working on this in [another branch](https://github.com/pyroscope-io/pyroscope/compare/comparison-view)... Interested to get your thoughts @petethepig :

So currently the flow for `FlameGraphRenderer` is (roughly):
1. Parameters change (and thus update state )
2. Updates to state trigger new API call for the new data (if `from`, `until`, `max-nodes`, or `name` changes) 
3. The response of the API call updates the redux store `flamebearer` key 

[...]

PR #72, opened 2021-02-13
[WIP] Add Comparison View for Two Time Ranges

Rperry2174:
Working on #16

some decent sized changes:

- Switch charting library to use [ApexCharts](https://github.com/apexcharts/react-apexcharts) so that we can do the colored "ranges"
- Add `leftFrom`, `leftUntil`, `rightFrom`, `rightUntil` to route so that each chart can have its own respective date range
- Move table to be below chart instead of to the side of chart
- Move state to be local to `FlameGraphRenderer` instead of in redux store

Here's a preview -- actually I found an interesting case where CPU load was consistent, then randomly dropped for a period of time before going back to its initial levels. The reason can be seen when comparing the flamegraph on the left to the one on the right:

- Left flamegraph has an icicle with a bunch of logging that the right flamegraph does not have
- On futher inspection the root of this logging icicle is actually specifically **error logging** meaning that the reason that load was low on the left flamegraph is because an error was preventing the CPU from doing any real work! 

ezgif com-gif-maker

We will probably wait a little while to merge this as there are a couple of moving pieces right now and this is more of a proof of concept


Rperry2174:
Closing because I split this into multiple prs and it is done being addressed in #90

PR #75, opened 2021-02-14
fix: unknown revision cosmtrek/air v1.21.2 => v1.12.2

yveshield:


Rperry2174:
Hi @yveshield, thanks for contributing! Could you please add some context on why you're proposing this change?

yveshield:
> Hi @yveshield, thanks for contributing! Could you please add some context on why you're proposing this change?

After import, prompted this package version number error, I went to check https://github.com/cosmtrek/air/releases after I found that v1.21.2 does not exist, which needed to be fixed anyway, 12 and 21,I suspect it is a typing error, and it works, so I changed it.

petethepig (approved):
Looks good! Thank you @yveshield! I found a relevant issue in air repo https://github.com/cosmtrek/air/issues/107

yveshield:
> Looks good! Thank you @yveshield! I found a relevant issue in air repo [cosmtrek/air#107](https://github.com/cosmtrek/air/issues/107)

It seems to be the case that cosmtrek/air had the wrong version (v1.21.2), which was quickly removed but still cached by goproxy. cosmtrek/air has been informed of this and has stated that a larger version number will be given soon, so that the problem no longer exists. As for now, I or anyone else who needs to can use go mod replace to cope with it. Then I can close this pr now.

PR #76, opened 2021-02-14
fix upload URL to include existing path

geoah:
Currently the upload URL overrides the whole path on the server URL given to the agent.
Assuming the server URL of `https://localhost/pyroscope/` is given to the agent, it will attempt to upload to `https://localhost/ingest` instead of the (imo at least) expected `https://localhost/pyroscope/ingest`.

If you think this might break things please let me know of any better ways of dealing with this issue and I can update the PR. :)

Thank you very much in advance!


petethepig (approved):
Looks good, thank you George!

PR #77, opened 2021-02-16
Move flamebearer to local state instead of redux store

Rperry2174:
Breaking out #72 into multiple PRs so its easier to follow...

Basically we eventually want to have situations where we have **two** `<FlameGraphRenderer />` on the same page in order to make a comparison view.

The problem with this is that as long as the `flamebearer` data is in the redux store, we'd need to add a new key to the store in order to handle it. In reality, the `flamebearer` data is a function of the input parameters. If the input parameters change the `flamebearer` key will always refresh. 

So the solution here is that instead of having `flamebearer` available to all components I'm moving it to be stored locally and refreshed whenever one of the parameters changes

Also added ability to turn Flamegraph / table in vertical or horizontal orientation


webapp/javascript/components/FlameGraphRenderer.jsx
```diff
@@ -87,19 +90,62 @@ class FlameGraphRenderer extends React.Component {
...
     }
-    this.props.actions.fetchJSON(this.props.renderURL);
+    this.fetchFlameBearerData(this.props.renderURL);
   }
 
-  componentDidUpdate(prevProps) {
-    if (prevProps.renderURL != this.props.renderURL) {
-      this.props.actions.fetchJSON(this.props.renderURL);
+  componentDidUpdate(prevProps, prevState) {
+    if (this.getParamsFromRenderURL(this.props.renderURL).name != this.getParamsFromRenderURL(prevProps.renderURL).name ||
```
Rperry2174, line 97: Explicitly defining params because in the future there will be three different set of params:
1. Global params the global timeline
2. Left `<FlameGraphRenderer />`
3. Right `<FlameGraphRenderer />

and we only want each component to update when its respective parameters change

petethepig (approved):
LGTM

Rperry2174:
Works almost perfectly but with small bug when rendering the flamegraph... closing for now and will reopen after its fixed:

LouisInFlow:
@Rperry2174 Try adding names, levels, numTicks, sampleRate to state. And use setState anytime you need to mutate those values.

LouisInFlow:
@Rperry2174 I believe what's happening is that the data fetching, re-rending of component and rendering of canvas aren't synced properly. When you pass flamebearer as props everything is in sync, however when you use componentDidUpdate/setState to handle flamebearer things get out of sync. 

One approach would be to add redux-saga which is a redux library for handling side effects. 

Another approach would be to add all the variables (topLevel, slectedLevel, name, level etc) into state and use setState to update them.

Another approach would be to do both of the above to clean up tech debt.

Rperry2174:
This makes sense I'll give that a shot thanks for taking a look @LouisInFlow

Rperry2174:
@LouisInFlow Just wanted to update that your solution worked here thanks!

PR #82, opened 2021-02-18
Revert "Move flamebearer to local state instead of redux store"

Rperry2174:
Reverts pyroscope-io/pyroscope#77

Found that the refresh button doesn't work


Rperry2174:
We could also just fix this by:
adding:

```
prevProps.refreshToken != this.props.refreshToken
```

https://github.com/pyroscope-io/pyroscope/blob/093052f6a2090927a25921ea5ab4d36f04288df4/webapp/javascript/components/FlameGraphRenderer.jsx#L92

PR #84, opened 2021-02-18
Import pyroscope version from package.json

abaali:


petethepig:
Hi @abaali !

Thanks for the PR. Could you also remove `PYROSCOPE_VERSION` from here https://github.com/pyroscope-io/pyroscope/blob/master/scripts/webpack/webpack.common.js#L181 ? That's where it was coming from before. I like your way much better, I just want to make sure we clean up the old one to avoid confusion.

petethepig (approved):
LGTM

petethepig (changes requested):
See the comment above

abaali:
@petethepig 
I had some issues adding commits to this pr so I will open a new one with the changes.

Issue #70, opened 2021-02-11
write: broken pipe


sujit-baniya:
Out of  requests (on local), one request end up with write: broken pipe

ERRO[0263] Error happened when uploading a profile:Post "http://localhost:4040/ingest?from=1613035910&name=Verify+Rest&sampleRate=100&spyName=gospy&until=1613035920": write tcp 127.0.0.1:58206->127.0.0.1:4040: write: broken pipe 

And another error:
ERRO[0453] Error happened when uploading a profile:Post "http://localhost:4040/ingest?from=1613036100&name=Verify+Rest&sampleRate=100&spyName=gospy&until=1613036110": EOF

petethepig:
Do most requests succeed, or does every single one fail?

sujit-baniya:
@petethepig Most of the request succeed.. and some are failed (I would say every 5 or 10 request, above error occurs

Issif:
I've same issue, and for all uploads:

```
ERRO[0029] Error happened when uploading a profile:Post "http://localhost:4040/ingest?from=1613826350&name=falcosidekick&sampleRate=100&spyName=gospy&until=1613826360": EOF 
```

petethepig:
I think intermittent issues were fixed by #87.

sujit-baniya:
@petethepig Still there are these issues:

```go
ERRO[0091] Error happened when uploading a profile:Post "http://localhost:4040/ingest?from=1614087850&name=Verify+Rest&sampleRate=100&spyName=gospy&until=1614087860": EOF 
```

petethepig:
@sujit-baniya  Are you using the latest version? we just released one

sujit-baniya:
@petethepig Yes I pulled latest docker image for server and latest agent for app

PR #87, opened 2021-02-19
fix connection reuse

wusphinx:
When I use pyroscope with golang, I got some problem, as the same with the issue [#70](https://github.com/pyroscope-io/pyroscope/issues/70), and I found something with wireshark as blow:
image
The Server closed connection, the reason is: without setting `IdleTimeout` in `http.Server`,  `ReadTimeout` will be used instead,  We can find instructions [here](https://github.com/golang/go/blob/release-branch.go1.14/src/net/http/server.go#L2560),  besides,  the agent will sent to server at 10s intervals, so that's what happens in the picture above。
So I set `IdleTimeout` with magic number `30`, so the connection can be reused.


petethepig (approved):
Thank you!

PR #143, opened 2021-03-29
Export flamegraph and table to pdf/png

cjsampson:
This PR adds the ability to export the flamegraph as a PNG/PDF and the table pane as a PDF.


webapp/sass/profile.scss
```diff
@@ -223,6 +224,13 @@ $pane-width: 6px;
     margin: 0 $pane-width;
     width: 50%;
 
+    .flamegraph-header {
+      display: flex;
+      justify-content: space-between;
+      align-items: center;
+      border: 2px solid lime;
```
Rperry2174, line 231: Lets remove the lime border here. I only have this in the screenshot to highlight the element of the page that I was referring to.

webapp/sass/profile.scss
```diff
@@ -223,6 +224,13 @@ $pane-width: 6px;
     margin: 0 $pane-width;
     width: 50%;
 
+    .flamegraph-header {
```
Rperry2174, line 227: Can you add padding bottom to this. I think `padding-bottom: 5px` probably would work here.
image

webapp/javascript/components/ExportData.jsx
```diff
@@ -0,0 +1,144 @@
...
+  };
+
+  const handleToggleMenu = (event) => {
+    event.preventDefault();
+    setToggleMenu(!toggleMenu);
+  };
+
+  return (
+    <div className="dropdown-container">
+      <button type="button" className="btn" onClick={handleToggleMenu}>
```
Rperry2174, line 99: Can we make this button a square as opposed to a rectangle? Probably 35x35 pixels would work here

webapp/javascript/components/FlameGraphRenderer.jsx
```diff
@@ -492,6 +495,11 @@ class FlameGraphRenderer extends React.Component {
         key={'flamegraph-pane'}
         className={clsx("pane", { hidden: this.state.view === "table", "vertical-orientation": this.props.viewType === "double" })}
       >
+        <div className='flamegraph-header'>
+          <span></span>
+          <span>Frame width represents CPU time per function</span>
```
Rperry2174, line 500: I think this is great. However the text is a really bright white that is kind of distracting. Could we make it a little more grey maybe something like this: `#a7a7a7` And lets make the font size slightly smaller as well maybe ~`14px`

webapp/javascript/components/ExportData.jsx
```diff
@@ -0,0 +1,144 @@
...
+        "JPEG",
+        pdfXOffset,
+        pdfYOffset,
+        canvas.width / count,
+        canvas.height / count
+      );
+
+      const textXOffset = pdfXOffset;
+      const textYOffset = 10;
+      pdf.text(textXOffset, textYOffset, "Flamegraph Visual");
```
Rperry2174, line 57: I think "Flamegraph Visual" could be more descriptive here. Could you create a function similar to this:
```
formatPdfTitle = () =>{ }
```

Where the function returns a title in the format:
```
{application_name} - from: {from} until: {until}
```

This will be more useful if you want to recreate the same flamegraph you can use these parameters to do so. I know sometimes the parameters will be named kind of funny (i.e. `from=now-1h` or `from=10299489403094`) but I think thats fine.

Rperry2174 (changes requested):
Thanks @cjsampson just a couple of small changes we appreciate you adding this for us!

PR #159, opened 2021-04-21
Update README.ch.md

cnych:
Correct Chinese translation content


cnych:
oh~~ I just commit a README File, then the go test workflow timeout~

Rperry2174:
I can't really read it but thanks for the edits @cnych!

Issue #168, opened 2021-04-28
Implement data retention in storage layer


petethepig:
### Feature Description

Sometimes it doesn't make sense to store data indefinitely so we need to implement a way to do data retention. 

Internally Pyroscope uses segment trees to store the data. 
See our [Storage Design Doc](https://github.com/pyroscope-io/pyroscope/blob/main/docs/storage-design.md) for more context.

![Page2 3](https://user-images.githubusercontent.com/662636/116452030-192a1c80-a812-11eb-9489-dd2660d43e50.png)

We need to add a new retention threshold parameter that would control how much data (in days) can be stored in pyroscope. Then we would run some sort of a `Cleanup` method every once in a while to delete old data:

![Page3](https://user-images.githubusercontent.com/662636/116452984-2dbae480-a813-11eb-8c4a-f31808136182.png)

### TODO

* [x] Implement a method that would delete data before a certain time on [segment tree](https://github.com/pyroscope-io/pyroscope/blob/main/pkg/storage/segment/segment.go) and return a list of trees that need also need to be deleted.
* [ ] cover it with specs
* [x] add a retention option to [config](https://github.com/pyroscope-io/pyroscope/blob/main/pkg/config/config.go#L31)
* [ ] in storage package, put conditions in [Put method](https://github.com/pyroscope-io/pyroscope/blob/main/pkg/storage/storage.go#L180) that would prevent data ingestions past the retention threshold
* [ ] in storage package add a method that would delete segment tree nodes as well as relevant trees
* [ ] create a separate goroutine that would call this cleanup method every X minutes

PR #174, opened 2021-05-02
Resolves #165 - implemented out of space check

### LINKED ISSUE #165: Make pyroscope handle lack of disk space gracefully
Currently when pyroscope runs out of disk space it might corrupt the data in the database. We should make it so that:
* [ ] pyroscope detects that it's close to running out of space
* [ ] when this happens, stop accepting writes
* [ ] have a way to notify users that this is close to happening, e.g we should have some notifications on the frontend and also probably a dedicated "server status" page (https://github.com/pyroscope-io/pyroscope/issues/164) — for v1 let's just have the controller respond with some 4xx error.

For more information about the storage design, see [storage design doc](https://github.com/pyroscope-io/pyroscope/blob/main/docs/storage-design.md)

---

ideally we should also make it so that pyroscope can survive some amount of data corruption

s4kibs4mi:


pkg/util/resource_usages/disk_usages.go
```diff
@@ -0,0 +1,68 @@
...
+	stat.Free = fs.Bfree * uint64(fs.Bsize)
+	stat.Used = stat.All - stat.Free
+
+	stat.UsedPercentage = (float64(stat.Used) / float64(stat.All)) * 100
+	stat.FreePercentage = (float64(stat.Free) / float64(stat.All)) * 100
+	return &stat, nil
+}
+
+func IsRunningOutOfSpace(cfg *config.Config) bool {
+	stats, err := DiskUsage("/")
```
petethepig, line 41: We should use `cfg.Server.StoragePath` here instead of `"/"` if possible

pkg/server/ingest.go
```diff
@@ -100,6 +101,12 @@ func ingestParamsFromRequest(r *http.Request) *ingestParams {
 }
 
 func (ctrl *Controller) ingestHandler(w http.ResponseWriter, r *http.Request) {
+	if resource_usages.IsRunningOutOfSpace(ctrl.cfg) {
```
petethepig, line 104: I think we should move this check into `storage.Put`, because we also do self-profiling for the `pyroscope server` itself where it invokes `storage.Put` directly and with this current implementation this check won't be performed for those ingestions.

pkg/config/config.go
```diff
@@ -59,6 +59,11 @@ type Server struct {
 
 	// currently only used in our demo app
 	HideApplications []string `def:"" desc:"please don't use, this will soon be deprecated"`
+
+	OutOfSpaceThreshold              float64 `def:"10" desc:"Threshold value to consider out of space in percentage"`
+	OutOfSpaceWarningThreshold       float64 `def:"20" desc:"Threshold value to consider to show out of space in warning percentage"`
+	OutOfSpaceStaticThreshold        int     `def:"512" desc:"Threshold static value to consider out of space. Useful when % value is very high than required value"`
+	OutOfSpaceWarningStaticThreshold int     `def:"1024" desc:"Threshold static value to consider to show out of space warning. Useful when % value is very high than required value"`
```
petethepig, line 66: I know we've decided earlier that it makes sense to have 4 values, but I thought about it some more (also talked to @Rperry2174) and I think we should just leave 1 and have it be the threshold value in bytes for when the writes should stop. We can set it at 512 MB and I will run some tests later to determine a perfect value for this. This will make code simpler and it will be easier for users to understand what this is (currently it's really hard to get the difference between the 4 values). And it will work on any system. And if users have too little space allocated for pyroscope this way we will be able to notify them about it.

And, after we implement #168 (data retention) we can set the whole thing up in a way that it would automatically delete old data when it's above the threshold. This way we also don't need the warning thresholds.

Sorry about the confusion.

petethepig (changes requested):

PR #182, opened 2021-05-09
Closes #168 - Implement data retention in storage layer (WIP)

### LINKED ISSUE #168: Implement data retention in storage layer

s4kibs4mi:
Data is important but how long? With this implementation, you can decide how much old data you want to keep and which you want to be removed.


pkg/cli/cli.go
```diff
@@ -365,3 +366,35 @@ func printDiskUsage(cfg *config.Config) {
...
+			depth = cfg.Server.StorageMaxDepth
+		}
+
+		labels := []string{
+			"pyroscope.server.cpu",
+			"pyroscope.server.alloc_space",
+			"pyroscope.server.alloc_objects",
+			"pyroscope.server.inuse_space",
+			"pyroscope.server.inuse_objects",
+		}
```
petethepig, line 388: I see you hardcoded these for now. We need this list to be generated dynamically and it should include all of the keys out there.

To get a list of all possible keys you can do this:
* call `labels.GetValues` with `__name__` as the key
* for each value, call `dimensions.Get(k+":"+v)` with key being `__name__`, and `v` being a value from GetValues ([example](https://github.com/pyroscope-io/pyroscope/blob/main/pkg/storage/storage.go#L266))

pkg/storage/storage.go
```diff
@@ -314,6 +314,47 @@ func (s *Storage) Get(gi *GetInput) (*GetOutput, error) {
...
+
+	if s.closing {
+		return errClosing
+	}
+
+	lg := logrus.WithField("key", ci.Key.Normalized())
+
+	sk := ci.Key.SegmentKey()
+	st := s.segments.Get(sk).(*segment.Segment)
+	st.Get(ci.TimeThreshold, ci.TimeThreshold, func(depth int, samples, writes uint64, t time.Time, r *big.Rat) {
```
petethepig, line 335: Couple of problems with this approach:
* `Get` will only return a limited number of nodes, because it optimizes for the smallest number of nodes to return as possible
```
|__________________________|
|XXXXXXXX|________|________|
|__|__|__|XX|XX|__|__|__|__|
    ^          ^-threshold
    missed nodes
```
* We should also delete the relevant nodes from the segment tree, not just `s.trees`

pkg/storage/cache/cache.go
```diff
@@ -136,3 +138,31 @@ func (cache *Cache) Get(key string) interface{} {
 	lg.Debug("storage hit")
 	return val
 }
+
+func (cache *Cache) Cleanup(key string) error {
+	lg := logrus.WithField("key", key)
+
+	if cache.lfu.UpperBound > 0 {
+		// TODO: Cleanup from cache
```
petethepig, line 146: this is pretty important too

pkg/config/config.go
```diff
@@ -59,6 +59,9 @@ type Server struct {
 
 	// currently only used in our demo app
 	HideApplications []string `def:"" desc:"please don't use, this will soon be deprecated"`
+
+	RetentionThresholdDays  int `def:"30" desc:"data retention threshold in days"`
+	RetentionThresholdDepth int `def:"3" desc:"data retention threshold in depth"`
```
petethepig, line 64: I don't think we need `RetentionThresholdDepth` option. I think `RetentionThresholdDays` might be enough.

pkg/storage/storage.go
petethepig: I think the way to solve this is to write a dedicated `Cleanup` function for the segment tree as well.

pkg/storage/dimension/dimension.go
```diff
@@ -144,3 +144,19 @@ func Intersection(input ...*Dimension) []key {
 		}
 	}
 }
+
+func Union(input ...*Dimension) []key {
```
petethepig, line 148: As it's implemented right now, this function will return a lot of duplicates I think. We should optimize it to take advantage of the fact that the input dimensions are all sorted arrays, similarly to how `Intersection` is implemented.

pkg/storage/segment/segment.go
```diff
@@ -144,6 +153,19 @@ func (sn *streeNode) get(st, et time.Time, cb func(sn *streeNode, d int, t time.
 	}
 }
 
+func (sn *streeNode) cleanup(timeThreshold time.Time, cb func(depth int, t time.Time)) {
+	rel := sn.cleanupRelationship(timeThreshold)
+	if sn.present && (rel == inside || rel == match) {
+		cb(sn.depth, sn.time)
```
petethepig, line 159: This node (`sn`) should also be deleted from the segment tree after we call `cb`

pkg/config/config.go
```diff
@@ -59,6 +59,8 @@ type Server struct {
 	HideApplications []string `def:"" desc:"please don't use, this will soon be deprecated"`
 
 	OutOfSpaceThreshold bytesize.ByteSize `def:"512MB" desc:"Threshold value to consider out of space in bytes"`
+
+	RetentionThresholdDays int `def:"3" desc:"data retention threshold in days"`
```
petethepig, line 63: We should make this a `time.Duration`. ParseDuration does not support setting the duration in days, so we'll need to implement that. For that we can make a wrapper, similarly to how we have a wrapper for `time.Time` values: https://github.com/pyroscope-io/pyroscope/blob/main/pkg/cli/cli.go#L51-L74

petethepig (approved):

petethepig (approved):

tuananh2508:
Does this feature has document yet ? Cause I can't find this feature on https://pyroscope.io/docs/ @petethepig

Issue #154, opened 2021-04-08
Add ability to change sampling rate


petethepig:
Currently it's hardcoded at 100 Hz.

Multiple people asked for this in both [Slack](https://pyroscope.io/slack) and here on Github, e.g #153

petethepig:
I’m thinking, when people run `pyroscope exec python` they should be able to set sample rate, e.g `pyroscope exec -sample-rate 50 python`

Same with the go profiler:
```
profiler.Start(profiler.Config{
    ApplicationName: "simple.golang.app",
    ServerAddress:   "http://pyroscope:4040", // this will run inside docker-compose, hence `pyroscope` for hostname
    SampleRate: 50,
})
```

qzaidi:
Using the custom profiler has somehow resulted in all of the cpu profiles vanishing from the display. Memory profiling data does show up.

alonlong:
can u help to show more details about the problem?

qzaidi:
changed sampling to 20 from 100 hz as it seems profiler had increased cpu usage significantly - this switched it to the custom profiler internally I guess,  and now only alloc and inuse traces are shown, no cpu traces are present.

qzaidi:
This issue happens always when custom profiler is used. In both cases, the agent seems to be sending the cpu samples correctly to server, but  the server won't display anything except when a sample rate of 100 hz is used.

Here's some example data that is being sent when sample rate is 20

[...]

PR #190, opened 2021-05-16
Fix issue #154: Add ability to change sampling rate

alonlong:
make the sample rate configured for python, ruby and go, but it doesn't take affect for go, because we use the the 'runtime profile' of golang, we need to find the way to build the customer profiler for go.

### DIFF
--- a/pkg/agent/profiler/profiler.go
+++ b/pkg/agent/profiler/profiler.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/pyroscope-io/pyroscope/pkg/agent"
 	"github.com/pyroscope-io/pyroscope/pkg/agent/spy"
+	"github.com/pyroscope-io/pyroscope/pkg/agent/types"
 	"github.com/pyroscope-io/pyroscope/pkg/agent/upstream/remote"
 )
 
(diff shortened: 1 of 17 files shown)


pkg/config/config.go
```diff
@@ -79,6 +79,7 @@ type DbManager struct {
 type Exec struct {
 	SpyName                string        `def:"auto" desc:"name of the profiler you want to use. Supported ones are: <supportedProfilers>"`
 	ApplicationName        string        `def:"" desc:"application name used when uploading profiling data"`
+	SampleRate             uint          `def:"100" desc:"sample rate for reading profile data. 100 means reading 100 times per second"`
```
petethepig, line 82:
```suggestion
	SampleRate             uint          `def:"100" desc:"sample rate for the profiler in Hz. 100 means reading 100 times per second"`
```

pkg/agent/profiler/profiler.go
```diff
@@ -20,12 +20,17 @@ var (
...
+	DefaultSampleRate = 100 // 100 times per second
+)
+
 var DefaultProfileTypes = []ProfileType{ProfileCPU, ProfileAllocObjects, ProfileAllocSpace, ProfileInuseObjects, ProfileInuseSpace}
 
 type Config struct {
 	ApplicationName string // e.g backend.purchases
 	ServerAddress   string // e.g http://pyroscope.services.internal:4040
 	AuthToken       string // specify this token when using pyroscope cloud
+	SampleRate      uint32
```
petethepig, line 33:
```suggestion
	# SampleRate      uint32
```

I think we should comment this out until we implement golang profiler support (https://github.com/pyroscope-io/pyroscope/issues/191)

Issue #163, opened 2021-04-26
nil pointer dereference in transporttrie


m110:
Hey Team 👋 

We've stumbled upon a panic following metrics ingestion. As far as I can tell it hasn't been reported yet. 🙂

This is on version `0.0.28`

It seems there's an unhandled error at https://github.com/pyroscope-io/pyroscope/blob/v0.0.28/pkg/convert/parser.go#L14

```
runtime error: invalid memory address or nil pointer dereference

goroutine 792501 [running]:
net/http.(*conn).serve.func1(0xc0373cfe00)
	net/http/server.go:1801 +0x147
panic(0x127cd60, 0x1b83d80)
	runtime/panic.go:975 +0x3e9
github.com/pyroscope-io/pyroscope/pkg/structs/transporttrie.(*Trie).Iterate(0x0, 0xc037e1ea20)
	github.com/pyroscope-io/pyroscope/pkg/structs/transporttrie/trie.go:181 +0x43
github.com/pyroscope-io/pyroscope/pkg/convert.ParseTrie(0x7f530db67258, 0xc057f7e680, 0xc02167b640, 0x7f530db67258, 0xc057f7e680)
	github.com/pyroscope-io/pyroscope/pkg/convert/parser.go:15 +0x6b
github.com/pyroscope-io/pyroscope/pkg/server.(*Controller).ingestHandler(0xc006288420, 0x15a04e0, 0xc057f92000, 0xc03f2ae800)
	github.com/pyroscope-io/pyroscope/pkg/server/ingest.go:90 +0x206
net/http.HandlerFunc.ServeHTTP(0xc0002921f0, 0x15a04e0, 0xc057f92000, 0xc03f2ae800)
	net/http/server.go:2042 +0x44
net/http.(*ServeMux).ServeHTTP(0xc0007f2200, 0x15a04e0, 0xc057f92000, 0xc03f2ae800)
	net/http/server.go:2417 +0x1ad
net/http.serverHandler.ServeHTTP(0xc00628c1c0, 0x15a04e0, 0xc057f92000, 0xc03f2ae800)
	net/http/server.go:2843 +0xa3
net/http.(*conn).serve(0xc0373cfe00, 0x15a1d60, 0xc0553fd3c0)
	net/http/server.go:1925 +0x8ad
created by net/http.(*Server).Serve
	net/http/server.go:2969 +0x36c
```

Rperry2174:
Thanks for reporting well add a fix for this in the next release!

[...]

PR #194, opened 2021-05-18
fix issue#163: nil reference when paring the tries

alonlong:
add the error handler for paring the tries


petethepig (approved):
Looks good, thanks for fixing this @alonlong . This closes #163

Issue #191, opened 2021-05-16
Build custom cpu profiler for golang


alonlong:
Now we use the StartCPUProfile and StopCPUProfile for reading the pprof data, and the default sample rate for runtime is 100 hz, there is no way for us to change and make it affected. So we need to find a way to build the custom cpu profiler for pyroscope.

The possible way is using the following runtime function to read the profile data:

//go:linkname readProfile runtime/pprof.readProfile
func readProfile() (data []uint64, tags []unsafe.Pointer, eof bool)

but still need to research how to parse the data......

Issue #198, opened 2021-05-19
When i configue a wrong format of ServerAddress


alonlong:
Let's see a code like this:
```
func main() {
        profiler.Start(profiler.Config{
                ApplicationName: "backend-gin",                
                ServerAddress: "192.168.0.10:4040",  // the right value should be "http://192.168.0.10:4040"
        })

        r := gin.Default()
        r.GET("/ping", func(c *gin.Context) {
                c.JSON(http.StatusOK, gin.H{
                        "message": "pong",
                })
        })

        r.Run()
}
```

After i build and run this code, it happens a panic:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x48 pc=0x9c5535]

goroutine 1 [running]:
github.com/pyroscope-io/pyroscope/pkg/agent/profiler.Start(0xaf242c, 0xb, 0xaf427e, 0xf, 0x0, 0x0, 0x0, 0x0, 0xf2ec20, 0x5, ...)
        $HOME/Project/go/pkg/mod/github.com/pyroscope-io/pyroscope@v0.0.30/pkg/agent/profiler/profiler.go:50 +0x115
main.main()
        $HOME/Project/go/src/project/scope/main.go:11 +0x58
```

i think the error is caused by the code, which will return an error when the address is not valid:
```
urlObj, err := url.Parse(cfg.UpstreamAddress)
if err != nil {
    return nil, err
}
```

And for the Start function in profiler.go, the code is like this:
```
        u, err := remote.New(remote.RemoteConfig{
		AuthToken:              cfg.AuthToken,
		UpstreamAddress:        cfg.ServerAddress,
		UpstreamThreads:        4,
		UpstreamRequestTimeout: 30 * time.Second,
	})

	u.Logger = cfg.Logger

	if err != nil {
		return nil, err
	}
```

u.Logger = cfg.Logger is the place to cause the panic: 
```
panic: runtime error: invalid memory address or nil pointer dereference
```

I think the better code for the Start function should be like this:
```
        u, err := remote.New(remote.RemoteConfig{
		AuthToken:              cfg.AuthToken,
		UpstreamAddress:        cfg.ServerAddress,
		UpstreamThreads:        4,
		UpstreamRequestTimeout: 30 * time.Second,
	})
	if err != nil {
		return nil, err
	}
        u.Logger = cfg.Logger
```

PR #199, opened 2021-05-19
fix the issues: #191 and #198

alonlong:
fix the issues:
1. Build custom cpu profiler for golang -> #191
1. When i configue a wrong format of ServerAddress -> #198


petethepig (approved):
I changed a few things, mostly around logging.

PR #200, opened 2021-05-19
Add .Net support

kolesnikovae:
The change introduces support for .Net CPU time profiling.

`dotnetspy` uses [Diagnostic IPC Protocol](https://github.com/dotnet/diagnostics/blob/main/documentation/design-docs/ipc-protocol.md) client to communicate to .Net runtime and gather trace events. It is important to note that only managed code is accounted. Refer to [dotnetdiag](https://github.com/pyroscope-io/dotnetdiag) package for implementation details.

Supported versions:
 - .Net 5.0
 - .Net Core 3.1

Examples:
 - web - a web app that listens on :5000/tcp and performs some calculations.
 - fast-slow - a single thread application similar to examples for other runtimes:
image


petethepig (approved):
Everything looks good, merging

PR #203, opened 2021-05-21
Fix dotnetspy panic on premature exit and add tests.

kolesnikovae:
There is a case when `dotnetspy` couldn't establish a connection to Diagnostics Server before the next snapshot/stop call (e.g, target process exited, or the socket file cannot be found, etc).

This fix adds a check whether the session has been actually created before accessing it. Given that there are no race conditions between reset/stop/snapshot calls, redundant mutexes were removed.


petethepig (approved):
Looks good, merging

PR #208, opened 2021-05-22
Closes: #192 Use the specific config structure for each command

### LINKED ISSUE #192: Use the specific config structure for each command
Now there are different flagsets for different commands, but for each command, it uses a global config structure:

```
type Config struct {
    Version bool
    Agent     Agent     `skip:“true”`
    Server    Server    `skip:“true”`
    Convert   Convert   `skip:“true”`
    Exec      Exec      `skip:“true”`
    DbManager DbManager `skip:“true”`
}
```

just like the server command, the function is like this:
```
func startServer(cfg *config.Config)
```

i think it's better to use its own config for each command, the function should be like this:
```
func startServer(cfg *config.Server)
```

AdrK:
#192

### DIFF
--- a/pkg/agent/cli/agent.go
+++ b/pkg/agent/cli/agent.go
@@ -15,19 +15,19 @@ import (
 )
 
 type Agent struct {
-	cfg            *config.Config
+	cfg            *config.Agent
 	cs             *csock.CSock
 	activeProfiles map[int]*agent.ProfileSession
 	id             id.ID
 	u              upstream.Upstream
 }
 
-func New(cfg *config.Config) *Agent {
+func New(cfg *config.Agent) *Agent {
 	// TODO: handle this error properly
 	r, _ := remote.New(remote.RemoteConfig{
-		UpstreamThreads:        cfg.Agent.UpstreamThreads,
-		UpstreamAddress:        cfg.Agent.ServerAddress,
-		UpstreamRequestTimeout: cfg.Agent.UpstreamRequestTimeout,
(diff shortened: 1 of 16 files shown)


pkg/agent/cli/agent.go
```diff
@@ -15,19 +15,19 @@ import (
...
-	cfg            *config.Config
+	cfg            *config.Agent
 	cs             *csock.CSock
 	activeProfiles map[int]*agent.ProfileSession
 	id             id.ID
 	u              upstream.Upstream
 }
 
-func New(cfg *config.Config) *Agent {
+func New(cfg *config.Agent) *Agent {
```
AdrK, line 25: @petethepig Seems that this is unused, is that possible?

pkg/agent/cli/agent.go
```diff
@@ -15,19 +15,19 @@ import (
 )
 
 type Agent struct {
-	cfg            *config.Config
```
AdrK, line 18: @petethepig 
Do we want to keep the config.Config struct?
If my understanding is correct, the lifetime of all the config structs are withing the `Start(cfg *config.Config)` function in cli.go, so maybe we don't need an extra container for all the configs?

pkg/agent/cli/agent.go
petethepig: Yeah, I think we'll delete it soon

pkg/agent/cli/agent.go
petethepig: Yeah, good point, we might do that in the future

PR #211, opened 2021-05-24
Disable inlining for golang example

kolesnikovae:
Otherwise, `work` function is inlined (go1.16.4 windows/amd64) and flamegraph becomes slightly less demonstrative.

Inlined:
image

With `go:noinline`:
image


petethepig (approved):
Oh, that's an interesting find, merging!

PR #212, opened 2021-05-25
Windows support (agent)

kolesnikovae:
This PR is the first step toward full-fledged Windows support. With this change, pyroscope can be run on Windows in `agent`, `exec` and `connect` modes.

Limitations:
 - Minimal supported version: Windows 10 Build 17063.
 - Only `gospy` and `dotnetspy` can be used. In contrast to Linux/macOS, on Windows, `dotnetspy` accounts time spent in unmanaged code.


pkg/agent/debugspy/placeholder.go
```diff
@@ -1,3 +1 @@
-// +build !debugspy
```
kolesnikovae, line 1: It seems the build tags in empty "placeholders" are redundant.

pkg/util/disk/usage_unix.go
```diff
@@ -1,3 +1,5 @@
+// +build !windows
```
kolesnikovae, line 1: `FreeSpace` on Linux/macOS relies on [statfs](https://man7.org/linux/man-pages/man2/statfs.2.html) bfree (free blocks in filesystem), consider use of bavail (free blocks available _to unprivileged user_).

pkg/agent/debugspy/placeholder.go
petethepig: Not exactly... I think I added those because otherwise imports here don't work: https://github.com/pyroscope-io/pyroscope/blob/main/pkg/agent/session.go#L14-L17

Maybe we should add a comment in this case because I this is not intuitive

petethepig: Oh, I think I misunderstood your comment. I see what you're saying. Yeah, we can probably drop them in some of these files.

petethepig (approved):
LGTM

Issue #167, opened 2021-04-27
Dynamically resize cache depending on the amount of RAM available


petethepig:
### Introduction

Pyroscope keeps most frequently used objects in an LFU cache layer in RAM. This way we get good write and read performance. 

See our [Storage design doc](https://github.com/pyroscope-io/pyroscope/blob/main/docs/storage-design.md) for more context on the overall architecture.

![image](https://user-images.githubusercontent.com/662636/116319718-dca3e580-a76b-11eb-956b-cb85bc70b162.png)

### Summary of the problem

Currently the size of this cache is hardcoded at 1,000 or 10,000 depending on the type of object, see [the relevant code here](https://github.com/pyroscope-io/pyroscope/blob/main/pkg/config/config.go#L44-L47).

This works for an average system with ~ 8GB of RAM, but fails if you want to use pyroscope on a smaller or larger machine. With smaller ones you end up with out-of-memory (OOM) events, and on larger systems you're just not utilizing all of RAM available, which is also bad.

![Page2 2](https://user-images.githubusercontent.com/662636/116324453-057ca880-a775-11eb-8a4a-bc746df51ede.png)
<p align="center">
  <strong>When cache gets too big it takes time to save it to disk</strong>
</p>

### Proposed solution

We need to change the way these caches work. We need to make the cache take into consideration the amount of RAM available. Currently we use this project https://github.com/dgrijalva/lfu-go for LFU implementation. We should probably fork it or write our own.

### TODO:

* [ ] modify LFU cache to evict based on amount of RAM available, not the number of objects in RAM, see [relevant code where cache initialization happens](https://github.com/pyroscope-io/pyroscope/blob/master/pkg/storage/cache/cache.go#L26-L29)

PR #213, opened 2021-05-25
Dynamically resizes cache depending on the amount of RAM available (issue #167)

alonlong:
Fix the issue: Dynamically resize cache depending on the amount of RAM available #167


pkg/storage/cache/cache.go
```diff
@@ -71,11 +76,18 @@ func (cache *Cache) saveToDisk(key string, val interface{}) {
...
 func (cache *Cache) Flush() {
+	// evict all the items in cache
 	cache.lfu.Evict(cache.lfu.Len())
+
 	close(cache.lfu.EvictionChannel)
+	// wait until cache flushing is finished
 	<-cache.cleanupDone
 }
 
+func (cache *Cache) Evit(percent float64) {
```
petethepig, line 87: Is this a typo? I think you meant `Evict`

pkg/storage/storage.go
```diff
@@ -163,6 +164,42 @@ func New(cfg *config.Config) (*Storage, error) {
 		return tree.New()
 	}
 
+	// load the total memory of the server
+	vm, err := mem.VirtualMemory()
```
petethepig, line 168: Can you please add tests for this?

For tests we could add a way to set this limit arbitrarily, to something like 1 MB, and then make sure the system still works.

pkg/storage/cache/cache.go
alonlong: right~ sorry

pkg/storage/cache/cache.go
```diff
@@ -71,11 +76,18 @@ func (cache *Cache) saveToDisk(key string, val interface{}) {
...
+	cache.lfu.Evict(cache.lfu.Len() / int(percent*100))
```
AdrK, line 88:
```suggestion
	cache.lfu.Evict(int(cache.lfu.Len() * percent))
```
Current version works for `percent == 0.10` by an accident, it will not work for any other value :P

pkg/storage/storage.go
```diff
@@ -163,6 +164,49 @@ func New(cfg *config.Config) (*Storage, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	interval := time.Duration(10)
+	// start a timer for checking if the memory used by application
+	// is more than 25% of the total memory, if so, trigger eviction
+	// with 10% to every cache
+	go func() {
```
kolesnikovae, line 177: Perhaps it would be better to have this as a separate function.

pkg/storage/storage.go
```diff
@@ -163,6 +164,49 @@ func New(cfg *config.Config) (*Storage, error) {
...
+		ticker := time.NewTimer(interval * time.Second)
```
kolesnikovae, line 178: I think a `Ticker` is more appropriate here.

pkg/storage/storage.go
```diff
@@ -163,6 +164,49 @@ func New(cfg *config.Config) (*Storage, error) {
...
+		defer ticker.Stop()
+
+		for range ticker.C {
+			if s.closing {
```
kolesnikovae, line 182: I guess this will cause races.

pkg/storage/storage.go
alonlong: good idea

pkg/storage/storage.go
alonlong: already add the test for this PR

pkg/storage/storage.go
alonlong: why?

pkg/storage/storage.go
alonlong: can u help to describe the races?

pkg/storage/storage.go
kolesnikovae: * [The `Timer` type represents a single event](https://golang.org/src/time/sleep.go?s=1658:1710#L40)
 * [A `Ticker` holds a channel that delivers “ticks” of a clock at intervals](https://golang.org/src/time/tick.go?s=271:373#L1)

Timers are for when you want to do something once in the future - tickers are for when you want to do something repeatedly at regular intervals.

pkg/storage/storage.go
kolesnikovae: There is a race condition on access to `closing` field, you should acquire `closingMutex` and release it only after `Evict` calls. See `Close`: basically, I think that we could replace the mutex with an `RWMutex`.

pkg/storage/cache/cache.go
alonlong: right, thank u very much

alonlong: cache.lfu.Evict(int(float64(cache.lfu.Len()) * percent))

pkg/storage/storage.go
alonlong: yes, but i reset the timer in the loop

// reset the timer
ticker.Reset(interval * time.Second)

pkg/storage/storage.go
alonlong: for range ticker.C {
			if s.closing {
				return
			}

alonlong: there is no closingMutex here

alonlong: actually, i think for the field closing, the mutex is not necessary. the code is weird for using the mutex for closing

kolesnikovae: I'm afraid, without a mutex, there is no other way to avoid the race (without changing too many things): eventually the call will conflict with `Close`.

pkg/storage/storage.go
kolesnikovae: For sure that works. However, I think it would be better to use more idiomatic approach with `Ticker` :)

pkg/storage/storage.go
alonlong: it's a bool variable, i think it's not a problem

kolesnikovae: Regardless of the type, when the same address accessed for read-write concurrently, a race condition occurs. The funny part is that it would be safer to not check `closing` field at all in this particular function, because `Evict` and `Flush` (in `Close`) methods can be called in arbitrary order with no issues :)

pkg/storage/storage.go
petethepig: @kolesnikovae I talked with @alonlong about it. I think the idea is that if this procedure takes longer than the timer we want to wait for another X seconds (10 right now I think) before flushing.

So I think it's fine. Although, I wonder if we should just add a time.Sleep there. What do you both think about that? I think that would make code more concise, but maybe there're some downsides I'm not thinking of.

kolesnikovae: `Ticker` also has `Reset` method for that.

Sorry guys if led you astray. Please don't get me wrong, I was just trying to say that `Timer` and `Ticker` have different semantics, and for this particular case the last one is more appropriate IMO. Taking into account the mismatch of the variable name and the type it's a bit confusing: is it supposed to be a timer and the variable name should be changed, or is it a ticker indeed and the type should be changed? Both options are okay.

alonlong: @kolesnikovae Got it, I think we can change variable name to timer

Issue #188, opened 2021-05-16
Add code line for console log


alonlong:
now there is no code line information when printing logs, as it's useful for developers to fix issues, also when users feedback question about pyroscope, code line will be very useful~ please consider adding the code line for logs

petethepig:
This was done in #214

PR #214, opened 2021-05-25
fix the issue #188: add the file name and code line

alonlong:
the output format is like:
INFO[25-05-2021 23:05:56] cli/cli.go:415 starting HTTP server


pkg/exec/cli.go
```diff
@@ -28,7 +28,6 @@ import (
 
 // used in tests
 var disableMacOSChecks bool
-var disableLinuxChecks bool
```
petethepig, line 31: why delete this?

PR #215, opened 2021-05-25
handle error when starting the server

alonlong:
1. handle error when starting the server, remove the panic
2. make the logger always available, remove the nil check
3. optimize some logics for remote uploading profile

### DIFF
--- a/pkg/agent/upstream/remote/remote.go
+++ b/pkg/agent/upstream/remote/remote.go
@@ -3,6 +3,7 @@ package remote
 import (
 	"bytes"
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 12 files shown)


pkg/agent/upstream/remote/remote.go
```diff
@@ -38,62 +40,67 @@ type RemoteConfig struct {
...
 }
 
-func (u *Remote) Stop() {
-	wg := sync.WaitGroup{}
-	wg.Add(u.cfg.UpstreamThreads)
-	for i := 0; i < u.cfg.UpstreamThreads; i++ {
-		u.done <- &wg
+func (r *Remote) Stop() {
+	if r.done != nil {
+		close(r.done)
```
petethepig, line 82: I like this solution :+1:

Issue #119, opened 2021-03-11
Feature: add ability to delete apps


petethepig:
See https://github.com/pyroscope-io/pyroscope/issues/118#issuecomment-796519891 for context

AdrK:
There is still front end to be done, so reopening ;)

PR #221, opened 2021-05-27
Closes: Feature: add ability to delete apps #119

AdrK:


pkg/storage/cache/cache.go
```diff
@@ -76,6 +76,24 @@ func (cache *Cache) Flush() {
 	<-cache.cleanupDone
 }
 
+func (cache *Cache) Delete(key string) {
+	cache.Flush()
```
AdrK, line 80: This causes an `EvictionChannel` to be closed. How to remove the entries from cache in different way?

petethepig: `Flush` will clear the whole cache if I understand it correctly. I don't see a `Delete` function in [lfu-go](https://github.com/dgrijalva/lfu-go/blob/master/lfu.go) which is a bummer. It should be easy to add though.

We can add it to our fork of lfu-go https://github.com/pyroscope-io/lfu-go The function should just do something like this:
```
c.lock.Lock()
defer c.lock.Unlock()
delete(c.values, key)
```

To make the project use our fork we'll need to make changes to go.mod the same way we use this revive fork:
https://github.com/pyroscope-io/pyroscope/blob/main/go.mod#L48

AdrK: allright, I have it done locally, but I need write access to lfu-go to commit it. I have updated cache.go, but I expect test to fail until the fork will be updated.

pkg/storage/storage.go
```diff
@@ -320,6 +320,58 @@ func (s *Storage) Get(gi *GetInput) (*GetOutput, error) {
...
+	StartTime time.Time
+	EndTime   time.Time
+	Key       *Key
+}
+
+func (s *Storage) Delete(di *DeleteInput) error {
+	s.closingMutex.Lock()
+	defer s.closingMutex.Unlock()
+
+	if s.closing {
```
kolesnikovae, line 333: It is not related to the change itself but I'm wondering if we really need to keep `closingMutex` acquired during the whole call. Do other functions actually compete? If so, should we rename it accordingly (just `m`/`mu`)?

kolesnikovae (commented):
LGTM.

pkg/storage/storage.go
petethepig: Yeah, this is a very good point. The idea was to make sure we don't try to read / write after `Close` is called.

But the way it is currently implemented, we basically have a lock on the whole db... I changed it to a `RWMutex`. 

Good news is that this means we'll get better performance. Bad news is that this might now expose some bugs that were hidden before because we had a lock on the whole db. Hopefully unit tests + running pyroscope server locally will expose those.

kolesnikovae: Oh, that's my biggest concern. We can mitigate the risk by using `RLock` in `Get` only.

Also, I think `Close` should acquire the mutex for the whole call.

petethepig: If Close acquires mutex for the whole call, then Get and Put will just block instead of returning errors... I think I'd rather them returning errors..

> We can mitigate the risk by using RLock in Get only.

I'll do some tests today. One thing I can tell already is that if I [benchmark it](https://github.com/pyroscope-io/pyroscope/tree/main/benchmark) nothing breaks. That's good, but is not a 100% guarantee.

petethepig (approved):

pkg/storage/storage.go
AdrK: I was wondering also about a situation when we remove from cache, but not yet from disk. If in the middle someone would call a read, cache would read from disk, reloading an item that was previously deleted. And at the end, we will in fact remove from disk, but we end up with not removed item in cache.

PR #222, opened 2021-05-28
Hotfix/segment cache panic

alonlong:
1. add the error handler for different cache
2. solve the panic when root node is not ready for segment

### DIFF
--- a/pkg/storage/cache/cache.go
+++ b/pkg/storage/cache/cache.go
@@ -1,6 +1,9 @@
 package cache
 
 import (
+	"errors"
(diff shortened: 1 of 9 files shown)


petethepig:
There's one test that's failing. Could you fix it please? https://github.com/pyroscope-io/pyroscope/pull/222/checks?check_run_id=2694916899

Other than that it looks good I think.

alonlong:
done, please help to check

petethepig (approved):
LGTM

PR #223, opened 2021-05-29
Part of: Lint Go Code #49

AdrK:


petethepig:
I like that you separated commits by package. We should not lint `pkg/agent/pprof/` package (and we should actually add it to some exclude list) because it's copied from golang source code and if we don't it's gonna get messy if we ever want to pull the changes from upstream.

pkg/storage/dimension/dimension.go
```diff
@@ -6,21 +6,21 @@ import (
 	"sync"
 )
 
-type key []byte
```
petethepig, line 9: why make this one public?

pkg/storage/tree/tree.go
```diff
@@ -60,15 +60,15 @@ func New() *Tree {
 	}
 }
 
-func (dstTrie *Tree) Merge(srcTrieI merge.Merger) {
```
petethepig, line 63: I know the linter says these names should be consistent, but I kinda feel in this particular case it's useful to have this src / dst distinction.

What do you guys thing about this? @AdrK @kolesnikovae

petethepig (changes requested):
Looks good. I added a couple of comments. We should definitely add agent/pprof to the list of exceptions

pkg/storage/dimension/dimension.go
AdrK: `Intersection()` is a public method which returns an array of items of `Key` type which itself is not public.

pkg/storage/tree/tree.go
AdrK: I would go with consistent approach. Reader of this file should quickly learn what `t` stands for. Also, naming as `dstTree` and `srcTree` might suggests that both those vars came as an arguments and this method can work on two arbitrary trees, while it can only merge a tree provided by argument with itself and `t` denotes it ( like `self` in Python ).

petethepig: Yeah, good points, I think I'm in favor of t now

petethepig (approved):
LGTM

Issue #100, opened 2021-02-26
Make a subcommand to upload profiles (pyroscope upload)


petethepig:
As I was working on #83 I figured it would be cool to be able to do something like:
```
printf "foo;bar 100\nfoo;baz 200\n" | pyroscope upload -from now-1m -until now
```

I think this would also enable faster adoption of other profilers.

---

It could also support a mode in which it chunks (10 second each) a stream of profiling data coming in that same kind of format.

---

There's already support for different profiling formats:
* https://github.com/pyroscope-io/pyroscope/blob/main/pkg/convert/parser.go#L20

---

Here's some documentation on agent / server communication:
![diagram](https://user-images.githubusercontent.com/662636/115311366-69bfbc80-a124-11eb-9891-8f94f9df0eac.png)

^ in this example `pyroscope exec` communicates with `pyroscope server`. For this issue we basically need to implement a way to upload profiling data via `pyroscope upload` instead of `pyroscope exec`

Rperry2174:
closing in favor of #652 #601

PR #235, opened 2021-06-04
Upload functionality for pyroscope

yuridevx:
Solves #100 
To be reviewed @petethepig

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,8 @@ pkged.go
 .vscode
 vendor/
 benchmark/git-info.env
+
+storage/
+playground/
+pyroscope
+LOCAL_TODO.md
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -15,3 +15,8 @@ pkged.go
 .vscode
 vendor/
 benchmark/git-info.env
+
+storage/
+playground/
+pyroscope
+LOCAL_TODO.md
```
petethepig, line 22: Can we remove these?

`storage/` and `pyroscope` are taken care of if you use `make` to build and start `pyroscope`:
* `make build`: https://github.com/pyroscope-io/pyroscope/blob/main/Makefile#L35
* `make server`: https://github.com/pyroscope-io/pyroscope/blob/main/Makefile#L59

^ these are also documented in [Developed Guide](https://pyroscope.io/docs/developer-guide)

RE `LOCAL_TODO.md` & `playground` — I'd say let's keep it if these were commonly used conventions, but afaik they're not. So I would suggest you put these in your local .gitignore, e.g. [~/.gitignore](https://git-scm.com/docs/gitignore)

petethepig: Also please set up your editor to add a new line in the end of files, it's a good practice. See [this stackoverflow answer for context](https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file).

PR #238, opened 2021-06-12
fix #226: support the readiness and liveness for kubernetes

### LINKED ISSUE #226: Create a liveness check endpoint
k8s has this feature for services to advertise that they are able to process requests. [Link to documentation](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/)

We should implement this for pyroscope. I think we can just add an endpoint somewhere here https://github.com/pyroscope-io/pyroscope/blob/main/pkg/server/controller.go#L63-L66

alonlong:


kolesnikovae (commented):
LGTM.

I guess https://github.com/pyroscope-io/helm-chart should be updated accordingly. And, perhaps, we don't need a separate deployment manifest - consider use of the helm chart instead.

alonlong:
> LGTM.
> 
> I guess https://github.com/pyroscope-io/helm-chart should be updated accordingly. And, perhaps, we don't need a separate deployment manifest - consider use of the helm chart instead.

ok, i will commit a PR for helm chart

petethepig (approved):
LGTM

beellz:
There should be two different endpoints for liveness and readiness.
The endpoint readiness is available (healthz) Wouldn't work for liveness.
If the readiness fails, it would not restart the pod, but it will restart the pod when the liveness fails.
It will be better to have two different endpoints for it.

kolesnikovae:
> There should be two different endpoints for liveness and readiness.
> The endpoint readiness is available (healthz) Wouldn't work for liveness.
> If the readiness fails, it would not restart the pod, but it will restart the pod when the liveness fails.
> It will be better to have two different endpoints for it.

That makes a lot of sense. But right now we don't have any particular reason (apart from crash) for a health controller to restart pyroscope server. The same for readiness: the only condition we should meet before starting accepting incoming requests is that the local storage is accessible - and http server won't start otherwise. Therefore the same endpoint serves both purposes: readiness probe and liveness probe.

Once we introduce more dependencies (e.g., connection to a remote storage backend), we should definitely split the endpoints and logic behind them.

@beellz @alonlong what do you think?

beellz:
Will work for now. 
Also added readiness and liveness in helm chart

PR #243, opened 2021-06-16
Graceful shutdown

### LINKED ISSUE #232: Relay POSIX signals in exec
As of now `exec` unconditionally kills spawned process on receiving `SIGINT` or `SIGTERM`:

https://github.com/pyroscope-io/pyroscope/blob/cfcc0a2ead608a15abfb5699dd0f49886b5afbfe/pkg/exec/cli.go#L179-L205

Instead, we should respect POSIX signals semantics wherever it is possible. I would rather relay those in the same way it is implemented in [go tool](https://github.com/golang/go/blob/dc8f87b7493e173d65d3587389cc41468ba16dc0/src/cmd/go/internal/tool/tool.go#L89-L101) with `signal.Notify(c)`.

kolesnikovae:
This pull request is aimed to improve pyroscope shutdown and initialisation.
* `controller` (http server) is drained at the very beginning of the shutdown to prevent storage reads and writes. Diagnostic endpoints (healthz, metrics, pprof) are handled till the storage closes, thus allowing clean exit.
* removed `IsClosing` check from `storage`: it is guaranteed that no calls to storage will be made after `Close`.
* `atexit` package was removed.
* Closes #232.


pkg/cli/cli.go
```diff
@@ -145,7 +141,15 @@ func generateRootCmd(cfg *config.Config) *ffcli.Command {
 			return nil
 		}
 
-		return exec.Cli(context.Background(), &cfg.Exec, args)
+		err := exec.Cli(&cfg.Exec, args)
+		// Normally, if the program ran, the call should return ExitError and
+		// the exit code must be preserved. Otherwise, the error originates from
+		// pyroscope and will be printed.
+		if e, ok := err.(*goexec.ExitError); ok {
```
petethepig, line 148: :+1:

petethepig (approved):
It is so much better now, love it!

pkg/cli/server.go
```diff
@@ -0,0 +1,132 @@
...
+	svc.selfProfiling = agent.NewSession(selfProfilingConfig, svc.logger)
+	if !c.AnalyticsOptOut {
+		svc.analyticsService = analytics.NewService(c, svc.storage, svc.controller)
+	}
+
+	return &svc, nil
+}
+
+func (svc *serverService) Start() error {
+	g, ctx := errgroup.WithContext(context.Background())
```
petethepig, line 73: 👍🏼

PR #244, opened 2021-06-18
Port can be defined by env var

AdrK:


pkg/cli/cli.go
```diff
@@ -51,6 +51,13 @@ func generateRootCmd(cfg *config.Config) *ffcli.Command {
 	dbmanagerSortedFlags := PopulateFlagSet(&cfg.DbManager, dbmanagerFlagSet)
 	rootSortedFlags := PopulateFlagSet(cfg, rootFlagSet)
 
+	port := os.Getenv("PORT")
+	if port != "" {
+		cfg.Server.APIBindAddr = fmt.Sprintf(":%s", port)
+		cfg.Agent.ServerAddress = fmt.Sprintf("http://localhost:%s", port)
+		cfg.Exec.ServerAddress = fmt.Sprintf("http://localhost:%s", port)
+	}
+
```
kolesnikovae, line 60: Some points for your consideration:
- `ffcli` is used to [bind environment variables to configuration](https://github.com/pyroscope-io/pyroscope/blob/4a429df29b6bff48e0f9f13a2d34f6e91535941f/pkg/cli/cli.go#L56) (and we have some plans to migrate to [cobra](https://github.com/spf13/cobra) - see https://github.com/pyroscope-io/pyroscope/issues/240).
- This affects how CLI usage message (and documentation) is rendered: https://github.com/pyroscope-io/pyroscope/pull/236.
- To keep things in order, it's better to preserve 1:1 mapping between CLI args, env vars, and configuration options (wherever it is possible).

To be very honest, I'm not 100% sure this option is required indeed.

AdrK: Thanks for the useful remarks. I wasn't aware of `ffcli`. I guess we need a decision if this is needed at all here. There might be cases when we need to have port configurable.

petethepig:
Let's not merge this, you can use `PYROSCOPE_API_BIND_ADDR` to set the address on the server side, and `PYROSCOPE_ SERVER_ADDRESS` to set it on clients.

PR #250, opened 2021-06-20
add the agent(many applications) to test the cache

alonlong:


examples/apps/Dockerfile.centrifugo
```diff
@@ -0,0 +1,12 @@
...
+RUN git clone https://github.com/centrifugal/centrifugo.git $GOPATH/src/centrifugo
+WORKDIR $GOPATH/src/centrifugo
+
+COPY main/centrifugo main.go
+RUN GO111MODULE=on go mod tidy
+RUN GO111MODULE=on go build
+RUN chmod +x centrifugo
+RUN ./centrifugo genconfig
+
+ENTRYPOINT ["./centrifugo"]
```
petethepig, line 12: Please make sure your text editor adds new lines at the end of each file. More info here: https://pyroscope.io/docs/style-guide#other-considerations

alonlong: ok

alonlong: done

examples/apps/Dockerfile.centrifugo
```diff
@@ -0,0 +1,13 @@
...
+
```
petethepig, line 13: Well, now there's 2 empty lines at the end of each file.

examples/apps/Dockerfile.centrifugo
```diff
@@ -0,0 +1,13 @@
+FROM golang:1.16
+
+RUN git clone https://github.com/centrifugal/centrifugo.git $GOPATH/src/centrifugo
+WORKDIR $GOPATH/src/centrifugo
+
+COPY main/centrifugo main.go
```
petethepig, line 6: Can we do something like this instead:
```
RUN inject-profiler-code.sh main.go
```

Reasons I think this would be better:
* when consul, centrifugo, etc make changes to their main.go files this still works
* from copyright perspective this seems like a cleaner solution, because this way we don't need to copy other people's code into our repo.


I don't want to copy

PR #252, opened 2021-06-21
Fixed wrong time calc in dotnetspy.

AdrK:


pkg/agent/dotnetspy/session.go
```diff
@@ -89,7 +89,7 @@ func (s *session) start() error {
 				for k, v := range p.Samples() {
 					s.ch <- line{
 						name: []byte(k),
-						val:  int(v.Milliseconds()) / 10,
+						val:  int(v.Milliseconds()),
```
AdrK, line 92: The `examples/dotnet/web` was not working and I found that this is due to wrong calculations here.
I wonder why we were doing `/10` ? For `v.Milliseconds() == 1`, this was resolving to `0` and no samples were reported in stacktrace.

kolesnikovae: It's a leftover. Thank you for fixing this!

kolesnikovae:
LGTM.

I've also updated `dotnetdiag` package - could you please bump its version to [v1.2.1](https://github.com/pyroscope-io/dotnetdiag/tree/v1.2.1)?

petethepig:
@kolesnikovae So that must mean we should be seeing more data now when we profile dotnet apps, right?

petethepig (approved):

kolesnikovae:
@petethepig, my bad, this denominator is not a mistake - https://github.com/pyroscope-io/pyroscope/pull/255:
```
// dotnet profiler reports total time v per call stack k.
// Meanwhile, pyroscope agent expects number of samples is
// reported. Every sample is a time fraction of second
// according to sample rate: 1000ms/100 = 10ms by default.
// To represent reported time v as a number of samples,
// we divide it by sample duration.
//
// Taking into account that under the hood dotnet spy uses
// Microsoft-DotNETCore-SampleProfiler, which captures a
// snapshot of each thread's managed callstack every 10 ms,
// we cannot manage sample rate from outside.
```

Thus, we cannot capture samples of duration less than `1s/sample-rate` (10ms by default). For dotnet that means we may be missing up to 0.1% of data for a 10s session.

@AdrK, the reason why `web` example didn't show anything is that we filter out "unmanaged" code (native): unless there are no incoming requests, collected profiles will be empty (though, sometimes on slow machines we can capture time spent in managed code that invokes native methods).

PR #255, opened 2021-06-22
Document dotnet magic denominator

kolesnikovae:
Due to how dotnet profiler is different comparing to other spies, there is an obscure transforming of reported CPU time, which may be misleading. This PR is aimed to shed light on why it takes place.


petethepig (approved):
Thanks for the explanation!

PR #256, opened 2021-06-23
Fixes max memory available detection on linux

petethepig:
I introduced a pretty big bug in the latest version where it would detect amount of RAM available incorrectly. E.g here we can see pyroscope thought it had 8EB of RAM. So that threw off cache evictions procedure and led to OOM events.

This PR fixes that issue.

This is before / after this fix on a linux machine:

Screen Shot 2021-06-22 at 6 31 54 PM

### DIFF
--- a/pkg/storage/mem.go
+++ b/pkg/storage/mem.go
@@ -1,44 +1,13 @@
+// +build !linux
+
 package storage
 
 import (
-	"io/ioutil"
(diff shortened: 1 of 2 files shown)


petethepig:
@kolesnikovae I'm merging it now because it's kind of a big bug, but please take a look when you have a chance.

PR #259, opened 2021-06-23
Adds new build and config endpoints

petethepig:
This is useful for debugging / ops people.

### DIFF
--- a/pkg/server/build.go
+++ b/pkg/server/build.go
@@ -1,0 +1,59 @@
+package server
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"runtime"
+
+	"github.com/pyroscope-io/pyroscope/pkg/build"
+)
+
+type buildInfoJSON struct {
+	GOOS              string `json:"goos"`
+	GOARCH            string `json:"goarch"`
+	Version           string `json:"version"`
+	ID                string `json:"id"`
+	Time              string `json:"time"`
+	GitSHA            string `json:"gitSHA"`
+	GitDirty          int    `json:"gitDirty"`
+	UseEmbeddedAssets bool   `json:"useEmbeddedAssets"`
+}
+
+func (ctrl *Controller) buildHandler(w http.ResponseWriter, r *http.Request) {
+	buildInfoStr, err := buildInfoJSONString(true)
+	if err != nil {
+		renderServerError(w, fmt.Sprintf("could not marshal buildInfoObj json: %q", err))
+		return
+	}
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(200)
+	w.Write([]byte(buildInfoStr))
+}
+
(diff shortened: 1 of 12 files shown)


pkg/server/controller.go
```diff
@@ -63,6 +62,8 @@ func New(c *config.Server, s *storage.Storage) (*Controller, error) {
 		{"/render", ctrl.renderHandler},
 		{"/labels", ctrl.labelsHandler},
 		{"/label-values", ctrl.labelValuesHandler},
+		{"/config", ctrl.configHandler},
+		{"/build", ctrl.buildHandler},
 	}
 
 	addRoutes(mux, routes, ctrl.drainMiddleware)
```
kolesnikovae, line 69: May be move it to other diagnostic routes that are not using drain middleware?

pkg/server/config.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/pyroscope-io/pyroscope/pkg/config"
+)
+
+func (ctrl *Controller) configHandler(w http.ResponseWriter, r *http.Request) {
+	configStr, err := configJSONString(ctrl.config)
```
kolesnikovae, line 12: I think it can be shortened to: 
```go
b, err := json.MarshalIndent(ctrl.config, "", "\t")
```

pkg/server/build.go
```diff
@@ -0,0 +1,59 @@
...
+		UseEmbeddedAssets: build.UseEmbeddedAssets,
+	}
+	var b []byte
+	var err error
+
+	if pretty {
+		b, err = json.MarshalIndent(buildInfoObj, "", "  ")
+	} else {
+		b, err = json.Marshal(buildInfoObj)
+	}
```
kolesnikovae, line 52: I would move this to `build` package and add necessary methods, like:
```go
func (b Info) String() string
func (b Info) JSON() []byte
func (b Info) PrettyJSON() []byte
```
Taking into account that there is no apparent reason for json marshaler to fail, an error may be omitted, or cause panic (+ optional `Must` prefix for method names).

pkg/server/build.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func (ctrl *Controller) buildHandler(w http.ResponseWriter, r *http.Request) {
+	buildInfoStr, err := buildInfoJSONString(true)
+	if err != nil {
+		renderServerError(w, fmt.Sprintf("could not marshal buildInfoObj json: %q", err))
+		return
+	}
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(200)
```
kolesnikovae, line 30: Here and below: `http.StatusOK` is set by default, this line may be safely omitted.

kolesnikovae (approved):
Looks great! I've left a couple of notes but those are just nits.

PR #266, opened 2021-06-28
improve tooltip rendering

Loggy:
- moved some values computing into RenderCanvas
- rewrite tooltip positioning from absolute top-left values to transform CSS property.

There's still a render lag. It's not so long but still visible.  As the next step, I can put tooltip rendering inside canvas

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "react-datepicker": "^3.4.1",
     "react-dom": "^16.13.1",
     "react-flot": "^1.3.0",
+    "react-fps-stats": "^0.1.3",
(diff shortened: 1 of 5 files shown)


petethepig:
@Loggy I think I made it even faster by skipping react virtual dom stuff. What do you think about this solution?

Loggy:
Works really better) But it reveals that state is pretty huge and needs optimization as we can't move things out of state every time to improve rendering performance))

But in this case, I think it fits the goal)

petethepig:
Alright, let's merge this for now and next time we touch that part we can split it.

PR #272, opened 2021-07-05
Oauth flow for Google, GitHub and GitLab

Skemba:
This version:
- Needs (google, github, gitlab) apps to be turned for verification to be rolled into production
- Doesn't include any check for who is allowed to login (it can be futher limited by email address for google, organization or team for github/gitlab)
- Google parameter for `allowed_domain` was not included as it doesn't provide much more security but can be returned back to improve UX for GSuite users
- Includes HTML files that can be replaced with a better JS-based frontend code

Needs review:
- What to do in case of errors during auth process? -> right now error is logged and user is redirected to /login again
- Do we want any other additional feature added in this version
- Which routes need auth?


pkg/server/controller.go
```diff
@@ -129,6 +207,65 @@ func (ctrl *Controller) drainMiddleware(next http.HandlerFunc) http.HandlerFunc
...
+		})
+
+		if err != nil {
+			ctrl.httpServer.ErrorLog.Printf("Error parsing jwt token: %v", err)
+			http.Redirect(w, r, "/login", http.StatusTemporaryRedirect)
+			return
+		}
+
+		claims, ok := token.Claims.(jwt.MapClaims)
+		if !ok || !token.Valid {
```
kolesnikovae, line 240: `jwt.Parse` carries out all the necessary verifications. I guess this code won't be reached.

pkg/server/controller.go
```diff
@@ -129,6 +207,65 @@ func (ctrl *Controller) drainMiddleware(next http.HandlerFunc) http.HandlerFunc
...
+			ctrl.httpServer.ErrorLog.Printf("Token not valid")
+			http.Redirect(w, r, "/login", http.StatusTemporaryRedirect)
+			return
+		}
+
+		if exp, ok := claims["exp"].(float64); ok && int64(exp) < time.Now().Unix() {
```
kolesnikovae, line 246: Why do we need to do this on our own? `token.Claims.Valid()` which is called in `jwt.Parse` does validate expiration as well.

pkg/server/controller.go
```diff
@@ -129,6 +207,65 @@ func (ctrl *Controller) drainMiddleware(next http.HandlerFunc) http.HandlerFunc
...
+			ctrl.httpServer.ErrorLog.Printf("Token no longer valid")
+
+			jwtCookie := &http.Cookie{
```
kolesnikovae, line 249: Should we add a helper like `invalidateCookie(w http.ResponseWriter, name string)`? Then, call it on any error along with redirect.

pkg/server/controller.go
```diff
@@ -145,6 +282,264 @@ type indexPage struct {
 	BaseURL       string
 }
 
+func (ctrl *Controller) decodeGoogleCallbackResponse(resp *http.Response) (name string, err error) {
```
kolesnikovae, line 285: Just a NB: I myself like naked return, but nowadays almost nobody uses this (except for deferred recovery use case). Perhaps we should do so as well.

pkg/server/controller.go
```diff
@@ -145,6 +282,264 @@ type indexPage struct {
...
+	err = json.NewDecoder(resp.Body).Decode(&userProfile)
+	if err != nil {
+		return
+	}
+
+	name = userProfile.Username
+	return
+}
+
+func (ctrl *Controller) callbacRedirectkHandler(getAccountInfoURL string, oauthConf *oauth2.Config, decodeResponse func(*http.Response) (string, error)) http.HandlerFunc {
```
kolesnikovae, line 339: 1. Typo: `callbackRedirectHandler`.
2. It also may make sense to define a type for `func(*http.Response) (string, error)` just to keep signature readable.
3. I guess it's better to move handlers to a separate file.

pkg/server/controller.go
```diff
@@ -145,6 +282,264 @@ type indexPage struct {
...
+	return func(w http.ResponseWriter, r *http.Request) {
+		cookie, err := r.Cookie(stateCookieName)
+		if err != nil {
+			ctrl.httpServer.ErrorLog.Printf("There seems to be problem with state cookie: %v", err)
```
kolesnikovae, line 343: The only returned error by `Cookie` is `ErrNoCookie`, this annotation may not be necessary.

pkg/server/controller.go
```diff
@@ -145,6 +282,264 @@ type indexPage struct {
...
+			http.Redirect(w, r, "/login", http.StatusTemporaryRedirect)
+			return
+		}
+
+		if cookie == nil {
```
kolesnikovae, line 348: Not needed.

pkg/server/controller.go
```diff
@@ -145,6 +282,264 @@ type indexPage struct {
...
+			ctrl.httpServer.ErrorLog.Printf("Missing state cookie")
+			http.Redirect(w, r, "/login", http.StatusTemporaryRedirect)
+			return
+		}
+
+		cookieState := cookie.Value
+		requestState := r.FormValue("state")
+
+		if requestState != cookieState {
+			ctrl.httpServer.ErrorLog.Printf("invalid oauth state, expected %v got %v", cookieState, requestState)
```
kolesnikovae, line 358: Perhaps we should avoid printing out cookies - it is barely useful. We should be extremely careful logging anything related to security matters to avoid disclosure.

pkg/server/controller.go
```diff
@@ -145,6 +282,264 @@ type indexPage struct {
...
+		}
+
+		code := r.FormValue("code")
+		if code == "" {
+			ctrl.httpServer.ErrorLog.Printf("Code not found")
+			http.Redirect(w, r, "/login", http.StatusTemporaryRedirect)
+			return
+		}
+
+		token, err := oauthConf.Exchange(oauth2.NoContext, code)
```
kolesnikovae, line 370: Use `r.Context()` instead of `NoContext` (all places).

pkg/server/controller.go
```diff
@@ -145,6 +282,264 @@ type indexPage struct {
...
+		defer resp.Body.Close()
+
+		name, err := decodeResponse(resp)
+		if err != nil {
+			ctrl.httpServer.ErrorLog.Printf("Decoding response body failed: %v", err)
+			http.Redirect(w, r, "/login", http.StatusTemporaryRedirect)
+			return
+		}
+
+		claims := jwt.MapClaims{
```
kolesnikovae, line 393: I think a helper method (e.g, `newJwtToken() *jwt.Token`) is more appropriate here.

pkg/server/controller.go
```diff
@@ -145,6 +282,264 @@ type indexPage struct {
...
+	b := make([]byte, length)
+	if _, err := rand.Read(b); err != nil {
+		return "", err
+	}
+	return hex.EncodeToString(b), nil
+}
+
+func (ctrl *Controller) oauthLoginHandler(oauthConf *oauth2.Config) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		URL, err := url.Parse(oauthConf.Endpoint.AuthURL)
```
kolesnikovae, line 453: We should perform all the validation steps before creating `oauth2.Config` instance. Consider use of shallow copy (https://github.com/golang/go/issues/41733).

 `URL` would be `authUrl` (`url` collides to the package name).

pkg/server/controller.go
```diff
@@ -145,6 +282,264 @@ type indexPage struct {
...
+		state, err := generateStateToken(16)
+		if err != nil {
+			ctrl.httpServer.ErrorLog.Printf("Generate token error: %v", err)
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		parameters.Add("state", state)
+		URL.RawQuery = parameters.Encode()
+		url := URL.String()
```
kolesnikovae, line 476: Do we need this variable?

kolesnikovae (commented):
Awesome! I left a couple of comments, most of them are just nits – please take a look.

pkg/config/config.go
```diff
@@ -80,6 +80,34 @@ type Server struct {
 	CacheDictionarySize int               `deprecated:"true"`
 	CacheSegmentSize    int               `deprecated:"true"`
 	CacheTreeSize       int               `deprecated:"true"`
+
+	GoogleEnabled      bool   `def:"false" desc:"enables Google Oauth"`
+	GoogleClientID     string `def:"<yourClientID>" desc:"client ID generated for Google API"`
+	GoogleClientSecret string `def:"<yourClientSecret>" desc:"client secret generated for Google API"`
+	GoogleRedirectURL  string `def:"<pyroscope.mycompany.org/google/callback>" desc:"url that google will redirect to after logging in. Has to be in form <pathToPyroscopeServer/google/callback>"`
+	GoogleScopes       string `def:"https://www.googleapis.com/auth/userinfo.email" desc:"scopes for Google API"`
```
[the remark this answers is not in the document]
Skemba: At this point probably no, I left it in just in case someone could think of a use-case, would be easy to remove if you don't think there's anything useful we could offer?

pkg/config/config.go
```diff
@@ -80,6 +80,34 @@ type Server struct {
...
+
+	GithubEnabled      bool   `def:"false" desc:"enables Github Oauth"`
+	GithubClientID     string `def:"<yourClientID>" desc:"client ID generated for Github API"`
+	GithubClientSecret string `def:"<yourClientSecret>" desc:"client secret generated for Github API"`
+	GithubRedirectURL  string `def:"<pyroscope.mycompany.org/github/callback>" desc:"url that Github will redirect to after logging in. Has to be in form <pathToPyroscopeServer/github/callback>"`
+	GithubScopes       string `def:"read:user user:email" desc:"scopes for Github API"`
+	GithubAuthURL      string `def:"https://github.com/login/oauth/authorize" desc:"auth url for Github API"`
+	GithubTokenURL     string `def:"https://github.com/login/oauth/access_token" desc:"token url for Github API"`
+
+	JWTSecret                string `def:"qC8=%q~'z'o'CBi" desc:"secret used to secure your JWT tokens"`
```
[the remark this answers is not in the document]
Skemba: I agree. Left it so that user would have less config to fill up, but better to make them be explicit about security-related things.

pkg/config/config.go
```diff
@@ -80,6 +80,34 @@ type Server struct {
...
+	LoginMaximumLifetimeDays int    `def:"0" desc:"amount of days after which user will be logged out. 0 means non-expiring."`
```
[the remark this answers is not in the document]
Skemba: I've mostly followed what grafana have done and I see their point of view as most people wouldn't want to be logged out implicitly after shorter than a day duration?

pkg/server/controller.go
```diff
@@ -71,7 +85,7 @@ func (ctrl *Controller) mux() http.Handler {
 		{"/label-values", ctrl.labelValuesHandler},
 	}
 
-	addRoutes(mux, routes, ctrl.drainMiddleware)
+	addRoutes(mux, routes, ctrl.drainMiddleware, ctrl.authMiddleware)
```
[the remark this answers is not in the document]
Skemba: Is that the only one that shouldn't be behind auth?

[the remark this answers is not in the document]
kolesnikovae: AFAIK, only this one.

pkg/server/controller.go
```diff
@@ -129,6 +207,65 @@ func (ctrl *Controller) drainMiddleware(next http.HandlerFunc) http.HandlerFunc
...
+		if err != nil {
+			ctrl.httpServer.ErrorLog.Printf("There seems to be problem with jwt token cookie: %v", err)
+			http.Redirect(w, r, "/login", http.StatusTemporaryRedirect)
+			return
+		}
+
+		if jwtCookie == nil {
+			ctrl.httpServer.ErrorLog.Printf("Missing jwt cookie")
+			http.Redirect(w, r, "/login", http.StatusTemporaryRedirect)
+			return
```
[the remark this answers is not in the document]
Skemba: that can be done, but I'm wondering is this even the error handling we want? Do we maybe want to show a different `Unauthorized` page to user instead of `/login`?

pkg/config/config.go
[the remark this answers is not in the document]
kolesnikovae: Aha, I see. Let's keep it as is.

pkg/server/controller.go
```diff
@@ -129,6 +207,65 @@ func (ctrl *Controller) drainMiddleware(next http.HandlerFunc) http.HandlerFunc
...
+		}
+
+		jwtToken := jwtCookie.Value
```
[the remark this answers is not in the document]
Skemba: Definitely not, left it accidentally after printing some things out :)

pkg/server/controller.go
Skemba: yes! totally missed that.

pkg/server/controller.go
Skemba: nope, left it on accident as well :)

pkg/server/controller.go
```diff
@@ -129,6 +207,65 @@ func (ctrl *Controller) drainMiddleware(next http.HandlerFunc) http.HandlerFunc
 	}
 }
 
+func (ctrl *Controller) authMiddleware(next http.HandlerFunc) http.HandlerFunc {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		jwtCookie, err := r.Cookie(jwtCookieName)
+		if err != nil {
+			ctrl.httpServer.ErrorLog.Printf("There seems to be problem with jwt token cookie: %v", err)
```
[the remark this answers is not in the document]
Skemba: ErrorLog is already created from `logrus` logger, do you mean keep it like that and instantiate separate logger for ErrorLogs or instantiate ErrorLog the way you described by passing it explicitly or something third?

pkg/server/controller.go
```diff
@@ -129,6 +207,65 @@ func (ctrl *Controller) drainMiddleware(next http.HandlerFunc) http.HandlerFunc
...
+
+		if jwtCookie == nil {
+			ctrl.httpServer.ErrorLog.Printf("Missing jwt cookie")
+			http.Redirect(w, r, "/login", http.StatusTemporaryRedirect)
+			return
+		}
+
+		jwtToken := jwtCookie.Value
+		token, err := jwt.Parse(jwtToken, func(token *jwt.Token) (interface{}, error) {
+			if _, ok := token.Method.(*jwt.SigningMethodHMAC); !ok {
```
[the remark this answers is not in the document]
Skemba: I'd add that one but also keep the current check as well. That way we check both type of the object and string returned by Alg(). What do you think?

pkg/server/controller.go
Skemba: you're right, I did this with a custom field first to test something and then just went along and wrote for exp for some reason as well. will remove it together with the above.

pkg/server/controller.go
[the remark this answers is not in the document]
kolesnikovae: Yeah, I mean we should probably have a separate logger, and `ErrorLog` should be created with `WriterLevel(ErrorLevel)` I think.

Apart from unification, I see two main reasons for that:
1. Wherever possible, log messages should be structured, i.e. have fields.
2. Diagnostic messages and audit (any login attempts should be recorded) – I suppose we should use appropriate log levels for that: e.g, debug and info, correspondingly.

pkg/server/controller.go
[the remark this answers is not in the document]
kolesnikovae: A very good question.

I'd rather add a flash message saying "not authorised" for such a case. There is one more point we should always keep in mind dealing with auth: user should not know the exact reason of the denial – in most of cases such a message with detailed explanation of what went wrong implies disclosure.

pkg/server/controller.go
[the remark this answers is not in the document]
kolesnikovae: Well, `Alg()` allows us to unambiguously distinguish signing methods. From this perspective a type check is redundant. I'd avoid that but it's up to you ;)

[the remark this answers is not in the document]
Skemba: I was just thinking if that string could somehow be tampered with, but just checked the jwt.Parse code and you're right it should be enough.

pkg/server/controller.go
Skemba: I'm not sure I understand what should I use shallow copy for? Also if we were to call url.Parse before building oauth2.Config we would have to either pass some sort of identifier and the create it in a switch or remove url parsing to above function which would triple the amount of code. URL parsing should rarely fail so it won't influence performance that much that we created instance before if it fails.

Or did I misunderstand something?

pkg/server/controller.go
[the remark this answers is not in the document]
Skemba: I have so far changed it to use `logrus` as mentioned but haven't added any diagnostic/audit messages. We should agree first on what these should be and then I can create a new PR for this?

pkg/server/controller.go
kolesnikovae: I agree. My point is that when we create `*oauth2.Config` we should validate user input (fail fast).

BTW, am I right that scopes are passed as a space delimited string? I'm wondering if it works - CLI may not interpret that in the right way.

Next are just my subjective thought :) Given that we already have parsed URL at validation, should we do that again for every request? For `oauthLoginHandler` we use `*oauth2.Config` as a data container - instead we could introduce our own type to bear prepared data. But it's up to you ;)

pkg/server/controller.go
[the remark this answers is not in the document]
kolesnikovae: Great! Let's back to audit later.

pkg/server/handler.go
```diff
@@ -0,0 +1,367 @@
...
+
+		code := r.FormValue("code")
+		if code == "" {
+			ctrl.logAndRedirect(w, r, "code not found", true)
+			return
+		}
+
+		token, err := oauthConf.Exchange(r.Context(), code)
+		if err != nil {
+			ctrl.logAndRedirect(w, r, "Exchanging auth code for token failed with "+err.Error(), true)
```
kolesnikovae, line 240: Well, we either can use `fmt.Sprintf`, or slightly modify the signature:
```
func (ctrl *Controller) logAndRedirect(w http.ResponseWriter, r *http.Request, format string, args ...interface{})
```
I think it's okay to invalidate state cookies on error regardless of anything (even if those are missing).

pkg/config/config.go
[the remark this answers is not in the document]
petethepig: Yeah, I'd say let's remove them. Less flags is always better.

pkg/config/config.go
```diff
@@ -80,6 +80,34 @@ type Server struct {
 	CacheDictionarySize int               `deprecated:"true"`
 	CacheSegmentSize    int               `deprecated:"true"`
 	CacheTreeSize       int               `deprecated:"true"`
+
+	GoogleEnabled      bool   `def:"false" desc:"enables Google Oauth"`
+	GoogleClientID     string `def:"<yourClientID>" desc:"client ID generated for Google API"`
+	GoogleClientSecret string `def:"<yourClientSecret>" desc:"client secret generated for Google API"`
+	GoogleRedirectURL  string `def:"<pyroscope.mycompany.org/google/callback>" desc:"url that google will redirect to after logging in. Has to be in form <pathToPyroscopeServer/google/callback>"`
```
petethepig, line 87: Can we have a way of determining this automatically from request Host header?

petethepig: I think ideally:
* it should still be configurable
* by default just be an empty string
* and if it's an empty string the server just guesses it from host header and current request scheme (http / https)
* and we could add something to `desc` mentioning that you should only set it if pyroscope is under some fancy load balancer

pkg/server/controller.go
```diff
@@ -129,94 +208,28 @@ func (ctrl *Controller) drainMiddleware(next http.HandlerFunc) http.HandlerFunc
...
+			return []byte(ctrl.config.JWTSecret), nil
+		})
 
-	var extraMetadataStr string
-	extraMetadataPath := os.Getenv("PYROSCOPE_EXTRA_METADATA")
-	if extraMetadataPath != "" {
-		b, err = ioutil.ReadFile(extraMetadataPath)
 		if err != nil {
-			logrus.Errorf("failed to read file at %s", extraMetadataPath)
+			ctrl.log.Errorf("Error parsing jwt token: %v", err)
```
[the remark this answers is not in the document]
Skemba: Totally forgot about logrus `WithError`, will add in next commit

pkg/server/controller.go
Skemba: Removed scope as config flag so that won't be a problem :) 

I really like the idea. Will include this.

pkg/server/handler.go
Skemba: In general only error can be passed so I'll remove this boolean parameter, pass error when it's available and log it with `WithError`

pkg/config/config.go
Skemba: Hm interesting idea ... In general Host header shouldn't be trusted as they can be spoofed, but in this case I can't think of a problem that would cause us. Worst case scenario I can think of is spoofing the Host header which leads to wrong redirect URL and "400" error page from google. I'll look into doing this.

pkg/config/config.go
```diff
@@ -80,6 +80,34 @@ type Server struct {
...
+	// TODO: remove skip: true when we enable these back
+	GoogleEnabled      bool   `skip:"true" def:"false" desc:"enables Google Oauth"`
+	GoogleClientID     string `skip:"true" def:"" desc:"client ID generated for Google API"`
+	GoogleClientSecret string `skip:"true" def:"" desc:"client secret generated for Google API"`
+	GoogleRedirectURL  string `skip:"true" def:"" desc:"url that google will redirect to after logging in. Has to be in form <pathToPyroscopeServer/google/callback>"`
+	GoogleAuthURL      string `skip:"true" def:"https://accounts.google.com/o/oauth2/auth" desc:"auth url for Google API (usually present in credentials.json file)"`
+	GoogleTokenURL     string `skip:"true" def:"https://accounts.google.com/o/oauth2/token" desc:"token url for Google API (usually present in credentials.json file)"`
+
+	GitlabEnabled bool `skip:"true" def:"false" desc:"enables Gitlab Oauth"`
+	// TODO: why is this one ApplicationID and not ClientID ?
```
Skemba, line 93: Gitlab calls it Application ID while others call it Client ID, but it's used in the same manner in both cases

Screenshot 2021-07-16 at 22 55 09

pkg/config/config.go
```diff
@@ -80,6 +80,34 @@ type Server struct {
...
+	GitlabAPIURL        string `skip:"true" def:"https://gitlab.com/api/v4/user" desc:"URL to gitlab API (keep default for cloud, usually https://gitlab.mycompany.com/api/v4/user for on-premise)"`
+
+	GithubEnabled      bool   `skip:"true" def:"false" desc:"enables Github Oauth"`
+	GithubClientID     string `skip:"true" def:"" desc:"client ID generated for Github API"`
+	GithubClientSecret string `skip:"true" def:"" desc:"client secret generated for Github API"`
+	GithubRedirectURL  string `skip:"true" def:"" desc:"url that Github will redirect to after logging in. Has to be in form <pathToPyroscopeServer/github/callback>"`
+	GithubAuthURL      string `skip:"true" def:"https://github.com/login/oauth/authorize" desc:"auth url for Github API"`
+	GithubTokenURL     string `skip:"true" def:"https://github.com/login/oauth/access_token" desc:"token url for Github API"`
+
+	// TODO: can we generate these automatically if it's empty?
```
Skemba, line 108: We can, we would just have to write it in some kind of persistent storage so that users don't get logged out if pyroscope is restarted. It's either that or making users explicitly generate it, which I think should be a good security practice (but it does make less of the things out-of-the-box)

Issue #216, opened 2021-05-25
Explore shared library approach


petethepig:
Some of the profilers we depend on use `process_vm_readv`, and it's disabled in some environments. We could maybe get around this if we added ability to integrate profilers as shared libraries. Then maybe the profiler could read its own memory. Then you'd be able to integrate by just adding a dependency to your code, something like

```
pip install pyroscope
```

TODO: add more details here

AdrK:
I have been trying to play around with the process capabilities. Using `prctl()` you can change the capabilities of the process. I have written a piece of code in which I have used different ways to modify the capabilities of current process to allow for ptrace'ing it:
```
#include <stdlib.h>
#include <stdio.h>
#include <time.h>
#include <sys/ptrace.h>
#include <sys/prctl.h>

int main(int argc, char* argv[])
{
	//ptrace(PTRACE_TRACEME, 0, 0);
	prctl(PR_SET_PTRACER, PR_SET_PTRACER_ANY , 0, 0, 0);
	struct timespec ts = {.tv_sec = 0, .tv_nsec = 1e8};
	for(int i = 0;; i++)
	{
		printf("i = %d\n", i);
		nanosleep(&ts, NULL);
	}
}
```

That didn't worked :wink: So I figured that the capabilities are not an issue, especially because this:
```
root@c9ce6049f22e:/workdir# cat /proc/sys/kernel/yama/ptrace_scope
0
```
Shows that ptrace'ing capabilities should be enabled by default.

I was looking for a solution, and this doc nailed the topic:
`https://jvns.ca/blog/2020/04/29/why-strace-doesnt-work-in-docker/`

TL;DR is:
Docker is using seccomp to allow and dissallow certain syscalls and `--cap-add` in Docker is not only just setting the capabilities, but also modifies the `seccomp` settings to allow for `process_vm_readv & writev`

So at the end, seems that there is no other option than, either use `--cap-add` or `--security-opt seccomp=unconfined` or in any other way change the seccomp settings

AdrK:
Since this commit ( Docker 19.03 ):
https://github.com/moby/moby/commit/1124543ca8071074a537a15db251af46a5189907
ptrace should work by default.

I will check it

AdrK:
I have found that although ptrace and process_vm_* calls are not working, the direct read from /proc/<pid>/mem is working :)

https://github.com/AdrK/ptrace_test/blob/master/src/test_on_child.c

AdrK:
[...]

PR #274, opened 2021-07-07
"native" integrations for ruby, python and php

petethepig:
This is work in progress PR for #216. This will allow to install pyroscope via pip, gem, etc.

/cc @AdrK

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,12 @@ else
 	THIRD_PARTY_DEPENDENCIES ?= "build-rust-dependencies"
 endif
 
+ifeq ("$(shell go env GOOS || true)", "linux")
+	ARCHIVE_EXT ?= "a"
+else
+	ARCHIVE_EXT ?= "bundle"
+endif
+
 EMBEDDED_ASSETS ?= ""
 EMBEDDED_ASSETS_DEPS ?= "assets-release"
 EXTRA_LDFLAGS ?= ""
@@ -34,12 +40,25 @@ all: build
 build:
 	$(GOBUILD) -tags $(ENABLED_SPIES) -ldflags "$(EXTRA_LDFLAGS) $(shell scripts/generate-build-flags.sh $(EMBEDDED_ASSETS))" -o ./bin/pyroscope ./cmd/pyroscope
 
+.PHONY: build-rbspy-static-library
+build-rbspy-static-library:
+	$(GOBUILD) -tags nogospy,rbspy -buildmode=c-archive -ldflags "$(EXTRA_LDFLAGS) $(shell scripts/generate-build-flags.sh $(EMBEDDED_ASSETS))" -o "./bin/libpyroscope.rbspy.$(ARCHIVE_EXT)" ./pkg/agent/clib
+
+.PHONY: build-pyspy-static-library
+build-pyspy-static-library:
+	$(GOBUILD) -tags nogospy,pyspy -buildmode=c-archive -ldflags "$(EXTRA_LDFLAGS) $(shell scripts/generate-build-flags.sh $(EMBEDDED_ASSETS))" -o "./bin/libpyroscope.pyspy.$(ARCHIVE_EXT)" ./pkg/agent/clib
+
+.PHONY: build-phpspy-static-library
+build-phpspy-static-library:
+	$(GOBUILD) -tags nogospy,phpspy -buildmode=c-archive -ldflags "$(EXTRA_LDFLAGS) $(shell scripts/generate-build-flags.sh $(EMBEDDED_ASSETS))" -o "./bin/libpyroscope.phpspy.$(ARCHIVE_EXT)" ./pkg/agent/clib
+
 .PHONY: build-release
 build-release: embedded-assets
 	EMBEDDED_ASSETS=true $(MAKE) build
 
 .PHONY: build-rust-dependencies
 build-rust-dependencies:
+# relocation-model=pic
 	cd third_party/rustdeps && RUSTFLAGS="-C target-feature=+crt-static" cargo build --release
 
 .PHONY: build-phpspy-dependencies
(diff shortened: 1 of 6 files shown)


Dockerfile.static-libs
```diff
@@ -0,0 +1,90 @@
...
+RUN wget https://github.com/libunwind/libunwind/releases/download/v1.3.1/libunwind-1.3.1.tar.gz && \
+    tar -zxf libunwind-1.3.1.tar.gz && \
+    cd libunwind-1.3.1/ && \
+    ./configure --disable-minidebuginfo --enable-ptrace --disable-tests --disable-documentation && make && make install
+
+COPY third_party/rustdeps /opt/rustdeps
+
+WORKDIR /opt/rustdeps
+
+RUN RUSTFLAGS="-C target-feature=+crt-static" /root/.cargo/bin/cargo build --release --target $(uname -m)-unknown-linux-musl
```
AdrK, line 25: If musl is used, we will not be able to use this library later in any dynamic library.

Dockerfile.static-libs
```diff
@@ -0,0 +1,90 @@
...
+# | |_ _ _ __   __ _| |  _ _ __ ___   __ _  __ _  ___
+# |  _| | '_ \ / _` | | | | '_ ` _ \ / _` |/ _` |/ _ \
+# | | | | | | | (_| | | | | | | | | | (_| | (_| |  __/
+# |_| |_|_| |_|\__,_|_| |_|_| |_| |_|\__,_|\__, |\___|
+#                                           __/ |
+#                                          |___/
+
+FROM scratch AS lib-exporter
+COPY --from=go-builder /opt/pyroscope/out/libpyroscope.rbspy.a /
+COPY --from=go-builder /opt/pyroscope/out/libpyroscope.pyspy.a /
```
AdrK, line 88: afaik, there is no way in Python to load a static lib, so I guess we have to go with dynamic lib.

AdrK, line 88: This, for some reason, is missing the definitions from rustdeps library:
image

Same here: https://github.com/AdrK/pyspy_so/blob/master/Dockerfile

AdrK: Strange thing is, that if you would compile it as c-shared, it has those symbols.
WA is to link librustdeps.a again, later in Python extension.

petethepig: Yeah, I don't understand why this is happening right now. I'm just adding `librustdeps.a` on ruby side right now as well. :(

AdrK, line 88: Headers are missing:
```
COPY --from=go-builder /opt/pyroscope/out/libpyroscope.phpspy.a /
COPY --from=go-builder /opt/pyroscope/out/libpyroscope.phpspy.h /
COPY --from=go-builder /opt/pyroscope/out/libpyroscope.pyspy.a /
COPY --from=go-builder /opt/pyroscope/out/libpyroscope.pyspy.h /
COPY --from=go-builder /opt/pyroscope/out/libpyroscope.rbspy.a /
COPY --from=go-builder /opt/pyroscope/out/libpyroscope.rbspy.h /
COPY --from=go-builder /opt/pyroscope/third_party/rustdeps/target/release/librustdeps.a /
```

pkg/agent/clib/clib.go
```diff
@@ -0,0 +1,136 @@
...
+
+func performOSChecks() error {
+	if !caps.HasSysPtraceCap() {
+		return errors.New("if you're running pyroscope in a Docker container,  add --cap-add=sys_ptrace." +
+			"See our Docker Guide for more information: https://pyroscope.io/docs/docker-guide")
+	}
+	return nil
+}
+
+var Sessions = map[int]*agent.ProfileSession{}
```
AdrK, line 30: Guys, any idea on how can we close all sessions that would be still open at unexpected exit?

petethepig: Yeah.. I was gonna add a check to see if the process is still alive every 10 seconds or something like that.

Makefile
```diff
@@ -37,18 +37,77 @@ all: build
...
+# ensure-modern-ld-version:
+# 	@echo "Not on Linux, so not checking for modern ld version."
+# endif
+
+
+
+.PHONY: build-rbspy-static-library
+build-rbspy-static-library:
+	mkdir -p ./out
+	$(GOBUILD) -tags nogospy,rbspy,clib -buildmode=c-archive -ldflags "$(EXTRA_LDFLAGS) $(shell scripts/generate-build-flags.sh $(EMBEDDED_ASSETS))" -o "./out/libpyroscope.rbspy.a" ./pkg/agent/clib
```
AdrK, line 55: are `EMBEDDED_ASSETS` needed?

pkg/agent/clib/clib.go
```diff
@@ -0,0 +1,111 @@
...
+	sessionsMapMutex.Lock()
+	defer sessionsMapMutex.Unlock()
+
+	pid := int(Pid)
+
+	if _, ok := sessionsMap[pid]; !ok {
+		logger.Errorf("session for pid: %d doesn't exists", pid)
+		return -1
+	}
+	sessionsMap[int(Pid)].Stop()
```
AdrK, line 91: We should check if the pid is alive.

petethepig: right, good point

Makefile
petethepig: yeah we can remove it from here

pkg/agent/clib/clib.go
AdrK: Or do we? :D If we check if the pid is alive, and then try to stop it, it might already be killed/stopped in meantime, so it will be a race condition. We should rather be immune to situation when we try to stop already non existent process.
Sorry for confusion!

pkg/agent/gospy/gospy.go
```diff
@@ -170,3 +172,7 @@ func (s *GoSpy) Reset() {
 
 	s.reset = true
 }
+
+func init() {
```
AdrK, line 176: I wonder, how come that tests didn't find this?

pkg/util/throttle/throttle.go
```diff
@@ -0,0 +1,33 @@
+package throttle
```
AdrK, line 1: Is this used anywhere?

PR #276, opened 2021-07-12
Add test for storage/tree.Merge

olvrng:
This PR adds tests for storage/tree.Merge to cover:

- merging 2 similar trees
- merging 2 trees with differences in structure

Issue #95, opened 2021-02-24
differential flamegraph for comparision view


YangKeao:
First thanks for the `pyroscope`. It's really a great project :beers: 

I have seen the comparision view in the online demo. It's awesome :+1: . Wouldn't it be better to show as a [Differential Flame Graphs](http://www.brendangregg.com/blog/2014-11-09/differential-flame-graphs.html), as it's really hard to tell the difference between two flamegraph :cry:.

Rperry2174:
Hi @YangKeao I think this is a great idea actually! Thanks for the resource as well it definitely lays out some of the parts where we got stuck as we were thinking about how to display this data. Just curious what your opinion is on which diff you think is best UI/UX and which one is most useful:

[Option 1:](http://www.brendangregg.com/blog/images/2014/zfs-flamegraph-diff.svg) Red blue differential
![image](https://user-images.githubusercontent.com/23323466/109081302-ee5b0500-76b6-11eb-9bf5-53ef58438729.png)

Option 2: Negation
![image](https://user-images.githubusercontent.com/23323466/109081431-24988480-76b7-11eb-81d3-4612d257301b.png)

[Option 3:](http://corpaul.github.io/flamegraphdiff/demos/dispersy/dispersy_diff.html) Before / After + differential
![image](https://user-images.githubusercontent.com/23323466/109081499-40038f80-76b7-11eb-8a31-7e99b8175227.png)

YangKeao:
I prefer the first two options :yum: . As described in the article, they are useful for different situations. The best solution may be making it selectable and let the users choose the best fit one.

petethepig:
First thing we need I think is a Diff function in `tree` package. It should take 2 trees and subtract values of one tree from the other one. I think it's gonna be pretty similar to the existing merge function:

https://github.com/pyroscope-io/pyroscope/blob/main/pkg/storage/tree/tree.go#L63

olvrng:
I think there is another option for drawing diff, as described in this [gist](https://gist.github.com/gelisam/f64934146a31194d71e2883c24f4c879) by [gelisam](https://github.com/gelisam).

- Option 1: Red/Blue
  Draw the structure with shape and size as the second profile, then use color to display the difference.

- Option 2: Negation
  Draw the structure with shape and size as the first profile, then use color to display the difference.

- Option 4: Red/Blue with sum
  Draw the structure with shape and size as the sum of both profiles, then use color to display the difference.

[...]

PR #277, opened 2021-07-12
WIP: Implement tree diff

olvrng:
This PR provides a sample implementation for tree diff as described in #95

Depends on: #276


olvrng:
replaced by #288

PR #280, opened 2021-07-13
Tags support

kolesnikovae:
Related:
 - https://github.com/pyroscope-io/pyroscope/issues/162
 - https://github.com/pyroscope-io/pyroscope/issues/264
 - https://github.com/pyroscope-io/pyroscope.io/pull/13 (FlameQL documentation)


petethepig (approved):
It all looks great. The only thing is I'm a little concerned about the removal of `runtime.GC` calls. I'm going to merge this PR, and then do some more tests to see if we need to bring `runtime.GC` back or not.

PR #283, opened 2021-07-15
fix(gospy.go): Snapshot add custom_pprof.StopCPUProfile()

hiyanxu:
Hello:
    There is a problem happen with me: when I use custom sampling, the pyroscope can not display CPU profiles.
    Discover by debug, there is an err: cpu profiling already in use
image

    I found that custom sampling did not stop, so unable to call startCPUProfile from the second time.


petethepig:
Thanks @hiyanxu !

PR #284, opened 2021-07-15
scripts/decode-resp: decode flamebearer response for debugging

olvrng:
This PR adds scripts/decode-resp for decoding response.

```
Decode response from pyroscope, for debugging

Usage: decode-resp -file <response.json>
  where <response.json> is the body response from /render API
  example: http://localhost:4040/render?from=now-1h&until=now&name=pyroscope.server.alloc_objects{}&max-nodes=1024&format=json

Flags:
  -file string
    	path to response file (required)
  -out string
    	name of output file (default to NAME.out.EXT)
```

### DIFF
--- a/scripts/decode-resp/decode_test.go
+++ b/scripts/decode-resp/decode_test.go
@@ -1,0 +1,70 @@
+package main
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/pyroscope-io/pyroscope/pkg/storage/tree"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Decode", func() {
+	Context("simple case", func() {
+		It("should decode correctly", func() {
+			names := strings.Split("total,a,b,c", ",")
+			levels := [][]int{
+				{0, 3, 0, 0},             // total
+				{0, 3, 0, 1},             // a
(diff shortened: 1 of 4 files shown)


scripts/decode-resp/decode_test.go
```diff
@@ -0,0 +1,66 @@
...
+        {
+          "name": "b",
+          "total": 2,
+          "self": 2
+        }
+      ]
+    ]
+  }
+}`
+	assert.Equal(t, expected, outJson)
```
petethepig, line 60: Please use ginkgo + gomega for testing stuff, here's an example: https://github.com/pyroscope-io/pyroscope/blob/main/pkg/storage/storage_test.go#L228

olvrng: I see. Will change that.

petethepig:
This looks good. My only question is why do we need it. Can you explain the usescase?

olvrng:
As a new contributor, I need to understand the response structure (especially `levels` with delta encoding), so I created this script:

- This will help other new contributors when joining the project, to easily decoding/understanding the response (with delta encoding).
- Also, when we add new data or modify the structure, as discussed in implementing the diff view, this will help to debug.

scripts/decode-resp/decode_test.go
olvrng: I updated the test with ginkgo + gomega.

PR #288, opened 2021-07-16
Implement mock ui for diff

olvrng:
This PR implements a mock UI for comparision-diff:

- [x] New sidebar button: Diff View
  - [ ] Currently using a mock icon
- [x] New URL: /comparision-diff
- [x] Mock UI for selecting from 2 time ranges (left & right)

This serves as a reference for discussing the UI and control. Other features will be pushed in other PRs:

- [x] FE: Request combined data for rendering from 2 time ranges
        #289
- [x] BE: Respond combined data for 2 time ranges
        New URL format: `/render?leftFrom=...&leftUntil=...&rightFrom=....&rightUntil=...`
        #290 
- [x] FE: Actually render the differences
        #289
- [ ] FE: Add more columns to the ProfilerTable
- [ ] FE: Add diff text to tooltip
- [ ] FE: Support different rendering options: left, right, both, percentage
- [ ] FE: Allow selecting between 3 different options

Issues: #95

Screen Shot 2021-07-16 at 22 36 15

### DIFF
--- a/pkg/server/handler.go
+++ b/pkg/server/handler.go
@@ -311,10 +311,11 @@ func (ctrl *Controller) callbackRedirectHandler(getAccountInfoURL string, info *
 func (ctrl *Controller) indexHandler() http.HandlerFunc {
 	fs := http.FileServer(ctrl.dir)
 	return func(rw http.ResponseWriter, r *http.Request) {
-		if r.URL.Path == "/" {
+		path := r.URL.Path
+		if path == "/" {
 			ctrl.statsInc("index")
 			ctrl.renderIndexPage(rw, r)
-		} else if r.URL.Path == "/comparison" {
+		} else if path == "/comparison" || path == "/comparison-diff" {
 			ctrl.statsInc("index")
 			ctrl.renderIndexPage(rw, r)
 		} else {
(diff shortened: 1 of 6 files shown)


Rperry2174:
I think this is a good start. Could you explain a little more whats going on in the picture?

The part thats confusing about this, as it currently stands, is that it is hard to see which timelines correspond with which components on the page. 

For example:
- does the timeline on the left correspond with the table? 
- why is there no "blue" selection in the timeline on the right?
- How will the page be rearranged once you are taking the "diff" of two flamegraphs?

olvrng:
It's the second way: 

> the resulting flamegraph + table is going to be the diff of the right timeline's flamegraph and the left timeline's flamegraph, then the resulting diff flamegraph should take up the full width and the table should go beneath that flamegraph (similar to the single view). That way it is more clear that the resulting diff belongs to both timelines' flamegraphs.

- I forgot to draw the blue section. Both sections are required for displaying the diff graph.
- Both sections should be of the same width. For example, if the red section is 1 hour, then the second section should also be 1 hour. This makes more sense when comparing/diff-ing them.

Screen Shot 2021-07-17 at 06 26 12

petethepig:
Closing this as this commit is also in https://github.com/pyroscope-io/pyroscope/pull/289

PR #289, opened 2021-07-16
implement diff rendering

olvrng:
This PR implements diff rendering:
- [x] send request with new format: `single | double`
- [x] add adapters for different formats
- [x] render diff graph
- [x] render color as red/green (just work for now, will need a better calculation later)
- [x] render table with 3 options: self(left), self(right) ; total(left), total(right) ; self(diff), total(diff)
- [x] split API to `/render` and `/render-diff`
- [x] add tooltip
- [x] add color legend 

Depends: #288

### DIFF
--- a/pkg/server/render.go
+++ b/pkg/server/render.go
@@ -17,13 +17,6 @@ import (
 	"github.com/pyroscope-io/pyroscope/pkg/util/attime"
 )
 
-type RenderFormat string
-
-const (
-	FormatSingle = "single"
-	FormatDouble = "double"
-)
-
 var (
 	errUnknownFormat   = errors.New("unknown format")
 	errLabelIsRequired = errors.New("label parameter is required")
@@ -55,9 +48,8 @@ func (ctrl *Controller) renderHandler(w http.ResponseWriter, r *http.Request) {
 	leftStartTime, leftEndTime, leftOK := parseRenderRangeParams(r.URL.Query(), "leftFrom", "leftUntil")
 	rghtStartTime, rghtEndTime, rghtOK := parseRenderRangeParams(r.URL.Query(), "rightFrom", "rightUntil")
 
-	format := FormatSingle
-	if rghtOK || leftOK {
-		format = FormatDouble
+	isFormatDouble := rghtOK || leftOK
+	if isFormatDouble {
 		out, leftOut, rghtOut, err = ctrl.loadTreeConcurrently(p.gi, p.gi.StartTime, p.gi.EndTime, leftStartTime, leftEndTime, rghtStartTime, rghtEndTime)
 	} else {
 		out, err = ctrl.storage.Get(p.gi)
@@ -75,7 +67,7 @@ func (ctrl *Controller) renderHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	var fs *tree.Flamebearer
-	if format == FormatDouble {
+	if isFormatDouble {
 		leftOut.Tree, rghtOut.Tree = tree.CombineTree(leftOut.Tree, rghtOut.Tree)
 		fs = tree.CombineToFlamebearerStruct(leftOut.Tree, rghtOut.Tree, p.maxNodes)
 	} else {
@@ -90,10 +82,10 @@ func (ctrl *Controller) renderHandler(w http.ResponseWriter, r *http.Request) {
 		"timeline":    out.Timeline,
 		"flamebearer": fs,
 		"metadata": map[string]interface{}{
+			"format":     fs.Format, // "single" | "double"
 			"spyName":    out.SpyName,
 			"sampleRate": out.SampleRate,
 			"units":      out.Units,
-			"format":     format, // "single" | "double"
 		},
 	}
 
--- a/pkg/server/render.go
+++ b/pkg/server/render.go
@@ -1,14 +1,16 @@
 package server
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
+	"runtime/debug"
 	"strconv"
 	"sync"
 	"time"
+
+	"github.com/sirupsen/logrus"
 
 	"github.com/pyroscope-io/pyroscope/pkg/flameql"
 	"github.com/pyroscope-io/pyroscope/pkg/storage"
@@ -18,8 +20,10 @@ import (
 )
 
 var (
-	errUnknownFormat   = errors.New("unknown format")
-	errLabelIsRequired = errors.New("label parameter is required")
+	errUnknownFormat         = errors.New("unknown format")
+	errLabelIsRequired       = errors.New("label parameter is required")
+	errNoData                = errors.New("no data")
+	errTimeParamsAreRequired = errors.New("leftFrom,leftUntil,rightFrom,rightUntil are required")
 )
 
 type renderParams struct {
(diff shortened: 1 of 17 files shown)


pkg/server/render.go
```diff
@@ -106,3 +129,53 @@ func (ctrl *Controller) renderParametersFromRequest(r *http.Request, p *renderPa
...
+	return treeOut, leftOut, rghtOut, nil
+}
+
+func (ctrl *Controller) loadTree(gi *storage.GetInput, startTime, endTime time.Time) (_ *storage.GetOutput, _err error) {
+	defer func() {
+		rerr := recover()
+		if rerr != nil {
+			_err = fmt.Errorf("panic: %v", rerr)
+		}
+	}()
```
kolesnikovae, line 168: Consider implementing `recoveryMiddleware` that would wrap all handlers and catch all panics.

pkg/storage/tree/treediff.go
```diff
@@ -0,0 +1,240 @@
+package tree
```
kolesnikovae, line 1: Perhaps `tree_diff` is more appropriate file name.

kolesnikovae:
I have a suggestion to merge it with https://github.com/pyroscope-io/pyroscope/pull/290. What is the benefit of having two separate branches?

pkg/server/render.go
```diff
@@ -39,7 +42,19 @@ func (ctrl *Controller) renderHandler(w http.ResponseWriter, r *http.Request) {
...
-	out, err := ctrl.storage.Get(p.gi)
+	var err error
+	var out, leftOut, rghtOut *storage.GetOutput
+
+	leftStartTime, leftEndTime, leftOK := parseRenderRangeParams(r.URL.Query(), "leftFrom", "leftUntil")
+	rghtStartTime, rghtEndTime, rghtOK := parseRenderRangeParams(r.URL.Query(), "rightFrom", "rightUntil")
+
+	isFormatDouble := rghtOK || leftOK
+	if isFormatDouble {
+		out, leftOut, rghtOut, err = ctrl.loadTreeConcurrently(p.gi, p.gi.StartTime, p.gi.EndTime, leftStartTime, leftEndTime, rghtStartTime, rghtEndTime)
```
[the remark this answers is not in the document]
olvrng: This `out` is required for getting some information later, especially `"timeline": out.Timeline`. The timeline data is required for rendering, well, timeline on the UI.

Screen Shot 2021-07-25 at 10 56 44

pkg/server/render.go
```diff
@@ -39,7 +42,19 @@ func (ctrl *Controller) renderHandler(w http.ResponseWriter, r *http.Request) {
...
 
-	out, err := ctrl.storage.Get(p.gi)
+	var err error
+	var out, leftOut, rghtOut *storage.GetOutput
+
+	leftStartTime, leftEndTime, leftOK := parseRenderRangeParams(r.URL.Query(), "leftFrom", "leftUntil")
+	rghtStartTime, rghtEndTime, rghtOK := parseRenderRangeParams(r.URL.Query(), "rightFrom", "rightUntil")
+
+	isFormatDouble := rghtOK || leftOK
+	if isFormatDouble {
```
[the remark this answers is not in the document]
olvrng: I generally follow the convention from the original code, where `/render` is `GET` with `from=...&until=...`.

Because `out` is required, hence `from` and `until` are required. So extending the query with `leftFrom`, `leftUntil`, `rightFrom`, `rightUntil` is a reasonable choice.

(personally, I prefer using `POST` in API, but I followed the project convention here)

pkg/server/render.go
olvrng: Actually, this method is intended to be called in `loadTreeConcurrently`:

```go
var treeErr, leftErr, rghtErr error
var wg sync.WaitGroup
wg.Add(3)
go func() { defer wg.Done(); treeOut, treeErr = ctrl.loadTree(gi, treeStartTime, treeEndTime) }()
go func() { defer wg.Done(); leftOut, leftErr = ctrl.loadTree(gi, leftStartTime, leftEndTime) }()
go func() { defer wg.Done(); rghtOut, rghtErr = ctrl.loadTree(gi, rghtStartTime, rghtEndTime) }()
wg.Wait()
```

So middleware is not an option here. We must catch all panics in each goroutine or the whole process will stop.

pkg/storage/tree/treediff.go
olvrng: Screen Shot 2021-07-25 at 11 08 48
Screen Shot 2021-07-25 at 11 09 06

Actually, there is a reason for calling it `treediff` instead of `tree_diff`:
- The code file and test files are placed together in the first way: tree.go -> tree_test.go -> treediff.go ...
- Instead of the second way: tree.go -> treediff.go -> tree_test.go ...

But I think it's a project preference, and I will follow the convention in our code.

olvrng:
> I have a suggestion to merge it with #290. What is the benefit of having two separate branches?

I generally split server and client implementation. As described the commit tree:

* 94dd1a9 - (23 hours ago) add diff color **(feature/diff-client)**
* 7965619 - (24 hours ago) simplify parseFlamebearerFormat
* 6061286 - (24 hours ago) render diff with table
* 5075196 - (25 hours ago) render diff with mousemove
* bc3f7a0 - (3 days ago) render diff with new levels format
* fb984a0 - (4 days ago) send requests to render api with new format
* bd2230c - (24 hours ago) add format to flamebearer response **(feature/diff-server)**
* d0bedc0 - (26 hours ago) query 3 trees for diff
* 1819adb - (4 days ago) implement diff api by combining 2 trees
* c1d1429 - (13 days ago) implement ui for diff

The client implementation (feature/diff-client) is made dependant on the server implementation (`feature/diff-server`). This way, each team can do review their side. Also, when there is another fix that needs to be applied to the server, I can add it to the `feature/diff-server` branch, and rebase `feature/diff-client` on it.

* 94dd1a9 - (23 hours ago) add diff color **(feature/diff-client)**
* 7965619 - (24 hours ago) simplify parseFlamebearerFormat
* 6061286 - (24 hours ago) render diff with table
* 5075196 - (25 hours ago) render diff with mousemove
* bc3f7a0 - (3 days ago) render diff with new levels format
* fb984a0 - (4 days ago) send requests to render api with new format
* **a00000** - (1 hour ago) _another commit to fix some thing on the server_ **(feature/diff-server)**
* bd2230c - (24 hours ago) add format to flamebearer response
* d0bedc0 - (26 hours ago) query 3 trees for diff
* 1819adb - (4 days ago) implement diff api by combining 2 trees
* c1d1429 - (13 days ago) implement ui for diff

pkg/server/render.go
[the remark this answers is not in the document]
kolesnikovae: If we don't use the tree from `out`, perhaps it worth extending `storage` with a method for timeline retrieval (without the resulting tree).

Also I'm wondering how we should handle diff/comparison with respect to queries/tags.

pkg/server/render.go
kolesnikovae: `recoveryMiddleware` is not supposed to be a replacement for this recovery. There are many other places in handlers routines that may cause panic.

pkg/server/render.go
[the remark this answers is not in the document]
kolesnikovae: I think we define conventions ourselves :)

I believe that every endpoint/handler should solve exactly one problem, therefore I'd introduce a new one for rending diff. And POST with JSON-encoded parameters would be the most appropriate here, IMO.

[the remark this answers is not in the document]
olvrng: I'm ok with that. Let's add @petethepig to this conversation so we can have another opinion.

pkg/server/render.go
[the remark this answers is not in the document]
olvrng: Maybe we should add another PR for extending `storage` and queries/tags.

pkg/server/render.go
olvrng: I'm not sure I understand your last comment. Should I keep that code or use something else?

kolesnikovae: If you are concerned with possible panics, in addition to this panic recovery (which scope is `loadTree`) consider implementing recovery middleware like this: https://github.com/pyroscope-io/pyroscope/blob/3bca77ab0930ef02021719b463a2009a31085450/pkg/server/controller.go#L303
that would catch all panics in all handlers (but, of course, not in goroutines created by them). Although a panic in an HTTP handler does not cause app crash (directly), recovering improves user experience, especially if server logs are collected and parsed.

Regarding panic recovery in general: it may make sense to capture stack trace, otherwise error message may be not informative.

olvrng: No, I mean the code is intended to be used in goroutines, not from an HTTP handler:

```go
func (ctrl *Controller) loadTreeConcurrently(...) {
  var treeErr, leftErr, rghtErr error
  var wg sync.WaitGroup
  wg.Add(3)
	                                       // call here ⤵
  go func() { defer wg.Done(); treeOut, treeErr = ctrl.loadTree(gi, treeStartTime, treeEndTime) }()
  go func() { defer wg.Done(); leftOut, leftErr = ctrl.loadTree(gi, leftStartTime, leftEndTime) }()
  go func() { defer wg.Done(); rghtOut, rghtErr = ctrl.loadTree(gi, rghtStartTime, rghtEndTime) }()
  wg.Wait()
  // ...
}

func (ctrl *Controller) loadTree(...) {
  defer func() { 
    rerr := recover() // <- capture here
    // ...
  }()
  // ...
}
```

We should always capture panic inside a goroutine, or the process will crash. This is different from HTTP handlers.

(for HTTP handlers, the project should already have recover middleware somewhere, it's out of scope for this PR)

kolesnikovae: Okay, thanks @olvrng.

Meanwhile, please take into account:
> Regarding panic recovery in general: it may make sense to capture stack trace, otherwise error message may be not informative.

olvrng: Yes, I see. A stack trace log should be good enough for now. Is it ok?

kolesnikovae: Yep, `debug.Stack()` should tell us everything.

petethepig:
@olvrng I closed the other 2 prs to reduce confusion and consolidate the discussion to just one PR

pkg/server/render.go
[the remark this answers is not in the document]
petethepig: @olvrng I generally agree with @kolesnikovae — I think it would make the API simpler if we split it. But I know with these kinds of things there might be other non-obvious reasons to doing things in one function, so I would like to talk to you about this before we change it.

olvrng:
Screen Shot 2021-07-28 at 22 04 37

I added diff rendering to the table with selecting between 3 options: 
1. self(left), self(right)
2. total(left), total(right) 
3. self(diff), total(diff)

Rperry2174:
Thanks for the explanation comment I think this looks really good -- we talked a little bit about this in slack but just some thoughts:

- We should use some shade of grey instead of orange to show the "base" metric and then either green or red in contrast to that to show the diff 
- We should probably remove the line that sums the two graphs and instead replace it with descriptive information i.e.: `base graph: left, comparison graph: right` 
- Here the scale should go from -100% to +100% (i believe) so that all colors are represented 
image

olvrng:
I updated the legend and tooltip based on your comment.

webapp/javascript/components/ProfilerTable.jsx
```diff
@@ -122,37 +209,93 @@ function TableBody({ flamebearer, sortBy, sortByDirection }) {
...
-    );
+    ) : viewDiff === 'self' ? (x, color, style) => (
+      <tr key={x.name + renderID}>
+        <td>
+          <span className="color-reference" style={style} />
+          <span title={x.name}>{x.name}</span>
+        </td>
+        {/* NOTE: it seems React does not understand multiple backgrounds, have to workaround:  */}
+        {/*   The `style` prop expects a mapping from style properties to values, not a string. */}
+        <td STYLE={backgroundImageDiffStyle(x.selfLeft, x.selfRght, maxSelf, color, 'L')}>
```
Rperry2174, line 246: Can you clarify this part a little more? I noticed the console has this error, but are you saying this is intended? 
image

webapp/javascript/util/flamebearer.js
```diff
@@ -17,12 +17,67 @@
...
-    for (let i = 0; i < level.length; i += 4) {
+    for (let i = start; i < level.length; i += step) {
       level[i] += prev;
       prev = level[i] + level[i + 1];
     }
   }
 }
+
+export function deltaDiffWrapper(format, levels) {
+  if (format === "double") {
```
Rperry2174, line 31: Can you explain why there are two `deltaDiff` here with different start / steps with one happening right before the other?

webapp/javascript/components/ProfilerTable.jsx
olvrng: I try to use multiple backgrounds here. The original code that I tested (and it didn't work) was:

```html
// tried "style" instead of "STYLE"
<td style={backgroundImageDiffStyleALT(x.selfLeft, x.selfRght, maxSelf, color, 'L')}>
```
```js
function backgroundImageDiffStyleALT(a, b, total, color, side) {
  // ...
  // tried object instead of string for css
  return {
    backgroundImage: `linear-gradient(${clr}, ${clr}), linear-gradient(${cld}, ${cld})`;
    backgroundPosition: `${neg(-k)}px 0px, ${neg(-kd)}px 0px`;
    backgroundRepeat: `no-repeat`;
  };
}
```

It didn't work. And I had to change to the current code:

```html
// workaround: use "STYLE"
<td style={backgroundImageDiffStyle(x.selfLeft, x.selfRght, maxSelf, color, 'L')}>
```
```js
function backgroundImageDiffStyle(a, b, total, color, side) {
  // ...
  // workaround: use string for css
  return `
    background-image: linear-gradient(${clr}, ${clr}), linear-gradient(${cld}, ${cld});
    background-position: ${neg(-k)}px 0px, ${neg(-kd)}px 0px;
    background-repeat: no-repeat;
  `;
}
```

If you can make the original code work or find a better workaround, I'm happy to learn about it.

webapp/javascript/util/flamebearer.js
olvrng: The corresponding code on the server is here: [diff](https://github.com/pyroscope-io/pyroscope/pull/289/files#diff-b8e94c44499b631822aa06d0e72f48a20addcf5a4cca15e4e55d14745fcf3065R188-R191)

```go
func CombineToFlamebearerStruct(leftTree, rightTree *Tree, maxNodes int) *Flamebearer {
	// ...
	// delta encoding
	deltaEncoding(res.Levels, 0, 7)
	deltaEncoding(res.Levels, 3, 7)
	return &res
}
```

With `format=double`, the response from the server combines values from 2 left and right trees into a single array for `levels`, start at `i+0` and `i+3` with `step=7`:

```
i+0 = x offset, left  tree
i+1 = total   , left  tree
i+2 = self    , left  tree
i+3 = x offset, right tree
i+4 = total   , right tree
i+5 = self    , right tree
i+6 = index in the names array
```

The `levels` array is encoded using delta encoding for x offset on the left and the right trees before being sent to the client. Therefore, the client needs to decode it before rendering.

webapp/javascript/components/ProfilerTable.jsx
olvrng: I fixed the first warning `Warning: Each child in a list should have a unique "key" prop.`. My mistake.

Rperry2174:
Looks great thanks @olvrng! After merging this I'll follow up with creating some new issues for some of the todo's / future improvements needed here. Feel free to do the same for anything you feel like you didn't get to on this PR but is worth addressing in the future

PR #293, opened 2021-07-21
Metrics exporter

kolesnikovae:
This PR should allow exporting some user-defined metrics.

Metrics of specific profiles (defined with `FlameQL` query expressions) can be exposed as prometheus counters. User may also define a filter to limit profile nodes (stacks/signatures) accounted for the metric. Example configuration:
```yaml
metric-export-rules:
  - name: my_app_cpu_abc
    expr: app.name.cpu{foo=~"bar"}
    node: a;b;c # actually, a regex
  - name: my_app_cpu_total
    expr: app.name.cpu{baz="qux",foo!~"ba.*"}
    node: total # optional
```

TODO:
 - [x] Configuration, CLI & unit tests
 - [x] Cleanup
 - [ ] Documentation

**Important**: this PR also restricts allowed application name characters - an application name may only contain ASCII letters, digits, `_`, `-`, and `.`.


petethepig:
This looks great so far, I like the format for setting the rules:
```
  - name: my_app_cpu_total
    expr: app.name.cpu{baz="qux",foo!~"ba.*"}
    node: total # optional
```

2 thoughts I have on this:
* [ ] we should ideally allow users to set up these rules from the UI, that means we'll a) need to store that information somewhere (badger?) b) have some frontend for it.
* [ ] do you think we should expose these metrics along with all other pyroscope metrics via `/metrics` or should we have a different endpoint for them? 

We can probably discuss all of this in more detail on a call on Thursday.

petethepig:
I played with it some, have a couple of comments:
* right now `Observe` method is called in the http controller. This means the `Direct` upstream thing doesn't work with the exported metrics. It's not super important but something to think about.
* the other thing is I'm wondering if we should make these metrics gauges or counters... For some reason I feel like gauge would make more sense for this kind of usecase, but I'm not super sure. Let's discuss on a call.

kolesnikovae:
Thank you @petethepig for the comments!

> right now `Observe` method is called in the http controller. This means the `Direct` upstream thing doesn't work with the exported metrics. It's not super important but something to think about.

Done.

> the other thing is I'm wondering if we should make these metrics gauges or counters... For some reason I feel like gauge would make more sense for this kind of usecase, but I'm not super sure. Let's discuss on a call.

As discussed: the metric value is always a cumulative one (and usually, aggregates multiple dimensions), use of gauge is problematic. I think `Histogram` or `Summary` could be a best fit for us.

> we should ideally allow users to set up these rules from the UI, that means we'll a) need to store that information somewhere (badger?) b) have some frontend for it.

Agree, without UI it is barely useful. Related issue: https://github.com/pyroscope-io/pyroscope/issues/298. I'll get into it.

> do you think we should expose these metrics along with all other pyroscope metrics via /metrics or should we have a different endpoint for them?

As discussed, let's expose it via the same endpoint for now.

PR #296, opened 2021-07-23
Feature/tags-frontend

Loggy:


Rperry2174:
Lets append application name to the beginning of the tags input box: 
`pyroscope.server.alloc_objects{iteration=0}`

And this part is fixed / uneditable via the text. The only way one can edit the application name is using the dropdown.

Two other thoughts here:
- Can we also make the dropdown box wider to show the full name? 
- Can we add some color to the query input box so that its easier to distinguish tags from application name

PR #299, opened 2021-07-24
Replaced no longer maintained jwt-go with golang-jwt/jwt

Skemba:


petethepig (approved):
Thanks for looking into it!

PR #300, opened 2021-07-26
Migration to Cobra/Viper

Skemba:
In this PR, migrations from current setup to Cobra/Viper is being done.

It is still a work in progress, and some parts are not working (those that depend on nested structure) but would love to hear thoughts on structure/design decisions.

To Do:
- [x] Default values read from config struct
- [x] Biding env vars
- [x] Binding flags
- [x] Unmarshal binded keys in viper to config struct
- [x] Populate nested flags
- [x] Fix old tests / Add new ones
- [x] access limiting through allowed-domains, allowed-groups and allowed-organizations
- [x] Fix size type not being unmarshaled
- [x] Duration support
- [ ] Fix generate-sample-config to work with nested structs


main.go
```diff
@@ -0,0 +1,7 @@
+package main
+
+import "github.com/pyroscope-io/pyroscope/cmd"
+
+func main() {
+	cmd.Execute()
```
kolesnikovae, line 6: The layout slightly violates [the standard](https://github.com/golang-standards/project-layout). I think we should move sub-commands to a sub-package:

```
cmd
 └──pyroscope
	├── command (or cmd)
	│	├── agent.go
	│	├── banner.go
	│	├── connect.go
	│	├── convert.go
	│	├── dbmanager.go
	│	├── server.go
	│	└── ...
	├── main.go
	├── main_unix.go
	└── main_windows.go
```

cmd/agent.go
```diff
@@ -0,0 +1,39 @@
...
+	// agentCmd.PersistentFlags().String("foo", "", "A help for foo")
+
+	// Cobra supports local flags which will only run when this command
+	// is called directly, e.g.:
+	// agentCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
+	agentCmd.SetUsageFunc(func(cmd *cobra.Command) error {
+		fmt.Println(gradientBanner() + "\n" + DefaultUsageFunc(cmd.Flags(), cmd))
+		return nil
+	})
+}
```
kolesnikovae, line 39: Would it be better to do initialise commands in a more explicit way without global vars and `init`s? Like:
```go
func newAgentCmd(cfg *config.Agent) *cobra.Command
```

main.go
Skemba: I was thinking about this as well. On one hand - cobra provides the CLI generation tool to do it the "cobra way" but on the other hand like you said it's not fully compatible to the standard. Of course neither of these two is an official way that we have to follow. In the end I went with the cobra recommendation as we don't have the "main" entry but rather each (sub) command can be an entry to the application itself. We can of course still change this if you think it better fits how pyroscope should be organized as whole.

cmd/agent.go
Skemba: Again something that cobra generates and something that made me feel a bit uneasy as well, especially with global vars which I always like to avoid. But when I thought about why they would do it that way a bit more I realized it might be okay in this case. These are just configuration variables that are created and initialized in the beginning and never changed again, so having them won't create tight coupling or decrease readability or even hinder our ability to test these things. So I'm not sure whether I should abandon cobra "recommendation" in this case unless there are some downfalls I'm not seeing?

pkg/cli/flags_test.go
```diff
@@ -1,56 +1,58 @@
...
-					"-foo-bar", "test-val-4",
-					"-foo-foo", "10.23",
-					"-foo-bytes", "100MB",
+				cfg := FlagsStruct{}
+				PopulateFlagSet(&cfg, exampleCommand.Flags())
+
+				b := bytes.NewBufferString("")
+				exampleCommand.SetOut(b)
+				exampleCommand.SetArgs([]string{
+					fmt.Sprintf("--foo=%s", "test-val-1"),
```
kolesnikovae, line 45: My biggest concern is backward compatibility. Is there a way to accept [single-dash args](https://github.com/spf13/cobra/issues/740)?

petethepig (commented):
@Skemba Looks pretty good!

A found a few minor issues:
* logging format changed in exec function
```
# before:
INFO[2021-08-03T16:56:59.236412] exec/cli.go:80 to disable logging from pyroscope, pass -no-logging argument to pyroscope exec
# after:
INFO[0000]github.com/pyroscope-io/pyroscope/pkg/exec/cli.go:80 github.com/pyroscope-io/pyroscope/pkg/exec.Cli() to disable logging from pyroscope, pass -no-logging argument to pyroscope exec
```
* single dashes don't work, Anton already mentioned it. I think this is the biggest issue right now.
* errors in commands, e.g `pyroscope exec non-existing-program` are printed twice. it also prints usage message... I think for some errors it makes sense to print usage message and for some it doesn't. E.g if you run a command with a flag that's not defined that should print usage message, but when you run `pyroscope exec non-existing-program` it shouldn't print usage message imo

pyroscope.go
```diff
@@ -1,4 +0,0 @@
-package pyroscope
```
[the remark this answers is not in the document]
Skemba: - Fixed logging
- Done the workaround to allow both single and double dash flags as discussed
- fixed so that errors are not printed twice

> Why did you delete this file? It's needed for pkger to work.

As for this, it is needed to be a `.go` file in the root directory. I just replaced that one with another one so pkger should still be working (it worked when I ran it with `pkger`)

cmd/pyroscope/command/agent.go
```diff
@@ -0,0 +1,54 @@
...
+
+				// If a config file is found, read it in.
+				if err := viper.ReadInConfig(); err == nil {
+					fmt.Fprintln(os.Stderr, "Using config file:", viper.ConfigFileUsed())
+				}
+
+				if err := viper.Unmarshal(cfg); err != nil {
+					fmt.Fprintln(os.Stderr, "Unable to unmarshal:", err)
+				}
+			}
```
kolesnikovae, line 30: The error flow is most likely broken. And this fragment is very repetitive. Let it DRY:

```Golang
func loadConfig(path string, v interface{}) error {
	if path == "" {
		return nil
	}
	// I guess we should also handle case when the default config file is not found.
	viper.SetConfigFile(path)
	if err := viper.ReadInConfig(); err != nil {
		return err
	}
	return viper.Unmarshal(v)
}
```

cmd/pyroscope/command/agent.go
```diff
@@ -0,0 +1,54 @@
...
+
+			err := cli.StartAgent(cfg)
+			if err != nil {
+				cmd.SilenceUsage = true
+			}
```
kolesnikovae, line 35: Not sure if it required indeed: `SilenceUsage` suppresses usage when an error occurs so that it can be set to `true` at command initialisation.

cmd/pyroscope/command/agent.go
```diff
@@ -0,0 +1,54 @@
...
+		},
+	}
+
+	cli.PopulateFlagSet(cfg, agentCmd.Flags(), cli.WithSkip("targets"))
+	viper.BindPFlags(agentCmd.Flags())
+
+	agentCmd.SetUsageFunc(func(cmd *cobra.Command) error {
+		fmt.Println(gradientBanner() + "\n" + DefaultUsageFunc(cmd.Flags(), cmd))
+		return nil
+	})
```
kolesnikovae, line 47: I'm wondering if specifying this at `root` level is sufficient.

cmd/pyroscope/command/agent.go
```diff
@@ -0,0 +1,54 @@
...
+
+	if err := viper.Unmarshal(cfg); err != nil {
+		fmt.Fprintln(os.Stderr, "Unable to unmarshal:", err)
+	}
```
kolesnikovae, line 51: It seems we can move it to a separate function – this snippet is used multiple times without any changes.

cmd/pyroscope/command/root.go
```diff
@@ -0,0 +1,85 @@
...
+	dbManagerCmd := newDbManagerCmd(&cfg.DbManager, &cfg.Server)
+	execCmd := newExecCmd(&cfg.Exec)
+	serverCmd := newServerCmd(&cfg.Server)
+
+	rootCmd.AddCommand(agentCmd)
+	rootCmd.AddCommand(connectCmd)
+	rootCmd.AddCommand(convertCmd)
+	rootCmd.AddCommand(dbManagerCmd)
+	rootCmd.AddCommand(execCmd)
+	rootCmd.AddCommand(serverCmd)
```
kolesnikovae, line 61:
```suggestion
	rootCmd := newRootCmd(&cfg)
	rootCmd.AddCommand(
		newAgentCmd(&cfg.Agent),
		newConnectCmd(&cfg.Exec),
		newConvertCmd(&cfg.Convert),
		newDbManagerCmd(&cfg.DbManager, &cfg.Server),
		newExecCmd(&cfg.Exec),
		newServerCmd(&cfg.Server),
	)
```

cmd/pyroscope/command/root.go
```diff
@@ -0,0 +1,85 @@
...
+
+	args := os.Args[1:]
+	for i, arg := range args {
+		if len(arg) > 2 && strings.HasPrefix(arg, "-") && !strings.HasPrefix(arg, "--") {
+			args[i] = fmt.Sprintf("-%s", arg)
+		}
+	}
+
+	rootCmd.SetArgs(args)
+	rootCmd.Execute()
```
kolesnikovae, line 84: It may be better to set `SilenceErrors` (is an option to quiet errors down stream) to `true` and handle error from this call. After all it gives us more transparent error handling and better control.

cmd/pyroscope/main.go
```diff
@@ -1,14 +1,7 @@
...
-	"github.com/pyroscope-io/pyroscope/pkg/config"
-)
+import "github.com/pyroscope-io/pyroscope/cmd/pyroscope/command"
 
 func main() {
-	if err := cli.Start(new(config.Config)); err != nil {
-		fatalf("%s %v\n\n", color.RedString("Error:"), err)
-	}
+	command.Initialize()
 }
```
kolesnikovae, line 7: It's essential to preserve `fatalf` behaviour. On Windows, Agent may run as a Windows Service in non-interactive mode, therefore any errors at start time should be written to Windows Event Log.

cmd/pyroscope/command/agent.go
petethepig: I think this just makes it so that it doesn't print usage twice or something like that. @Skemba can you please clarify? And I guess a comment here would be great because it's not obvious.

cmd/pyroscope/command/agent.go
petethepig: +1 on the DRYing

Skemba: ah yes, didn't notice it was repeating. 
As for 
`	if err := viper.ReadInConfig(); err != nil {
		return err
	}`
I am not sure what we are to do in case of error. As it will fallback to flags and env vars if they exist. Should we just log this error and continue? Or should we actually stop further execution if config file is not found or has an error?

cmd/pyroscope/command/agent.go
Skemba: I will add comment for this as well, but it was added because we don't want usage to be printed when an error occurs running the subcommand, but we still want it printed when parsing flags and similar fails.

cmd/pyroscope/command/agent.go
Skemba: great catch! it defaults first to its own usage func, then the one set for a parent and in the end to the default cobra one. In our case specifying it at `root` level is sufficient.

cmd/pyroscope/command/agent.go
kolesnikovae: Good point. I'd advocate for fail fast strategy – if user specified a config file but it can not be read, it clearly says that this is a mistake/misconfiguration. Silent failures are pure evil.

cmd/pyroscope/command/agent.go
kolesnikovae: > we don't want usage to be printed when an error occurs running the subcommand, but we still want it printed when parsing flags and similar fails.

Aha, I see. I should say that we can achieve exactly this behaviour by setting some additional handlers:

```golang

func newRootCmd() *cobra.Command {
	var version bool
	rootCmd := &cobra.Command{
		Use:           "pyroscope [flags] <subcommand>",
		SilenceUsage:  true,
		SilenceErrors: true,
		Run: func(cmd *cobra.Command, args []string) {
			if version {
				printVersion(cmd, args)
				return
			}
			_ = usageFn(cmd)
		},
	}
	rootCmd.SetUsageFunc(usageFn)
	rootCmd.SetFlagErrorFunc(flagParseErrFn)
	rootCmd.Flags().BoolVarP(&version, "version", "v", false, "")
	return rootCmd
}

func usageFn(cmd *cobra.Command) error {
	cmd.Println(gradientBanner())
	cmd.Println(DefaultUsageFunc(cmd.Flags(), cmd))
	return nil
}

func flagParseErrFn(cmd *cobra.Command, err error) error {
	_ = usageFn(cmd)
	cmd.Println("Error:", err)
	return nil
}
```

```Golang

func newVersionCmd() *cobra.Command {
	return &cobra.Command{
		Use:    "version",
		Hidden: true,
		Run:    printVersion,
	}
}

func printVersion(_ *cobra.Command, _ []string) {
	fmt.Println(gradientBanner())
	fmt.Println(build.Summary())
}

```

----

Note that the snippet also affects how `version` command/version works: now it is printed on:
 - `pyroscope -v`
 - `pyroscope -version`
 - `pyroscope --version`
 - `pyroscope version`

`Version` member of `config.Config` can be removed, I think.

----

We also need to make sure command descriptions are okay. Now it looks as the following:
```
  agent       starts pyroscope agent.
  completion  generate the autocompletion script for the specified shell
  connect     connects to an existing process and profiles it
  exec        starts a new process from arguments and profiles it
  help        Help about any command
  server      starts pyroscope server. This is the database + web-based user interface
```

`completion` and `help` specify the action imperatively, meanwhile other just describe. Please check capitals and punctuation marks.

----

> Although I'm not sure if I should make it even more specific as we might want usage to still be printed for example if exec.Cli failes with "no arguments passed", but not if any other error happens?

We could also use cobra's validation.

----

Another thing is that we log some messages at start time which makes the output a bit inconsistent :/ I think we could omit this.

> INFO[2021-08-19T12:39:50.540138] scope/pkg/exec/cli.go:80 to disable logging from pyroscope, pass -no-logging argument to pyroscope exec 
Error: on macOS you're required to run the agent with sudo

cmd/pyroscope/command/exec.go
```diff
@@ -0,0 +1,49 @@
...
+		Short: "starts a new process from arguments and profiles it",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if cfg.NoLogging {
+				logrus.SetLevel(logrus.PanicLevel)
+			} else if l, err := logrus.ParseLevel(cfg.LogLevel); err == nil {
+				logrus.SetLevel(l)
+			}
+			if len(args) == 0 || args[0] == "help" {
+				fmt.Println(gradientBanner())
+				fmt.Println(DefaultUsageFunc(cmd.Flags(), cmd))
```
kolesnikovae, line 27:
```suggestion
				_ = cmd.Help()
```

But I'm not sure if we should doing this at all: we have `--help` flag, it is sufficient.

cmd/pyroscope/command/usage.go
```diff
@@ -32,44 +32,45 @@ var hiddenCommands = []string{
...
 
-	if len(c.Subcommands) > 0 {
+	if c.HasSubCommands() {
 		headerClr.Fprintf(&b, "SUBCOMMANDS\n")
 		tw := tabwriter.NewWriter(&b, 0, 2, 2, ' ', 0)
-		for _, subcommand := range c.Subcommands {
-			if !slices.StringContains(hiddenCommands, subcommand.Name) {
-				fmt.Fprintf(tw, "  %s\t%s\n", itemClr.Sprintf(subcommand.Name), subcommand.ShortHelp)
+		for _, subcommand := range c.Commands() {
+			if !slices.StringContains(hiddenCommands, subcommand.Name()) {
```
kolesnikovae, line 57: We should use `cobra.Command.Hidden` attribute. Instead of global `hiddenCommands` every such a command should be marked accordingly.

```suggestion
			if !subcommand.Hidden {
```

cmd/pyroscope/command/exec.go
Skemba: I agree with the second statement, it's probably redundant in this case, will remove it, unless @petethepig think there's a reason to keep it?

cmd/pyroscope/command/agent.go
Skemba: > Aha, I see. I should say that we can achieve exactly this behaviour by setting some additional handlers:

That's true, just don't know if we only want usage printed when there's a flag error or is there some other case we want this to cover as well? "We could also use cobra's validation" -> for example this is a good idea if we care about flags+arguments, just wondering if there is something else too?

> Note that the snippet also affects how version command/version works: now it is printed on:

Not sure what is this version part in reference to? 

> completion and help specify the action imperatively, meanwhile other just describe. Please check capitals and punctuation marks.

these are the default python ones, while the others are ours for which I have copied out what text was there previously. How do we want it to be?

> Another thing is that we log some messages at start time which makes the output a bit inconsistent :/ I think we could omit this.

not sure what do you mean exactly exactly? Remove the logs that happen before error checking?

kolesnikovae: 0. Usage. I believe we should print out usage message only when subcommand or flag is not valid. Argument value most likely does not have anything to do with CLI.

1. Version. Right now `pyroscope version` and `pyroscope --version` return `Error: unknown command "version" for "pyroscope"`. I'd suggest introducing a separate subcommand. And optionally a flag – just for convenience.

2. Help. What caught my eye is that descriptions for subcommands are written in different fashions:
```
  agent       starts pyroscope agent.
  help        Help about any command
```
I think we should rely on some naming convention. I'd suggest to use cobra's one:
```
  agent       Start pyroscope agent
  help        Help about any command
```

3. Log messages at start. As you can see this log message is not quite appropriate here. Perhaps we should use logger only after user input is validated.

cmd/pyroscope/command/agent.go
kolesnikovae: @Skemba at this point pyroscope fails to start if the default config file is not found. I'm wondering if it is intended.

cmd/pyroscope/command/common.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+func loadFlags(cfg interface{}, cmd *cobra.Command, opts ...cli.FlagOption) {
+	cli.PopulateFlagSet(cfg, cmd.Flags(), opts...)
+	viper.BindPFlags(cmd.Flags())
+
+	if err := viper.Unmarshal(cfg); err != nil {
+		fmt.Fprintln(os.Stderr, "Unable to unmarshal:", err)
+	}
+}
```
kolesnikovae, line 32: I'm afraid but use of global `viper` instance is not we really want. In `PopulateFlagSet` we set default values by parameter name, like:
```golang
	val := fieldV.Addr().Interface().(*string)
	for old, n := range o.replacements {
		defaultValStr = strings.ReplaceAll(defaultValStr, old, n)
	}
	flagSet.StringVar(val, nameVal, defaultValStr, descVal)
	viper.SetDefault(nameVal, defaultValStr)
```

The problem is that it happens during a command initialisation (when `loadFlags` is called) so that _all_ flag sets for _all_ commands are processed regardless of the subcommand user invokes. And due to name collisions (e.g. `config`), everything goes really bad. For example, `pyroscope agent` configuration file is resolved to `C:\Program Files\Pyroscope/etc/pyroscope/server.yml`.

The problem can be solved by using separate `viper` instances for every subcommand.

cmd/pyroscope/command/agent.go
kolesnikovae: Just to clarify, it's a breaking change and we should be very careful.

if user specified a config file but it can not be read, it clearly says that this is a mistake/misconfiguration. Silent failures are pure evil.

petethepig: yeah it shouldn't fail if there's no config file

cmd/pyroscope/command/exec.go
petethepig: Let's keep help subcommand, the easier it is for people to find help the better imo. From user perspective it's frustrating when you install some command and you don't know if it's `pyroscope -h`, `pyroscope -help`, `pyroscope --help` or `pyroscope help`. So why not just handle all of them.

kolesnikovae:
LGTM

petethepig (approved):
Looks good, merging it. Thank you @Skemba for leading this huge effort and thank you @kolesnikovae for thorough testing!

PR #302, opened 2021-07-26
fix(agent): use the ProfileTypes from the configuration

vbehar:
instead of the default list of ProfileTypes


petethepig:
Thank you @vbehar !

PR #303, opened 2021-07-26
Performance optimisations

kolesnikovae:
The PR is aimed to improve pyroscope server and profiler performance. Primarily by reducing the amount and size of allocations.

1. Use byte buffer pool for serialization.

Although it might be not so apparent in synthetic benchmarks, this should alleviate GC work. The benchmark shows that with a buffer pool, we decrease size of allocations by more than 20% (example for an empty `Segment`):
```
BenchmarkSegment_Serialize/Before-4            370618              2828 ns/op            1288 B/op         22 allocs/op
BenchmarkSegment_Serialize/After-4             397603              2807 ns/op            1032 B/op         19 allocs/op
```

2. Refactor `GoSpy` pprof profile traversal.

This change affects both server and agent ends. Before the change, the function could constitute up to 20-50% of all allocations (by size).
```
BenchmarkProfile_Get/Before-4              77934             14211 ns/op            8094 B/op        105 allocs/op
BenchmarkProfile_Get/After-4              350104              3637 ns/op             429 B/op          3 allocs/op
```

Related: https://github.com/pyroscope-io/pyroscope/issues/294


pkg/convert/profile_extra.go
```diff
@@ -0,0 +1,73 @@
...
+			if b.Len() > 0 {
+				_ = b.WriteByte(';')
+			}
+			_, _ = b.WriteString(name)
+		}
+		cb(b.Bytes(), int(s.Value[valueIndex]))
+		b.Reset()
+	}
+
+	bytebufferpool.Put(b)
```
petethepig, line 38: would it be better to do `defer bytebufferpool.Put(b)` after `b := bytebufferpool.Get()`?

kolesnikovae: It was so initially :) I was playing with benchmark and decided to not put `defer` here: although its cost is negligible, it is a hot path. From my perspective, there is no much difference, if you think it looks better with `defer`, I'll return it back.

petethepig: No, this makes sense, I was just wondering why you decided not to do defer. I think in these cases it'd be useful to add comments, e.g:
```suggestion
	// don't use defer here as it's a hot path
	bytebufferpool.Put(b)
```

---

Now that I think about this, is this a hot path though? This will run ~ 5 times every 10 seconds, right?

kolesnikovae: Okay, I'll add a comment.

> Now that I think about this, is this a hot path though? This will run ~ 5 times every 10 seconds, right?

The function is called from `session.Snapshot()`. To my understanding it is called `profile_types` x `sample_rate` times per second (defaults to 500).

petethepig: Default sample rate is 100 Hz

And for in gospy case that API is... deceiving. the function is called Snapshot and for rbspy and pyspy profilers it means it will take a snapshot 100 times per second. In go case it almost always returns early and only does any work every 10 seconds (https://github.com/pyroscope-io/pyroscope/blob/e0f799a732de44679a9366cb7379e61f8163a08f/pkg/agent/gospy/gospy.go#L108-L115)

kolesnikovae: Oh, right. I'm wondering why then it consumes so much CPU (https://github.com/pyroscope-io/pyroscope/issues/294#issuecomment-887180487). I'll take a closer look.

kolesnikovae: I decided to put `defer` so as to not confuse anyone.

olvrng:
I tested `sync.Pool` and `bytebufferpool` by running benchmark from this repository: [omgnull/go-benchmark](https://github.com/omgnull/go-benchmark) (the benchmark is recommended by [bytebufferpool](https://github.com/valyala/bytebufferpool)).

Here is the result on my machine with Go 1.14 (the version from our [go.mod](https://github.com/pyroscope-io/pyroscope/blob/main/go.mod))

```
$ go test ./... -bench=. -benchmem
?   	github.com/omgnull/go-benchmark/buffer	[no test files]
goos: darwin
goarch: amd64
pkg: github.com/omgnull/go-benchmark/buffer/main
cpu: Intel(R) Core(TM) i7-6567U CPU @ 3.30GHz
BenchmarkGenericBuf-4                       	 2474088	       486.2 ns/op	    2576 B/op	       4 allocs/op
BenchmarkGenericStackBuf-4                  	 2242417	       644.1 ns/op	    2576 B/op	       4 allocs/op
BenchmarkAllocBuf-4                         	 2161522	       488.1 ns/op	    2576 B/op	       4 allocs/op
BenchmarkSyncPoolBuf-4                      	21519672	        49.74 ns/op	       0 B/op	       0 allocs/op
BenchmarkBpoolPoolBuf-4                     	 5466226	       218.2 ns/op	       0 B/op	       0 allocs/op
BenchmarkByteBufferPoolBuf-4                	19147827	        54.19 ns/op	       0 B/op	       0 allocs/op
BenchmarkEasyJsonBuffer-4                   	 4705569	       247.9 ns/op	     600 B/op	       4 allocs/op
BenchmarkEasyJsonBuffer_OptimizedConfig-4   	11684805	        97.13 ns/op	      24 B/op	       1 allocs/op
PASS
ok  	github.com/omgnull/go-benchmark/buffer/main	12.111s
```

It seems `sync.Pool` is faster than `bytebufferpool`. So I suggest removing the package and just use `sync.Pool` from the standard library.

kolesnikovae:
To me it looks more like fluctuations - see my results below:

`go 1.14.15`:
```
goos: darwin
goarch: amd64
BenchmarkGenericBuf-4                            1550229               765 ns/op            2576 B/op          4 allocs/op
BenchmarkGenericStackBuf-4                       1504646               793 ns/op            2576 B/op          4 allocs/op
BenchmarkAllocBuf-4                              1524476               815 ns/op            2576 B/op          4 allocs/op
BenchmarkSyncPoolBuf-4                          16666263                81.6 ns/op             0 B/op          0 allocs/op
BenchmarkBpoolPoolBuf-4                          4400764               269 ns/op               0 B/op          0 allocs/op
BenchmarkByteBufferPoolBuf-4                    16411653                75.8 ns/op             0 B/op          0 allocs/op
BenchmarkEasyJsonBuffer-4                        3181826               371 ns/op             608 B/op          4 allocs/op
BenchmarkEasyJsonBuffer_OptimizedConfig-4        7567078               139 ns/op              32 B/op          1 allocs/op
PASS
ok      go-benchmark/buffer/main     13.804s
```

`go 1.16.3`:
```
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
BenchmarkGenericBuf-4                            1578630               738.7 ns/op          2576 B/op          4 allocs/op
BenchmarkGenericStackBuf-4                       1441327               695.4 ns/op          2576 B/op          4 allocs/op
BenchmarkAllocBuf-4                              1619761               755.8 ns/op          2576 B/op          4 allocs/op
BenchmarkSyncPoolBuf-4                          17595960                74.40 ns/op            0 B/op          0 allocs/op
BenchmarkBpoolPoolBuf-4                          4571432               264.8 ns/op             0 B/op          0 allocs/op
BenchmarkByteBufferPoolBuf-4                    18865262                58.98 ns/op            0 B/op          0 allocs/op
BenchmarkEasyJsonBuffer-4                        3780392               309.9 ns/op           600 B/op          4 allocs/op
BenchmarkEasyJsonBuffer_OptimizedConfig-4        9238102               121.0 ns/op            24 B/op          1 allocs/op
PASS
ok      go-benchmark/buffer/main     12.559s
```

`go1.17-b8ca6e5` (tip):
```
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
BenchmarkGenericBuf-4                            1604217               799.3 ns/op          2576 B/op          4 allocs/op
BenchmarkGenericStackBuf-4                       1449591               813.2 ns/op          2576 B/op          4 allocs/op
BenchmarkAllocBuf-4                              1501189               796.2 ns/op          2576 B/op          4 allocs/op
BenchmarkSyncPoolBuf-4                          16659342                61.95 ns/op            0 B/op          0 allocs/op
BenchmarkBpoolPoolBuf-4                          4727404               246.3 ns/op             0 B/op          0 allocs/op
BenchmarkByteBufferPoolBuf-4                    17938633                58.09 ns/op            0 B/op          0 allocs/op
BenchmarkEasyJsonBuffer-4                        3227756               344.5 ns/op           600 B/op          4 allocs/op
BenchmarkEasyJsonBuffer_OptimizedConfig-4        8777935               128.8 ns/op            24 B/op          1 allocs/op
PASS
ok      go-benchmark/buffer/main     12.178s
```

But this benchmark suite uses a static data set which does not really allow evaluating `bytebufferpool` advantages. One great think about `bytebufferpool` (and the reason to replace `sync.Pool`) is that due to its internal calibrating mechanism it is more adaptive, which results in limited memory fragmentation and more efficient allocations. 

BTW, re go version - I think it's time to switch to newer one.

petethepig:
@kolesnikovae re go version — I updated it for linux and windows https://github.com/pyroscope-io/pyroscope/commit/e44f8c8d23467528a7ccb9846fee9d033529020f https://github.com/pyroscope-io/pyroscope/commit/94d5c4323dc8c3f18214bbde8ff1a903e551b993

petethepig (approved):

PR #309, opened 2021-07-28
Fix lint error

ayeniblessing101:


pkg/server/handler.go
```diff
@@ -194,6 +196,7 @@ func (ctrl *Controller) decodeGithubCallbackResponse(resp *http.Response) (strin
 	var userProfile callbackResponse
 	err := json.NewDecoder(resp.Body).Decode(&userProfile)
 	if err != nil {
+		ctrl.log.WithError(err).Error("Decoding Github Callback Response failed")
 		return "", err
```
kolesnikovae, line 200: The error is logged later in `callbackRedirectHandler`, no need to log it twice. Same for `decodeGitLabCallbackResponse` and `decodeGoogleCallbackResponse`.

pkg/server/handler.go
```diff
@@ -63,7 +63,8 @@ func invalidateCookie(w http.ResponseWriter, name string) {
 func (ctrl *Controller) logoutHandler() http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		if r.Method != "POST" && r.Method != "DELETE" {
-			ctrl.writeErrorMessage(w, http.StatusMethodNotAllowed, "only POST and DELETE are allowed")
+			renderServerError(w, "you can only logout via a POST or DELETE")
+			ctrl.log.Error("you can only logout via a POST or DELETE")
 			return
```
kolesnikovae, line 68: `renderServerError` has been removed not so long ago. I guess, 405 code (`http.StatusMethodNotAllowed`) is more appropriate here.

We also should avoid method name strings and use `http.MethodPost` and `http.MethodGet` instead:

```go
func (ctrl *Controller) logoutHandler() http.HandlerFunc {
	return func(w http.ResponseWriter, r *http.Request) {
		switch r.Method {
		case http.MethodPost, http.MethodGet:
			invalidateCookie(w, jwtCookieName)
			http.Redirect(w, r, "/login", http.StatusTemporaryRedirect)
		default:
			ctrl.writeErrorMessage(w, http.StatusMethodNotAllowed, "only POST and DELETE methods are allowed")
		}
	}
}

```

scripts/windows/generate-windows-version-info/main.go
```diff
@@ -22,16 +29,16 @@ func main() {
 	flag.Parse()
 
 	if version == "" {
-		fatalf("version is required")
+		err = fmt.Errorf("%s", "version is required")
```
kolesnikovae, line 32: I'm afraid this approach does not stop execution as `fatalf` did. This will result in unpredictable failures.

Instead, we could wrap these steps with another function, and handle errors as usual:

```go
func main() {
	var version string
	flag.StringVar(&version, "version", "", "Version in semver format.")

	var outputPath string
	flag.StringVar(&outputPath, "out", "", "Output file path.")

	var iconPath string
	flag.StringVar(&iconPath, "icon", "", "Icon file path.")
	flag.Parse()

	if err := generateVersionInfo(version, outputPath, iconPath); err != nil {
		_, _ = fmt.Fprintln(os.Stderr, err)
		os.Exit(1)
	}
}

func generateVersionInfo(version, outputPath, iconPath string) error {
 	version = strings.Trim(version, `"`)
 	v, err := semver.Parse(strings.TrimPrefix(version, "v"))
 	if err != nil {
 		return fmt.Errorf("invalid version %q: %w", version, err)
 	}
 	if outputPath == "" {
 		return fmt.Errorf("output path is required")
 	}
 	// ... all remaining steps.
 	return nil
}
```

scripts/windows/generate-windows-version-info/main.go
```diff
@@ -11,6 +11,13 @@ import (
 )
 
 func main() {
+	var err error
+	defer func(){
+		if err != nil {
+			os.Exit(1)
+		}
+	}()
+
```
kolesnikovae, line 20: I guess the error should be printed to stderr.

pkg/server/build.go
```diff
@@ -8,5 +8,9 @@ import (
 
 func (ctrl *Controller) buildHandler(w http.ResponseWriter, _ *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
-	_, _ = w.Write([]byte(build.PrettyJSON()))
+	_, err := w.Write([]byte(build.PrettyJSON()))
+
+	if err != nil {
+		ctrl.log.WithError(err).Error("Build Handler failed")
+	}
```
kolesnikovae, line 15: By convention, log messages should not begin with a capital letter (except acronyms). Logging an error without a proper annotation might be confusing – consider "writing response body" or "could not write response".

By the way, there are very limited number of cases when `http.ResponseWriter` (`*http.response`, actually) returns an error. In vast majority of them those are related to network (e.g., client closed connection) and usually are suppressed.

kolesnikovae (commented):
Thank you @ayeniblessing101! I've left a couple of comments, please take a look.

scripts/windows/generate-windows-version-info/main.go
```diff
@@ -82,11 +91,8 @@ func main() {
 	versionInfo.Walk()
 
 	if err = versionInfo.WriteSyso(outputPath, "amd64"); err != nil {
-		fatalf("failed to write output file %s: %v", outputPath, err)
+		err = fmt.Errorf("failed to write output file %s: %w", outputPath, err)
```
kolesnikovae, line 94:
```suggestion
		return fmt.Errorf("failed to write output file %s: %w", outputPath, err)
```

scripts/windows/generate-windows-version-info/main.go
```diff
@@ -21,17 +21,26 @@ func main() {
...
+		os.Exit(1)
 	}
+}
+
+func generateVersionInfo(version, outputPath, iconPath string) error {
 	version = strings.Trim(version, `"`)
 	v, err := semver.Parse(strings.TrimPrefix(version, "v"))
+  
+	if version == "" {
+		return fmt.Errorf("%s", "version is required")
```
kolesnikovae, line 35:
```suggestion
		return fmt.Errorf("version is required")
```

kolesnikovae (commented):
Looks pretty good! A couple of minor notes.

kolesnikovae (approved):
LGTM. Thanks @ayeniblessing101!

petethepig (approved):
Looks great, thank you @ayeniblessing101

PR #311, opened 2021-08-02
Some small improvement

olvrng:
- [x] improve tree.Insert
- [x] add prependBytes
- [x] change slice pre-allocation from 100 to 128

### DIFF
--- a/pkg/storage/tree/tree.go
+++ b/pkg/storage/tree/tree.go
@@ -136,20 +136,25 @@ func (n *treeNode) insert(targetLabel []byte) *treeNode {
 	return n.ChildrenNodes[i]
 }
 
+func (n *treeNode) insertChild(label []byte, value uint64) *treeNode {
+	n.Total += value
+	childNode := n.insert(label)
+	return childNode
+}
+
 func (t *Tree) Insert(key []byte, value uint64, _ ...bool) {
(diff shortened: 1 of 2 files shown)


pkg/storage/tree/tree.go
```diff
@@ -108,6 +108,13 @@ func prepend(s []*treeNode, x *treeNode) []*treeNode {
 	return append(s, prev)
 }
 
+func prependBytes(s [][]byte, x []byte) [][]byte {
+	s = append(s, nil)
+	copy(s[1:], s)
+	s[0] = x
+	return s
+}
```
kolesnikovae, line 116: I'm sure `prepend` can be done in the same way, without unnecessary checks, like:
```go
func prependTreeNode(s []*treeNode, x *treeNode) []*treeNode {
	s = append(s, nil)
	copy(s[1:], s)
	s[0] = x
	return s
}
```

kolesnikovae (commented):
LGTM.

pkg/storage/tree/tree.go
olvrng: That's right. I added a commit for `prependTreeNode`.

petethepig (approved):
Looks good. Thanks @olvrng !

PR #315, opened 2021-08-04
Improvement/dpr style

eh-am:
Hello, I've improved the Date Range Picker aesthetics/UX (subjective yes)

Previous:
2021-08-03-220157_413x604_scrot

After changes:
2021-08-03-221638_419x612_scrot

* changed the background/icon on hover
* setup spacing to be more consistent
* replace the from/until paragraphs for labels (so as a bonus clicking on the label opens the calendar)
* changed the button color to black (white text contrast wasn't very good according to chrome dev tools)
* moved .btn to its own sass file, maybe as a push for a design system? :P


Rperry2174:
Hey Eduardo! Thanks for the contribution :) -- this looks great and I like the general direction definitely an improvement. There are a couple of comments that I have here:

I'm not a huge fan of the checkmark (and mine looks different than yours)
image

Could we try maybe just doing a simple  white "bullet" instead?

image

I definitely think the background change is really nice and makes this easier to see. I also think the button contrast  is fine with white. I have the "dark theme" github and they use close to the same colors with their buttons

Standard:
image
On hover:
image

could you also alter the colors of our buttons to match these and add the "on hover" change?

eh-am:
Hmmm, I think the safest way is to use a svg.

I've updated the button, let me know what you think. Keep in mind right now it's scoped exclusively to this component, so we may require further updates before merging.

Rperry2174:
I think this looks great! I'm not sure what you mean though about requiring other updates before merging. Can you clarify?

eh-am:
>I think this looks great (both checkmark and button)! I'm not sure what you mean though about requiring other updates before merging. Can you clarify?
Ah, the change are scoped to the date range picker only, so other buttons aren't affected. 

After you give the go we can apply the same style to the other buttons, probably creating a component.

Rperry2174:
Yeah lets do this then. Can you change the color of this button to match the color of the new button? 
image

And then there is one JS test that is failing (I think a space got added to one of the presets)

With those two fixed I would be comfortable merging

webapp/javascript/components/CustomDatePicker.jsx
```diff
@@ -52,8 +53,9 @@ function CustomDatePicker({ setRange, dispatch, setDateRange }) {
         />
       </div>
       <div className="until">
-        <p style={{ marginBottom: "10px", color: "white" }}>Until: </p>
+        <label htmlFor="datepicker-from">Until: </label>
```
Rperry2174, line 56: should this be `htmlFor="datepicker-until"`?

PR #324, opened 2021-08-09
fix diff tree

olvrng:
This PR fixes `CombineToFlamebearerStruct()` when there are `other` nodes from both side.


petethepig (approved):
Looks good, merging.

PR #327, opened 2021-08-10
Revert "improve tree.Insert"

kolesnikovae:
`v0.0.37` has a problem with memory consumption: some tree nodes are not garbage collected:

image

This is more than 50% of available memory, which causes endless cache evictions (and consequent GC calls) with no results.

(pprof snapshot has been taken for a different time window)
```
(pprof) list wrap  
Total: 18.30GB
ROUTINE ======================== github.com/pyroscope-io/pyroscope/pkg/server.wrapConvertFunction.func1.1 in github.com/pyroscope-io/pyroscope/pkg/server/ingest.go
   16.10GB    16.10GB (flat, cum) 88.00% of Total
         .          .    132:
         .          .    133:func wrapConvertFunction(convertFunc func(r io.Reader, cb func(name []byte, val int)) error) func(io.Reader) (*tree.Tree, error) {
         .          .    134:	return func(r io.Reader) (*tree.Tree, error) {
         .          .    135:		t := tree.New()
         .          .    136:		if err := convertFunc(r, func(k []byte, v int) {
   16.10GB    16.10GB    137:			t.Insert(k, uint64(v))
         .          .    138:		}); err != nil {
         .          .    139:			return nil, err
         .          .    140:		}
         .          .    141:		return t, nil
         .          .    142:	}
(pprof) 
```

The problem is caused by the fact that at insertion we share the same byte buffer (underlying memory) for all labels. This is very wise from the performance respective but prevents GC from freeing such tightly coupled objects.

I've reverted the commit without any additional changes: right now `tree.Insert` does not seem to be the perfect candidate for optimisations.

### DIFF
--- a/pkg/storage/tree/tree.go
+++ b/pkg/storage/tree/tree.go
@@ -136,25 +136,20 @@ func (n *treeNode) insert(targetLabel []byte) *treeNode {
 	return n.ChildrenNodes[i]
 }
 
-func (n *treeNode) insertChild(label []byte, value uint64) *treeNode {
-	n.Total += value
-	childNode := n.insert(label)
-	return childNode
-}
-
 func (t *Tree) Insert(key []byte, value uint64, _ ...bool) {
-	buf := make([]byte, len(key))
-	copy(buf, key)
+	// TODO: can optimize this, split is not necessary?
+	labels := bytes.Split(key, []byte(";"))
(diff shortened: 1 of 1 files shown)


olvrng:
Thanks for your fix. I need to learn more from this.

PR #328, opened 2021-08-10
Fix CPU check

kolesnikovae:
On Linux, if there are no file descriptors available, the call returns an empty slice, which results in server crash.


petethepig (approved):
Thanks for fixing this.

PR #338, opened 2021-08-16
Add a POST handler for `render-diff`

### LINKED ISSUE #321: Separate 'comparison' endpoint from 'diff' endpoint
In #289 we added Diff rendering to be able to show the diff between two flamegraphs. Currently this diff is consolidated into one endpoint. This will become harder to maintain over time. See this comment for more information:

https://github.com/pyroscope-io/pyroscope/pull/289#discussion_r675988905

> It may be way more convenient to introduce a new endpoint. Let's say, render-diff (or query-diff), that would accept POST with the following payload:
> 
> type RenderDiffParams struct {
> 	Format      string
> 	MaxNodes    int
> 	Left, Right RenderTreeParams
> }
> 
> type RenderTreeParams struct {
> 	From  string
> 	To    string
> 	Query *string
> 	Name  *string
> }

yashrsharma44:
Added a POST handler for `/render-diff`
Fixes #321 
## Verification
Verified the endpoint using Postman - 

* Request Body
```json
{
	"query":"pyroscope.server.alloc_objects{}",
	"from":"now-1h",
	"until":"now",
	"max-nodes":1024,
	"left-params" : {
		"from":"1629120696",
		"until":"1629121946"
	},
	"right-params" : {
		"from":"now-30m",
		"until":"now"
	},
	"format":"json"
}
```

TODO - 
- [x] Tests for POST method


pkg/server/render.go
```diff
@@ -59,19 +61,53 @@ func (ctrl *Controller) renderHandler(w http.ResponseWriter, r *http.Request) {
...
+		leftStartTime time.Time
+		leftEndTime   time.Time
+		rghtStartTime time.Time
+		rghtEndTime   time.Time
+
+		leftOK bool
+		rghtOK bool
+	)
+
+	if r.Method == "GET" {
```
kolesnikovae, line 77: What do you think about introducing a function that would take care of diff parameters regardless of the method, something like this:
```
func (ctrl *Controller) renderDiffParametersFromRequest(r *http.Request, p *RenderDiffParams) error
```

pkg/server/render.go
```diff
@@ -198,3 +270,23 @@ func (ctrl *Controller) loadTree(gi *storage.GetInput, startTime, endTime time.T
...
+
+	From  string `json:"from"`
+	Until string `json:"until"`
+
+	Format   string `json:"format"`
+	MaxNodes *int   `json:"max-nodes,omitempty"`
+
+	Left  RenderTreeParams `json:"left-params"`
+	Right RenderTreeParams `json:"right-params"`
+}
```
kolesnikovae, line 287: Taking into account that we already follow [mixed-case naming convention](https://google.github.io/styleguide/jsoncstyleguide.xml?showone=Property_Name_Format#Property_Name_Format) in responses (e.g., `sampleRate`), perhaps we should do so for requests as well.

kolesnikovae (commented):
Looks good! I've left a couple of notes, please take a look :)

pkg/server/render.go
```diff
@@ -59,19 +61,53 @@ func (ctrl *Controller) renderHandler(w http.ResponseWriter, r *http.Request) {
...
+		}
+
+		leftStartTime, leftEndTime, leftOK = parseRenderRangeParams(r.URL.Query(), "leftFrom", "leftUntil")
+		rghtStartTime, rghtEndTime, rghtOK = parseRenderRangeParams(r.URL.Query(), "rightFrom", "rightUntil")
+		if !leftOK || !rghtOK {
+			ctrl.writeInvalidParameterError(w, errTimeParamsAreRequired)
+			return
+		}
+
+	} else if r.Method == "POST" {
```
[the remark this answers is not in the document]
yashrsharma44: Sure, let me take a note

pkg/server/render.go
yashrsharma44: Sure, I have shifted to camel case.

pkg/server/render.go
yashrsharma44: I think since we would be deprecating the GET method, I would like to keep the request struct for GET and POST different, so I am using different methods.
Let me know if this makes sense 😄

kolesnikovae: Huh, that's right :) Let's leave it as it is

petethepig (approved):
Looks good, merging

PR #340, opened 2021-08-19
switch pkger to goembed

ayeniblessing101:
_This PR switch from pkger to go:embed_


webapp/webapp.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"embed"
+	"io/fs"
+	"net/http"
+)
+
+// Asterisk here is important, see https://github.com/golang/go/issues/42328 for more details
+//go:embed public/*
+var assets embed.FS
```
kolesnikovae, line 11: Just out of curiosity – why do we need to include all files (including "hidden" ones)? I may only assume that this is for backward compatibility with `pkger` but not sure.

Taking into account that there is `webapp/public/.gitkeep` and `public/*` glob pattern, embedding will never fail regardless of the actual presence of assets.

petethepig: I added a comment there, but basically:
* hidden files (e.g .gitkeep) don't count
* so when you try to run this with no assets in `public/assets` it breaks (fails to compile, complains that there's no files to embed)
* if you add asterisk it treats it does include hidden files, so .gitkeep now works

I think I'll also add this package under a tag so that it's only included in release builds..

petethepig:
I think I'll also add this package under a tag so that it's only included in release builds..

PR #347, opened 2021-08-23
Http and info metrics

eh-am:
* tracking http metrics (request duration, size of response)
* added `pyroscope_build_info` to expose version as a metric, similar to https://www.robustperception.io/exposing-the-software-version-to-prometheus
* added latency/error rate/throughput (per endpoint) pane in the dashboard

screencapture-localhost-3000-d-tsWRL6ReZQkirFirmyvnWX1akHXJeHT8I8emjGJo-pyroscope-server-2021-08-23-18_06_44


pkg/util/debug/reporter.go
```diff
@@ -34,6 +36,21 @@ type Reporter struct {
...
+	promauto.With(reg).NewGaugeFunc(
+		prometheus.GaugeOpts{
+			Name: "pyroscope_build_info",
+			Help: fmt.Sprintf(
+				"A metric with a constant '1' value labeled by version and git hash from which %s was built.",
+				"pyroscope",
+			),
+			ConstLabels: prometheus.Labels{
+				"version": build.Version,
+				"gitSHA":  build.GitSHA,
```
petethepig, line 48: love it! Can we add all the other things from `build`.

eh-am:
exposing the following fields
```
pyroscope_build_info{GOARCH="amd64",GOOS="linux",revision="fb579a0e",time="2021-08-24T15:01:13Z",use_embedded_assets="false",version="v0.0.37"} 1
```

pyroscope_build_info

PR #349, opened 2021-08-24
Benchmark dashboard

eh-am:
Creates a benchmark dashboard based on the user facing one.
The only difference so far is an additional row at the top with benchmark metrics:
Screenshot from 2021-08-24 18-43-08

Plus updates `examples/grafana-integration`, although it's not currently functional, since it refers to the `pyroscope:latest` docker image, which doesn't contain the new metric names yet.

### DIFF
--- a/benchmark/Dockerfile
+++ b/benchmark/Dockerfile
@@ -3,7 +3,7 @@ FROM golang:1.15.6
 WORKDIR /go/src/app
 
 COPY go.mod go.sum ./
-RUN go get -d ./
+RUN go mod download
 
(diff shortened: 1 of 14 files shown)


petethepig:
Looks good, I have 3 comments:

* Could you make the run progress be a gauge?
Screen Shot 2021-08-24 at 3 29 39 PM
* what is `pyroscope_benchmark` metric? Second panel in the first row
* the first panel in the first raw was my hacky attempt at implementing the table you have in the second row. If it's not too hard it would be nice to just add all those things from the text panel to the table.

PR #362, opened 2021-08-31
reduces bundle size

petethepig:
Before: 546.1 KB
After: 471.44 KB

^ these are gzipped production numbers from webpack bundle analyzer

This is what it currently looks like:
Screen Shot 2021-08-31 at 1 27 26 AM

### DIFF
--- a/webapp/javascript/redux/store.js
+++ b/webapp/javascript/redux/store.js
@@ -1,5 +1,4 @@
 import thunkMiddleware from "redux-thunk";
-import promiseMiddleware from "redux-promise";
 
 import { createStore, applyMiddleware } from "redux";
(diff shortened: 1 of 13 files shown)


webapp/javascript/redux/store.js
```diff
@@ -23,7 +22,7 @@ import {
 import { parseLabels, encodeLabels } from "../util/key";
 
 const enhancer = composeWithDevTools(
-  applyMiddleware(thunkMiddleware, promiseMiddleware)
+  applyMiddleware(thunkMiddleware)
```
petethepig, line 25: looks like we're not using promiseMiddleware

PR #363, opened 2021-08-31
Refactor CLI

kolesnikovae:
- Minor changes aimed to improve readability and maintainability.
- Removed unused and redundant code.
- Added some tests.


cmd/pyroscope/command/command.go
```diff
@@ -0,0 +1,70 @@
...
+				return fmt.Errorf("loading configuration file: %w", err)
+			}
+		}
+		if err = cli.Unmarshal(vpr, cfg); err != nil {
+			return err
+		}
+		if err = fn(cmd, args); err != nil {
+			cmd.SilenceUsage = true
+		}
+		return err
```
yashrsharma44, line 35:
```suggestion
		return nil
```
should we return an unintialized variable?

pkg/config/config.go
```diff
@@ -28,25 +43,23 @@ type Agent struct {
...
+	Targets []Target `yaml:"targets" desc:"list of targets to be profiled" mapstructure:"-"`
 
 	// Note that in YAML the key is 'tags' but the flag is 'tag'.
-	Tags map[string]string `yaml:"tags" name:"tag" def:"" desc:"tag key value pairs" mapstructure:"tags"`
+	Tags map[string]string `yaml:"tags" name:"tag" def:"" desc:"tag key value pairs" mapstructure:"-"`
 }
 
 type Target struct {
-	ServiceName string `yaml:"service-name" desc:"name of the system service to be profiled"`
+	ServiceName string `yaml:"service-name" mapstructure:"service-name" desc:"name of the system service to be profiled"`
```
yashrsharma44, line 53: Didn't knew we could pass in multiple tags of yaml 💯

yashrsharma44 (commented):
LGTM!

cmd/pyroscope/command/command.go
kolesnikovae: Not really :) The check above is just to suppress "usage" message if an error is not related to CLI stuff (e.g. invalid flag/missing args, etc.), and should not affect error handling flow: the error itself (a nil, or not - does not matter) will be handled by the caller.

Alternatively, it could be put as follows:
```go
		if err = fn(cmd, args); err != nil {
			cmd.SilenceUsage = true
			return err
		}
		return nil
```

I don't have any preferences on this. If you find this option more appropriate, make a suggestion - I'll accept it :)

yashrsharma44: Oh great, this makes sense.

petethepig (approved):

petethepig:
Thank you @kolesnikovae !

PR #366, opened 2021-08-31
249 server benchmark

eh-am:
this is just prep work for 249, but I thought it would be a good idea to have a pr early on

this is strictly for the case where we want to run 2 instances at the same time (a reference one, probably `:main` and the PR, but technically could be anything) and compare its metrics (thinking only `throughput` for now), ie. **there's nothing regarding agent benchmarking yet**

# the cli
* started creating a cli (named `pyrobench`) to help run the tests and whatnot (worst case scenario it becomes a kitchen sink for helper code)
* why another cli instead of having as a hidden subcommand (as we do with dbmanager)? honestly there's no big argument, just to have it more scoped
* tried to reuse as much as possible from our existing cli, due to good timing managed to get to use cobra as well.
* unfortunately, there's still some code that was copy/pasted (eg common.go) and already got out of sync (for example, I missed the refactor from https://github.com/pyroscope-io/pyroscope/pull/363)
* copied the existing /benchmark/main.go code  there (env var ergonomics are not the best), command is called `loadgen` (load generator)

# plan
(stuff that's relatively clear in my mind)
* the plan is to run 2 instances `:main` and whatever is in the pr at the same time in docker (using docker-compose)
* run `pyrobench loadgen` with exact same parameters against both instances for TBD amount of time
* generate a dashboard that compares both instances, take a screenshot of the throughput pane and post in the PR
* use pushgateway to push loadgen metrics? (this was previously in `main.go`, but it required the load generator to be a server that was scraped by prometheus)

# plan part 2
(still cloudy)
* calculate the throughput (median? average?) using a prometheus api client (https://pkg.go.dev/github.com/prometheus/client_golang/api/prometheus/v1) and also post in the PR
* be scrapped by a long-storage prometheus (the one in the k8s cluster)?
* if that's the case, why not make the dashboard public?
* for how long to run these load generator? 5 minutes? 15? 30? 1 hour? maybe having different "checks" in the PR for each of the deadlines?
* should we run it multiple times and do some more refined statistical analysis? 

I've kept `main.go` since the new code is not fully in par with yet.


benchmark/cmd/command/banner.go
```diff
@@ -0,0 +1,58 @@
+// Copied from github.com/pyroscope-io/pyroscope/cmd/command/banner.go
```
petethepig, line 1: This code is in the same module so we should be able to reuse it relatively easily. Maybe make these functions public?

petethepig:
The code looks good, the only thing I thought we could improve is those two copied files, I commented above.

Regarding the plan part 1, everything makes sense except for running two copies of server. I wonder if we *have to* run both at the same time. I feel like that would complicate the setup quite a bit without providing much of a benefit.

I'm thinking for our internal benchmarking we'll always have results from `main` branch to compare the PR results to. For users, comparing main and something else is even less relevant, they would probably just want to run one copy anyway.

So yeah, I think for the server I would keep just 1 instance for simplicity.

It's different for agent benchmarking where the difference in performance between `pyroscope enabled` and `pyroscope disabled` modes is the important part. I don't think we can get away with running just 1 instance there.

---

Regarding the plan part 2, everything looks good, to answer the questions:

> be scrapped by a long-storage prometheus (the one in the k8s cluster)?

yes, that sounds good. we might have to push them in which case we'll need to set up push-gateway, not sure, /cc @kolesnikovae 

> if that's the case, why not make the dashboard public?

Yeah, I think it should be public. We probably need a separate grafana instance for that? /cc @kolesnikovae 

> for how long to run these load generator? 5 minutes? 15? 30? 1 hour? maybe having different "checks" in the PR for each of the deadlines?

I'm thinking 5 minutes is good, I feel like when it comes to CI it is important to have these things run quickly. But there should be a way to override it when you run it locally.

> should we run it multiple times and do some more refined statistical analysis?

This is maybe v3. This would depend a lot on how much variance we get between runs.

benchmark/cmd/command/banner.go
eh-am: maybe i misunderstood you, but it's two different modules (`pyroscope-io/pyroscope/benchmark/cmd/command`, and `pyroscope-io/pyroscope/cmd/command`)
correct me if i am wrong, but i think modules under `cmd` are not exported publicly?

so what we can do is to just move that gradient code to `pyroscope-io/pyroscope/pkg/cli`, doing it now.

eh-am:
> Regarding the plan part 1, everything makes sense except for running two copies of server. I wonder if we have to run both at the same time. I feel like that would complicate the setup quite a bit without providing much of a benefit.

Hmm why complicated? Just add another server to the docker-compose.

The advantage is to have easily generated panels such as
Screenshot from 2021-09-01 11-12-35


> I'm thinking for our internal benchmarking we'll always have results from main branch to compare the PR results to. For users, comparing main and something else is even less relevant, they would probably just want to run one copy anyway.

Valid point, my answer is simply that those are 2 different use cases (PR comparison against main in CI and user local benchmarking) which should be treated differently (ie 2 docker-composes and whatnot).

eh-am:
> Regarding the plan part 1, everything makes sense except for running two copies of server. I wonder if we have to run both at the same time. I feel like that would complicate the setup quite a bit without providing much of a benefit.

Hmm why complicated? Just add another server to the docker-compose.

The advantage is to have easily generated panels such as
Screenshot from 2021-09-01 11-12-35


> I'm thinking for our internal benchmarking we'll always have results from main branch to compare the PR results to. For users, comparing main and something else is even less relevant, they would probably just want to run one copy anyway.

Valid point, my answer is simply that those are 2 different use cases (PR comparison against main in CI and user local benchmarking) which should be treated differently (ie 2 docker-composes and whatnot).

> yes, that sounds good. we might have to push them in which case we'll need to set up push-gateway, not sure, /cc @kolesnikovae

Yeah that's one of the cases running within the cluster makes everything easier.  As per my comment right above, maybe we should just assume there's different cases (local user driven, CI driven) and have separate workflows for them. Docker-compose when running locally and k8s when running in CI. Plus running in k8s makes some of the things somewhat easier (provisioning a brand new node just to run pyroscope/clients for example, to try to minimize as much interference as possible)

Another thing is, do users care much about reproducing server benchmarking? In my mind they are most interested in the agent overhead.

eh-am:
(looking into why cypress is failing)

petethepig (approved):
LGTM, feel free to merge if you're done with it @eh-am

PR #380, opened 2021-09-05
cypress: cleanup nasty waitInDevMode

eh-am:
Initially the cypress tests where flaky, as they were failing due the `.then` callbacks of `fetch` trying to access an unmounted component (think `setState`).
I added a nasty `waitInDevMode` to avoid that.
This PR cleans it up.

Since I was at it (ie. these weren't necessarily affecting the tests), I decided to add code to abort other requests too.


petethepig (approved):
Looks good to me. Thanks for fixing this @eh-am !

Issue #342, opened 2021-08-20
Serve assets gzipped


petethepig:
I think pyroscope should serve assets gzipped

gawicks:
Hi it's @petethepig it's @gawicks from slack. Please assign me to this issue.

petethepig:
Done. Thanks for the help, it's much appreciated!

gawicks:
@petethepig Looks like there's some housekeeping left. Could you link PR #381 to this issue and close 😊

petethepig:
@gawicks Right, I always forget, thanks for the reminder!

petethepig:
Fixed in #381

PR #381, opened 2021-09-06
serve assets gzipped #342

gawicks:


gawicks:
Setting this to WIP till I've written some tests

gawicks:
@petethepig I'd like some feedback on the use of [klauspost/compress](https://github.com/klauspost/compress) library.

petethepig:
It looks good! I tried running it too and it seems to be working as expected. My original plan was to only enable it for everything in `/assets`, but then I thought about it and I figured this even better actually.

petethepig:
@gawicks Oh, and regarding the library, I do think it makes sense to use it here.

gawicks:
@petethepig Thanks! I've finished adding tests. The PR is ready for review.

petethepig (approved):

gawicks:
Yay!

Issue #356, opened 2021-08-26
Add e2e tests (cypress)


petethepig:
This is a placeholder for e2e tests

PR #386, opened 2021-09-09
tests for app change dropdown (issue #356)

ruslanpascoal2:
https://github.com/pyroscope-io/pyroscope/issues/356

Tests if when selecting an app from the dropdown, an api call to '/render' is being triggered, and if the response matches the data displayed in the table.


Rperry2174:
Thanks a ton! This is a great test to have -- a while back we introduced a bug that this test would have solved. Could you rebase from main?

ruslanpascoal2:
> Thanks a ton! This is a great test to have -- a while back we introduced a bug that this test would have solved. Could you rebase from main?

Done!

petethepig:
Looks great! Thank you so much @ruslanpascoal2 !

PR #389, opened 2021-09-13
fix race conditions that were crashing cypress tests

eh-am:
cypress tests were flaky due to a race condition (eg
https://github.com/pyroscope-io/pyroscope/actions/runs/1219594752)

just added checks to not access an unitialized variable


petethepig (approved):
Looks good!

PR #392, opened 2021-09-14
generate a random prefix for server bench pr docker-compose

eh-am:
the reasoning here is that we may want to run multiple jobs
at the same time with different profiles (different parameters)

that requires a custom prefix, since docker-compose assumes
the name of the directory

also remove the ports binding, which generally aren't needed
unless one wants to run locally, which in that case can be added manually


petethepig (approved):
Makes a lot of sense, looks good!

petethepig:
@eh-am I don't know if it's fully finished, feel free to merge when you want to

PR #395, opened 2021-09-15
support taking screenshot of dashboard with rows

eh-am:
previously we assumed the dashboard we are taking screenshots doesn't have any rows

that, however, makes generating it with jsonnet a bit difficult (see the `gridPos` math below)

in this pr we support dashboards with rows, which then allow us to just add panels to a row and they will be placed correctly


petethepig (approved):
Looks good. Feel free to merge if / when it's done

PR #396, opened 2021-09-15
benchmark: add cmd to generate a meta report

eh-am:
as previously discussed in slack, added a command to generate a `meta` (using meta here since table is already taken) to post metadata

this will make it easier to visualize different executions with different profiles

since there's a concern on revealing sensitive data, we decided to add an allowlist
which currently is somewhat silly, since the code that generates the allowlist and the code that uses the allowlist is in the same codebase
but it will make sense, where in the future we distribute the cli separately from the code

currently allowing the following fields
```
"BENCH_RUN_FOR",
"PYROBENCH_RAND_SEED",
"PYROBENCH_PROFILE_WIDTH",
"PYROBENCH_PROFILE_DEPTH",
"PYROBENCH_PROFILE_SYMBOL_LENGTH",
"PYROBENCH_APPS",
"PYROBENCH_CLIENTS",
"PYROBENCH_REQUESTS",
```


petethepig (approved):
@eh-am Looks good, feel free to merge if it's ready.

PR #413, opened 2021-09-22
Fix search bar

### LINKED ISSUE #409: Search bar is not working
Not sure where we introduced this, but currently when I search the search bar, it no longer highlights the node with the searched phrase:

Before:
image

Now:
image

- [x] Fix the bug
- [ ] Add tests to make sure it works for single view
- [ ] Add tests to make sure it works for comparison view 
- [ ] Add tests to make sure it works for diff

eh-am:
Fixes #409 

Ended up adding a debounce to the search to make it faster (otherwise on every keystroke we would rerender the flamegraph)

No tests yet since they require visual testing, @ruslanpascoal2 will look into it later.

I think we should merge this as soon as possible and worry about the tests later.

BTW reset view is also broken, working on it in a different PR.

### DIFF
--- a/cypress/integration/basic.ts
+++ b/cypress/integration/basic.ts
@@ -59,6 +59,13 @@ describe('basic test', () => {
   });
 
   it('view buttons should change view when clicked', () => {
+    // mock data since the first preselected application
+    // could have no data
+    cy.intercept('**/render*', {
+      fixture: 'render.json',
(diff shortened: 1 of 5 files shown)


Rperry2174:
@eh-am did this fix it for you? It still looks like its not working for mine. Am I missing something? 
image

PR #414, opened 2021-09-22
Fix reset view

eh-am:
The "Reset View" button wasn't working properly. This PR fixes it

image

Also added a simple test that clicks on the flamegraph, so maybe we could move that to a bigger test that tests the drill effect and whatnot.

@ruslanpascoal2 I've created a new fixture based on real profiling data (the go example), I think it's best to use this one from now on, since it makes testing easier since you have real, consistent data to base on (for example the previous `render.json` file produces a broken flamegraph).

### DIFF
--- a/cypress/fixtures/simple-golang-app-cpu.json
+++ b/cypress/fixtures/simple-golang-app-cpu.json
@@ -1,0 +1,74 @@
+{
+  "flamebearer": {
+    "names": [
+      "total",
+      "runtime.main",
+      "main.slowFunction",
+      "main.work",
+      "main.main",
+      "main.fastFunction"
+    ],
+    "levels": [
+      [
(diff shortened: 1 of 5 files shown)


ruslanpascoal2:
Ok, thanks. > The "Reset View" button wasn't working properly. This PR fixes it
> 
> 
> 
> image
> 
> 
> 
> Also added a simple test that clicks on the flamegraph, so maybe we could move that to a bigger test that tests the drill effect and whatnot.
> 
> 
> 
> @ruslanpascoal2 Created a new fixture based on real profiling data (the go example), I think it's best to use this one from now on, since it makes testing easier since you have real, consistent data to base on (for example the `render.json` file produces a broken flamegraph).
> 
>

PR #417, opened 2021-09-23
frontend: apply prettier to cypress/webapp

eh-am:
Ran `eslint` and `prettier` against ALL files. That is so that everything is consistent (from js files to yamls)
This ends up breaking `git blame` (since I modified all files).
To solve that added a `postinstall` script that runs  `git config blame.ignoreRevsFile .git-blame-ignore-revs`, which makes `git blame` ignore commits from the `.git-blame-ignore-revs` file.
Therefore, we require existing code to run `yarn` again.

Plus added `husky` and `lint-staged` integration. Which makes us run `eslint` and `prettier` to staged files when committing. That will guarantee all new modifications are formatted.

Also added a ci check to validate files have been formatted (let's say the husky hook didn't work for some reason).


cypress/integration/basic.ts
```diff
@@ -1,13 +1,13 @@
-import { BAR_HEIGHT } from '../../webapp/javascript/components/FlameGraph/FlameGraphComponent/index.jsx'
+import { BAR_HEIGHT } from '../../webapp/javascript/components/FlameGraph/FlameGraphComponent';
```
Rperry2174, line 1: Does this cause a problem or is the `/index.jsx` not needed?

eh-am: That's an `eslint` error
```
Unexpected use of file extension "jsx" for "../../webapp/javascript/components/FlameGraph/FlameGraphComponent/index.jsx"  import/extensions
```

We weren't catching it before because our lint rule was only applied to `js/jsx` files (this one is a `ts` file).

But overall it's not needed, AFAIK all tools follow convention that `foo/index.{js,jsx,ts,tsx}` equal to `/foo`

eh-am:
retest this please

petethepig (approved):
@eh-am Thank you for doing this, this is great!

PR #419, opened 2021-09-23
add basic visual testing test

eh-am:
* Add the [jaredpalmer/cypress-image-snapshot](https://github.com/jaredpalmer/cypress-image-snapshot) for visual testing.
* Create a command `yarn cy:ss` that spawns a docker container to update the snapshots (to reduce test flakiness).
* Update the `app selection` test to do visual testing. Let's keep it an experiment to see how flaky it is.
* Add more fixtures based on real data.


cypress/README.md
```diff
@@ -15,3 +15,45 @@ Or run `yarn cy:ci` to run all tests.
...
+To just run without updating the snapshots, run
+```
+CYPRESS_updateSnapshots=false yarn cy:ss
+```
+
+## Why
+Part of our core functionality revolves around rendering a flamegraph **canvas**,
+which is not straightforward to test.
+
+We've decided to test it using [visual testing](https://docs.cypress.io/guides/tooling/visual-testing#Functional-vs-visual-testing).
```
Rperry2174, line 39: So am I correct in understanding you set a threshold somewhere of how different the pictures can be before failing the test?

Rperry2174 (approved):
LGTM! I also am slightly worried about these being more flaky than helpful, but I think this is a good test anyway and if its too flaky we can always remove! Thanks for adding :)

cypress/README.md
eh-am: That is correct. However, I didn't set anything, so it's using the plugin's default, [which is 0](https://github.com/jaredpalmer/cypress-image-snapshot/blob/0510252bc5f192c410bcca45b511f07737bce4a2/src/plugin.js#L61).

PR #425, opened 2021-09-29
Fix intersection

kolesnikovae:
Although the new implementation is pretty expensive, it should not be an issue given that it is still in order of microseconds (and will never be the main factor of querying performance). I also think we should migrate to bitmaps for better set operations.


petethepig (approved):
Thanks for fixing this Anton!

PR #423, opened 2021-09-29
docs: update the translation to keep up with the original

Cluas:
I noticed that many of the Chinese translations were lagging behind, so I submitted this PR to update them.

PR #426, opened 2021-09-29
Update readme

Rperry2174:
We made some improvements to pyroscope and the readme needed some updating to match.


Rperry2174:
After seeing #423 I think we should make more of an effort to keep translations up-to-date

PR #427, opened 2021-09-30
Make go lint fail build

eh-am:
* Cleaned up the straightforward problems and annotated the non-trivial ones.
* Consider all warnings as errors and fail CI if any of them happen

Suggestions are welcome.


benchmark/internal/cireport/upload.go
```diff
@@ -38,22 +33,24 @@ func (rs *fs) WriteFile(dest string, data []byte) (string, error) {
 	// not really useful since we don't have the full path
 	if _, err := os.Stat(dest); err == nil {
 		// file exists
+		logrus.Debugf("file %s exists\n", dest)
 	} else if os.IsNotExist(err) {
 		// file does not exist
+		logrus.Debugf("file %s does not exist\n", dest)
```
kolesnikovae, line 39: It seems I'm missing something. Why do we need to check the file it has been successfully written?

pkg/agent/session.go
```diff
@@ -169,6 +169,7 @@ func mergeTagsWithAppName(appName string, tags map[string]string) (string, error
 	return k.Normalized(), nil
 }
 
+// revive:disable-next-line:cognitive-complexity complexity is fine
```
kolesnikovae, line 172: Agree, although we could move `Snapshot` callback to a separate function.

pkg/cli/command.go
```diff
@@ -136,7 +136,7 @@ func NewViper(prefix string) *viper.Viper {
 	return v
 }
 
-func loadConfigFile(path string, cmd *cobra.Command, vpr *viper.Viper, v interface{}) error {
+func loadConfigFile(path string, cmd *cobra.Command, vpr *viper.Viper, _ interface{}) error {
```
kolesnikovae, line 139: Let's just remove the unused argument

pkg/flameql/flameql.go
```diff
@@ -21,6 +21,7 @@ type TagMatcher struct {
 
 type Op int
 
+//revive:disable:var-naming Those names mimics constants names
```
kolesnikovae, line 24: Actually, I used this notation for no good reason. Perhaps, it would be better to refactor names (e.g:
`OpEqual`, `OpEqualRegex`, etc),

kolesnikovae (commented):
Great work!

benchmark/internal/cireport/upload.go
eh-am: Good point, this is left out code. Removing.

pkg/storage/query.go
```diff
@@ -21,23 +21,23 @@ func (s *Storage) exec(_ context.Context, qry *flameql.Query) []dimension.Key {
 
 	for _, m := range qry.Matchers {
 		switch m.Op {
-		case flameql.EQL:
+		case flameql.OpEqual:
```
petethepig, line 24: :+1:

petethepig (approved):
Thank you so much @eh-am for doing all of this! It looks great, merging.

PR #429, opened 2021-09-30
Update Ruby Example with new Ruby gem

Rperry2174:
This is a more practical example of how one could use Pyroscope's ruby gem. Would love to get some feedback on the README and if the example makes sense


examples/ruby/README.md
```diff
@@ -0,0 +1,89 @@
...
+![ruby_example_architecture_05](https://user-images.githubusercontent.com/23323466/135726784-0c367d3f-c9e5-4e3f-91be-761d4d6d21b1.gif)
+
+Note: For documentation on the Pyroscope ruby gem visit [our website](https://pyroscope.io/docs/ruby/)
+## Backround
+In this example we show a simplified, basic use case of Pyroscope. We simulate a "ride share" company which has three endpoints found in `server.rb`:
+- `/bike`    : calls the `order_bike(search_radius)` function to order a bike
+- `/car`     : calls the `order_car(search_radius)` function to order a car
+- `/scooter` : calls the `order_scooter(search_radius)` function to order a scooter
+
+We also simulate running 3 distinct servers in 3 different regions (via [docker-compose.yml](https://github.com/pyroscope-io/pyroscope/blob/update-ruby-example/examples/ruby/docker-compose.yml))
```
Rperry2174, line 11: will need to update this once merged

Rperry2174:
Hey @eh-am I wonder what you think about this... I noticed that for Ruby there is often very tall stack traces... I'm wondering if there is a good UI where: When a user selects a "node" we could **_ether_** dim _or collapse_ the stacktrace above the selected node... 

I supposed this might get confusing, but maybe with some good UI/UX we could find a way to do that:
Screen Shot 2021-09-30 at 1 17 43 PM (2)

Any ideas?

PR #431, opened 2021-10-01
Segment level-based retention

kolesnikovae:
The PR introduces a new feature: segment level-based retention policies. This also may be treated as a downsampling mechanism.

The feature will provide users an ability to manage data retention based on the segment node level. For example, default policies:
- level 0 nodes (every node represents 10s) are stored 7d
- level 1 nodes (100s) are stored 30d
- level 2 nodes (1000s) – 180d

A demo pyroscope server with historical data: https://data-retention-testing.pyroscope.cloud (uses policies listed above). After some testing on the real data, I've come to the conclusion that in the vast majority of cases, setting RP for level 2 and higher doesn't make sense: it doesn't free up a lot of disk space and doesn't significantly reduce the number of objects on the heap.

TODO:
 - [x] Add metrics, update dashboards
 - [x] Configuration UX
 - [x] [Document how, why and when to use](https://github.com/pyroscope-io/docs/pull/1)

---

There are some points I'd like to clarify:
1. The timeline diagram may look broken:

image

I think we need to somehow highlight downsampled periods. For example, make bar width dependent on the number of samples in the range. At least, me may put some hints in the timeline itself:
image 

Backend provides with so-called _watermarks_ in the `timeline` member of the render request response. These watermarks describe periods precision.

```go
// Watermarks map contains down-sampling watermarks (Unix timestamps)
// describing resolution levels of the timeline.
//
// Resolution in seconds is calculated as 10^k, where k is a map key.
// Meaning that any range within these 10^k seconds contains not more
// than one sample. Any sub-range less than 10^k shows down-sampled data.
//
// Given the map:
//  1: 1635508310
//  2: 1635507500
//  3: 1635506200
//
// This should be read as follows:
//  1. Data after 1635508310 is as precise as possible (10s resolution),
//     down-sampling was not applied.
//  2. Data before 1635508310 has resolution 100s
//  3. Data before 1635507500 has resolution 1000s
//  4. Data before 1635506200 has resolution 10000s
//
Watermarks map[int]int64 `json:"watermarks"`
```

---

2. Tree downsampling may "truncate" (and omit completely) a tree if there are too few samples. Example:
image

Given a tree with 100 samples total representing 100s time range. A sampled tree for 10s within this 100s will only have 10 samples and most likely some nodes will be dropped (0 values are not preserved). Basically, this is totally fine for downsampling but may be confusing.

There must be a way to avoid this (or I overlooked something). @petethepig what do you think?

---

3. Configuration options.

Despite the fact that the storage design is documented, it may be not apparent what values should be set in a particular case, and the configuration itself may be not obvious. I came up with the following:

```yaml
retention: 365d # not required, here for consistency.
retention-levels:
 10s: 7d
 100s: 30d
 1000s: 180d
```
(can't decide which is more appropriate: `downsampling`, `downsampling-rules`, `retention-policies`)

CLI:
```shell
> pyroscope server --retention-levels.10s 7d --retention-levels.100s 30d --retention-levels.1000s 180d
```

---

Update: I came to the following:

```yaml
retention-levels:
  0: 24h
  1: 720h # 30 days
  2: 4320h # 180 days
```

CLI:
```shell
> pyroscope server --retention-levels.0 24h --retention-levels.1 720h
```

Env vars are also supported.

Only levels 0-2 are eligible for configuration (we can extend the list in future). Due to the fact that nodes above L2 constitute less than 0.01% of the segment, and represent relatively long periods (> ~2h40m), the effect of setting retention policy for those levels is marginally low. Users to be advised to only configure levels 0 and 1.


eh-am:
still digesting, but couple questions:
1 - 10s, 100s vs "level 1" x "level 2" and so on. would it be better to expose this "level" abstraction instead? so that we can tweak what these levels actually represent
2 - that being said, is it a good idea to expose that of a low level to end users? maybe we could expose "retention" profiles? "aggressive" | "conservative" | "normal" etc, so that in the feature we can tweak these values (for any reason) 


---
my reasoning here is that as a user that want to experiment with pyroscope (not a sre/ops kinda person, but a dev that just "installed it"), I may not have interest on its inner workings, I just want to start with something that works ™️ . Then maybe after some time i notice that it's using too much disk! I don't care tooooo much about old data, since I only use pyroscope from time to time. 

"Is there a flag to reduce that disk usage? "

"there is! it's `downsampling-policy: aggressive`"
vs 
"there is! it's"
```
retention-levels:
 10s: TBD
 100s: TBD
 1000s: TBD
```
"hmm, what values should I use? no idea..."

tl;dr [make me make fewer decisions](https://en.wikipedia.org/wiki/Decision_fatigue)

kolesnikovae:
Along with level-based retention policy, I took an attempt to implement size-based retention. It turned out, it will require significant changes in BadgerDB.

The approach is very simple – to remove oldest trees calculating the volume of the removed items until the requested space is reclaimed. It is essential to have accurate size estimations of the removed items, the estimation must precisely match occupied disk space.

I faced the following problems:

1. BadgerDB garbage collector reclaims disk space eventually (this is not a problem per se but an aspect of LSM implementation in Badger), making it not possible estimating how much space a particular item occupies on the disk. A naive attempt to _fix_ GC (basically, get rid of many and many heuristics) only slightly improved results. I suspect that SS tables compaction should be also adjusted (an explicit `Flatten` call is not enough).
It should be noted that BadgerDB v3 has many improvements in GC area and consumes 2-3 times less disk space. Unfortunately, with BadgerDB v3 server crashes with OOM after short period of time (under standard PR review app load). Perhaps, we should migrate to v3 first; although, this may be complicated by the fact that it is not backward compatible.

2. There are 5 databases, 5 independent GC jobs. This fact biases size estimations badly and causes storage to remove way more trees than actually has to. We should unite all databases.

The PR contains the code to remove trees from the segment based on the size specified. When we figure out how to make BadgerDB to reclaim disk space on-demand, we can revisit size-based retention feature.

petethepig (approved):
Everything looks except for a couple of extra functions :)

Issue #408, opened 2021-09-22
Switch from Samples to Percentages for Diff view


Rperry2174:
in #289 we added diff rendering.

Right now when we look at a diff like [this one](https://demo-dev.pyroscope.io/comparison?query=cart.service.dotnet.cpu%7B%7D&from=1632254386&leftFrom=1632257992&leftUntil=1632261549&rightFrom=1632261595&rightUntil=1632265226&until=1632268786):
![image](https://user-images.githubusercontent.com/23323466/134264153-300281b9-1a3e-4dc7-9857-f8f2a380443a.png)

Here we can see that the difference between these two is clearly the `GetCartAsync` function. So we would assume that going directly to the comparison view should show this clear difference between the two flamegraphs. However, I believe we are using _absolute difference in number of samples_ as opposed to using _absolute difference in percentage_. 

So rather than comparing 30% vs 60% (a difference of +100%) and showing the diff that way, we are instead comparing 36,423 samples vs 296,720 samples (a difference of +714%) which doesn't give as useful of a flamegraph since the increase in cpu utilization for the right side will _always_ have more than 100% more samples.

![image](https://user-images.githubusercontent.com/23323466/134264586-ebf9fd35-bb20-4a87-a064-6dcb92944010.png)

We should definitely switch this to use percentages instead of absolute number of samples. Let us know if you have any thoughts on how this can be implemented @olvrng...

Rperry2174:
closed in #432

PR #432, opened 2021-10-01
408 switch from samples to percentages for diff view

eh-am:
Addresses #408


pkg/server/render.go
```diff
@@ -120,10 +120,39 @@ func (ctrl *Controller) renderDiffHandler(w http.ResponseWriter, r *http.Request
...
+		"flamebearer": merged,
+		"metadata": map[string]interface{}{
+			"format":     fs.Format, // "single" | "double"
+			"spyName":    out.SpyName,
+			"sampleRate": out.SampleRate,
+			"units":      out.Units,
+		},
+	}
+
+	//	res := renderResponse(merged, out)
```
kolesnikovae, line 146: Oh, we definitely have some tech dept here :) In fact `Flamebearer` type should be refactored to keep only related info:

```go
type Flamebearer struct {
	Names    []string `json:"names"`
	Levels   [][]int  `json:"levels"`
	NumTicks int      `json:"numTicks"`
	MaxSelf  int      `json:"maxSelf"`
	// TODO: see note in render.go
	// These fields are not related to flamebearer and should be removed.
	// Instead, the fields should be put to the response struct.
	SpyName    string `json:"spyName"`
	SampleRate uint32 `json:"sampleRate"`
	Units      string `json:"units"`
	Format     Format `json:"format"`
}
```

Thus, I wouldn't "wrap" `Flamebearer` type with two more fields but put them at the response struct root level.

Rperry2174 (approved):
Tested it and couldn't find any issues from frontend. LGTM 🔥 !

PR #436, opened 2021-10-04
[Exec] Don't use waitpid for waiting for a spawned process to exit.

AdrK:
Don't use waitpid for waiting for a process to exit.


pkg/exec/cli.go
```diff
@@ -172,19 +171,13 @@ func Cli(cfg *config.Exec, args []string) error {
 	defer session.Stop()
 
 	if isExec {
-		return waitForSpawnedProcessToExit(c, cmd)
+		<-c
+		_ = sendSignal(cmd.Process, syscall.SIGTERM)
+		return nil
+	} else {
+		waitForProcessToExit(c, pid)
+		return nil
```
kolesnikovae, line 179: As we discussed before, I'd like to split this into two parts:

**1. Freezing in the blocking mode (pyspy, rbspy)**

We presume that `cmd.Wait` call conflicts with a process `wait()` called by the rust library. Experiments showed that w/o `cmd.Wait`, pyroscope does not hang.

The code in PR blocks until a signal is received: when the child process exits, pyroscope does not. I think we can handle this in a better way. I'd suggest using pipes (not tested extensively):

```go
func main() {
	cmd := exec.Command("ping", "-c", "10", "localhost")
	signals := make(chan os.Signal, 10)
	signal.Notify(signals)
	defer close(signals)
	if err := execute(cmd, signals); err != nil {
		panic(err)
	}
}

func execute(cmd *exec.Cmd, signals <-chan os.Signal) error {
	stdout, err := cmd.StdoutPipe()
	if err != nil {
		return err
	}
	defer func() {
		_ = stdout.Close()
	}()

	stderr, err := cmd.StderrPipe()
	if err != nil {
		return err
	}
	defer func() {
		_ = stderr.Close()
	}()

	r := io.MultiReader(stdout, stderr)
	if err = cmd.Start(); err != nil {
		return err
	}

	// Signal relay routine.
	go func() {
		for s := range signals {
			if s != syscall.SIGCHLD {
				_ = cmd.Process.Signal(s)
			}
		}
	}()

	// Blocking call: releases once the process exits.
	if _, err = io.Copy(os.Stdout, r); err != nil {
		return err
	}

	// Just to get the error.
	return cmd.Wait()
}
```

---

**2. Signal relying**

For me it's not entirely clear whether the signal relay interferes with the first issue. Although, I guess it's definitely makes sense to exclude `SIGCHLD` from processing (Child stopped or terminated).

I think that we need to relay at least "common" signals (e.g. SIGUSR1/2, SIGABRT, etc) that are used widely.

AdrK: 1. Thanks for the proposal, it clarifies a lot. I have checked, and it seems to be working as expected, except that the io.Copy() blocks on r.Read(), as Read() is performing the read until EOF. The copying itself is performed after the traced process is being killed, so there is no print on stdout in real time. Do you know of other solution that will allow for real time copying to stdout? Another pipe? TeeReader?

kolesnikovae: Hm, interesting.

> I have checked, and it seems to be working as expected, except that the io.Copy() blocks on r.Read(), as Read() is performing the read until EOF. The copying itself is performed after the traced process is being killed, so there is no print on stdout in real time.

True, `Copy` blocks until the pipe descriptors are closed (and EOF is encountered). To me it sounds like we have buffering problem – basically, `Copy` does buffer data (default size is 32K). I'd try two things:
1. Set smaller buffer size via `CopyBuffer` (4K as this is the size of stdout buffer).
2. For testing, remove `MultiReader` and copy stdout only. Just to localise the problem.

> Do you know of other solution that will allow for real time copying to stdout? Another pipe? TeeReader?

This is the correct way, I believe :)

Issue #433, opened 2021-10-01
Dynamic tags support for go


petethepig:
We recently added support for dynamic tags in Ruby and Python integrations. We need to do the same in our Go app.

Here's an example from our python docs:
```python
# You can use a wrapper:
with pyroscope.tag_wrapper({ "controller": "slow_controller_i_want_to_profile" }):
  slow_code()

# Or you can add / remove tags manually:
pyroscope.tag({ "controller": "slow_controller_i_want_to_profile" })
slow_code()
pyroscope.remove_tags("controller")
```

Some pointers:
* [Profiler class](https://github.com/pyroscope-io/pyroscope/blob/88aaed0bf72ee627cc666bde402e51f1e835a449/pkg/agent/profiler/profiler.go#L84-L85) — this is the the place where we need to add the new methods
* [ProfileSession.SetTag method](https://github.com/pyroscope-io/pyroscope/blob/main/pkg/agent/profiler/profiler.go#L86-L90) this is the method they will be calling

### Proposed API

I'm thinking we could just copy Ruby / Python API for consistency:

```
func (p *Profiler) TagWrapper(map[string]string, cb func()) 
func (p *Profiler) Tag(map[string]string) 
func (p *Profiler) RemoveTags(tags ...string) 
``` 

### TODO

* [ ] add the API
* [ ] add tests

deepto98:
@petethepig  Hi, I'd like to work on this issue, can you please assign this to me?

petethepig:
Hi @deepto98 ,

Actually, @awwalker said he's interested in working on this one too. @awwalker can you let us know if you're still planning to work on this so that we can plan for our roadmap.

Otherwise I'll assign to @deepto98

awwalker:
Hey @petethepig and @deepto98 yeah I’m still committed to taking this on. I have a pretty vested interest in getting this feature in ASAP

PR #437, opened 2021-10-04
profiler: add support for dynamic tags

awwalker:
For #433 

The issue mentions adding tests but I was a little unsure about the best way to add tests to the `profiler` because of the `upstream` instantiation. I added a couple of unit tests to the `session` which the profiler wraps instead but happy to revisit tests if y'all have some guidance on the best way to test the profiler


petethepig (approved):
Looks good, thank you @awwalker !

Issue #348, opened 2021-08-24
Implement out of space notification


petethepig:
We already have support on the frontend https://github.com/pyroscope-io/pyroscope/pull/334

Now we need to add backend part that will notify users about things like lack of disk space

yashrsharma44:
Can I pick this issue? Would love some pointers regarding this 😄

petethepig:
@yashrsharma44 The error happens here: https://github.com/pyroscope-io/pyroscope/blob/main/pkg/storage/storage.go#L335-L337

When it happens we need to pass a message to the frontend via this function https://github.com/pyroscope-io/pyroscope/blob/3227700ced61f47cb68c9d2c921c3c66b975c809/pkg/server/handler.go#L393-L396

I'm thinking we could maybe implement it so that this notification appears if an error like that has happened within the last minute or so.

kolesnikovae:
I think we should implement a more generic thing that just one health check. Let me know if you think we don't need this complication :)

First, let's define the domain. We have some checks to be performed on the interval (at this point I think it is the same for all probes):

```go
// pkg/server/health/health.go

package health

// Condition represents an aspect of pyroscope server health.
type Condition interface {
	State() State
	Message() string
	MakeProbe() error
}

type State int

const (
	NoData State = iota
	Healthy
	Warning
	Critical
)
```

```go
// pkg/server/health/controller.go

package health

import "sync"

// Controller performs probes of health conditions.
type Controller struct{
	m sync.Mutex
	c []Condition
}

func (*Controller) Start() {}
func (*Controller) Stop()  {}

func (*Controller) NotificationText() string {
	// This method should not access Conditions. Instead this message
	// should be built by controller after all the probes polled.
	return ""
}

```

Controller should also be properly initialized and stopped at shutdown. `server.Controller` depends on it, therefore initialization order matters. `server.Controller` should accept it by an interface so that it could be mocked in tests, e.g (later we may want to elaborate this to leverage live notifications):

```go
// pkg/server/controller.go

type Notifier interface {
	NotificationText() string
}
```

Having done with the health controller we can implement various health `Condition`s. In particular, free disk space:

```go
// pkg/server/health/disk.go

package health

import "sync"

type DiskPressure struct {
	m     sync.Mutex
	state State

	Available uint64
	Used      uint64

	WarningThreshold  uint64
	CriticalThreshold uint64
}

func (d *DiskPressure) State() State     { return NoData }
[...]

PR #456, opened 2021-10-10
Implement out of space notification #348

gawicks:
- Implements health probes
- Implement disk out of space notification


eh-am:
Nice work!

Question: would it be possible to add an endpoint to query these notifications? I think it would be handy on the client side. (It doesn't have to be in this PR)

PR #470, opened 2021-10-18
feat(golang): support for golang profiling labels

petethepig:


petethepig:
@eh-am yeah good point, I need to add that too before merging.

PR #473, opened 2021-10-20
Golang profiler API

kolesnikovae:
Following the discussion, an example API that does propagate labels without a context:

```go
func WithLabels(LabelSet, func())

// WithLabelsContext is just a counterpart of pprof.Do.
func WithLabelsContext(context.Context, LabelSet, func(context.Context))
```

The problem is that this can not be used with `pprof.Do` due to the difference in semantics: `pprof.Do` (as well as `WithLabelsContext`) replaces existing label set, meanwhile `WithLabels` appends new labels to it: see tests for details. This discrepancy jeopardizes the very idea of "simple" API.

Althout, it could be mitigated with an additional call that would create a new context with the current labels:
```go
// Context creates a new context with labels of the current goroutine.
func Context() context.Context
```

```go
profiler.WithLabels(profiler.Labels("foo", "bar"), func() {
    pprof.Do(context.Background(), pprof.Labels("baz", "qux"), func(context.Context) {
        // Does not have foo=bar label.
    })
    pprof.Do(profiler.Context(), pprof.Labels("baz", "qux"), func(context.Context) {
        // Both foo=bar and baz=qux pairs are preserved.
    })
    profiler.WithLabelsContext(profiler.Context(), profiler.Labels("baz", "qux"), func(context.Context) {
        // Both foo=bar and baz=qux pairs are preserved.
    })
})
```

Not quite sure this complication really outweighs benefits of the simplification.

### DIFF
--- a/examples/golang/main.go
+++ b/examples/golang/main.go
@@ -17,15 +17,15 @@ func work(n int) {
 	// revive:enable:empty-block
 }
 
-func fastFunction(c context.Context) {
-	profiler.TagWrapper(c, profiler.Labels("function", "fast"), func(c context.Context) {
+func fastFunction(ctx context.Context) {
+	profiler.WithLabelsContext(ctx, profiler.Labels("function", "fast"), func(context.Context) {
 		work(20000000)
 	})
 }
 
-func slowFunction(c context.Context) {
+func slowFunction(ctx context.Context) {
 	// standard pprof.Do wrappers work as well
-	pprof.Do(c, pprof.Labels("function", "slow"), func(c context.Context) {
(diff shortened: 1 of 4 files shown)


kolesnikovae:
Decided to postpone the PR. The most problematic part concerns the cast of a private type returned from the linked runtime function (see `GetGoroutineLabels`):
```go
(*map[string]string)(runtime_getProfLabel())
```

I believe this is too dangerous for public use. Moreover, the new API contradicts ideas expressed in [the original proposal](https://go.googlesource.com/proposal/+/master/design/17280-profile-labels.md#motivation).

PR #480, opened 2021-10-27
Feature/441 "focus on this subtree"

eh-am:
This PR introduces new functionality called `focus on this subtree`, based on https://github.com/pyroscope-io/pyroscope/issues/441

1 - It allows to change the flamegraph root:
Peek 2021-10-29 16-02

Currently the state is tored locally, but in the future people could use to share a specific subtree.

2 - It still allows zooming
Peek 2021-10-29 16-04

3 - Clicking on the same node undoes the zoom:
Peek 2021-10-29 16-05

4 - You can mix and match zoom/focus, here for example we zoom and then we focus:
Peek 2021-10-29 16-06

This currently is scoped under a `contextmenu` event, up next this feature would be available under a toolbar as well.


Rperry2174 (approved):
This is awesome! Everything seems to be running pretty smoothly and  couldn't find any edge cases thanks for adding the double-click functionality too!

Rperry2174:
Two potential ideas for how we can let people know about the right click functionality:
1. We could put one of these: ⓘ somewhere where when you hover over it it says instructions about how right click adds functionality
2. We could add it to the shortcuts list 
image

Up to you do I'd be fine merging this and thinking more about this part later too

eh-am:
@Rperry2174 Great ideas! Let me merge this and do that work in a separate PR.

PR #487, opened 2021-10-31
Improve text when there is no data available

clovis1122:
If the selected application or range does not have data to show, we'll render a message saying so. Additionally, we won't show the "Frame width represents" text if there's nothing to show.

closes https://github.com/pyroscope-io/pyroscope/issues/483

image

### DIFF
--- a/webapp/javascript/components/FlameGraph/FlameGraphComponent/index.tsx
+++ b/webapp/javascript/components/FlameGraph/FlameGraphComponent/index.tsx
@@ -174,6 +174,9 @@ export default function FlameGraphComponent(props: FlamegraphProps) {
     }
   }, [flamegraph]);
 
+  const dataUnavailable =
+    !flamebearer || (flamebearer && flamebearer.names.length <= 1);
+
   return (
     <>
       <div
@@ -188,7 +191,18 @@ export default function FlameGraphComponent(props: FlamegraphProps) {
           ExportData={ExportData}
         />
 
-        <div>
+        {dataUnavailable ? (
+          <div className="error-message">
+            <span>
+              No profiling data available for this application / time range.
+            </span>
+          </div>
+        ) : null}
+        <div
+          style={{
+            opacity: dataUnavailable ? 0 : 1,
+          }}
+        >
           <canvas
             height="0"
             data-testid="flamegraph-canvas"
--- a/webapp/javascript/components/FlameGraph/FlameGraphComponent/index.tsx
+++ b/webapp/javascript/components/FlameGraph/FlameGraphComponent/index.tsx
@@ -192,7 +192,7 @@ export default function FlameGraphComponent(props: FlamegraphProps) {
         />
 
         {dataUnavailable ? (
-          <div className="error-message">
+          <div className={styles.error}>
             <span>
               No profiling data available for this application / time range.
             </span>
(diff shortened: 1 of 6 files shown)


webapp/javascript/components/FlameGraph/FlameGraphComponent/index.tsx
```diff
@@ -139,7 +142,18 @@ export default function FlameGraphComponent(props: FlamegraphProps) {
...
+        {dataUnavailable ? (
+          <div className="error-message">
+            <span>
+              No profiling data available for this application / time range.
+            </span>
+          </div>
+        ) : null}
+        <div
+          style={{
+            opacity: dataUnavailable ? 0 : 1,
```
eh-am, line 154: I'm a bit curious, why was this needed, as opposed to simply not rendering the canvas?

webapp/javascript/components/FlameGraph/FlameGraphComponent/index.tsx
```diff
@@ -139,7 +142,18 @@ export default function FlameGraphComponent(props: FlamegraphProps) {
           ExportData={ExportData}
         />
 
-        <div>
+        {dataUnavailable ? (
+          <div className="error-message">
```
eh-am, line 146: If this class is used only here, would you mind using css modules? We are now trying to scope the css as much as possible.

Just to give you some direction, you have to create a `index.css.module`, for example
https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/components/FlameGraph/FlameGraphComponent/Highlight.tsx#L2
https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/components/FlameGraph/FlameGraphComponent/Highlight.module.css

eh-am (commented):
Awesome, thank you for your contribution!

Would you mind adding a unit test for this case? I'm thinking something like passing an empty`flamebearer` and assert that the "No profiling data" message is shown and the flamegraph is not visible.

webapp/javascript/components/FlameGraph/FlameGraphComponent/index.tsx
clovis1122: Because of the logic attached to `canvasRef`. I imagine simply not rendering the canvas is possible, though it would require some larger refactor across different components.

webapp/javascript/components/FlameGraph/FlameGraphComponent/index.tsx
clovis1122: Done!

clovis1122:
hmmm, I added some tests but had to move it to the bottom... it looks like some of the tests in this file have dependency among themselves.

eh-am:
Nice!

> hmmm, I added some tests but had to move it to the bottom... it looks like some of the tests in this file have dependency among themselves.

Almost! The [`findBy` family returns a promise](https://testing-library.com/docs/queries/about/) and the test isn't waiting for it. I suggest to just replace by `getBy`. Plus, you can use a regex there, to avoid matching the whooooole text, eg. `screen.getByText(/No profiling data available/);`


Another thing, can you fix the test that's breaking? We've updated the `FlamegraphComponent` API, it should be pretty straightforward to fix it, `{ i: -1, j: -1}` becomes `Option.none()`

clovis1122:
@eh-am done, this should be good for a second pass!

PR #488, opened 2021-10-31
Fix sidebar getting out of sync

clovis1122:
This can happen if the location changes without using the sidebar (pressing back/next in browser).

The sidebar component will now subscribe to changes.

closes https://github.com/pyroscope-io/pyroscope/issues/320.


eh-am (commented):
Nice one!

Would you mind adding a test for this case? Honestly I don't know if it's better to unit test or simply a test with cypress. Up to you which one is easier!

clovis1122:
@eh-am done! Let me know if that's good enough!

eh-am:
Beautiful! Merging it now

PR #498, opened 2021-11-03
fix: highlight in diff view

eh-am:
Basically just check if we are highlighting first before checking if it's a diff flamegraph.

This now works:
Peek 2021-11-03 16-44


Rperry2174:
Looks good to me! 

Also I know it's small but wondering if we could sneak in adding a couple pixels to the width of the legend. For some reason 
with the default width of `35px`
image
If I change to width of `37px`
image

It looks as though yours doesn't have this issue though

Rperry2174 (approved):
LGTM (see comment about legend)

eh-am:
Done! @Rperry2174

PR #499, opened 2021-11-04
fix: linking flags in the linux build system

abeaumont:
Fix the linking flags problem that prevents local Pyroscope builds on Linux with the `rbspy` and `pyspy` to succeed. Check each commit individually for further details.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -84,7 +84,9 @@ RUN apk upgrade elfutils
 WORKDIR /opt/pyroscope
 
 RUN mkdir -p /opt/pyroscope/third_party/rustdeps/target/release
-COPY --from=rust-builder /opt/rustdeps/librustdeps.a /opt/pyroscope/third_party/rustdeps/target/release/librustdeps.a
+# TODO: find a way to determine the architecture in the copy command, we copy once per architecture for now.
+COPY --from=rust-builder /opt/rustdeps/librustdeps.a /opt/pyroscope/third_party/rustdeps/target/x86_64-unknown-linux-musl/release/librustdeps.a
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -84,7 +84,9 @@ RUN apk upgrade elfutils
 WORKDIR /opt/pyroscope
 
 RUN mkdir -p /opt/pyroscope/third_party/rustdeps/target/release
-COPY --from=rust-builder /opt/rustdeps/librustdeps.a /opt/pyroscope/third_party/rustdeps/target/release/librustdeps.a
+# TODO: find a way to determine the architecture in the copy command, we copy once per architecture for now.
```
abeaumont, line 87: This is a hack, a better solution is welcome here.

eh-am:
Works wonderfully here!

Dockerfile
petethepig: I'm thinking let's do the opposite, let's move files from `/opt/pyroscope/third_party/rustdeps/target/x86_64-unknown-linux-musl/release/librustdeps.a` to `/opt/pyroscope/third_party/rustdeps/target/release/librustdeps.a` in the Makefile.

I think the culprit here is that on Linux we're adding this `--target` option https://github.com/pyroscope-io/pyroscope/blob/main/Makefile#L110 and when you specify a `--target` it always puts it into a "target-specific" directory. So if we do something like `mv third_party/rustdeps/target/x86_64-unknown-linux-musl/release third_party/rustdeps/target/release` right after the build that should work and is arguably a slightly cleaner solution.

petethepig (changes requested):

Dockerfile
abeaumont: Indeed, it's a much simpler solution! I was a bit wary as both the generic and the target-specific `release` dirs are non-empty, but copying the resulting lib file should be fine :+1:

petethepig (approved):
LGTM

Issue #228, opened 2021-06-01
Add support for ad-hoc profiling.


petethepig:
Clarify what I'm talking about

eh-am:
Here's an example of workflow that the `adhoc` functionality should provide (ignoring that's that's a nodejs repo)

https://github.com/mapbox/flamebearer/issues/7#issuecomment-449841832

Rperry2174:
Adding this as a mockup for what a potential end UI would look like. 
![adhoc_flow](https://user-images.githubusercontent.com/23323466/140575239-174a1cc0-b401-4c7a-b666-7726349741a4.gif)

For continuous profiling, the input is the "start time" and "end time" which comes from the timeline. This makes sense for continuous profiling because part of the appeal here is being able to drill down into particular time ranges.

With adhoc profiling the idea is to be able to take a _file_ as input and then output a flamegraph. Then over time, you can revisit previous ad-hoc profiles that you have created via a file-selector-type-system. This would allow us to recreate all three of our main views:
- single view
- comparison view
- diff view

By simply replacing the timeline input component, with a file-selector input component. 

We likely would output the files to a specific folder so this mockup is slightly inaccurate in that sense, but the general idea is there.

PR #504, opened 2021-11-05
feat: initial support for adhoc mode

abeaumont:
Initial support for `adhoc` subcommand, related to #228.

There have been several iterations and experiments about how `adhoc` should work, this PR presents an initial support for the latest iteraction, in which:
- A new adhoc subcommand is created, which allows the user to:
  - Launch a new process if needed (it also works with already running processes).
  - Profile the process in every way pyroscope knows how to profile:
    - through spies for supported languages, both by launching a new process or attaching to an existing one.
    - through native integrations, in applications that support them  (a new process is needed in this case to override the server address).
    - through pull-mode, in applications that support it, both by launching a new process or attaching to an existing one.
  - Store the results in a local file, in different formats (Pyroscope native JSON, pprof, collapsed). These results will be eventually available through the UI (an initial support for this is already available at https://github.com/pyroscope-io/pyroscope/pull/546).

The feature is not yet enabled and has some rough edges, but I think it's in a mergeable state, so it may be good to do so to prevent it from diverging too much.


eh-am:
Love this!

This is more of an idea, maybe we could open automatically (`xdg-open` on linux) the pyroscope page with the time timestamp?

petethepig:
> This is more of an idea, maybe we could open automatically (xdg-open on linux) the pyroscope page with the correct time timestamp (something like 1~2 minutes around the uploaded time)?

Love the idea. I think we can open it sooner (first time any data is available).

And for cross-compatibility between operating systems we could add reuse something like this: [code from git/git](https://github.com/git/git/blob/master/git-web--browse.sh#L115)

cmd/pyroscope/command/adhoc.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/pyroscope-io/pyroscope/pkg/config"
+)
+
+func newAdhocCmd(cfg *config.Adhoc) *cobra.Command {
+	vpr := newViper()
+
+	cmd := &cobra.Command{
+		Use:   "adhoc [flags] [args]",
+		Short: "Profile a process and save the results to be used in adhoc mode",
+		Long: `adhoc command is a complete toolset to profile a process and save the profiling
```
abeaumont, line 17: A proposal of how to integrate the different profiling modes with the `adhoc` command in the CLI.

abeaumont:
The feature is not yet enabled and has some rough edges, but I think it's in a mergeable state, so it may be good to do so to prevent it from diverging too much.

cmd/pyroscope/command/adhoc.go
```diff
@@ -0,0 +1,56 @@
...
+func newAdhocCmd(cfg *config.Adhoc) *cobra.Command {
+	vpr := newViper()
+
+	cmd := &cobra.Command{
+		Use:   "adhoc [flags] [<args>]",
+		Short: "Profile a process and save the results to be used in adhoc mode",
+		Long: `adhoc command is a complete toolset to profile a process and save the profiling
+results.
+
+These results are then available to be visualized through the 'Adhoc Profiling'
```
eh-am, line 20: I know we kinda dropped this idea, but now that I am thinking again, a `pyroscope adhoc report/analyze pyroscope.data` which opens the browser in the adhoc only mode would be kinda useful.

cmd/pyroscope/command/adhoc.go
```diff
@@ -0,0 +1,56 @@
...
+section in the UI (available through 'pyroscope server').
```
eh-am, line 21: I know we kinda dropped this idea, but now that I am thinking again, `pyroscope adhoc report/analyze pyroscope.data` which opens the browser in the adhoc only mode would be kinda useful.

cmd/pyroscope/command/adhoc.go
```diff
@@ -0,0 +1,56 @@
...
+
+There are multiple ways to gather the profiling data, and not all of them are
+available for all the languages.
+Which way is better depends on several factors: what the language supports,
+how the profiled process is launched, and how the profiled process provides
+the profiled data.
```
eh-am, line 27: This is good, I think we can expand further and for each command give a suggestion for when we would use it (for example, `pull` works well with the `go` integration)

cmd/pyroscope/command/adhoc.go
```diff
@@ -0,0 +1,56 @@
...
+- exec. In this case, pyroscope creates a different process for the profiled
+program and uses a spy to directly gather profiling data. It's a useful way
+to profile a whole execution of some program that has no other pyroscope
+integration or way of exposing profiling data.
+It's the default mode for languages with a supported spy when either the spy
+name is specified (through '--spy-name' flag) or when the spyname is
+autodetected.
+- connect. Similar to exec, pyroscope uses a spy to gather profiling data,
+but instead of creating a new profiled process, it spies an already running
+process, indicated through '--pid' flag.
```
eh-am, line 39: I think this can be simpler:

- connect. Similar to exec, but it spies an already running process, indicated through '--pid' flag.

cmd/pyroscope/command/adhoc.go
```diff
@@ -0,0 +1,56 @@
...
+- push. In this case, pyroscope creates a different process for the profiled
+program and launches an HTTP server with an ingestion endpoint. It's useful
+to profile programs already integrated with Pyroscope using its HTTP API.
+Push mode is used by default when no spy is detected and no '--url' flag is
+provided. It can also be override the default exec mode with the '--push' flag.
+- pull. In this case, pyroscope periodically connects to the URL specified
+thorugh '--url' where it tries to retrieve profiling data in any of the
+supported formats. In this case arguments are optional, and if provided,
+they are used to launch a new process before polling the URL.`,
```
eh-am, line 48: >Similar to exec
>In this case
I suggest to drop these, since they don't add much to the sentence.

cmd/pyroscope/command/adhoc.go
abeaumont: That's a good point. I don't think we should talk about integrations in the CLI help, as integrations can evolve independently of pyroscope itself, but giving a link to the documentation may be good (and we also need to write that documentation where this is all explained, of course).

cmd/pyroscope/command/adhoc.go
abeaumont: I think this kind be handled by an HTML export too, depending on how we want the server to work.

PR #508, opened 2021-11-05
Decouple internal config from public CLI config in exec and server/storage

abeaumont:
Internal components' configuration is coupled to public CLI configuration, which makes evolving public and internal configurations independently is harder than it should. This goes both
ways, e.g.:
- Deprecated configuration variables are still accessible internally.
- Adding internal configuration variables without exposing them publicly is hard.

The case of the storage component is worse, as it has access to a lot of unneeded server configuration variables.

While this problem already shows in the code (in how `dbmanager` was implemented, or exec/connect distinction, for example), it may become worse with `adhoc` mode (where we may want to use in-memory storage, direct upload, remove server app, etc.).

This refactor tries to solve this problem by isolating public and private configurations through a public API to build the internal config (through NewConfig and With* methods, as needed) for the `Exec` and `Storage` components, but it could also be extended to other components as needed.

### DIFF
--- a/cmd/pyroscope/command/connect.go
+++ b/cmd/pyroscope/command/connect.go
@@ -16,7 +16,7 @@ func newConnectCmd(cfg *config.Exec) *cobra.Command {
 
 		DisableFlagParsing: true,
 		RunE: cli.CreateCmdRunFn(cfg, vpr, func(_ *cobra.Command, args []string) error {
-			return exec.Cli(cfg, args)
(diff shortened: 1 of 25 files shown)


petethepig (approved):
Looks great! Don't have much to say here — I like the approach, this definitely makes the system better.

I'm thinking at some point we should probably move the old config package closer to cli packages so that it's clear that those configs belong to the public api part.

PR #510, opened 2021-11-07
Fix snapshotting error handling

### LINKED ISSUE #502: Snapshot error on exec mode
With https://github.com/pyroscope-io/pyroscope/pull/436, the exec mode uses polling to check whether the target process is running or not, but the sampling rate is (usually) higher than the polling rate, so usually, the snapshot taking code detects the process has finished. This case is covered in the code but `process.PidExists` seems to fail to detect that the process is gone, and instead, a different error is raised: `ERRO[2021-11-05T09:14:38.151745] agent/session.go:224 error taking snapshot: Permission denied (os error 13)`

The execution seems to work fine after that:
```
ERRO[2021-11-05T09:26:03.517763] agent/session.go:224 error taking snapshot: Permission denied (os error 13)
DEBU[2021-11-05T09:26:04.477188] exec/cli.go:192 child process exited
```

kolesnikovae:
The PR is aimed to make snapshot error handling more clear: no error should be reported by `Session` when the target process exits. The actual error returned from the app is handled by `exec`, including stderr/stdout message and the returned exit code.

Closes #502


abeaumont:
LGTM :+1:

PR #511, opened 2021-11-08
simplify local dev

eh-am:
Starting from a clean env

```
make install-dev-tools
make build-third-party-dependencies
make install-web-dependencies

make dev
``` 
Should work now

@abeaumont pleeease can you check it works for you?

If it works I will update the documentation later.

---
Just some explanation, by moving `WebpackPluginServe` to the front, we wait for the build to be finished (`waitForBuild`), which generates `webapp/public/assets`, which the go server serves. The `HtmlWebpackPlugin` hits the go server and serves that under the port `4041`. Whew.

### DIFF
--- a/scripts/webpack/webpack.dev.js
+++ b/scripts/webpack/webpack.dev.js
@@ -18,25 +18,9 @@ module.exports = merge(common, {
   },
 
   plugins: [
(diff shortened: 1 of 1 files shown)


abeaumont:
@eh-am it works perfectly here, thanks!

PR #515, opened 2021-11-09
Feat/delete apps

eh-am:
WIP, as always, a PR early on to get some feedback

It's currently hidden behind a flag, I recommend setting the following env var `PYROSCOPE_ENABLE_EXPERIMENTAL_ADMIN=true`

The only added functionality is to get the list of app names, since you need to know what apps exist to be able to delete them!

There's no CLI yet, so a simple way to test it is running with curl
```
curl --unix-socket /tmp/pyroscope.sock http://bla/v1/apps 
```
---
Edit 1: 
This is working now:

1. Have a server running:
`make dev` or `./bin/pyroscope server`

2. Run the cli

(to get the list the apps)
```
./bin/pyroscope admin app get
```

Then delete the app. For example:
```
./bin/pyroscope admin app delete mytest2
Are you sure you want to delete the app 'mytest2'? This action can not be reversed.

Keep in mind the following:
a) If an agent is still running, the app will be recreated.
b) The API is idempotent, ie. if the app already does NOT exist, this command will run just fine.

Type 'mytest2' to confirm (without quotes).
mytest2
Deleted app 'mytest2'.
```

---
Some thoughts:
a)  We went back and forth with where to create the socket file. I was initially afraid of having to deal with permissions issues.
Turns out [/tmp is POSIX compliant](https://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap10.html)

b) By having the socket in a well-known location, we simplify the client (here, the cli), by just assuming the socket is on that well-known place. 

Compare this with using `storagePath`, in that case the cli doesn't know where the `storagePath` is!

c) Of course, it's less straightforward if you have multiple instances running at the same time. But there's nowhere to run in that case. You need to manually pass `--socket-path`.


pkg/config/config.go
```diff
@@ -212,3 +217,8 @@ type Exec struct {
 
 	Tags map[string]string `name:"tag" def:"" desc:"tag in key=value form. The flag may be specified multiple times" mapstructure:"tags"`
 }
+
+type Admin struct {
+	// TODO: use <storagePath> somehow
+	SocketPath string `def:"/tmp/pyroscope.sock" desc:"path where the admin UDS will be created." mapstructure:"admin-socket-path"`
```
petethepig, line 223: If I understand this right, this will look like this in the config file:
```
admin:
  admin-socket-path: 
```

I think we can drop `admin-` prefix, because it will be clear from context in this case

petethepig: Also, I don't know what UDS stands for. We should probably spell it out more, because it's not super clear

pkg/admin/server.go
```diff
@@ -0,0 +1,48 @@
...
+		ctrl: ctrl,
+	}
+	as.HTTPServer = httpServer
+
+	mux := http.NewServeMux()
+	as.Mux = mux
+
+	// Routes
+	// TODO maybe use gorilla?
+	mux.HandleFunc("/v1/apps", as.ctrl.HandleApps)
```
petethepig, line 37: One thing we should perhaps talk about is weather or not we want to reuse the existing handler that we use for http endpoint. I think the way you're doing it is better, but let's discuss it next time we talk in person.

petethepig (approved):
I think it looks good. I added a couple of minor comments / suggestions.

pkg/config/config.go
eh-am: Good points, renamed the property and removed UDS from there since it's an implementation detail. Also explained what UDS stands for in the pkg code.

It means Unix Domain Socket BTW

Issue #471, opened 2021-10-18
Support other formats in /render endpoint


petethepig:
Currently /render endpoint only returns data in flamebearer format. Users have been asking for other formats as well.

We already have support for different ingestion formats. I think it's time to extend the render endpoint with support for these formats as well. [Here's the code for ingestion part](https://github.com/pyroscope-io/pyroscope/blob/bac117131abd01ebde688c440b15c832bee9987d/pkg/server/ingest.go#L95-L107).

For rendering we should do something similar. Extra formats we should add support for are:
* [ ] collapsed (also called "groups" in our source code), it's basically stacktraces grouped by how many times a certain stacktrace was seen, e.g:
```
foo;bar 100
foo;baz 200
```
* [ ] pprof

PR #518, opened 2021-11-10
Support other formats in /render endpoint #471

gawicks:
Support other formats in /render endpoint

- [x] pprof
- [x] collapsed
 
GETing the render endpoint with the corresponding format specified (/render?format=pprof) would download these as files


pkg/server/render.go
```diff
@@ -60,9 +61,26 @@ func (ctrl *Controller) renderHandler(w http.ResponseWriter, r *http.Request) {
...
+		res := renderResponse(fs, out)
+		ctrl.writeResponseJSON(w, res)
+	case "pprof":
+		pprof := out.Tree.PprofStruct(&tree.PprofMetadata{
+			Type: out.SpyName, // TODO
+			Unit: out.Units,
+		})
+		out, err := proto.Marshal(pprof.Pprof())
+		if err == nil {
+			ctrl.writeResponseFile(w, "profile.pprof", out)
```
petethepig, line 76: I'm thinking it would be kind of nice to have a full name for a profile, for example it would be cool if it was something like:
```
<query> - <start time>.pprof
```
so for example:
```
pyroscope.server.cpu{} - 2021-11-11 15:00:00Z.pprof
```

pkg/server/render.go
```diff
@@ -60,9 +61,26 @@ func (ctrl *Controller) renderHandler(w http.ResponseWriter, r *http.Request) {
...
+		} else {
+			ctrl.writeInternalServerError(w, err, "")
+		}
+	case "collapsed":
+		collapsed := out.Tree.Collapsed()
+		ctrl.writeResponseFile(w, "profile.collapsed", []byte(collapsed))
```
petethepig, line 82: I would maybe call it `profile.collapsed.txt` so that it's clear this is plaintext. And I guess the same comment about name from above applies here.

pkg/storage/tree/pprof.go
```diff
@@ -0,0 +1,106 @@
...
+		value := []int64{int64(self)}
+		loc := []uint64{}
+		for _, l := range stack {
+			loc = append(loc, uint64(p.newLocation(l)))
+		}
+		sample := &Sample{LocationId: loc, Value: value}
+		p.profile.Sample = append(p.profile.Sample, sample)
+	})
+
+	/* TODO: Remove */
```
petethepig, line 54: can we remove this?

pkg/storage/tree/tree.go
```diff
@@ -181,6 +181,33 @@ func (t *Tree) Iterate(cb func(key []byte, val uint64)) {
 	}
 }
 
+func (t *Tree) Iterate2(cb func(name string, self uint64, stack []string)) { // TODO
```
petethepig, line 184: Can you change the name to something more descriptive?

pkg/storage/tree/profile.proto
```diff
@@ -0,0 +1,211 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
```
petethepig, line 1: We already have this file in a different package (`convert`), so ideally we should avoid creating another copy.

petethepig (changes requested):

pkg/storage/tree/profile.proto
petethepig: Same with profile.pb.go

pkg/server/render.go
```diff
@@ -60,9 +61,26 @@ func (ctrl *Controller) renderHandler(w http.ResponseWriter, r *http.Request) {
...
+	case "pprof":
+		pprof := out.Tree.PprofStruct(&tree.PprofMetadata{
+			Type: out.SpyName, // TODO
+			Unit: out.Units,
+		})
+		out, err := proto.Marshal(pprof.Pprof())
+		if err == nil {
+			ctrl.writeResponseFile(w, "profile.pprof", out)
+		} else {
+			ctrl.writeInternalServerError(w, err, "")
```
abeaumont, line 78: It may be good to give a bit more context about the error in the response

```suggestion
			ctrl.writeInternalServerError(w, err, "failed to serialize data")
```

pkg/storage/tree/collapsed.go
```diff
@@ -0,0 +1,22 @@
+package tree
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+)
+
+func (t *Tree) Collapsed() string {
```
abeaumont, line 9: `Collapsed` seems quite close to `String` method. Would it make sense to reuse/adapt `String` instead?

pkg/storage/tree/pprof.go
```diff
@@ -0,0 +1,106 @@
...
+		profile: &Profile{
+			StringTable: []string{""},
+		},
+		tree:     t,
+		metadata: metadata,
+	}
+	return pprof
+}
+
+func (p *Pprof) Pprof() *Profile {
```
abeaumont, line 40: Maybe splitting pprof format generation is not needed. An instance of `Pprof` cannot be used to generate the `Profile` more than once (i.e. calling `Pprof()` multiple times would result in different `Profile`s for the same tree), so it may be better to collapse the functionality into a single method with the following signature: `func (t *Tree) Pprof(metadata *PprofMetadata) *Profile`. What do you think?

pkg/storage/tree/pprof.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func (p *Pprof) newFunction(function string) uint64 {
+	id, ok := p.functions[function]
+	if !ok {
+		id = uint64(len(p.profile.Function) + 1)
+		newFn := &Function{
+			Id:         id,
+			Name:       int64(p.newString(function)),
+			SystemName: int64(p.newString(function)),
```
abeaumont, line 100: It's a bit confusing now to understand that  `Name` and `SystemName` have the same value. Making it explicit may be helpful:

```suggestion
		name := int64(p.newString(function))
		newFn := &Function{
			Id:         id,
			Name:       name,
			SystemName: name,
```

pkg/storage/tree/tree.go
```diff
@@ -181,6 +181,33 @@ func (t *Tree) Iterate(cb func(key []byte, val uint64)) {
...
+	t.RLock()
+	defer t.RUnlock()
```
abeaumont, line 186: Note that in `Collapsed` method locking is done explicitly before calling `Iterate2` while in `Pprof` is not, better to make this consistent. Since `Iterate` already leaves locking to the caller, the same approach could be used here.

abeaumont (commented):
It looks good so far! I'll leave some minor improvement comments/ideas.

pkg/storage/tree/profile.proto
gawicks: This is something I wanted bring up with you, If I use the proto file in `convert` it gives a circular imports error. What's a good way to handle this you think?

abeaumont: That's a good point. It seems the tree parser (`ParseTreeNoDict`) is in the `convert` package causing the cycle. It probably makes sense to move the parser to the `tree` package`.

pkg/server/render.go
```diff
@@ -60,9 +61,26 @@ func (ctrl *Controller) renderHandler(w http.ResponseWriter, r *http.Request) {
...
-	res := renderResponse(fs, out)
-	ctrl.writeResponseJSON(w, res)
+	switch p.format {
+	case "json":
+		fs := out.Tree.FlamebearerStruct(p.maxNodes)
+		res := renderResponse(fs, out)
+		ctrl.writeResponseJSON(w, res)
+	case "pprof":
+		pprof := out.Tree.PprofStruct(&tree.PprofMetadata{
+			Type: out.SpyName, // TODO
```
gawicks, line 71: @petethepig @abeaumont Don't think it's correct to use `SpyName` here. Is there a way to extract the type of the profile?

pkg/storage/tree/profile.proto
gawicks: Good suggestion

pkg/storage/tree/pprof.go
gawicks: Agree. PprofStruct() is redundant

pkg/storage/tree/collapsed.go
abeaumont: Yes, I think it makes sense to have a single version and adapt unit tests if needed, as you did :+1:

pkg/storage/tree/tree.go
```diff
@@ -109,17 +108,7 @@ func prependInt(s []int, x int) []int {
...
-
-	res := ""
-	t.Iterate(func(k []byte, v uint64) {
-		if v > 0 {
-			res += fmt.Sprintf("%q %d\n", k[2:], v)
-		}
-	})
-
-	return res
+	return t.Collapsed()
```
abeaumont, line 111: I wonder why `String()` is built on top of `Collapsed()`... it may be simpler to just have a single implementation under `String` and just call `String()` from the controller, right? No need to have 2 different methods for the same thing.

I'm also curious about why the original implementation has been discarded and a new iteration function has been created. Is there some problem with the original code and, if there is, could the old code be replaced by the new one instead of having two similar methods?

pkg/storage/tree/profile.proto
```diff
@@ -38,9 +38,9 @@
 
 syntax = "proto3";
 
-package perftools.profiles;
+package perftools.profiles.tree;
 
-option go_package = "pkg/convert";
+option go_package = "pkg/tree";
```
abeaumont, line 43: I'm not sure the whole `Profile` type should be moved to the `tree` package, moving only the specific Tree parser would have been enough I think, but maybe I'm missing something.

abeaumont (commented):
I think the changes look much better now! I left a couple of comments about things I'm not sure about.

pkg/storage/tree/tree.go
gawicks: Yes I thought this might raise some eyebrows. The way I thought about this is , collapsed is a standardized format. String incidentally uses something very close to this format, but it could very well be something else. This makes that independence explicit.
Why not use the original Iterate - It returns a semicolon delimited list of frames , which is not very flexible. So you end up having to join then the output split to manipulates the frames, and it has other idiosyncrasies like needing to check for len(name) > 2 for empty stacks etc. It was probably written to Stringify the tree and hasn't grown out of it. I don't want to change it too much without having a firm grasp on how it's used in the ingestion code.

pkg/storage/tree/profile.proto
gawicks: This is again somewhat based on my opinion. The convert package only uses the single exported type `Profile` from the proto generated types, while the tree package uses a majority of its types. So the code is less noisy by keeping the types unprefixed. It also keeps the trio of related functions `ParseTreeNoDict`, `ParseIndividualLines`, `ParseGroups` in the same package.

petethepig (approved):
Looks good to me

PR #520, opened 2021-11-11
Added windows tests CI ,rename linux tests CI

Dzalevski:
- Added go-tests CI running on windows
- Rename Linux tests for consistency

### DIFF
--- a/.github/workflows/tests-go-linux.yml
+++ b/.github/workflows/tests-go-linux.yml
@@ -1,0 +1,35 @@
+name: Go Tests Linux
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  go-tests:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


Dzalevski:
@petethepig  Hey we can use 1 workflow, 
fist run the Linux tests then wait( windows will be depend on Linux passing ) and maybe rework little bit CI with matrix for version/OS so the code(checkout,gotest) will be reusable :)

petethepig:
These packages we can skip in windows tests:
```
FAIL	github.com/pyroscope-io/pyroscope/benchmark/internal/cireport	0.052s
FAIL	github.com/pyroscope-io/pyroscope/pkg/cli	0.235s
FAIL	github.com/pyroscope-io/pyroscope/pkg/server	7.413s
FAIL	github.com/pyroscope-io/pyroscope/pkg/storage	10.593s
FAIL	github.com/pyroscope-io/pyroscope/pkg/storage/cache	0.272s
```

PR #521, opened 2021-11-11
Fix Highlighting on Mobile Devices

ruslanpascoal2:
https://github.com/pyroscope-io/pyroscope/issues/468

Fixes highlight on mobile devices.


webapp/javascript/components/FlameGraph/FlameGraphComponent/Highlight.spec.tsx
```diff
@@ -21,7 +21,11 @@ describe('Highlight', () => {
   it('works', () => {
     const xyToHighlightData = jest.fn();
     render(
-      <TestComponent barHeight={50} xyToHighlightData={xyToHighlightData} />
+      <TestComponent
+        barHeight={50}
+        xyToHighlightData={xyToHighlightData}
+        zoom={undefined}
```
eh-am, line 27:
```suggestion
        zoom={Option.none()}
```

webapp/javascript/components/FlameGraph/FlameGraphComponent/Highlight.tsx
```diff
@@ -5,7 +5,7 @@ import styles from './Highlight.module.css';
 export interface HighlightProps {
   // probably the same as the bar height
   barHeight: number;
-
+  zoom: any;
```
eh-am, line 8:
```suggestion
  zoom: Option<DeepReadonly<{ i: number; j: number }>>
```
Which is the same type as https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/components/FlameGraph/FlameGraphComponent/Flamegraph.ts#L41

webapp/javascript/components/FlameGraph/FlameGraphComponent/Highlight.tsx
```diff
@@ -18,17 +18,26 @@ export interface HighlightProps {
...
 }
 export default function Highlight(props: HighlightProps) {
-  const { canvasRef, barHeight, xyToHighlightData } = props;
+  const { canvasRef, barHeight, xyToHighlightData, zoom } = props;
   const [style, setStyle] = React.useState<React.CSSProperties>({
     height: '0px',
     visibility: 'hidden',
   });
 
+  React.useEffect(() => {
```
eh-am, line 27: Can you add a comment explaining a little bit more what's going on here?

From my understanding, after the `zoom` prop is set, we hide by default.
Then after moving any pixel it will trigger the eventListener, which will then highlight again.

Can you just put that into words? I think it makes it easier to reason about later.

webapp/javascript/components/FlameGraph/FlameGraphComponent/Highlight.tsx
```diff
@@ -18,17 +18,26 @@ export interface HighlightProps {
...
+    if (!zoom) return;
```
eh-am, line 28: I believe this branch is never reached! Since `zoom` is an `Option`, which is never falsy.

So what this code is actually doing is EVERY time we change the zoom, we stop highlighting.
Which makes sense IMO, after all, the issue (stale highlighting) happens when BOTH ZOOMING and UNZOOMING.

Therefore I don't think this `if` is necessary.

eh-am (changes requested):
Hey Ruslan, thank you for your contribution!

I've added a few comments, nothing much. 

Let me know if you need any clarifications.

webapp/javascript/components/FlameGraph/FlameGraphComponent/Highlight.tsx
ruslanpascoal2: comments added, pretty much what you said

eh-am (approved):
LGTM!

PR #524, opened 2021-11-12
fix: base element position in templates

jonathanza:
According to the HTML spec, "A base element, if it has an href attribute, must come before any other elements in the tree that have attributes defined as taking URLs, except the html element (its manifest attribute isn't affected by base elements)." - https://html.spec.whatwg.org/multipage/semantics.html#the-base-element

If using pyroscope with a reverse proxy, the CSS does not load and returns an HTTP 404 status code. Moving the `base` element above the 'link' element in the files from 'webappss/templates' resolves the issue.


eh-am:
Makes sense to me!

petethepig:
Hi @jonathanza! Thank you for the contribution. 

I tried this patch out but I think I'm missing something. 

I'm testing it locally with pyroscope behind a reverse proxy with a subpath [link to a PR with an nginx config and docker-compose file](https://github.com/pyroscope-io/pyroscope/compare/base-element-test?expand=1) and I'm still getting 404s on assets because we use absolute paths for assets (they all start with `/assets`). So to me it looks like the problem still persists.

Can you tell us a bit more about your setup? Maybe show your nginx (or other reverse proxy) config and the value for `base-url` parameter?

zhuanxuhit:
I met the same question, and i modify
`<link rel="stylesheet" href="assets/styles.<%= webpack.hash %>.css" />
    <link rel="icon" href="assets/images/favicon.ico" />`
and then we can get the right css and js. 
but when i enter the http://xxxxx/pyroscope/, js does not work

jonathanza:
As @zhuanxuhit confirmed, the links to the assets are root relative (`<a href="/foo">`) and not document relative (`<a href="foo">`). 

I'm testing to see what the consequences are of dropping the leading forward slash in the templates.

zhuanxuhit:
> 

I find why it does not work. in "webapp/javascript/index.jsx", because i use base url is "pyroscope", so i modify it as below
image

petethepig:
We could use docker-compose and Caddyfile from [this PR](https://github.com/pyroscope-io/pyroscope/pull/69) to test this.

petethepig:
Closing this, going to finish in #611

PR #527, opened 2021-11-13
feat: pull mode

kolesnikovae:
The implementation is based on the Prometheus scrape and discovery packages.

 - [x] Basic pull-mode support
 - [x] Discovery support:
    - [x] Static configuration 
    - [x] Kubernetes
 - [x] ~Hot reload~
 - [x] [Update chart](https://github.com/pyroscope-io/helm-chart/pull/32)
 - [x] [Example setup](https://github.com/pyroscope-io/pyroscope/tree/feat/scraping/examples/pull)
 - [x] [Documentation](https://github.com/pyroscope-io/docs/pull/13)

Depends on https://github.com/pyroscope-io/pyroscope/pull/517: this PR introduces a large number of dependencies.

---

### Future work

- Targets status endpoint + FE.
- Port existing Prometheus discovery mechanisms (TBC).
- Support for pull-mode in Agent: agent should be able to scrape remote targets and reliably send collected data to server(s).

---

### Known bugs

1. Go CPU samples may be labeled incorrectly. See https://github.com/golang/go/issues/48577 for details.

---

### Example config

See [`examples/pull`](https://github.com/pyroscope-io/pyroscope/tree/feat/scraping/examples/pull) directory: there are two options:
 - kubernetes service discovery (note that it references chart from https://github.com/pyroscope-io/helm-chart/pull/32 - before merging you can specify it by path; the same for the image used: make sure you've overridden `image.repository` and `image.tag` accordingly)
 - static targets configuration

TODO:
 - make `hotrod-golang` repo public
 - update pyroscope repo/image references to public one


examples/pull/kubernetes/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+<details>
+    <summary>values.yaml</summary>
+
+```yaml
+---
+pyroscopeConfigs:
+  log-level: debug
+  scrape-configs:
+  - job-name: 'kubernetes-pods'
```
yeya24, line 35: Why not make this snake case? Like `job_name`, `kubernetes_sd_configs` etc

yeya24: It would be good to keep it consistent

petethepig: @yeya24 Honestly, this is just the convention we started with in our main config.

@kolesnikovae How hard do you think it would be to support both kebab-case and snake-case?

PR #528, opened 2021-11-15
feat: add completion script generation command

eh-am:
We already have support for this ever since we migrated to `cobra` (https://github.com/pyroscope-io/pyroscope/pull/300), but we were missing this last step, which is adding a command to generate the shell completion scripts.

It will be useful with eg https://github.com/pyroscope-io/pyroscope/pull/515


petethepig:
I think we already have this enabled: https://asciinema.org/a/nTZyHvnfG0NUAOcBk641XzJE4 Maybe it's enabled by default?

eh-am:
Whoops, my mistake.

PR #529, opened 2021-11-15
Remove labels on delete

kolesnikovae:
The PR is aimed to solve a bug when labels are not being deleted.


eh-am:
Thank you for working on this so promptly!

PR #531, opened 2021-11-16
Downgrade minimal node engine version required from 11 to 10.

abeaumont:
Downgrading supports older v10 versions, which seem to work fine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -159,7 +159,7 @@
     "size-limit": "^6.0.3"
   },
   "engines": {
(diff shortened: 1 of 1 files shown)


eh-am (approved):
Cool! Since nodejs is just for local dev stuff/builds I think it's ok to allow a certain discrepancy.

https://github.com/pyroscope-io/pyroscope/blob/main/Dockerfile#L50

PR #535, opened 2021-11-16
feat: add shell completion for 'admin app delete' cmd

eh-am:
Since the delete API is idempotent, ie, it won't fail if delete a non existent app, the cli ux could be a bit weird, since typos won't be immediately noticed.
We decided that as an easy improvement is to add autocomplete to the delete command, now at least it's less prone to typo.

For illustration, on my machine, `pyroscope admin app delete [tab][tab]` yields
image


petethepig (approved):
This is really cool :)

PR #538, opened 2021-11-16
chore: improve docker build speed by not pulling unnecessary dependencies

eh-am:
This was previously discussed in slack.

Basically it moves out all non essential dependencies to BUILD the application to `devDependencies`. Then, when building the docker image,  we install only `dependencies`.

For a data point, running the `node` build on my machine (extracted that whole "Assets" block into a different Dockerfile) took initially:

```
Executed in  216.58 secs
```

Then, after the cleanup:
```
Executed in  111.50 secs
```


petethepig (approved):
Awesome! Love these improvements!

PR #542, opened 2021-11-17
Fix race condition in version update.

abeaumont:
The testsuite (sometimes) fails due to this race condition.
Even if runOnce guarantees that there's a single writer, there can
still be concurrent read and writes. A read/write lock prevents this
from happening.


petethepig (approved):
Ah, I've hit this one locally a few times before. Thanks for fixing it!

PR #544, opened 2021-11-17
chore: increase cypress tests retries

eh-am:
Seems 2 times is not enough (I had a PR failing due to that), let's see if 5 is a good number.


petethepig (approved):
I can't say I'm happy about this change, but I also know that there's no easy solutions for this kind of stuff so let's try 5.

PR #546, opened 2021-11-17
feat: adds adhoc single view

eh-am:
just a quick poc


eh-am:
In an effort to **not** have long lived branches, I've introduced a new feature flag `--enable-experimental-adhoc-ui` and some foundation for exposing that to the frontend. Now this PR should be in a mergeable state (even if the code there is not optimal).

With `PYROSCOPE_ENABLE_EXPERIMENTAL_ADHOC_UI=true`:
image

When the flag is false/not set:
image

pkg/config/config.go
```diff
@@ -117,6 +117,7 @@ type Server struct {
 
 	AdminSocketPath         string `def:"/tmp/pyroscope.sock" desc:"path where the admin server socket will be created." mapstructure:"admin-socket-path"`
 	EnableExperimentalAdmin bool   `def:"false" desc:"whether to enable the experimental admin interface" mapstructure:"enable-experimental-admin"`
+  EnableExperimentalAdhocUI bool   `def:"false" desc:"whether to enable the experimental adhoc ui interface" mapstructure:"enable-experimental-adhoc-ui"`
```
petethepig, line 120: Same comment as here https://github.com/pyroscope-io/pyroscope/pull/561#discussion_r756818839

petethepig (approved):
@eh-am It looks good. Feel free to merge after we agree on the naming convention for boolean flags.

PR #549, opened 2021-11-18
Fix/comparison view

eh-am:
There was a bug where left/right values in comparison view would be duplicated.

This PR removes that bug by removing the `both` case, which simply does not occur, since `left` and `right` are loaded separately.

Also add a simple smoke test to validate that.
(At some point we will have to refactor all these cypress tests, since some of them are somewhat redundant/unnecessarily)


petethepig (approved):
:+1:

PR #551, opened 2021-11-18
Fix/delete apps

eh-am:
* Allow deleting apps even if their segment doesn't exist
* Increase timeout since the server may be busy
* Add a `--skip-verification` flag to delete to make it easy to run a command with for example scripts


kolesnikovae (approved):
Looks good!

As for repetitive flags/options - I think [embedding with mapstructure `squash`](https://pkg.go.dev/github.com/mitchellh/mapstructure#hdr-Embedded_Structs_and_Squashing) can help. Also, in cobra there are so-called persistent flags, though this can be hard/impossible in our setup.

BTW, consider `force` (as in many many other tools) or `auto-approve` (as in terraform, where a mistake can ruin the whole world) instead of `skip-verification`

eh-am:
@kolesnikovae good idea, done!

petethepig (approved):
LGTM

PR #553, opened 2021-11-19
fix: comparison view timeline

eh-am:
After a refactor we did, we ended up breaking the timeline for comparison when you just load the page.

So after spending some time reviewing the code, here's what I understood:

The [double render](https://github.com/pyroscope-io/pyroscope/issues/257) happened because we 
a) Loaded the flamegraph (which calls /render)
 b) Loaded the timeline (which also calls /render)

After that refactor, we now load just once, since we can infer the timeline from the /render response. However, that doesn't work with `comparison`, since there are 2 flamegraphs (what is the timeline coming from in this case?).

Summarizing it:

1 - For single view, we can infer the timeline. So just a single request is required.
2 - For diff, same thing. 
3 - For comparison, the timeline is actually decoupled from the flamegraphs. So we need to load it separately.

Honestly not very happy with the existing state. In the near future I would like to refactor it.


Rperry2174:
@eh-am definitely great that you got to the bottom of this... Yeah lets refactor this at some point

Rperry2174 (approved):
LGTM fixes the issue I was seeing earlier

PR #554, opened 2021-11-19
chore: use gorilla mux for admin server

eh-am:
This was previously discussed in a meeting. In the long run we want to use a framework (we chose gorilla)  for our http servers.

This PR adds gorilla for the admin server, which is a bit more scoped and therefore easier to understand its impacts.

Unfortunately the API isn't exactly 1:1 with `net.ServeMux` so it required a little refactor (to receive `http.Handler` instead).

Also add a logging middleware as 
a) an example of how to add a middleware 
and b) I often find useful to check in the logs whether the request got received or not.


petethepig (approved):
Looks good, merging

PR #560, opened 2021-11-24
fix: ignore ErrServerClosed when admin server is closed

eh-am:
While debugging code with @kolesnikovae we noticed the admin socket server wasn't handling shutdown properly, which was affecting other things such as the cache writeback.

### DIFF
--- a/pkg/cli/server.go
+++ b/pkg/cli/server.go
@@ -223,6 +223,12 @@ func (svc *serverService) Stop() {
 
 //revive:disable-next-line:confusing-naming methods are different
 func (svc *serverService) stop() {
+	if svc.config.EnableExperimentalAdmin {
+		svc.logger.Debug("stopping admin server")
+		if err := svc.adminServer.Stop(); err != nil {
+			svc.logger.WithError(err).Error("admin server stop")
+		}
+	}
 	svc.controller.Drain()
 	svc.logger.Debug("stopping discovery manager")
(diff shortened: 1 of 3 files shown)


pkg/cli/server.go
```diff
@@ -239,19 +245,15 @@ func (svc *serverService) stop() {
...
-	if svc.config.EnableExperimentalAdmin {
-		svc.logger.Debug("stopping admin server")
-		if err := svc.adminServer.Stop(); err != nil {
-			svc.logger.WithError(err).Error("admin server stop")
-		}
+
+	// storage needs to be the last thing to shutdown
+	svc.logger.Debug("stopping storage")
+	if err := svc.storage.Close(); err != nil {
+		svc.logger.WithError(err).Error("storage close")
```
kolesnikovae, line 256: I remembered why `controller` is stopped in the last turn:
1. We may want to observe metric values while storage is closing.
2. I think `/healthz` endpoint (which is used as liveness probe) should be still responding in order let storage flush everything to disk before server gets killed (although, I'm not 100% sure: e.g in kubernetes there is `terminationGracePeriodSeconds` option for this).

eh-am: interesting points, for 2) I wonder if we shouldn't leverage the `preStop` hook instead, so we have more confidence that our shutdown is going to finish

but for now I've undone the order change

PR #561, opened 2021-11-24
chore: add a flag to enable/disable self profiling

eh-am:
As previously discussed in slack.
Motivations are:
a) reduce clutter when running locally (specially the logs)
b) reduce disk/mem usage for certain cases
c) maybe incorporate in adhoc? cc @abeaumont 

I was considering adding a `noopAgent` or something like that, but I guess simply not launching `selfProfiling` is the easiest way to go about this.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -120,7 +120,7 @@ type Server struct {
 
 	ScrapeConfigs []*scrape.Config `yaml:"scrape-configs" mapstructure:"-"`
 
-	EnableSelfProfiling bool `def:"true" desc:"enable profiling of pyroscope itself" mapstructure:"enable-self-profiling"`
+	NoSelfProfiling bool `def:"false" desc:"disable profiling of pyroscope itself" mapstructure:"no-self-profiling"`
 }
 
 type MetricsExportRules map[string]MetricsExportRule
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -119,6 +119,8 @@ type Server struct {
 	EnableExperimentalAdmin bool   `def:"false" desc:"whether to enable the experimental admin interface" mapstructure:"enable-experimental-admin"`
 
 	ScrapeConfigs []*scrape.Config `yaml:"scrape-configs" mapstructure:"-"`
+
+	EnableSelfProfiling bool `def:"true" desc:"enable profiling of pyroscope itself" mapstructure:"enable-self-profiling"`
 }
 
 type MetricsExportRules map[string]MetricsExportRule
--- a/pkg/cli/server.go
+++ b/pkg/cli/server.go
@@ -125,7 +125,7 @@ func newServerService(c *config.Server) (*serverService, error) {
 	svc.debugReporter = debug.NewReporter(svc.logger, svc.storage, prometheus.DefaultRegisterer)
 	svc.directUpstream = direct.New(svc.storage, metricsExporter)
 
-	if svc.config.EnableSelfProfiling {
+	if !svc.config.NoSelfProfiling {
 		svc.selfProfiling, _ = agent.NewSession(agent.SessionConfig{
 			Upstream:       svc.directUpstream,
 			AppName:        "pyroscope.server",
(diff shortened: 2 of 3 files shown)


pkg/config/config.go
```diff
@@ -119,6 +119,8 @@ type Server struct {
 	EnableExperimentalAdmin bool   `def:"false" desc:"whether to enable the experimental admin interface" mapstructure:"enable-experimental-admin"`
 
 	ScrapeConfigs []*scrape.Config `yaml:"scrape-configs" mapstructure:"-"`
+
+	EnableSelfProfiling bool `def:"true" desc:"enable profiling of pyroscope itself" mapstructure:"enable-self-profiling"`
```
petethepig, line 123: This is a relatively minor, but what do you think about doing the opposite, e.g `DisableSelfProfiling`, and making it `false` by default?

The advantage is that when you're running it from command line you can just do `pyroscope --disable-self-profiling`. Otherwise you'd have to do something like `pyroscope --enable-self-profiling false` I think, which is slightly more wordy and less common.

petethepig: This is slightly separate but I'm also thinking maybe we should come up with some unified way of naming booleans. Right now we have multiple different styles, e.g some booleans start with `no-`, others start with `disable-`. I would probably choose to go with `no-` prefix, because then if you need to have a variable that's negative by default you can just name it without a prefix, e.g `--enable-experimental-admin` becomes just `-experimental-admin`. 

/cc @abeaumont @kolesnikovae  @Rperry2174 what do you all think about this?

petethepig (changes requested):

pkg/config/config.go
eh-am: These are the reasons I used `--enable`:

a) for better or for worse it's a boolean, and IMO booleans are better (in most cases) in the affirmative position. do you prefer `isDisabled` or `isEnabled`? what about `!isDisabled`? to me that becomes slightly confusing. then you can also have that classic double negative case of `!isNotPresent`, "we don't need no education" vibes

b) it's a toggle, a switch. It makes it easier to think that you are turning on that feature, even the UIs work better that way https://cdn.pulse2.com/cdn/2020/01/LaunchDarkly.jpg

c) RE `--enable-self-profiling false` vs `--disable-self-profiling` I don't think the brevity influences that much, since it's such an uncommon flag.

d) The point of feature flags is to add them early and them just flip them over. They are not attached to the usage patterns. With this proposal (some things are --disable and some are --enable), w we also have to think "what the default state is going to be" to come up with the name. 

e) Standardization. Goes a bit with point d) but I prefer to have everything following the same pattern.


I am sure I may be able to find even papers on this subject, but my experience says booleans should be used on the affirmative case:
https://www.google.com/search?q=booleans+afirmative

https://softwareengineering.stackexchange.com/questions/196830/boolean-method-naming-affirmative-vs-negative
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members?redirectedfrom=MSDN#names-of-properties
https://stackoverflow.com/questions/1445836/positive-or-negative-boolean-field-names
https://google.github.io/styleguide/jsguide.html#naming-method-names

kolesnikovae: I prefer short flags, vote for `no-` and omitting `enable-`.

petethepig: a,c,d) I agree with you that generally when programming affirmative names are better and it's a good practice. But in this particular context (command line interface) it seems to be a common practice for boolean flags to be inverted with --no prefix when the default is positive.

b) I agree that for some flags adding `enable-` prefix actually improves the name because it adds extra context making it clear that this is a boolean flag.

I still think `no-self-profiling` would be a better name here.

---

We're also mixing this conversation with the one about feature flags. I could see how for feature flags in particular a whole different approach might be better. I think that should be a separate converstaion.

eh-am: Ok will change it to `--no`.

>it seems to be a common practice for boolean flags to be inverted with --no prefix when the default is positive.
and when we figure out that users actually don't like having self profiling turned on and we decide to make it off by default?

eh-am:
When deploying for the first time it's possible to not have any appname:

```
Uncaught TypeError: window.initialState.appNames is null
```

petethepig (approved):

PR #566, opened 2021-11-28
docs: remove Go example obsolete comment

mhansen:
The comment refers to a hostname that was removed in 3286f05dae2235cc84248b81caa89a6af5e60e9d

### DIFF
--- a/examples/golang/main.go
+++ b/examples/golang/main.go
@@ -33,7 +33,7 @@ func slowFunction(c context.Context) {
 func main() {
 	profiler.Start(profiler.Config{
 		ApplicationName: "simple.golang.app",
-		ServerAddress:   "http://localhost:4040", // this will run inside docker-compose, hence `pyroscope` for hostname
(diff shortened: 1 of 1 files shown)


mhansen:
Dang, this looks like a custom CLA, dunno if I can sign this without
lawyercats getting involved; maybe you should just do a clean-room
implementation of removing/updating this comment yourselves haha

petethepig:
@mhansen I actually think I accidentally committed that with `localhost:4040`, it should be `pyroscope:4040`. The idea is that we have a docker compose file for each integration so that they all work the same way basically.

eh-am:
Can we take an env var there instead? I admit I also run that example app locally :)

petethepig:
@eh-am Maybe we could auto-detect if the app is running in docker or not?

mhansen:
It probably doesn't matter for this patch, but I didn't accept the CLA 5 days ago. I'm not sure why the CLA bot says I did. I visited the page to read the terms but didn't accept.

PR #567, opened 2021-11-29
fix: typo in readme

cfbolz:
flamgraph -> flamegraph


petethepig:
@cfbolz Thank you for the contribution!

I saw you're a major contributor to PyPy, I'm curious, how do profilers work with PyPy? Do standard tools work? Or do you have to write custom tooling for it?

cfbolz:
@petethepig hey :-). py-spy unfortunately doesn't work, because our frames are quite different (and with the JIT sometimes there are none at all). all the in-process profilers like scalene, cprofile work. we also wrote vmprof, but it's a bit languishing at the moment.

If you ever want to have a Zoom call and talk options, drop me a mail! cfbolz@gmx.de

PR #569, opened 2021-11-29
feat: delete apps with tags

eh-am:
As part of https://github.com/pyroscope-io/pyroscope/issues/287

Now we can also delete apps with tags.

Added few tests, but feel free to suggest any other kind tests if you find there's something missing.

Big thanks to @kolesnikovae for guiding me through this!


petethepig (approved):
Looks good, works great.

Thanks for the extra comments!

PR #570, opened 2021-11-30
fix: Prevent byte buffer pool copy.

abeaumont:
Since we are using a value receiver and a value pool, it was
efectively being copied.

Thanks @kolesnikovae for spotting it.


petethepig (approved):
:+1:

PR #571, opened 2021-11-30
Add a PoC for WebAssembly

abeaumont:
This is just a minimal working code to show that it works and try to get it integrated with the webpack/react ecosystem, without success so far.


cmd/pyroscope/main_js.go
```diff
@@ -0,0 +1,32 @@
+package main
```
petethepig, line 1: We could just move all of this to a whole separate package

petethepig (approved):

petethepig:
LOL at package-lock.json

.eslintignore
```diff
@@ -1,4 +1,5 @@
 /webapp/public/**
 /webapp/javascript/util/**
 /webapp/__tests__/**
+/scripts/**
```
petethepig, line 4: had to add this because otherwise I think linter uses webpack config which has fallbacks for `fs` and that breaks everything. at least that's how I understand the issue

babel.config.js
```diff
@@ -1,4 +1,4 @@
 module.exports = {
   presets: ['@babel/preset-env', '@babel/preset-react'],
-  plugins: ['transform-class-properties'],
+  plugins: ['transform-class-properties', '@babel/plugin-transform-runtime'],
```
petethepig, line 3: I do not know why we need this. I just know that without it the app errors out with "regeneratorRuntime is not defined"

abeaumont: I think this may contain some further details about the cause: https://github.com/babel/babel/issues/9849

PR #574, opened 2021-11-30
chore(grafana-panel): move grafana panel plugin to this repo

eh-am:
Basically this moves the grafana panel plugin from https://github.com/pyroscope-io/grafana-panel-plugin/ to here.

The idea is to make it easier to develop, test and release.

---
Had to ditch `@grafana/toolkit`, since a) it effectively redoes our existing setup (with ts, jest, eslint etc)  and b) doesn't allow for much extensability (even though it seems [they are hinting at going through that route](https://github.com/grafana/grafana/discussions/40568). I plan to clean that up, maybe even reusing `webpack.common.js`.

---

There are a plethora of other issues to be dealt with. They will be done in separate PRs.


petethepig (approved):
Looks good! This is a huge amount of work, good job @eh-am!

PR #576, opened 2021-12-01
feat: Add adhoc push mode support in clib.

abeaumont:
For native applications to work, they need to support the overriding
of the upstream address that ahdoc uses for this mode.

Instead of reusing PYROSCOPE_SERVER_ADDRESS enviroment variable,
a new variable PYROSCOPE_ADHOC_SERVER_ADDRESS has been used.
The users shouldn't be bothered with this override, using a separate
name should make it harder to see some undesired interactions.

Note that once this is merged the native integrations should be
upgraded to make use of this new clib for adhoc push to work.

### DIFF
--- a/pkg/adhoc/push.go
+++ b/pkg/adhoc/push.go
@@ -55,8 +55,9 @@ func (p push) Run() error {
 	// Note that we don't specify which signals to be sent: any signal to be
 	// relayed to the child process (including SIGINT and SIGTERM).
 	signal.Notify(c)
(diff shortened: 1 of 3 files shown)


eh-am:
I think you already thought this through, but there's no harm double checking it: can you think of a possible use case where `PYROSCOPE_ADHOC_SERVER_ADDRESS` would be used that's NOT adhoc? 

I'm asking that to check if we should name it something else, like `PYROSCOPE_OVERRIDE_SERVER_ADDRESS`

abeaumont:
I cannot think of any use case outside adhoc mode.

petethepig (approved):

PR #580, opened 2021-12-02
feat: adhoc comparison UI

abeaumont:
Initial work on polishing the UI for pyroscope adhoc:
- Cleanup of Adhoc Single.
- Initial support for Adhoc Comparison.


eh-am:
Nice one!

Just one thing: I'd suggest to put the uploaded flamebearer in the redux state, as we do with the other views.

abeaumont:
@eh-am done! please take another look :pray:

webapp/javascript/components/Sidebar.jsx
```diff
@@ -96,6 +96,21 @@ function Sidebar(props) {
     </>
   );
 
+  const adhocComparison = (
+    <>
+      <SidebarItem tooltipText="Adhoc Comparison View">
+        <NavLink
+          activeClassName="active-route"
+          data-testid="adhoc-comparsion"
```
eh-am, line 104:
```suggestion
          data-testid="adhoc-comparison"
```

webapp/javascript/components/AdhocComparison.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { bindActionCreators } from 'redux';
+import FlameGraphRenderer from './FlameGraph';
+import Footer from './Footer';
+import {
+  abortTimelineRequest,
+  setLeftFile,
+  setRightFile,
+} from '../redux/actions';
+
+// See docs here: https://github.com/flot/flot/blob/master/API.md
```
eh-am, line 14:
```suggestion
```

webapp/javascript/components/AdhocComparison.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+function AdhocComparison(props) {
+  const { actions, leftFile, leftFlamebearer, rightFile, rightFlamebearer } =
+    props;
+  const { setLeftFile } = actions;
+  const { setRightFile } = actions;
+
+  useEffect(() => {
+    return actions.abortTimelineRequest;
+  });
```
eh-am, line 24: I'm thinking this is not needed?

webapp/javascript/components/FileUploader.tsx
```diff
@@ -30,8 +30,14 @@ export default function FileUploader({ onUpload }: Props) {
           const s = JSON.parse(
             String.fromCharCode.apply(null, new Uint8Array(binaryStr))
           );
```
eh-am, line 32: I would handle this a little bit more carefully, since users may 

This is something we already discussed internally, we want in the future to validate ALL input using something like [zod](https://github.com/colinhacks/zod). But for now I think you can just do something quick and dirty and check if the required fields of a flamebearer exist.

webapp/javascript/components/FileUploader.tsx
```diff
@@ -30,8 +30,14 @@ export default function FileUploader({ onUpload }: Props) {
...
-          setFile({ file, ...s });
-          onUpload({ file, ...s });
+          const { flamebearer } = s;
+          const calculatedLevels = deltaDiffWrapper(
+            flamebearer.format,
+            flamebearer.levels
+          );
```
eh-am, line 37: Ugh I gate this deltaDiffWrapper`, this is more of a note to self, I think sometime in the near future we shall unify this initial processing. Or move to the backend.

webapp/javascript/components/FileUploader.tsx
```diff
@@ -30,8 +30,14 @@ export default function FileUploader({ onUpload }: Props) {
...
+
+          flamebearer.levels = calculatedLevels;
+          setFile(file, flamebearer);
         } catch (e) {
           console.log(e);
```
eh-am, line 42:
```suggestion
```

webapp/javascript/components/FileUploader.tsx
```diff
@@ -47,8 +53,7 @@ export default function FileUploader({ onUpload }: Props) {
   });
 
   const onRemove = () => {
-    setFile(null);
-    onUpload(null);
+    setFile(null, null);
```
eh-am, line 56: I am, personally, not a fan of `null`s. I would much rather use `Optional`s. But this is one of the cases it's fine IMO, up to you.

webapp/javascript/index.jsx
```diff
@@ -48,6 +49,11 @@ ReactDOM.render(
               <AdhocSingle />
             </Route>
           )}
+          {enableAdhoc && (
```
eh-am, line 52: IIRC this flag is hardcoded to `true`. You can use `isExperimentalAdhocUIEnabled` instead I think.

eh-am (commented):
We gonna have to create some kind of structure for turning feature flags on in PRs. CC @kolesnikovae 

For now can you just flip it in `config.go` @abeaumont?

webapp/javascript/components/FileUploader.tsx
abeaumont: Using optionals properly would require to also handle optional flamebearer I think, maybe something that can be considerend independently from this PR?

webapp/javascript/components/FileUploader.tsx
abeaumont: What would be the proposed approach? I have been looking if we could take advantage of the type definition and convert/cast the type, but I didn't find any clear way to do this, not sure if you know any. Or may be you are proposing something like:

```js
for field in ['names', 'levels', 'numTicks', 'spyName', 'format'] assert(field in flamebearer);
```

webapp/javascript/index.jsx
abeaumont: https://github.com/pyroscope-io/pyroscope/pull/580/commits/6e20912d59bd4369dfecaf3ce6b362d453139eb4

webapp/javascript/components/FileUploader.tsx
eh-am: Yeah, something like that! 

Just don't forget we need something to tell users how the file does not have the expected format. 

I know we have a notification component but I never got to use it. If you find it too out of your way, a classic `alert` can do the work TBH.

abeaumont: Added some basic checks in https://github.com/pyroscope-io/pyroscope/pull/580/commits/9b3bb8089b4b2cdd83f1936209f41610a83e752a. Once https://github.com/pyroscope-io/pyroscope/pull/596 is merged these checks should be moved to flamebearer decoder I think.

abeaumont:
I have integrated the new sidebar, I think this is good to review / merge

webapp/javascript/components/FlameGraph/FlameGraphRenderer.jsx
```diff
@@ -259,6 +260,14 @@ class FlameGraphRenderer extends React.Component {
         })}
       >
         <div className="canvas-container">
+          {this.props.uploader ? (
```
eh-am, line 263: Do we really need this here? Can we move this up to the parent component?

I don't like it having here, since for all effects the `FlamegraphRenderer` is the main Flamegraph component, as in this is the component other users* will interact with.
*By others users I mean the grafana plugin, eventually in the documentation, etc.

abeaumont: Right, with the rest of components out, there's no reason to keep the file uploader there either! Changed at https://github.com/pyroscope-io/pyroscope/pull/580/commits/165e7112c970d7c19c844f720814794658f0dd9e

eh-am (approved):
LGTM!

PR #581, opened 2021-12-02
feat: new sidebar

eh-am:
Implements a new sidebar 
image

It has a collapsed mode, which can be toggled on/off manually, but by default it's collapsed in smaller screens (windowWidth < 1200px)
image

Under the good it uses the `react-pro-sidebar` library, but tried to add indirection layers for the inevitable moment where we implement it ourselves/replace for another lib.

Also removed unused stuff we discussed in private (shortcuts, alerts) etc.


Rperry2174:
image

I think when the sidebar is wide that the logo should be aligned on the left side... also I think we should make the "Pyroscope" text be non-bold, a little bigger, and possibly aligned with the bottom? Wdyt?

eh-am:
Hey @Rperry2174! 
Eh, since it uses different font depending on the OS it's a bit difficult to get a stable experience here. I removed the font weight and aligned to the left.

PR #596, opened 2021-12-09
feat: improve pyroscope data format.

abeaumont:
Current data format used for rendering is not defined formally and has
a couple of issues:
- There's no format versioning, which makes it hard to evolve the
format or migrate between versions.
- Current format has redudancy, with duplicated metadata.

This may become a problem especially once adhoc mode is released, as
this format is directly exported to JSON.

I have created a new package with the flamebearer profile versioned
model, which should make the format more explicit and easier to
evolve. I have also removed the duplicated metadata.

### DIFF
--- a/webapp/javascript/redux/reducers/filters.js
+++ b/webapp/javascript/redux/reducers/filters.js
@@ -74,7 +74,23 @@ function decodeTimelineData(timelineData) {
   });
 }
 
-export default function (state = initialState, action) {
+function decodeFlamebearer({ flamebearer, metadata, leftTicks, rightTicks }) {
+  let fb = {
+    ...flamebearer,
+    format: metadata.format,
+    spyName: metadata.spyName,
+    sampleRate: metadata.sampleRate,
+    units: metadata.units,
+  };
+  if (fb.format == 'double') {
+    fb.leftTicks = leftTicks;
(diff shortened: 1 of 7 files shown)


webapp/javascript/redux/reducers/filters.js
```diff
@@ -74,6 +74,22 @@ function decodeTimelineData(timelineData) {
...
+    spyName: metadata.spyName,
+    sampleRate: metadata.sampleRate,
+    units: metadata.units,
+  };
+  if (fb.format === 'double') {
+    fb.leftTicks = leftTicks;
+    fb.rightTicks = rightTicks;
+  }
+  fb.levels = deltaDiffWrapper(fb.format, fb.levels);
+  return fb;
```
eh-am, line 90: Can we somehow parse/indicate in the frontend that's a `version: 1` that we are dealing with?

eh-am (approved):
LGTM!

webapp/javascript/redux/reducers/filters.js
abeaumont: I added the version field to the flamebearer model in https://github.com/pyroscope-io/pyroscope/pull/596/commits/39fc8f04295f8374b3e37b21d6777d59c2a6ffef

PR #598, opened 2021-12-09
refactor: allow using notifications with redux

eh-am:
For completeness, now allow notifications to be triggered via redux.

For example, server notifications:
```javascript
      dispatch(
        addNotification({
          message: notificationText,
          type: 'danger',
          dismiss: {
            duration: 0,
            showIcon: true,
          },
        })
      );
```

It will trigger two actions: `notifications/pending` and `notifications/fullfilled`. 
Currently their only purpose is to make it easier to visually inspect in the redux history, but in the future we can be fancy with it, for example, distinguishing between "Ok" and "Cancel" dismissals, or persisting which notifications we don't ever want to show again (for example one warning about new version available), etc.


petethepig (approved):
@eh-am Looks good! I I'm not sure how to fix the conflict, but it's good to go after that!

PR #604, opened 2021-12-10
chore (frontend): make flamegraphrenderer render children between toolbar and canvas

eh-am:
Currently the `FlamegraphRenderer` component conditionally renders stuff like the timeline, which makes it a bit difficult to integrate in other environments like storybook or grafana, since in these cases the flamegraph should be rendered alone, without a timeline. 

We could add conditionals to not render the timeline, but that still requires importing the required dependencies (eg Jquery) is not optimal.

This PR then allows the `FlamegraphRenderer` to take children components that will be rendered between the toolbar and the flamegraph canvas. Making the resulting DOM bit more explicit.


abeaumont (approved):
It looks good!

PR #607, opened 2021-12-13
refactor: move from io/ioutil to io and os packages

Juneezee:
The `io/ioutil` package has been deprecated in Go 1.16 (See https://golang.org/doc/go1.16#ioutil). This PR replaces the existing `io/ioutil` functions with their new definitions in `io` and `os` packages.

### DIFF
--- a/benchmark/internal/cireport/tablereport.go
+++ b/benchmark/internal/cireport/tablereport.go
@@ -5,13 +5,14 @@ import (
 	"context"
 	"embed"
 	"fmt"
-	"golang.org/x/sync/errgroup"
-	"gopkg.in/yaml.v2"
-	"io/ioutil"
 	"math"
+	"os"
 	"sync"
 	"text/template"
 	"time"
+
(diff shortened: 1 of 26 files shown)


petethepig (approved):
@Juneezee Thanks for contributing, it looks great! Do you know if there's any way to enforce this? Maybe there's some linter or something like that we could use to make sure we don't use ioutil going forward?

Juneezee:
> @Juneezee Thanks for contributing, it looks great! Do you know if there's any way to enforce this? Maybe there's some linter or something like that we could use to make sure we don't use ioutil going forward?

@petethepig Hi! Thanks for reviewing this PR. We can use the `depguard` in `golangci-lint`, see https://golangci-lint.run/usage/linters/#depguard

petethepig:
@Juneezee Awesome, that would be quite useful! I'm going to merge this one, if you want to add `depguard` from `golangci-lint` to github actions feel free to do so in a separate PR, but no worries if you don't :)

PR #608, opened 2021-12-13
chore(frontend): create box component

eh-am:
Creates a box component, which simply creates a box around it. Example:
image

It was extracted from the flamegraph, now making it declarative:
```js
<Box>
  <FlamegraphRenderer />
</Box>
```

The motivation is that we don't always want to render that box, so it's left to the caller. 
(An example of not requiring the box is the grafana panel, another example is rendering the Flamegraph into the documentation)

Since I was at it, also fixed the alignment/spacing of pages. For example, previously:
Previously             |  Now
:-------------------------:|:-------------------------:
image | image 

Previously             |  Now
:-------------------------:|:-------------------------:
image | image

We achieve that by moving the spacing into the parent component. So now the children don't have to set up the margin themselves.


petethepig (approved):
:+1:

Rperry2174:
this is awesome @eh-am thanks for cleaning all this up!

PR #609, opened 2021-12-14
fix: avoid converting args to double dash and disable flag parsing for adhoc.

abeaumont:


cmd/pyroscope/command/adhoc.go
```diff
@@ -46,6 +46,7 @@ provided. It can also be override the default exec mode with the '--push' flag.
 thorugh '--url' where it tries to retrieve profiling data in any of the
 supported formats. In this case arguments are optional, and if provided,
 they are used to launch a new process before polling the URL.`,
+		DisableFlagParsing: true,
```
petethepig, line 49: this seems extreme, how does it parse flags like --url now? Do we do it explicitly somewhere now?

abeaumont: Yeah, this is quite unintuitive, and I don't yet get all the details of how our parsing works, but I'll try to elaborate a bit on this.
There seems to be internal support for flag/argument parsing (in the [cli](https://github.com/pyroscope-io/pyroscope/blob/main/pkg/cli/) package) and this implementation already supports the expected _flags should appear before args_ logic, while cobra doesn't understand it. This means there's no need for cobra to do the parsing, but if we do so, it'll break whenever a command with their own flags is used.
This support seems to have been introduced in https://github.com/pyroscope-io/pyroscope/commit/50351b1dddd13fc488c322939961a489aad72368#diff-9a73f16f33e1ba62848f39fe5e3171a461913778f019d27fbd9bba6ea12d0ea2, possibly to solve the same problem `adhoc` is currently facing, and currently most commands have this flag set (`agent`, `server`, `exec`, `convert` and `dbmanager`).

petethepig (approved):

PR #611, opened 2021-12-15
fix: assets handling when using base-url argument

### LINKED ISSUE #610: `base-url` is not used correctly for all web resources
The `base-url` parameter doesn't seem to prefix links to assets (CSS/JS,etc).

We run a central nginx instance for each cluster which reverse-proxies to various services. The URL `https://<cluster>/directory-loki/pyroscope/loki-dev-001` results in a blank page, even if I specify the `base-url` as `/directory-loki/pyroscope/loki-dev-001`; all of the assets 404 because they're not using that prefix.

petethepig:
Fixes #610

As I was debugging the assets issue I also noticed that oauth flow was broken in base-url mode so I fixed that in this PR as well.

Here's a list of main changes:
* I made it so that we include BaseURL when referring to js and css files in html templates, this makes URLs absolute when using with BaseURL and solves the issue of missing assets
* I changed the images `publicPath` in webpack config to be relative. The idea here is that this way the image urls we'll always be relative to the CSS file and so this will work in all cases
* for 1 image that was loaded from JS I moved that class to CSS for now as well. The fundamental issue there is that after the previous change (`publicPath`) loading images from HTML (as opposed to from CSS) via relative paths no longer works. So this is a workaround. @eh-am For now I'm thinking let's just agree to only reference to images from CSS, and for a long term plan, let's discuss this on a call soon.
* for oauth flow, I had to make changes so that any time we do redirects we include base URL in the URL we're going to.

Overall I'm not super happy with this solution because I feel like it's going to be hard to maintain. These tweaks I made are not obvious and suddle, and I'm worried when we add new pages / features / images we're likely to break base URL functionality again. @eh-am this is not urgent, but I think having cypress tests would help massively in this case so let's discuss this soon as well.


pkg/server/controller.go
```diff
@@ -81,6 +83,13 @@ type Notifier interface {
 }
 
 func New(c Config) (*Controller, error) {
+	if c.Configuration.BaseURL != "" {
+		_, err := url.Parse(c.Configuration.BaseURL)
+		if err != nil {
+			logrus.Error("base URL is invalid, some redirects might not work as expected")
```
Dzalevski, line 89: @petethepig Maybe you should return the err coz if you use that controller after this log this will lead to nil pointer error

pkg/server/controller.go
```diff
@@ -310,6 +319,19 @@ func (ctrl *Controller) isAuthRequired() bool {
 	return ctrl.config.Auth.Google.Enabled || ctrl.config.Auth.Github.Enabled || ctrl.config.Auth.Gitlab.Enabled
 }
 
+func (ctrl *Controller) redirectPreservingBaseURL(w http.ResponseWriter, r *http.Request, urlStr string, status int) {
+	if ctrl.config.BaseURL != "" {
```
Dzalevski, line 323: @petethepig you are already doing that in Constructor of Controller New()?

pkg/server/controller.go
petethepig: agreed, I changed it a bit. Thanks for the tip @Dzalevski !

pkg/server/controller.go
petethepig: yeah, but I need to modify the url so I'm parsing it again. it's cheap and prevents any potential issues related to reusing an object (url object in this case)

PR #615, opened 2021-12-16
fix: use the correct controller variable.

abeaumont:


eh-am (commented):
Good catch but I don't think this is even being called :thinking:

abeaumont:
Yeah, you are right, it's not, let's do some clean-up then.

PR #616, opened 2021-12-16
feat(frontend): allow to export flamegraph json

eh-am:
Not optimal but it works.
Currently scoped to `single` flamegraphs (ie. singel view, comparison diff but NOT diff view, since we don't support importing that yet).

Had to store the raw response from /render in the store since that's what adhoc expects. Yadda yadda `deltaDiffWrapper` is troublesome.

Peek 2021-12-16 11-20

### DIFF
--- a/webapp/javascript/components/ExportData.tsx
+++ b/webapp/javascript/components/ExportData.tsx
@@ -5,8 +5,13 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faBars } from '@fortawesome/free-solid-svg-icons/faBars';
 
 import clsx from 'clsx';
+import { Flamebearer } from '@models/flamebearer';
 
-function ExportData() {
+interface ExportDataProps {
+  exportFlamebearer?: Flamebearer;
+}
+function ExportData(props: ExportDataProps) {
+  const { exportFlamebearer } = props;
   const [toggleMenu, setToggleMenu] = useState(false);
 
   const formattedDate = () => {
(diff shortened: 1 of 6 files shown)


webapp/javascript/components/ExportData.tsx
```diff
@@ -44,6 +49,21 @@ function ExportData() {
     setToggleMenu(!toggleMenu);
   };
 
+  const downloadFlamebearer = function (
+    exportObj: Flamebearer,
+    exportName: string
+  ) {
+    const dataStr = `data:text/json;charset=utf-8,${encodeURIComponent(
+      JSON.stringify(exportObj)
```
petethepig, line 57: for large files, could we do use something like this? https://github.com/eligrey/FileSaver.js

petethepig:
Love it! How hard would it be to also add pprof export option? You can use the same render url and set `format=pprof` to make it work. I guess a slight complication is that you might also have to replace `from` and `until` parameters (because default ones are relative to `now`).

Rperry2174:
After looking more at this I'm wondering would it be easier to add in the "metadata" that you were mentioning as part of this PR? We could also do separately if that's going to complicate things somewhere else

webapp/javascript/components/ExportData.tsx
eh-am: Good idea, I'm going to be bold here and leave as it is until it become a problem though. I believe our flamegraphs are small enough for that to become an issue.

eh-am:
@petethepig @Rperry2174 Let's do pprof in a diff PR, since that requires updates to the metadata.

PR #620, opened 2021-12-17
feat: export pprof format

eh-am:
Add a POC of exporting to pprof format

Peek 2021-12-16 21-17

CC: @abeaumont where to put that metadata. As you can see it's used to generate the download URL, so I enhanced the response instead of adding to the `Flamebearer` interface. Let me know what approach you find best.


abeaumont:
It looks good!

eh-am:
Now that I am thinking, this may pose a problem when running in Grafana, since the /render URL can be completely different.

gawicks:
@eh-am  didn't realize there was already an open PR on exporting pprof. Closing #645 in favour of this!

webapp/javascript/components/ExportData.tsx
```diff
@@ -64,6 +65,21 @@ function ExportData(props: ExportDataProps) {
...
+    const url = `${buildRenderURL({
+      from: flamebearer.metadata.startTime,
+      until: flamebearer.metadata.endTime,
+      query: flamebearer.metadata.query,
+      maxNodes: flamebearer.metadata.maxNodes,
+    })}&format=pprof`;
+
+    const downloadAnchorNode = document.createElement('a');
+    downloadAnchorNode.setAttribute('href', url);
+    document.body.appendChild(downloadAnchorNode); // required for firefox
```
gawicks, line 78: Just curious, instead of doing this with javascript wouldn't it be simpler to use an anchor tag in the export menu?

eh-am: Hey!

That's a really good question, there are a few reasons (one stronger, the others meh):

1 - We are already doing that for png/JSON: 
https://github.com/pyroscope-io/pyroscope/blob/3435a212aa659d5240f84702926833d351a9d44f/webapp/javascript/components/ExportData.tsx#L53-L64
https://github.com/pyroscope-io/pyroscope/blob/3435a212aa659d5240f84702926833d351a9d44f/webapp/javascript/components/ExportData.tsx#L27-L44

By maintaining consistency it becomes easier to a) understand the code b) refactor c) test

2 - There's an argument to be made that the anchor tag being used for navigating to another page and button for actions. Here we aren't really navigating to anywhere.

3 - The operation is delayed and obscured, which currently does not seem to matter >much<, but it will for cases like grafana where you need bit more involvement than simply pointing to a URL.

But honestly, both ways work. I woulnd't care too much about this.

lizthegrey:
Oh hell yes. Being able to examine accumulated profiles in the language native tooling for local analysis with local binaries (e.g. for disassembling to asm) would be handy.

eh-am:
Unfortunately not all features are there, like accessing the source or disassembling :(

petethepig (approved):

PR #624, opened 2021-12-17
Adhoc API

abeaumont:
Initial support for server side profile listing and rendering in adhoc mode.


pkg/adhoc/server/server.go
```diff
@@ -0,0 +1,119 @@
...
+
+func New(log logrus.FieldLogger, enabled bool) Server {
+	return &server{log: log, enabled: enabled, profiles: make(map[string]profile)}
+}
+
+func (s *server) AddRoutes(r *mux.Router) http.HandlerFunc {
+	if s.enabled {
+		r.HandleFunc("/v1/profiles", s.Profiles)
+		r.HandleFunc("/v1/profile/{id:[0-9a-f]+}", s.Profile)
+		r.HandleFunc("/v1/diff/{id1[0-9a-f]+}/{id2[0-9a-f]+}", s.Diff)
```
eh-am, line 46: I would love to start prefixing all of these with `/api`, I think you can do that globally with gorilla.

One immediate case is https://github.com/pyroscope-io/pyroscope/blob/main/scripts/webpack/webpack.dev.js#L31

pkg/adhoc/server/server.go
```diff
@@ -0,0 +1,119 @@
...
+//
+// The profiles are retrieved every time the endpoint is requested,
+// which should be good enough as massive access to this auth endpoint is not expected.
+func (s *server) Profiles(w http.ResponseWriter, _ *http.Request) {
+	profiles := make(map[string]profile, 0)
+	err := filepath.Walk(util.DataDirectory(), func(path string, info fs.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if info.Mode().IsRegular() {
```
eh-am, line 64: Maybe we should check if the file is valid. For example I created a `foo` file in the $PYROSCOPE_HOME_DIR

image

pkg/adhoc/server/server.go
```diff
@@ -0,0 +1,119 @@
...
+	s.profiles = profiles
+	s.mtx.Unlock()
+	w.Header().Set("Content-Type", "application/json")
+	s.mtx.RLock()
+	defer s.mtx.RUnlock()
+	if err := json.NewEncoder(w).Encode(s.profiles); err != nil {
+		s.log.WithError(err).Error("Unable to encode the profile list")
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
```
eh-am, line 88: Maybe we should split looking up files and sending responding

webapp/javascript/redux/actions.js
```diff
@@ -158,6 +167,28 @@ export const setRightFile = (file, flamebearer) => ({
   payload: { file, flamebearer },
 });
 
+export const requestProfiles = () => ({ type: REQUEST_PROFILES, payload: {} });
```
eh-am, line 170: I would suggest to qualify these actions a bit more, "requesting profiles" is too broad, maybe "request adhoc profiles"?

webapp/javascript/redux/actions.js
```diff
@@ -395,3 +426,67 @@ export function abortFetchNames() {
...
+    dispatch(requestProfiles());
+    return fetch('/adhoc/v1/profiles', {
+      signal: profilesController.signal,
+    })
+      .then((response) => response.json())
+      .then((data) => {
+        dispatch(receiveProfiles(data));
+      })
+      .catch((e) => {
+        if (!isAbortError(e)) {
```
eh-am, line 447: Let's dispatch an error notification here!

pkg/adhoc/server/server.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/gorilla/mux"
+	"github.com/sirupsen/logrus"
+
+	"github.com/pyroscope-io/pyroscope/pkg/adhoc/util"
+)
+
+type profile struct {
+	ID        string    `json:"id"`
+	Name      string    `json:"name"`
+	UpdatedAt time.Time `json:"updated_at"`
```
eh-am, line 24:
```suggestion
	UpdatedAt time.Time `json:"updatedAt"`
```

Based on the other fields I believe we are a camelCase team :laughing:

eh-am (commented):
I am digging it!

Couple frontend nitpicks:
There's not much space for the file list itself, which is the most important bit. I suggest to remove the spacing beteen table/tab, also make the rows height smaller. 
image


The other thing is setting a fixed height for both "Pyroscope data" and "Upload" tabs, that way you don't have the that content shift illustrated here:
Peek 2021-12-17 10-46

Apart for that I am loving it!!

webapp/javascript/redux/actions.js
eh-am: I think this can be done more easily once this PR is merged https://github.com/pyroscope-io/pyroscope/pull/625

pkg/adhoc/server/server.go
abeaumont: Done at https://github.com/pyroscope-io/pyroscope/pull/624/commits/f1d205dbbdf35945f8acdd803d1c7529ac863992

webapp/javascript/redux/actions.js
abeaumont: Right, once that's merged I'll change these accordingly :+1:

webapp/javascript/redux/actions.js
abeaumont: I specialized all the adhoc related actions at https://github.com/pyroscope-io/pyroscope/pull/624/commits/8eb4b96bcff07204009d9fc46383bc8442edc088

pkg/adhoc/server/server.go
abeaumont: What do you mean?

eh-am: I am suggesting a more MVC approach, splitting the controller and the service, instead of doing everything (business logic and response sending) directly in the handler.

abeaumont: Ah, yes. There's so little logic right now that it's a bit overkill, but as it grows we may want to separate the business logic from the formatting :+1:

webapp/javascript/redux/actions.js
abeaumont: I ended up overlapping the functionality with changes in  https://github.com/pyroscope-io/pyroscope/pull/624/commits/75bcb8f9e7caaa87b3c41b42d9e45b7a77251ecd and https://github.com/pyroscope-io/pyroscope/pull/624/commits/53188ec9436ec0c9e304d967a8158a56702db8a1 :facepalm:

abeaumont: I combined the approaches when merging at https://github.com/pyroscope-io/pyroscope/pull/624/commits/f9d0fdac3555d9f77c2fbeb00173b8fcade0bbef
Also added cancellation support in https://github.com/pyroscope-io/pyroscope/pull/624/commits/2471f56976b5b41ad087d3d5aa7021a3155d9f46

pkg/adhoc/server/server.go
abeaumont: Added some basic server-side validation in https://github.com/pyroscope-io/pyroscope/pull/624/commits/dea2544a7413afd5f5b0be67662fe08dc9dd8d42

eh-am:
Hey, this is looking pretty good!

One immediate thing I found missing is a hover effect/indicate what version you are selecting. I've added something quick, check it out:
Peek 2021-12-21 14-07

petethepig (approved):
Looks good. We should probably just mkdir the data directory for users if it's empty. Right now if you run it without the directory it shows an error.

PR #628, opened 2021-12-17
perf: optimize pprof parsing in pull mode.

abeaumont:
Our profiling data suggested that a lot of time was invested in teh
binary search to find locations and functions while writing scraped
profiles. This is an attempt to improve the performance by
preprocessing the functions and locations and putting them into a
table.

A benchmark is included to showcase the results with smaller and
bigger profiles. As expected, there's no gain with small profiles,
quite the opposite, as there's now an extra preprocessing.
On the other hand, there are big gains as profiles get bigger (2x for
the bigger case).

While it'd be possible to find some heuristic to disable the
optimization below a certain threshold, the absolute difference
between small and big profiles is so big that I think it's not worth
it, at least as a first approach.


petethepig:
Going to try this with #630

petethepig:
Ran it with #630 — it looks better, but now there's this mapaccess function that takes up a lot of time.
Screen Shot 2021-12-17 at 5 36 28 PM

abeaumont:
Yes, that's the tradeoff, using a map instead of the binary search. Looking deeper, the binary search already expects the identifiers to be consecutive numbers starting at 1, this means we can get rid of the map and use a slice instead. The results should be better now even for smaller profiles.

Using a map for caching (first change https://github.com/pyroscope-io/pyroscope/pull/628/commits/4aaec20d5fa6ca93ade206cc6db6fc39b023b414):

```
BenchmarkProfile_ParseNoCache/Locations:_29,_functions_28-16              248778              4098 ns/op
BenchmarkProfile_ParseWithCache/Locations:_29,_functions_28-16            170401              6764 ns/op
BenchmarkProfile_ParseNoCache_Big/Locations:_1145,_functions_1145-16         120           9858162 ns/op
BenchmarkProfile_ParseWithCache_Big/Locations_1145,_functions_1145-16                216           5477587 ns/op
```

Using a slice for caching (second change https://github.com/pyroscope-io/pyroscope/pull/628/commits/14eaf48cc659ff0fcdf23a7d97c954d7355ef588):
```
BenchmarkProfile_ParseNoCache/Locations:_29,_functions_28-16              290997              4151 ns/op
BenchmarkProfile_ParseWithCache/Locations:_29,_functions_28-16            467185              2513 ns/op
BenchmarkProfile_ParseNoCache_Big/Locations:_1145,_functions_1145-16         109           9495172 ns/op
BenchmarkProfile_ParseWithCache_Big/Locations_1145,_functions_1145-16                314           3826945 ns/op
```

abeaumont:
I have run the second change with #630 too, and it can be seen how there's no downside now:

issue-628

petethepig:
@abeaumont Awesome. I also looked at other metrics like memory usage and everything is looking great there as well. I'm going to finish #630, make it post one of those summaries and then merge this one.

abeaumont:
I have checked the pprof spec and unfortunately, consecutive IDs are not guaranteed:  https://github.com/google/pprof/blob/master/proto/profile.proto#L164-L165. I'll merge with both the slice and map approaches and make the slice part of the fast path.

abeaumont:
I added generic support to any valid location/function format as specified in the pprof format in https://github.com/pyroscope-io/pyroscope/pull/628/commits/217066ee6bfac37e26c8ed2cf357624eee3fa433, while still retaining the slice-based fast path. As suggested by @kolesnikovae, we don't even create a new slice now, as the IDs are usually already sorted (and we can sort them when they are not).

I also added some tests as the implementation got quite a bit more complex.

petethepig (approved):

PR #630, opened 2021-12-18
perf: benchmarking code improvements

petethepig:
improves existing benchmarking code, adds pull mode benchmarks. 

---

To run pull benchmarks run:
```bash
# 100 is number of clients that are being simulated
./start 100
```

You should be able to see:
* grafana dashboard with metrics at http://localhost:8080
* pyroscope instance at http://localhost:4040

/cc @abeaumont @kolesnikovae you all might be interested in pull mode benchmark


benchmark/pull/server.yml.example
```diff
@@ -0,0 +1,11 @@
...
+# log-level: debug
+scrape-configs:
+  - job-name: testing
+    enabled-profiles: [cpu, mem]
+    static-configs:
+      - application: pull-target
+        targets:
+          - pull-target:4042
+        labels:
+          pod: pod-0
```
kolesnikovae, line 11: I think we could even create a fake discovery mechanism that would produce targets (with unique scrape url).

petethepig: Yeah, that would be useful here, although generating these configs a is also pretty easy.

What do you mean by unique scraping URL? Is it important for them to have different ones?

abeaumont:
This is a great help when working on performance improvements!

PR #633, opened 2021-12-20
chore(dev): improve live reload server setup

eh-am:
previously we were handling responses, but not timeouts
turns out the go server will either respond with 2xx or won't respond at
all

this commit changes the logic to add retries for timeouts


Rperry2174 (approved):
works for me now -- thanks!

petethepig:
@eh-am feel free to merge after resolving the conflict

petethepig (approved):

PR #635, opened 2021-12-20
fix(frontend): fixes golang package name coloring

petethepig:
the important part that we decide what color to show is `packageName`. tested with pyroscope stacktraces here: https://regex101.com/r/99KReq/1


eh-am:
@petethepig To update the snapshots run `yarn cy:ss`

PR #637, opened 2021-12-21
chore(frontend): add tests for momentjs usage

eh-am:
Add tests for moment js usage, which will make things easier when we replace it for `date-fns`
RE: https://github.com/pyroscope-io/pyroscope/issues/447


petethepig (approved):
Looks good, thanks for adding this!

PR #646, opened 2021-12-23
fix: try to create data directory if it doesn't exist.

abeaumont:


eh-am:
Curious why not doing that in `adhoc/util`, something like `ensureDataDir()` which can then be called in both `adhoc/writer.go` and here.

abeaumont:
I considered it wasn't worth creating the new function just a single line, but it may be more clear that way :+1:

PR #650, opened 2021-12-23
fix(frontend): add tab panel styles to adhoc comparison component.

abeaumont:
Otherwise, the file list component's height is unbounded. cc: @eh-am


Rperry2174 (approved):
lgtm!

PR #654, opened 2021-12-23
build: speed up docker builds

eh-am:
This PR does 2 things:

* Move installing yarn dependencies at the top of the Dockerfile. That way we don't have to pull dependencies if `package.lock/yarn.lock` haven't changed.
* Use [`RUN --mount=type=cache`](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md#run---mounttypecache) to cache dependencies between builds. This required updating our CI scripts to use the env var `DOCKER_BUILDKIT=1`

There's more room for improvements, but this should be a good start.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,11 +55,14 @@ WORKDIR /opt/pyroscope
 
 COPY scripts ./scripts
 COPY webapp ./webapp
-COPY package.json yarn.lock babel.config.js .eslintrc .eslintignore .prettierrc tsconfig.json Makefile ./
+COPY package.json yarn.lock Makefile ./
+# we only need the dependencies required to BUILD the application
+RUN make install-build-web-dependencies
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -54,13 +54,17 @@ RUN apk add --no-cache make
 WORKDIR /opt/pyroscope
 
 COPY scripts ./scripts
+COPY package.json yarn.lock Makefile ./
+
+# we only need the dependencies required to BUILD the application
+RUN --mount=type=cache,target=/usr/local/share/.cache/yarn/v6 make install-build-web-dependencies
```
petethepig, line 60: Didn't know about `--mount=type=cache`. Looks like exactly what we need, awesome find!

petethepig (approved):

PR #661, opened 2021-12-26
feat(frontend): sync search in comparison view

GorvGoyl:


GorvGoyl:
reference https://github.com/pyroscope-io/pyroscope/issues/286

Rperry2174:
Reposting thoughts regarding tests to get input from others:
image

Here there are two issues I think:
```
      <DebounceInput
        data-testid="flamegraph-search"
        data-testname={`flamegraph-search-${viewSide}`}
        className={`${styles.search} ${
          showMode === 'small' ? styles['search-small'] : ''
        } ${
          searchInputType === 'linked' ? styles['linked-search-input'] : ''
        } ${isSearchLinked === true ? styles.active : ''}`}
```
So here there is a space between each ternary. so even when the ternaries return `''` they really return  `[space]''`  for each ternary. Might be a cleaner way to add these additional classes in this case.

For the other part `data-testname="flamegraph-search-undefined"`  did you change anything that would cause for `viewSide` to go from being left or right to being undefined?

It looks from this PR like you added `data-testname` not sure if we need it (?) 

cc/ @eh-am

GorvGoyl:
> So here there is a space between each ternary. so even when the ternaries return '' they really return [space]'' for each ternary. Might be a cleaner way to add these additional classes in this case.

extra space at the end in `class` shouldn't cause any issues in UI. Similar scenario for `Single View` page when the search bar is not small.

```
className={`${styles.search} ${
          showMode === 'small' ? styles['search-small'] : ''
        }
``` 
I did some refactoring but extra space is still there unless I do some more string manipulation. Is it possible to update the snapshot so it matches the latest test result? 


> It looks from this PR like you added data-testname not sure if we need it 

correct, I refactored the code to use data-testid for linked search test case

webapp/javascript/components/ComparisonApp.jsx
```diff
@@ -24,6 +24,51 @@ function ComparisonApp(props) {
...
+    setLinkedSearch((x) => {
+      return {
+        ...x,
+        linkedSearchQuery: query,
+      };
+    });
+  };
+  const toggleLinkedSearch = (side) => {
+    const { isSearchLinked } = linkedSearch;
+    if (isSearchLinked) {
```
eh-am, line 43: Can we return early here if `isSearchLinked` is false? The nesting here doesn't add much value IMO.

webapp/javascript/components/ComparisonApp.jsx
```diff
@@ -24,6 +24,51 @@ function ComparisonApp(props) {
...
+      if (side === 'left') {
```
eh-am, line 44: Can we use `switch/case` here? This file isn't using typescript yet, but in the future we can update it to use exhaustive case checks.

eh-am: Plus these cases are pretty similar, so we can just build an `args` parameter and then declaratively use it?

Eg

```js
let args;
if (side === 'right') {
  args = { resetLinkedSearchSide: 'left' }
}
if (side === 'both') {
  args = { sSearchLinked: false, resetLinkedSearchSide: '' }
}
(...)

setLinkedSearch((x) => {
   return { ...x,  ...args };
});
```

eh-am: Also I think having all this logic here bloats this file. These files (Coparison, SingleView, ComparisonDiff) are the entrypoints for our pages, ie the first or second thing people should read to understand what the page is about. Having this logic makes it harder for people to understand it.

webapp/javascript/components/ComparisonApp.jsx
```diff
@@ -61,6 +106,11 @@ function ComparisonApp(props) {
               data-testid="flamegraph-renderer-left"
               display="both"
               rawFlamegraph={rawLeft}
+              isSearchLinked={linkedSearch.isSearchLinked}
+              setSearchQuery={setSearchQuery}
+              linkedSearchQuery={linkedSearch.linkedSearchQuery}
+              toggleLinkedSearch={toggleLinkedSearch}
+              resetLinkedSearchSide={linkedSearch.resetLinkedSearchSide}
```
eh-am, line 113: This is case where lifting state up makes the parent component too bloated. I think you can infer based on the toolbar `viewType` whether to show the link, and then dispatch a redux action to enable/disable the link functionality.

webapp/javascript/components/FlameGraph/FlameGraphRenderer.jsx
```diff
@@ -57,6 +57,22 @@ class FlameGraphRenderer extends React.Component {
     if (prevState.flamegraphConfigs !== this.state.flamegraphConfigs) {
       this.updateFlamegraphDirtiness();
     }
+
+    if (this.props.isSearchLinked) {
+      if (this.props.linkedSearchQuery !== prevState.highlightQuery) {
+        // eslint-disable-next-line react/no-did-update-set-state
```
eh-am, line 63: Any special reason for disabling the linting rule?

webapp/javascript/components/Toolbar.tsx
```diff
@@ -179,22 +193,88 @@ function FocusOnSubtree({
...
+            isSearchLinked === true ? 'Unsync search term' : 'Sync search term'
+          }`}
+          onClick={() => {
+            toggleLinkedSearch();
+          }}
+          className={`${styles['linked-search-btn']} ${
+            isSearchLinked === true ? styles.active : ''
+          }`}
+        >
+          <svg
```
eh-am, line 262: Isn' there a fontAwesome version of this? Example https://github.com/pyroscope-io/pyroscope/blob/main/stories/Button.stories.tsx#L26

webapp/javascript/components/Toolbar.tsx
```diff
@@ -179,22 +193,88 @@ function FocusOnSubtree({
...
+          onHighlightChange(e.target.value);
+        }}
+      />
+      {searchInputType === 'linked' && (
+        <button
+          data-testid={`link-search-btn-${viewSide}`}
+          type="button"
+          title={`${
+            isSearchLinked === true ? 'Unsync search term' : 'Sync search term'
+          }`}
```
eh-am, line 254: Not a big fan of using titles for this cases, but I guess until we have proper tooltip support this is fine.

eh-am (commented):
Hey, that's pretty cool! I have a few comments:

---
Maybe it's just me, but I found a bit weird how removing the link also deletes the search. For comparison, in grafana, when you unsync 2 time ranges, they continue the same, just unlinked.

---
Functionality is missing on the Adhoc Comparison page.
image

---
How is the button going to work when we add a button to clear the search, like  ?

---
I would love to see some unit tests here!

---
The button is missing a hover cursor.

GorvGoyl:
> Functionality is missing on the Adhoc Comparison page.

Is this different from `/comparison` page? what's the URL here?

GorvGoyl:
- [ ] add hover cursor to button.

webapp/javascript/components/ComparisonApp.jsx
GorvGoyl: done.

webapp/javascript/components/FlameGraph/FlameGraphRenderer.jsx
GorvGoyl: eslint rule disallows updating state inside componentdidupdate. but, it's actually fine if we're updating the state conditionally. https://stackoverflow.com/a/30529826/3073272

eh-am:
> > Functionality is missing on the Adhoc Comparison page.
> 
> Is this different from `/comparison` page? what's the URL here?

You have to run the server with `--enable-experimental-adhoc-ui` enabled, or via env var `PYROSCOPE_ENABLE_EXPERIMENTAL_ADHOC_UI`. We will release after @abeaumont finishes the implementation, which is almost done here https://github.com/pyroscope-io/pyroscope/pull/652

Then it will be enabled in the sidebar ->
image



> > How is the button going to work when we add a button to clear the search
> 
> clear input text button is already there on hover. image

Sorry, I don't see it, am I missing something?
image

But anyway, that answers my question, it will be inside the input, so it will still look good!


> > would love to see some unit tests here!
> 
> I added some tests in `basic.ts` file. Where do I add more tests?

That's a e2e/integration test. For that I would simply 1 - Type the search 2 - Click the link button 3 - Check if the query was copied to the other field.
There's no need to test the highlight again IMO.

Now for adding the unit tests, it depends how the code is organized. You can try putting 2 toolbars side by side and seeing if the sync works. Or if that's too difficult, to put 2 flamegraphs side by side. The advantage of unit test here is that it forces you to check the API. If it's too difficult to test, then it's too difficult to (re)use the component.

webapp/javascript/components/FlameGraph/FlameGraphRenderer.jsx
eh-am: Fair enough, can you somehow make it clear on the code? Otherwise the next person reading will (probably) have the same question, and that time they probably won't be scavenging through github issues to understand it.

webapp/javascript/components/ComparisonApp.jsx
Rperry2174: @eh-am where would you put this instead?

webapp/javascript/components/FlameGraph/FlameGraphRenderer.jsx
GorvGoyl: done

webapp/javascript/components/ComparisonApp.jsx
GorvGoyl: > Can we use switch/case here
> declarative way

done.

webapp/javascript/components/Toolbar.tsx
GorvGoyl: It looks like that icon is different from the link one. @Rperry2174 let me know if we need to use that one instead. https://pyroscope.io/storybook/?path=/story/components-button--icon-only

GorvGoyl: done. Found one matching link icon (solid though) https://fontawesome.com/v5.15/icons?d=gallery&p=2&q=link

image

image

webapp/javascript/components/Toolbar.tsx
GorvGoyl: agree. we can replace title once we have a generic tooltip.

GorvGoyl:
> You have to run the server with --enable-experimental-adhoc-ui enabled, or via env var PYROSCOPE_ENABLE_EXPERIMENTAL_ADHOC_UI. We will release after @abeaumont finishes the implementation, which is almost done here #652

So, I was trying it and it looks like the profiler data is not able to load for the other view.

image

GorvGoyl:
> How is the button going to work when we add a button to clear the search. Sorry, I don't see it, am I missing something?

I'm using chromium browser (edge) and it's visible when I type something into the search box.

GorvGoyl:
@eh-am I've moved the logic to redux as discussed as per my understanding. `Toolbar` component is used to access redux store but it's now causing issues with `toolbar` unit tests:

```bash
ProfileHeader › ViewSection › small mode › changes to Both view

    could not find react-redux context value; please ensure the component is wrapped in a <Provider>

      204 | }) {
      205 |   const [text, setText] = useState('');
    > 206 |   const isSearchLinkedState = useSelector(isSearchLinked);
```

Should we also include the `provider` for unit tests or is there an alternative way? I've limited experience with redux so not sure about the right approach.

PR #663, opened 2021-12-27
perf: optimize ingestion

kolesnikovae:
The PR aims to improve ingestion performance, mainly for pull-mode. See https://github.com/pyroscope-io/pyroscope/issues/572 for details.
 - [x] Eliminate intermediate `transporttrie.Trie` in scrape flow. See https://github.com/pyroscope-io/pyroscope/issues/572 for details.
 - [x] (Arguable implementation) Embed a simple queue to the storage. Despite the fact that before the change scrape manager also used buffered queue (`upstream.Direct`), it was using a single consumer goroutine.
 - [x] Refactor pprof profile parsing to avoid repetitive allocations for sample stacks. Instead of iterating over samples for every profile type and allocating space for the same stack multiple times, tree nodes are created with the same _stack_ allocating memory only when a new tree node emerges.
 - [x] Force scraper to use configured profile timings. It was discovered that due to latencies and time lags a profile can overlap multiple 10s nodes of the segment hereby causing extra work for the storage: in fact, every profile could be written twice. Now scraper sets profile timings according to the scrape interval (or duration [`seconds` parameter], if applicable).
 - [x] Replace standard go ProtoBuf generator with [vtprotobuf](https://github.com/planetscale/vtprotobuf) (for pprof Profile type). We observe that significant part of allocations come from ProtoBuf unmarshal call. Use of optimised generator is an experiment; instead, a custom ProtoBuf parser for `pprof.Profile` will be implemented.
 - [x] Scrape metrics. The PR also adds a couple of basic profiling-specific scrape metrics (per job and profile path): profile size in bytes, samples count, and scrape duration.

---

## Further work

1. Reuse pprof processing in `/ingest` handler.
2. Custom pprof parser. Use of vtprotobuf has demonstrated that we still have some room for improvements here.

## Failed attempts

1. Discard insignificant pprof profile samples at parsing: avoid allocations and decrease tree size. It was observed that some profiles (e.g Go pprof heap) can include very high number of samples, meanwhile we only keep 2k tree nodes. The experiment has not shown any significant improvements in practice, despite very promising benchmarks:
```
BenchmarkProfile/Read-4     45	  22720437 ns/op	 1886226 B/op	   40638 allocs/op
BenchmarkProfile/ReadX-4   130	   8792172 ns/op	  479494 B/op	   10100 allocs/op
```
image
Similar impact on allocated objects. Taking into account that the change may cause notable loss of precision, I decided to exclude it from the set. Probably, we should revisit it again separately.

2. [Rework tree node insertion](https://github.com/pyroscope-io/pyroscope/pull/663#discussion_r777476770). It seems that the current implementation is slightly more performant: 
```
BenchmarkProfile/InsterStackBytes-4	24	  45021303 ns/op	 1885480 B/op	   40638 allocs/op
BenchmarkProfile/InsertStackString-4	43	  24825030 ns/op	 1886226 B/op	   40638 allocs/op
```
image

---

# Intermediate results

### CPU time

The overall CPU consumption decreases 2x times:
image
Some nodes are red simply because their proportion has increased. For example GC worker (~identical time ranges):
image

### Allocations

By eliminating `transporttrie.Trie` from scrape flow we drastically reduced allocations:

Objects (~2x decrease):
image

Space (~5x decrease):
image

Another notable factor is vtprotobuf generator which allocates ~2x less objects and space comparing to the standard one:
Objects:
image
Space:
image

## Heap size (in-use-space)

One more interesting point here is that memory consumption has also been reduced:
image

I can't say I fully understand where it comes from but assume that the heap has decreased in size due do reduced allocations; the number of objects is the same, though.


kolesnikovae:
I'll also try to make cache maintenance tasks to use Badger WriteBatch API to minimise cache lock duration.

pkg/storage/tree/tree_profile.go
```diff
@@ -0,0 +1,181 @@
...
+		for _, sampleType := range sampleTypes {
+			if v := uint64(s.Value[sampleType]); v != 0 {
+				c.getOrCreate(sampleType, s.Label).InsertStack(stack, v)
+			}
+		}
+		stack = stack[:0]
+	}
+}
+
+func unsafeStrToSlice(s string) []byte {
```
abeaumont, line 125: This is a bit scary, I think it may be possible to remove the copy without relying on `unsafe` with a bit of extra work in tree insertion (e.g. adding some `func (n *treeNode) insertString` that iterates over the bytes of the string (which should then be [optimized by the compiler](https://github.com/golang/go/wiki/CompilerOptimizations#range-over-bytes)).
Just thinking out loud here.

pkg/storage/storage_put.go
```diff
@@ -24,10 +24,6 @@ type PutInput struct {
 }
 
 func (s *Storage) Put(pi *PutInput) error {
-	// TODO: This is a pretty broad lock. We should find a way to make these locks more selective.
```
abeaumont, line 27: I haven't seen any finer grained mutex in this PR, is this no longer necessary?

pkg/scrape/pprof.go
```diff
@@ -15,267 +15,93 @@
...
-}
-
-func (w *pprofWriter) writeProfile(b []byte) error {
-	var p tree.Profile
-	if err := proto.Unmarshal(b, &p); err != nil {
+func (w *pprofWriter) writeProfile(st, et time.Time, b []byte) error {
+	p := tree.ProfileFromVTPool()
+	defer p.ReturnToVTPool()
+	p.Reset()
+	if err := p.UnmarshalVT(b); err != nil {
```
abeaumont, line 50: I'm curious about performance improvement here (in case you have measured it).

abeaumont (commented):
This looks great! I commented on some doubts I had while looking at this.

pkg/storage/storage_put.go
kolesnikovae: I failed to came up with a case when this mutex is needed, neither I found a race condition. It still requires more deep and thorough testing with race detector but to me it seems we can drop the mutex.

pkg/storage/tree/tree_profile.go
kolesnikovae: Thank you for the suggestion - this an awesome idea.

I think the same - instead of standard binary search (which does not even seem to be inlined) we could implement something more specific for our case. Regarding safety: although this is a widely adopted technique, I would rather not use it. I put it as it is the easiest way to avoid allocation here.

pkg/scrape/pprof.go
kolesnikovae: Working on it right now :)

kolesnikovae: I've added some measurements - not final ones, just to give the idea.

pkg/storage/tree/tree_profile.go
abeaumont: Coincidentally I ended up needing something like this for the flamebearer to tree conversion, so I have added some initial support for it in https://github.com/pyroscope-io/pyroscope/pull/652/commits/603201fed214b294b8737614d1bdb77b2d27679a.
I went with the standard binary search to validate it quickly, but I'll change that (and also add tests). Let me know if you have further insights :)

pkg/scrape/pprof.go
abeaumont: Awesome improvements! :clap::clap:

pkg/storage/tree/tree_profile.go
abeaumont: Unsurprisingly, the implementation had several bugs :sweat_smile:
Changed to use custom binary search and added tests in https://github.com/pyroscope-io/pyroscope/pull/652/commits/672d2156fd92aae5d88310e8f28e9898a96aa63c
A further improvement would be to use BSTs instead of slices for tree nodes, but that may be overkill for this thread, we can discuss that elsewhere.

pkg/storage/storage_put.go
kolesnikovae: Decided to revert this: too many invasive changes at a time; should be part of https://github.com/pyroscope-io/pyroscope/pull/685

kolesnikovae:
Investigating benchmark reports.

lizthegrey:
Very nice dogfooding!

Issue #584, opened 2021-12-07
add persistance to analytics module


petethepig:
Some of pyroscope analytics metrics are cumulative, or in other words, they are counters, not gauges. When the server starts they start with the value of 0 and increase continuously every time some event happens. 

The problem is that when a user restarts pyroscope server all of those counters reset to zero and we end up losing data. I think the simple way of handling this issue is implementing some very basic persistence for analytics.

Rough TODO:

* [ ] implement a function that would save data from [analytics module](https://github.com/pyroscope-io/pyroscope/blob/main/pkg/analytics/analytics.go#L72-L104) to badger. Analytics service already has an instance of `Storage` object, so maybe we could just add some methods to `Storage` (something like `SaveAnalyticsData` / `LoadAnalyticsData`) and then call them from analytics service. I think [`InstallId()` method](https://github.com/pyroscope-io/pyroscope/blob/main/pkg/storage/installid.go#L1) is a pretty good example of a similar pattern
* [ ] implement a function that would load saved data and populate [metrics struct](https://github.com/pyroscope-io/pyroscope/blob/main/pkg/analytics/analytics.go#L72-L103) with saved values
* [ ] make pyroscope perform this persistance call right before the shutdown
* [ ] in case shutdown doesn't succeed we should also periodically (every 10m ?) persist this data anyways

PR #665, opened 2022-01-02
feat: add persistence to analytics module #584

gawicks:


pkg/storage/analytics.go
```diff
@@ -0,0 +1,81 @@
...
+	MemAlloc             int       `json:"mem_alloc"`
+	MemTotalAlloc        int       `json:"mem_total_alloc"`
+	MemSys               int       `json:"mem_sys"`
+	MemNumGC             int       `json:"mem_num_gc"`
+	BadgerMain           int       `json:"badger_main"`
+	BadgerTrees          int       `json:"badger_trees"`
+	BadgerDicts          int       `json:"badger_dicts"`
+	BadgerDimensions     int       `json:"badger_dimensions"`
+	BadgerSegments       int       `json:"badger_segments"`
+	ControllerIndex      int       `json:"controller_index" kind:"cumulative"`
```
gawicks, line 28: wasn't really sure what to name this key. I've struck with kind

pkg/storage/analytics.go
```diff
@@ -0,0 +1,81 @@
...
+	ControllerDiff       int       `json:"controller_diff" kind:"cumulative"`
+	ControllerIngest     int       `json:"controller_ingest" kind:"cumulative"`
+	ControllerRender     int       `json:"controller_render" kind:"cumulative"`
+	SpyRbspy             int       `json:"spy_rbspy" kind:"cumulative"`
+	SpyPyspy             int       `json:"spy_pyspy" kind:"cumulative"`
+	SpyGospy             int       `json:"spy_gospy" kind:"cumulative"`
+	SpyEbpfspy           int       `json:"spy_ebpfspy" kind:"cumulative"`
+	SpyPhpspy            int       `json:"spy_phpspy" kind:"cumulative"`
+	SpyDotnetspy         int       `json:"spy_dotnetspy" kind:"cumulative"`
+	SpyJavaspy           int       `json:"spy_javaspy" kind:"cumulative"`
```
gawicks, line 39: possibly some of these other metrices are also cumulative?

pkg/storage/analytics.go
petethepig: Yeah, this is good, and I like this approach of using struct tags.

pkg/storage/analytics.go
```diff
@@ -0,0 +1,81 @@
...
+		err := txn.SetEntry(entry)
+		if err != nil {
+			return err
+		}
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	return nil
```
kolesnikovae, line 61:
```suggestion
	return s.main.Update(func(txn *badger.Txn) error {
		return txn.SetEntry(badger.NewEntry([]byte(analyticsKey), v))
	})
```

pkg/storage/analytics.go
```diff
@@ -0,0 +1,81 @@
...
+			return err
+		}
+		err = v.Value(func(val []byte) error {
+			err = json.Unmarshal(val, a)
+			if err != nil {
+				return err
+			}
+			return nil
+		})
+		return nil
```
kolesnikovae, line 78:
```suggestion
		return v.Value(func(val []byte) error {
			return json.Unmarshal(val, &a)
		})
```

gawicks: thanks @kolesnikovae

PR #670, opened 2022-01-03
chore: rename and fix typos from previous version in oauth

Skemba:


pkg/server/oauth_github.go
```diff
@@ -11,18 +11,18 @@ import (
 	"golang.org/x/oauth2"
 )
 
-type oauthHanlderGithub struct {
+type oauthHandlerGithub struct {
```
petethepig, line 14: this is funny, I looked at them so many times and never noticed

go.mod
```diff
@@ -44,6 +44,7 @@ require (
 	github.com/klauspost/compress v1.13.5
 	github.com/kyoh86/richgo v0.3.3
 	github.com/kyoh86/xdg v1.2.0 // indirect
+	github.com/markbates/pkger v0.17.1
```
petethepig, line 47: we no longer use pkger, can you run `go mod tidy`?

petethepig (changes requested):
see comments above

pkg/server/oauth_github.go
Skemba: me too, it was only once I installed VSCode extension for spelling errors that I noticed

PR #674, opened 2022-01-03
fix(frontend): keep query param when changing routes

### LINKED ISSUE #671: Changing views loses query params
When changing views from comparison, to diff, then back to comparison view we lose the query params needed in order to share a link.

Expected bahavior: the params should not be removed from the URL when switching views

https://user-images.githubusercontent.com/23323466/147954446-c7aab8c8-b710-45d1-89c9-9be3208c70b3.mov

eh-am:
Closes #671 

This was introduced when we added baseUrl support, where we replaced the barebones `Router` for `BrowserRouter`, which broke the `react-query-sync` library, since it doesn't expose the `history` object which is handled under the hood.
(see more details at https://github.com/Treora/redux-query-sync/issues/13#issuecomment-327361957)

 I spoke with @petethepig and the only reason we did that  was to allow passing a `basename`, which `BrowserRouter` support.

So I just replaced `BrowserRouter` for `Router` to get `redux-query-sync` working (which is the may issue this PR resolves), and then adds the basename to the history object.

The only thing here is that `basename` was removed from `history` from version 5 and onwards (https://github.com/remix-run/history/issues/810), so that may become a problem. But it should be fine for now since we have a simple test that covers the functionality.

Some other thoughts:
* This is where having the file with typescript would be useful, since we would spot that `BrowserRouter` does not take a `history` parameter (https://v5.reactrouter.com/web/api/BrowserRouter) which would probably raise an eyebrown.
* We need to test the `baseURL` stuff. Already spoken with @petethepig previously about doing it when we have more time, but honestly, this is the kind of thing we will forget if we don't do it soon.

### DIFF
--- a/webapp/javascript/index.jsx
+++ b/webapp/javascript/index.jsx
@@ -2,7 +2,7 @@ import ReactDOM from 'react-dom';
 import React from 'react';
 
 import { Provider } from 'react-redux';
-import { BrowserRouter, Switch, Route } from 'react-router-dom';
+import { Router, BrowserRouter, Switch, Route } from 'react-router-dom';
 import FPSStats from 'react-fps-stats';
 import { isExperimentalAdhocUIEnabled } from '@utils/features';
 import Notifications from '@ui/Notifications';
@@ -17,7 +17,6 @@ import AdhocComparison from './components/AdhocComparison';
 import ServerNotifications from './components/ServerNotifications';
 
 import history from './util/history';
-import basename from './util/baseurl';
 
 let showFps = false;
 try {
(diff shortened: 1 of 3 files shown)


Rperry2174:
@petethepig wasn't there a reason you had to use `browserrouter`?

webapp/javascript/index.jsx
```diff
@@ -30,7 +29,7 @@ try {
 
 ReactDOM.render(
   <Provider store={store}>
-    <BrowserRouter history={history} basename={basename()}>
```
Rperry2174, line 33: @petethepig wasn't there a reason you felt you needed `BrowserRouter` instead of `Router` ?

PR #675, opened 2022-01-03
ci: run cypress against an instance with baseURL

eh-am:
Runs the exact same cypress tests, but against an instance with `PYROSCOPE_BASE_URL` set

One of the tests will fail until https://github.com/pyroscope-io/pyroscope/pull/674 is merged


petethepig:
Thanks @eh-am, this is going to be very useful!

PR #676, opened 2022-01-03
fix(frontend) tag selection when used with base-url

### LINKED ISSUE #667: Tag selection doesn't use `base-url` value
I'm running v0.5.1, and the `base-url` changes seem to be largely working (thanks!).
The only problem I've run into so far is that selecting tags yields a `404` since it is not using the base URL.

I believe the bug is here:
https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/redux/actions.js#L444-L462

tags

petethepig:


scripts/webpack/webpack.common.ts
```diff
@@ -86,7 +86,8 @@ module.exports = {
         // However, we still need to prefix the public URL with /assets/static/img
         options: {
           outputPath: 'static/img',
-          publicPath: '/assets/static/img',
+          // using relative path to make this work when pyroscope is deployed to a subpath (with BaseURL config option)
+          publicPath: '../assets/static/img',
```
eh-am, line 90: This is going to work as long as images are being loaded from the CSS, ie if we add an image fromthe html it will fail to load since it's coming from the root.

But it should be fine for now, since we want to [address this in the future anyway](https://github.com/pyroscope-io/pyroscope/tree/base-url-test)?

eh-am (approved):

Issue #591, opened 2021-12-08
Persist frontend data


eh-am:
Currently there's no frontend persistence, which means every time one refreshes the page, configurations like Fit Mode, Table/Flamegraph/View etc have to be set again, which makes the experience annoying.

There's been some internal discussions about using [notifications](https://github.com/pyroscope-io/pyroscope/pull/586) to inform about new versions. In that case, after dismissing the notification for a specific version, that notification should never show up again. Which requires again, some kind of frontend persistence.

The idea is to store that data in the redux store and then persist/retrieve to/from localstorage.

Other potentially persistable items:
- [x] Sidebar collapsed state
- [x] Selected application (or application that loads when you load the dashboard)

![Peek 2021-12-08 10-34](https://user-images.githubusercontent.com/6951209/145217237-74e10f75-f1f1-4833-8dc9-07967f2eb9b8.gif)

Rperry2174:
Should we merge this with #587 @eh-am

eh-am:
We can use https://github.com/rt2zz/redux-persist

PR #681, opened 2022-01-05
feat: added persistence to frontend data

shaleynikov:
In order to accomplish #591 there were
- added redux-persts
- moved several UI states away from components to redux store (sidebar & flamebar views)


shaleynikov:
Current state
https://monosnap.com/file/BuXqrhH6Jhr5nRJ0an2BgPs7NcyHeq

webapp/javascript/components/FlameGraph/enchancers.tsx
```diff
@@ -0,0 +1,72 @@
...
+    (state: RootState) => ({
+      [valueName]: selectUIState(state)(savePath, defaultValue),
+    }),
+    (dispatch) => ({
+      [`set${capitalize(valueName)}`]: (value) => {
+        dispatch(setUIValue(savePath, value));
+      },
+    })
+  );
+
```
eh-am, line 32: Do we need any of this code? Except for `capitalize` I don't see the rest being used?

webapp/javascript/components/enchaners.tsx
```diff
@@ -0,0 +1,22 @@
+import { connect } from 'react-redux';
+import { RootState } from '@pyroscope/redux/store';
+import { selectUIState } from '../redux/reducers/views';
+
+import { setUIValue } from '../redux/actions';
+
+export interface ICollapsible {
```
eh-am, line 7:
```suggestion
export interface Collapsible {
```

eh-am (commented):
Wow I love this functionality! The app feels so much better!


I have a few thoughts:
# general
* since this is new code, IMO it's the perfect situation to start using `redux-toolkit` (see that `newRootState`)
* I like how it's scoped to the flamegraph, however I can see that being a problem in the future. We may want to persist more stuff that's not necessarily scoped to the `flamegraph` component. Do you think it makes sense persisting per route, with a possible alias (for when we rename a route for instance)?
* Expanding on the previous point, just to give you some idea, one thing we were thinking implementingn in the future is to persist data like the `flamebearer` for the last 5 queries, the idea is to make it feel fast.
* Another example we were thinking is silencing certain kinds of notifications that come from the server, for example: the server sends a notification telling the server is running out of space. The client can acknowledges that and persists it (since the server sends the same damn notification on every refresh).
* I also see we are persisting the whole store, which for a quick test here is around `~100KB`. Do we need to persist the whole thing if we are not really doing anything with it?
* I think it would be useful to have an e2e test (probs with cypress) where you can a) check sidebar is uncollapsed b)  collapse the sidebar c) refresh the page d) assert the sidebar is collapsed. It would give me some peace of mind

# enchancers.tsx
* I think you meant `enhancers` (webapp/javascript/components/FlameGraph/enchancers.tsx)
* Can you get rid of all the linting warnings for  `enchancers.tsx` and `implicit any`s? 
* Does it have to a be a `tsx` file?

webapp/javascript/components/FlameGraph/enchancers.tsx
shaleynikov: No we don't. Just as an example of a single value being saved to state

webapp/javascript/components/enchaners.tsx
shaleynikov: Yeah, I was bond to i-prefixed interfaces. Don't need that for sure

webapp/javascript/redux/store.ts
```diff
@@ -22,23 +24,40 @@ import {
...
+  combineReducers({
     newRoot: newRootStore,
     root: rootReducer,
     views: viewsReducer,
-  },
+  })
+);
+const store = configureStore({
+  reducer,
+  // Ignore non-serialiable redux-persis actions
```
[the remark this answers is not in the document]
shaleynikov: It's based on this issue
https://github.com/rt2zz/redux-persist/issues/988
and it's part of this guide

https://redux-toolkit.js.org/usage/usage-guide#use-with-redux-persist

Rperry2174:
@eh-am all great points.. I think many of these can be addressed separately.. The ones I'd say we should do before merging:

> I think it would be useful to have an e2e test (probs with cypress) where you can a) check sidebar is uncollapsed b) collapse the sidebar c) refresh the page d) assert the sidebar is collapsed. It would give me some peace of mind

> I think you meant enhancers (webapp/javascript/components/FlameGraph/enchancers.tsx)

> Can you get rid of all the linting warnings for enchancers.tsx and implicit anys?

Then I think we can make separate issues to address some of your other comments... Anything else you think is a must-have for this pr @eh-am ?

shaleynikov:
> I think you meant enhancers (webapp/javascript/components/FlameGraph/enchancers.tsx)
Yeah, somehow I misspell it every time

> Can you get rid of all the linting warnings for enchancers.tsx and implicit anys?
OFC

> Does it have to a be a tsx file?
OFC no ;)

webapp/javascript/redux/store.ts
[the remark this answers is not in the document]
eh-am: Yeah I get it, I am suggesting to make it clear to people reading the code in the future.

webapp/javascript/components/enchaners.tsx
shaleynikov: Fixed

webapp/javascript/redux/store.ts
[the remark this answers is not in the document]
shaleynikov: Added more actions and comments section

webapp/javascript/components/FlameGraph/enchancers.tsx
shaleynikov: Removed unused code

shaleynikov:
- [x]  I think it would be useful to have an e2e test (probs with cypress) where you can a) check sidebar is uncollapsed b) collapse the sidebar c) refresh the page d) assert the sidebar is collapsed. It would give me some peace of mind
- [x] en(c)hancer(s).ts(x). All 3 items
- [x] all code comments was resolved

I would be happy to deliver first PR :-)

shaleynikov:
There is an issue with cypress tests trying to interact with `detached` controls, e.g. throttled input or flamegraph. By `detached` I mean at some period of time during render the component isn't attached to DOM.

So tests may fail randomly due to race condition between cypress `.find()` command and actual rendering thread. Sometimes it may receive, for instance, canvas which is detached from DOM. 

In order to avoid that Cypress developers recommend to either `.click({force: true})` (looks wierd to me) or to directly ensure it's attached to dom `Cypress.dom.isAttached($el)`

More on details on that [here](https://github.com/cypress-io/cypress/issues/7306)

Rperry2174:
thanks for the work @shaleynikov closing in favor of #699

PR #685, opened 2022-01-07
perf: cache optimization

kolesnikovae:
The PR aims to improve storage cache write performance with focus on the ingestion latency. The work was started in https://github.com/pyroscope-io/pyroscope/pull/663 and then moved to a separate PR to avoid interference. There are still some open questions and the change itself should be carefully tested with benchmarks.

It has been discovered that due to buffering without delivery guaranties, significant part of profiling data could be discarded by scraper. This is mainly caused by _starvation_ of the upload goroutine – direct upstream only uses a single goroutine which prevents us from utilising available CPU resources, and we don't have a "pool" of scrape worker goroutines: scrape manager spawns a goroutine for each target. Thus, according to `go tool trace`, the upload loop spends most of its time in _scheduler wait_ and _sync block_. Another significant factor here is that the storage cache can be locked for relatively long time by maintenance tasks (eviction, write-back, retention).

Taking into account that both push and pull ingestion flows use in-memory queue (a buffered channel) and we will lose profiling data in case of an overflow, cache lock time becomes critical:
- [x] Split cache into buckets and get rid of storage.Put mutex. In order to prevent locking the cache for long time during maintenance tasks (`Eviction` and `WriteBack` first of all), the cache was split into buckets. Now the cache is partially available for reads and writes even during maintenance tasks run.
- [x] Use badger WriteBatch API in maintenance tasks. Typically, `Eviction` and `WriteBack` tasks write multiple items from the cache to the DB at once, therefore it is natural to do it in batches.
- [ ] Get rid of the `storage.Put` mutex.
- [ ] Benchmark under higher load on a bigger machine (c6i 2xlarge).


petethepig:
Decided to close for now

PR #686, opened 2022-01-08
chore: speed up local dev

eh-am:
* `maxWorkers=1` as per https://github.com/kulshekhar/ts-jest/issues/1044
* a scoped `includes` to only care about the webapp
* a `tsconfig` for tests where we manually set which `types` are required
* different eslint rules for tests
* use `esbuild` instead of babel
* use `webpack-livereload-plugin` to add a livereload server to the html files served by go (removing the need for the double server shenanigans)
* in local dev only build the `index.html` since that's the most common case (see webapp/README.md for more info)

---
# Local build
Before
```
21:32:05 webpack | 0 (webpack 5.65.0) compiled with 1 warning in 3296 ms
```

After
```
21:33:46 webpack | webpack 5.65.0 compiled with 1 warning in 446 ms
```

# Tests
Before
```
Test Suites: 15 passed, 15 total
Tests:       193 passed, 193 total
Snapshots:   4 passed, 4 total
Time:        25.552 s
Ran all test suites.
Done in 26.96s.
```

After
```
Test Suites: 15 passed, 15 total
Tests:       193 passed, 193 total
Snapshots:   4 passed, 4 total
Time:        17.276 s
Ran all test suites.
Done in 18.88s.
```

But the biggest improvement seems to be when running tests with `--watch`. 

Before
```
Test Suites: 1 failed, 1 total
Tests:       3 failed, 3 passed, 6 total
Snapshots:   0 total
Time:        8.744 s, estimated 10 s
Ran all test suites matching /Sidebar.spec.tsx/i.
```

After
```
Test Suites: 1 failed, 1 total
Tests:       3 failed, 3 passed, 6 total
Snapshots:   0 total
Time:        1.753 s, estimated 2 s
Ran all test suites matching /Sidebar.spec.tsx/i.
```

# Prod Build

Before
```
$ time yarn build 
(...) 
Done in 21.36s.

________________________________________________________
Executed in   21.53 secs    fish           external
   usr time   38.73 secs    0.51 millis   38.73 secs
   sys time    0.95 secs    1.05 millis    0.95 secs

````

After
```
$ time yarn build  
(...)
Done in 11.45s.

________________________________________________________
Executed in   11.61 secs    fish           external
   usr time   19.26 secs  811.00 micros   19.26 secs
   sys time    0.66 secs  909.00 micros    0.66 secs
```

### CODE REVIEW

**Comment by petethepig** [comment]
@eh-am `maxWorkers` one is quite unexpected

**Comment by eh-am** [comment]
One thing I noticed is that disabling `WebpackPluginServe` makes the build much much faster, but then you need to manually refresh the page

`WebpackPluginServe` turned on
```
0 (webpack 5.65.0) compiled with 1 warning in 3533 ms
```

`WebpackPluginServe` turned off
```
webpack 5.65.0 compiled with 1 warning in 305 ms
```

File: scripts/webpack/webpack.common.ts [hunk 1]
```diff
@@ -5,12 +5,23 @@ import fs from 'fs';
...
+  // Most of the cases we will be developing the SPA
+  // So it makes sense only building it
+  // We go from
+  // [webpack.Progress]  |  | 1131 ms asset processing > HtmlWebpackPlugin
+  // To  [webpack.Progress]  |  | 215 ms asset processing > HtmlWebpackPlugin
+  .sync(
+    process.env.NODE_ENV === 'production'
+      ? './webapp/templates/*.html'
+      : './webapp/templates/index.html'
+  )
```
eh-am, line 22: process.env.NODE_ENV isn't really reliable, we could use
https://webpack.js.org/configuration/mode/#mode-none instead

petethepig: Does this mean in dev mode you won't be able to test other pages? I understand the rationale, but I think this might be confusing when we do want to edit those pages.

It looks like the only thing we do with `HtmlWebpackPlugin` is this `<%= webpack.hash %>` Interpolation, and 215ms to do some string replacements is still very very slow imo.

Maybe we could git rid of HtmlWebpackPlugin altogether and instead export `webpack.hash` to some sort of a file, and then on the go side we would read this file each time we need to render a page and do interpolation via go templates?

I'm not sure how this would work with things like livereload, though.

@eh-am What do you think?

petethepig (approved):
This is awesome! I thought if you want to use `esbuild` you have to ditch `webpack`, this is a good middleground.

This also speeds up docker builds (2x improvement for `make assets-release` step on my machine) which I'm very happy about.

I left a comment about the html pages, but I'm going to merge this one anyway.

scripts/webpack/webpack.common.ts
petethepig: I guess I just remembered we also wanted to get rid of templating on the go side, let's maybe discuss this part on a call.

eh-am: >but I think this might be confusing when we do want to edit those pages.
100% agree, I hope to get rid of this "optimization" soon.

>and 215ms to do some string replacements is still very very slow imo.
I read on the [webpack docs](https://webpack.js.org/guides/build-performance/#avoid-production-specific-tooling) that stuff like hash isn't very good for production. I didn't profile but I trust them (actually I did but couldn't pin point to the exact problem). I guess we can just not hash for dev? Unless HMR use it for refreshing somehow, but we don't have it so it's a problem for the future.

PR #687, opened 2022-01-08
ci: new dockerhub build action

petethepig:
The idea here is to move this out of jenkins for more visibility, source controlled scripts, better permissions system, etc


petethepig:
closing for now

PR #691, opened 2022-01-09
feat: export standalone html

eh-am:
Illustration:
https://user-images.githubusercontent.com/6951209/148690255-ba541c5b-250e-4f72-8e1d-f908dba19a59.mp4

First example shows how to access it: via the `/render` endpoint with a `format=html` query parameter.

Second example shows how one can pipe that output to a file and open it in the browser.
The `adhoc` command (with a possible `--html` flag) should do the same.

---
* Created a separate wepback config called `webpack.standalone.ts`, the reason is that the parameters here are very different (we inline js/css/svg for example)
* When running `make dev` also run webpack for that standalone script. Unfortunately this makes the output a little bit more bloated 
* Another problem is that we need to minify the output (not entirely sure why), which makes debugging quite diffficult (since variable names are mangled)
* Allow running the standalone interface without a server by running `yarn dev:standalone` and enabling the `defaultFlamegraph` in `standalone.tsx`
* Add live reload when `--watch` flag is turned on. Unfortunately it's quite slow (~5s), let's see if it gets better after is merged https://github.com/pyroscope-io/pyroscope/pull/686


webapp/templates/standalone.html
```diff
@@ -0,0 +1,19 @@
...
+    <title>Pyroscope</title>
+
+    <script type="text/javascript">
+      try {
+        eval('window.flamegraph = {{ .Flamegraph }}');
+      } catch(e) {}
+    </script>
+  </head>
+  <body>
+    <div id="root"></div>
```
eh-am, line 17: Shall we add a pixel tracker? @petethepig @Rperry2174

petethepig: @eh-am that would be too invasive imo. we could add a new metric to analytics module, something like `html_exports`, and update the counter every time someone exports an html

eh-am: I was hoping to somehow know if people are **sharing** the export HTML around, but fair point, it's very intrusive yes.

eh-am:
Related to https://github.com/pyroscope-io/pyroscope/issues/525

eh-am:
One important bit when exporting via UI is that this is only available for single flamegraphs, ie, it doesn't work for `diff`.

petethepig (approved):
This is great! Awesome work @eh-am!

PR #693, opened 2022-01-10
feat: persistent app selection for default app

shaleynikov:
Issue: app name doesn't persist once it's default value

Now app is forced to default once it's empty.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
   },
   "devDependencies": {
     "@babel/core": "7.8.4",
+    "@clarketm/saga-monitor": "^2.0.1",
     "@fortawesome/fontawesome-common-types": "^0.2.36",
     "@size-limit/file": "^6.0.3",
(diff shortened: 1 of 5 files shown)


eh-am:
Hey! We currently don't use sagas. Do you think it's possible to simplify and solve this issue without adding new dependencies?

eh-am:
(This may not be relevant if you drop the saga stuff) 

The review app is failing to be built with
```
#119 13.59 Module not found: Error: Can't resolve '@clarketm/saga-monitor' in '/opt/pyroscope/webapp/javascript/redux'
#119 13.59 resolve '@clarketm/saga-monitor' in '/opt/pyroscope/webapp/javascript/redux'
#119 13.59   Parsed request is a module
```

Since '@clarketm/saga-monitor' was added as a dev dependency, it's not available when we `NODE_ENV=production yarn install`, which breaks the build.

PR #694, opened 2022-01-10
fix: close response body in traffic-duplicator to fix resource leak.

abeaumont:


petethepig (approved):
Great catch, thank you for fixing this!

PR #695, opened 2022-01-10
perf: optimize segment tree serialization.

abeaumont:
Segment tree serialization was done traversing the tree in depth
iteratively, using slices.
This causes a lot of slice insertions which caused an important amount
of allocations to grow these slices.

Serialization is now changed to do a recursive traversal instead,
which doesn't require any explicit data structure and thus all these
allocations are removed. Since the depth of segment trees is, in
practice, both bounded and small, it should be more efficient in all
cases, but especially as the segment trees grow.

A benchmark has also been included to be able to compare results
between the old and new implementations.
`benchstat` shows good results:

```
$ benchstat old.txt new.txt
name                            old time/op  new time/op  delta
Serialize/serialize_10-16       5.13µs ± 2%  3.77µs ± 4%  -26.60%  (p=0.008 n=5+5)
Serialize/serialize_100-16      29.4µs ± 4%  11.7µs ± 5%  -60.18%  (p=0.008 n=5+5)
Serialize/serialize_1000-16      299µs ± 1%    83µs ± 5%  -72.24%  (p=0.008 n=5+5)
Serialize/serialize_10000-16    4.34ms ± 9%  1.03ms ± 3%  -76.19%  (p=0.008 n=5+5)
Serialize/serialize_100000-16   35.6ms ± 4%   7.7ms ± 4%  -78.28%  (p=0.008 n=5+5)
Serialize/serialize_1000000-16   331ms ± 2%    76ms ± 6%  -77.15%  (p=0.008 n=5+5)
```

Using pyroscope to benchmark these optimizations is less useful as the
improvement becomes more relevant as the size of the segment trees
increases, and thus it's harder to appreciate the improvements in
short runs. On longer runs, these improvements should become more
significant (e.g. `pyroscope.server.cpu{}` at
https://demo.pyroscope.io/ shows over 11% time invested in
`runtime.growslice` inside segment serialization:
segment-serialize

### DIFF
--- a/pkg/storage/segment/serialization.go
+++ b/pkg/storage/segment/serialization.go
@@ -54,38 +54,39 @@ func (s *Segment) Serialize(w io.Writer) error {
 		return nil
 	}
 
-	nodes := []*streeNode{s.root}
-	for len(nodes) > 0 {
-		n := nodes[0]
-		vw.Write(w, uint64(n.depth))
-		vw.Write(w, uint64(n.time.Unix()))
-		vw.Write(w, n.samples)
-		vw.Write(w, n.writes)
-		p := uint64(0)
-		if n.present {
-			p = 1
-		}
-		vw.Write(w, p)
-		nodes = nodes[1:]
-
-		// depth
-		// time
-		// keyInChunks
-		// children
-		l := 0
-		r := []*streeNode{}
-		for _, v := range n.children {
-			if v != nil {
-				l++
-				r = append(r, v)
-			}
-		}
-
-		vw.Write(w, uint64(l))
-		nodes = append(r, nodes...)
-	}
+	s.serialize(w, vw, s.root)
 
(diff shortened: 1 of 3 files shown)


abeaumont:
The problem becomes more clear when looking at the number of allocated objects: 
segment-serialize-allocs

kolesnikovae:
Brilliant improvement! 🔝

abeaumont:
I deployed the branch to dev (thanks @kolesnikovae) to check the optimization in https://demo-dev.pyroscope.io.

This is how CPU consumption changes, with 1h ranges in both left and right:
dev-demo-cpu

Even the red runtime values are better in absolute terms.

For allocation objects, I used the comparison view which shows better how allocations mostly disappear now in segment serialization:

dev-demo-alloc-objects

petethepig (approved):
This is awesome!

PR #698, opened 2022-01-11
testing self hosted runners with cypress

petethepig:
I suspect cypress might be flaky because it's running in github actions so I'm just trying to run it on a self hosted runner, and see if that changes anything.


petethepig:
it was a fun experiment, we learned some things about self hosted runners, closing for now

PR #699, opened 2022-01-11
feat(frontend): persist sidebar collapsed state

eh-am:
Implements sidebar performance (related to https://github.com/pyroscope-io/pyroscope/issues/591)

The code is based on https://github.com/pyroscope-io/pyroscope/pull/681, except we only do the sidebar here.

The logic is similar to what twitch does:
* If the user has never interacted with the sidebar (state called 'pristine'):
   * If the window is small, we set to collapsed
   * If the window width is large, we set to uncollapsed
   * If the user resizes the screen, these rules should be still be valid.
 * If the user HAS interacted with the sidebar
   * then we assume they know what they are doing and ALWAYS use what've chosen
   * for example, if they had manually collapsed, we will always have it collapsed, even in bigger screens

---
Also added migrations, which currently don't do much but with new versions this is useful. The idea is that what you got from localStorage may not be what your application code is expecting (let's say if you had stored a boolean but now your app code expects an object).

---
Create a `ui` store that currently only holds the sidebar state. Also that's the only state serialized.
Preferred a more straightforward approach to make the code easier to understand, the component interacts directly with the store via the selector and the dispatcher.

### DIFF
--- a/webapp/javascript/components/Sidebar.tsx
+++ b/webapp/javascript/components/Sidebar.tsx
@@ -17,6 +17,13 @@ import Sidebar, {
   SubMenu,
   Menu,
 } from '@ui/Sidebar';
+import { useAppSelector, useAppDispatch } from '@pyroscope/redux/hooks';
+import {
+  selectSidebarCollapsed,
+  collapseSidebar,
+  uncollapseSidebar,
+  recalculateSidebar,
+} from '@pyroscope/redux/reducers/ui';
 import { useLocation, NavLink } from 'react-router-dom';
 import { isExperimentalAdhocUIEnabled } from '@utils/features';
 import Icon from '@ui/Icon';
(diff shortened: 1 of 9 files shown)


webapp/javascript/redux/reducers/ui.ts
```diff
@@ -0,0 +1,69 @@
...
+};
+
+export const uiSlice = createSlice({
+  name: 'ui',
+  initialState,
+  reducers: {
+    recalculateSidebar: (state) => {
+      if (state.sidebar.state === 'pristine') {
+        state.sidebar.collapsed = window.innerWidth < 1200;
+      }
```
eh-am, line 49: window width should come as an action parameter

webapp/javascript/components/Sidebar.tsx
```diff
@@ -110,7 +112,10 @@ export default function Sidebar2(props: SidebarProps) {
     </SubMenu>
   );
 
-  const toggleCollapse = () => setCollapsed(!collapsed);
+  const toggleCollapse = () => {
+    const action = collapsed ? uncollapseSidebar : collapseSidebar;
```
eh-am, line 116: Although this could've be written as a single action `toggleCollapse`, we prefer our actions to be idempotent, which makes it easier to understand.

PR #710, opened 2022-01-12
docs: show example using base-url

eh-am:


Rperry2174:
Where do you think we might be able to add this to the docs repo @eh-am @petethepig ?

eh-am (commented):
Just noticed we don't need to expose the pyroscope server use host network.

examples/base-url/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
...
+    image: 'pyroscope/pyroscope:latest'
+    ports:
+      - '4040:4040'
+    command:
+      - 'server'
+    environment:
+      PYROSCOPE_BASE_URL: /pyroscope
+  nginx:
+    image: nginx
+    network_mode: host
```
petethepig, line 13: this won't work on mac (and windows too I think)

it's more portable to refer to hosts using the hostnames:
```
proxy_pass http://pyroscope:4040;
```

and then you don't need `network_mode: host` either

eh-am: Yeah I was using that setup to test this base url stuff against a local pyroscope instance, and forgot to update that.

PR #717, opened 2022-01-13
chore: set -race flag in ci and windows go tests

eh-am:
* As pointed by @kolesnikovae, we were not enabling race condition detector when running `make coverage`, which is what we run for CI.
* Also don't run storage delete tests on windows.
* Ended up skipping `analytics_test.go` when `-race` is active, since it seems to contain a data race, and I am not familiar enough with that part of the code to fix it :(
* Bump to go 1.17


pkg/storage/storage_delete_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build !windows
+// +build !windows
 package storage
```
kolesnikovae, line 3:
```suggestion
//go:build !windows
// +build !windows

package storage
```

PR #718, opened 2022-01-13
chore: install yarn dependencies when running 'make dev'

eh-am:
This was brought up by @petethepig, it's annoying how you run `make dev` then to only realize the app is not working because there are new dependencies. 

This PR makes  the `install-web-dependencies` run before `make dev`.

Also cleans up other tasks that don't seem to be used anymore.


petethepig (approved):
I made the old commands `exit 1` in case someone depends on them

PR #719, opened 2022-01-13
chore: don't calculate webpack hash when in local dev

eh-am:
That makes local builds faster https://webpack.js.org/guides/build-performance/#avoid-production-specific-tooling

We went from (dev, incremental build)

```
webpack 5.65.0 compiled with 1 warning in 1257 ms
```
to
```
webpack 5.65.0 compiled with 1 warning in 223 ms
```

@petethepig I even removed that ["optimization" we didn't like](https://github.com/pyroscope-io/pyroscope/pull/686#discussion_r780840358) since it doesn't seem to be relevant here anymore :)


petethepig (approved):
Thanks for improving this!

PR #723, opened 2022-01-13
fix(frontend): don't allow selecting empty apps

eh-am:
Due to circumstances, some of our old apps have empty strings as name. This PR ashamedly skips those when fetching from the backend.

Highlighted an app with empty name (which this PR skips)
image


Rperry2174 (approved):
Lgtm 🎉

PR #725, opened 2022-01-14
Add html export

Rperry2174:
Adds html export button and changes the way we do filenames for it.

Before we were just doing current date, but I think its more helpful for debugging if we do the start and enddate of the query so that different queries taken on the same day will reflect that as opposed to having the same file name.

### DIFF
--- a/webapp/javascript/components/ExportData.tsx
+++ b/webapp/javascript/components/ExportData.tsx
@@ -146,7 +146,8 @@ function ExportData(props: ExportDataProps) {
               className="dropdown-menu-item"
               type="button"
               onClick={() => exportHtml(exportFlamebearer)}
-            >   Html
+            >
+              Html
             </button>
           )}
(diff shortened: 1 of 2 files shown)


webapp/javascript/components/ExportData.tsx
```diff
@@ -91,11 +113,11 @@ function ExportData(props: ExportDataProps) {
...
           <button
             className="dropdown-menu-item"
-            onClick={() => exportCanvas('png')}
-            onKeyPress={() => exportCanvas('png')}
+            onClick={() => exportCanvas(exportFlamebearer, 'png')}
+            onKeyPress={() => exportCanvas(exportFlamebearer, 'png')}
             type="button"
           >
-            PNG
+            Png
```
petethepig, line 120: Screen Shot 2022-01-17 at 10 20 47 PM

PR #726, opened 2022-01-14
feat: add  http discovery mechanism

linthan:
we need to scrape some targets temporarily. this way would be a convient way to add targets.

### DIFF
--- a/pkg/scrape/discovery/http/http.go
+++ b/pkg/scrape/discovery/http/http.go
@@ -1,0 +1,196 @@
+// Copyright 2021 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package http
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/url"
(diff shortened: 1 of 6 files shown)


pkg/scrape/discovery/http/http.go
```diff
@@ -0,0 +1,195 @@
...
+	"io/ioutil"
+	nhttp "net/http"
+	nurl "net/url"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/prometheus/common/model"
```
kolesnikovae, line 30: Consider avoiding use of `github.com/prometheus/common/model` in favour of `github.com/pyroscope-io/pyroscope/pkg/scrape/model` for better readability.

pkg/scrape/discovery/http/http.go
```diff
@@ -0,0 +1,195 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	// DefaultSDConfig is the default HTTP SD configuration.
+	DefaultSDConfig = SDConfig{
+		RefreshInterval:  model.Duration(60 * time.Second),
+		HTTPClientConfig: config.DefaultHTTPClientConfig,
+	}
+	userAgent        = fmt.Sprintf("Pyroscope/%s", "0.5.1")
```
kolesnikovae, line 45: I think we should use `build.Version`.

pkg/scrape/discovery/http/http.go
```diff
@@ -0,0 +1,195 @@
...
+// limitations under the License.
+package http
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	nhttp "net/http"
```
kolesnikovae, line 22: I'm wondering if import aliases are really needed here. Perhaps, it's better to avoid them.

kolesnikovae (commented):
Thank you so much @linthan! Well done 👍🏻 The only change I'd ask for is the HTTP client version.

kolesnikovae (approved):

PR #728, opened 2022-01-14
feat: output standalone HTML files for adhoc profiles.

abeaumont:


pkg/structs/flamebearer/html.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"html/template"
+	"io"
+	"net/http"
+)
+
+// FlameberarerToHTML converts and writes a flamebearer into HTML
+func FlameberarerToHTML(fb *FlamebearerProfile, dir http.FileSystem, w io.Writer) error {
```
eh-am, line 12:
```suggestion
func FlameberarerToStandaloneHTML(fb *FlamebearerProfile, dir http.FileSystem, w io.Writer) error {
```
I would suggest for consistency to use the `standalone` terminology. The reason is that it's not that we are really converting a flamegraph to HTML, we are converting a flamegraph to a "standalone HTML", which has different attributes. 

A simple flame->HTML conversion could be just the flamegraph itself for example, one possible use is for a hypothetical embed widget.

Where a standalone HTML (should) contain more stuff, like a nice header, footer, links to other places etc.

eh-am (commented):
This is nice!

This is more of a question: my understanding is that we would have a flag such as `--output=html`, which creates a file locally. That flag could then be expanded to `pprof`, `json` etc. Any reason to not go with this route?

abeaumont:
We discussed this a bit and concluded that both things are actually complementary, not exclusive. So it may make sense to store data locally for inspection through the UI (so comparison and diffing are possible) while also having an HTML file for easy inspection without having to run the server.

pkg/structs/flamebearer/html.go
abeaumont: Good point. Changed in https://github.com/pyroscope-io/pyroscope/pull/728/commits/a04b3607757e541381367874c86a0d62284a8aa7

abeaumont:
Changed how data is exported so both JSON and any external format can be combined in https://github.com/pyroscope-io/pyroscope/pull/728/commits/98d9a14efaf27672a6b31c63e1591f0df64d277f
cc: @eh-am

abeaumont:
https://github.com/pyroscope-io/pyroscope/pull/728/commits/18b7d5d6006f305a339a58b7b1c9bf68ece71aa6 makes adhoc GA by removed its experimental status. The adhoc UI is still a feature that can be disabled (with `--no-adhoc-ui`, but it's no longer experimental and it's enabled by default.

eh-am:
> [18b7d5d](https://github.com/pyroscope-io/pyroscope/commit/18b7d5d6006f305a339a58b7b1c9bf68ece71aa6) makes adhoc GA by removed its experimental status. The adhoc UI is still a feature that can be disabled (with `--no-adhoc-ui`), but it's no longer experimental and it's enabled by default.

Wait, can we do that? I love it, but I got the impression we were super zealous of not breaking the API, even if it's still experimental.  Just to be clear, I am all in favor for this since I want to kill `EnableExperimentalAdmin` haha.

abeaumont:
Hmmm, I thought the whole process of adding the `experimental` work was to make it clear that it was something not GA yet and that it'd be broken at any moment... so under this assumption, I think we can ditch it. Once it's no longer experimental, then we should be cautious about not breaking the API.

Issue #99, opened 2021-02-25
Add Tooltip to Timeline Chart


Rperry2174:
We currently use the [Flot](http://www.flotcharts.org/flot/examples/) library for our timeline. We'd like to add a tooltip that moves horizontally along the X axis and tells you what timestamp is currently being hovered:

This is an example taken from ApexCharts, but we'd like to recreate this (tooltip only) in Flot library
![ezgif com-gif-maker](https://user-images.githubusercontent.com/23323466/107838059-f0d56a80-6d58-11eb-8c57-52c4e0196080.gif)

Here is link to [Flot tooltip plugin](https://github.com/krzysu/flot.tooltip) 

The biggest goal is to just be able to better understand what is being hovered. I'm open to other "styles" of tooltip it just has to exist in some way that lets you know what particular time you are hovering over

shaleynikov:
Hey, got deeper into how timeline is rendered and here is my findings:

Flot tooltip plugin has a lot of logic built in. Using it for this purpose will require either:
1. Tooltip "target" to be hoverable (which isn't possible now as currently data chunks are displayed as "vertical lines" and not, for instance, bars) (https://github.com/krzysu/flot.tooltip/blob/master/js/jquery.flot.tooltip.js#L177)
2. Certain draw options to be enabled (https://github.com/krzysu/flot.tooltip/blob/master/js/jquery.flot.tooltip.js#L179)

For now I don't see an easy way to achieve required functionality with current tooltip plugin.

I would suggest to create a new plugin relying on Flot plugin system (https://github.com/flot/flot/blob/master/docs/PLUGINS.md) rather than re-using existing flot-plugin

Rperry2174:
Yeah I suspected this might be the better path too so if you think that would work better / be cleaner than definitely happy to go that route @shaleynikov  

cc @eh-am

PR #730, opened 2022-01-14
feat: added tooltip for timeline selection

shaleynikov:
This is a contribution for  #99


petethepig:
Looks great! The only thing I could find is it should add a zero before 5 in `10:05`
Screen Shot 2022-01-14 at 10 55 09 AM

Rperry2174:
Hey @shaleynikov everything looks good here. Could you add a simple unit test that:
- Creates a hover event
- Checks that the tooltip exists on-hover

Rperry2174:
hey @shaleynikov currently when selecting a time range on the graph, there are two tooltips on the bottom while the mouse is down. 

Can you make it so that there is only one (at the current location of the mouse)

PR #737, opened 2022-01-17
fix: incorrect reads when downsampling

kolesnikovae:
The PR fixes downsampling. Closes https://github.com/pyroscope-io/pyroscope/issues/715.


pkg/storage/segment/segment.go
```diff
@@ -158,23 +158,22 @@ func (sn *streeNode) get(ctx context.Context, st, et time.Time, cb func(sn *stre
...
-
-	// inside or overlap
-	if sn.present && sn.isLeaf() {
-		// TODO: I did not test this logic as extensively as I would love to.
-		//   See https://github.com/pyroscope-io/pyroscope/issues/28 for more context and ideas on what to do
-		trace.Log(ctx, traceCatNodeGet, "leaf")
-		cb(sn, sn.overlapRead(st, et))
-		return
-	}
-
```
kolesnikovae, line 170: I'm confused with this snippet. The trace in https://github.com/pyroscope-io/pyroscope/issues/715#issuecomment-1013752622 has shown that we have a side effect when data outside of the query time range appear in the response:

```
region segment.Get started (duration: 801.706µs)
--
7.236367021 | .     7041 | 3005 | node.get=D=5 T=1641403200 P=true M=false R=inside
7.236371117 | .     4096 | 3005 | node.get=drill down
7.236375597 | .     4480 | 3005 | node.get=D=4 T=1642203200 P=true M=false R=inside
7.236379822 | .     4225 | 3005 | node.get=leaf
7.236397647 | .    17825 | 3005 | storage.Get.Callback=tree_found=true time=1642203200 r=361/10000
```

Here is the bug: when a node does not have children, we generate downsampled tree from the current one. To my understanding, the only case when it should happen is when child nodes were removed by level-based retention policy, and not if the nodes were removed because of the absolute retention period policy.

So I'm wondering if there are other cases when `overlapRead` should be called.

kolesnikovae: I've added a fuzz-like test for segment retention policy: I deliberately left it in a failed state - although I'm guessing the problem is in the test itself, it seems that `DeleteNodesBefore` may not work as expected: perhaps we should account for decreasing node `samples` counter when its child node is getting deleted. Not sure about the side effects, but the resulting numbers may be inaccurate.

kolesnikovae: I confirm that with [the sample data](https://github.com/pyroscope-io/pyroscope/issues/715#issuecomment-1022488782) the issue is not reproduced. But can be reproduced easily with the latest release. Added the sample to test cases.

Another issue was solved by the way:

> I've added a fuzz-like test for this case: I deliberately left it in a failed state - although I'm guessing the problem is in the test itself, it seems that DeleteNodesBefore may not work as expected: perhaps we should account for decreasing node samples counter when its child node is getting deleted. Not sure about the side effects, but the resulting numbers may be inaccurate.

We have a segment and the retention policy:
```
         │              │         │             │
 ┌───────┼──────────────┼─────────┼──┐─ ─ ─ ─ ─ ┤
 │       │              │         │  │          │
 └───────┼───┬──────────┼┬────────┼──┤─ ─ ─ ─ ─ ┤
         │   │ N        ││        │  │          │
         │   └────────┬─┼┼────────┼──│─ ─ ─ ─ ─ ┤
         │            │ ││        │  │          │
         │            └─┼┴────────┼──┘─ ─ ─ ─ ─ ┤
         │              │         │             │
         │◄─────────────┼────────►│             │
         │ query range  │         │             │
                        │◄─────────────────────►│
                        │  retention period     │
                                               now
```

The problem is that on query, node N will be taken as is because the *relationship* to the query range assumes that it
*contain*s the node. Taking into account that node counters (`samples`, `writes`) are not updated when its children are removed, a tree for node N will be retrieved and incorrect numbers will be reported.

Updating counters would not solve the problem as we'd have to update related trees as well (to calculate average, for example). Therefore it's better truncating the query time range lower boundary to the node watermark (the lowest time mark of the retained data). By this we change the relationship making it traverse child nodes.

petethepig (approved):

Issue #657, opened 2021-12-24
Diff view is not colorblind friendly


lizthegrey:
Use of red and green is considered problematic -- https://webaim.org/articles/visual/colorblind

lizthegrey:
Relates to #24

petethepig:
Great point, thanks for bringing this up! That's another good reason to prioritize #24 

Another thing I'm thinking is when we work on this we might want to look closer at existing colors we use to distinguish between packages and see if we can tweak them to be more universally distinguishable.

Rperry2174:
Another example for color palettes https://venngage.com/blog/color-blind-friendly-palette/#4

Rperry2174:
For this I think the first step is to slightly refactor the code to take arbitrary colors as inputs. For the v1 of this, and for simplicity, we should just focus on the diff view and then potentially expand out in into the other views later. 

Right now the red and green colors are hardcoded [here](https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/components/FlameGraph/FlameGraphComponent/color.ts#L34-L35)

We should start by un-hardcoding these values and instead passing them into a function that then renders both the flamegraph and the legend. 

Then we should create a couple different color options in a dropdown that when selected will change the diff view to be rendered with the new colors (similar to picture below with just two colors per selection):
![image](https://user-images.githubusercontent.com/23323466/147499269-24fe4af7-de04-4f4f-b5d6-12434f271cb7.png)

Eventually once #591 is completed we can also save these preferences for users

eh-am:
@Rperry2174 

Few points I am thinking
* this is not strictly for the comparison diff
* for v1 we can have a simple dropdown (as you pointed) with 2 modes: Default | ColorBind. It's a bit reductive yes (as you pointed with the palettes above), but it's a good start.
* that however makes the toolbar more bloated, which I guess requires us to create versions of the buttons with icons only? That would require us to add a tooltip component (otherwise how am I supposed to know what this button means)?
* suffice to say, tooltips should also be a11y friendly.
* code wise, I'm thinking accepting a custom ColorRenderer or similar here https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/components/FlameGraph/FlameGraphComponent/Flamegraph_render.ts#L72 , which implements `getColor` https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/components/FlameGraph/FlameGraphComponent/Flamegraph_render.ts#L225-L246

[...]

PR #740, opened 2022-01-18
Feat/diff palette

eh-am:
https://user-images.githubusercontent.com/6951209/150017644-0187e2de-aa91-42f9-ad0c-18f0cca49d3a.mp4

Related to #657 

Comments:
* Ended up doing for diff only, altough created a FlamegraphPalettes abstraction that apply to ALL flamegraphs. 
* went with the naive approach and lifted the state up to `FlamegraphRenderer`, so you gonna see a lot of `palette` being passed around. In a future refactor we can use [Context](https://reactjs.org/docs/context.html)
* not persisting the preference to local storage


webapp/javascript/components/FlameGraph/FlameGraphComponent/Flamegraph_render.ts
```diff
@@ -79,6 +82,7 @@ export default function RenderCanvas(props: CanvasRendererConfig) {
   const { units } = props;
   const { rangeMin, rangeMax } = props;
   const { tickToX } = props;
+  const { palette } = props;
```
Rperry2174, line 85: just curious, could all these be combined into one? Any particular reason they are not? 

i.e. 
```
const {
    canvas,
    numTicks,
    sampleRate ,
    pxPerTick,
    fitMode,
    units,
    rangeMin ,
    rangeMax,
    tickToX ,
    palette
} = props;
```

webapp/javascript/components/FlameGraph/FlameGraphComponent/colorPalette.ts
```diff
@@ -0,0 +1,58 @@
...
+    Color.rgb(128, 110, 183),
+  ],
+};
+
+export const ColorBlindPalette: FlamegraphPalette = {
+  ...DefaultPalette,
+
+  name: 'Color Blind',
+  goodColor: Color('rgb(26,133,255)'),
+  neutralColor: Color.rgb(148, 142, 142),
```
Rperry2174, line 56:
```
  goodColor: Color('rgb(26,133,255)'),
  neutralColor: Color.rgb(148, 142, 142),
```
why is one in quotes and the other is not?

Rperry2174:
Hey @eh-am this looks great so far... can you elaborate on this:

> went with the naive approach and lifted the state up to FlamegraphRenderer, so you gonna see a lot of palette being passed around. In a future refactor we can use Context


1. Why did you choose this approach?
2. As a followup to this PR could you write more details on what that would look like to switch to using Context?

eh-am:
image
The tooltip should use the same palette

webapp/javascript/components/FlameGraph/FlameGraphComponent/color.ts
```diff
@@ -1,6 +1,8 @@
 /* eslint-disable camelcase */
 import Color from 'color';
+import { scaleThreshold, scaleLinear } from 'd3-scale';
 import murmurhash3_32_gc from './murmur3';
+import type { FlamegraphPalette } from './colorPalette';
 
 const colors = [
```
Rperry2174, line 7: do we still need this array here now?

Rperry2174: nvm will address in future PR

PR #749, opened 2022-01-20
feat(frontend): export comparison diff standalone html

eh-am:
Ended up having to refactor the whole `ExportData` component.
Also, since there are so many variations (single | diff x standalone | continuous | adhoc | grafana), I made them clear. So each `FlamegraphRenderer` needs to declare explicitly what it exports.

Other issues:
* Adhoc doesn't bring all the fields `/render` and `/render-diff` endpoint bring https://github.com/pyroscope-io/pyroscope/blob/main/pkg/server/render.go#L114-L131, which is used to define the file names. So I took the shortcut and disabled exporting everything except JSON for adhoc. 
* ADDITION: actually to be able to export adhoc we need to add another endpoint. Right now it works for continuous, since we hit either `/render` or `/render-diff`. For adhoc that doesn't make sense, so we need an endpoint that takes a JSON (`FlamebearerProfile`) and returns a HTML. cc @abeaumont how feasible this is
* Not exporting png for comparison since the functionality was always broken (it only exports the first flamegraph), ideally we would somehow access the ref or something similar
* No exporting in grafana, but for different reasons: a) the datasource doesn't bring a rawFlamegraph (which is required to be exported) and b) for `html/pprof` the panel doesn't know about the server url.

Both issues can be solved in the future.


webapp/javascript/components/FileUploader.tsx
```diff
@@ -20,6 +20,7 @@ export default function FileUploader({ file, setFile, className }: Props) {
       throw new Error('Only a single file at a time is accepted.');
     }
 
+    console.log(acceptedFiles);
```
Rperry2174, line 23: Did you mean to keep this here?

Rperry2174 (approved):
This looks great -- functionality works as expected. Only one comment but feel free to merge when ready

abeaumont:
Adding an endpoint to render adhoc profiles is doable. I'd put that into the same set of features that can be enabled if we upload adhoc profiles to the server: (save uploads in the pyroscope data directory, support diff of uploads, etc.)

PR #751, opened 2022-01-21
fix: return disk space check

kolesnikovae:
The check was mistakenly removed in https://github.com/pyroscope-io/pyroscope/pull/456. Meanwhile, this is essential in some scenarios: pyroscope server won't start if it can't create a LOCK file.


petethepig (approved):
Looks great! Thanks for quick turnaround!

PR #757, opened 2022-01-24
chore: make conversion functions public.

abeaumont:


eh-am:
Do we need to version these functions? Eg what if tomorrow we start returning `flamebearer.FlamebearerProfileV2`?

PR #759, opened 2022-01-24
fix(frontend) fixing diff issues 

shaleynikov:
If you go to the diff page, there are two issues:

- [x] The color of the text in the tooltip doesn't match the color 
- [x] The legend dropdown is now always in "small" mode. It used to be wider when the page is wider and only go to the "small" mode when the page is shrunk https://github.com/pyroscope-io/pyroscope/pull/754/files#diff-3e4cfc0789971eae1cf8878eeb1ea49b445785a6f030900bd9768f9501034342R42-R51 is not working anymore so we are always showing the small version


shaleynikov:
I had to move resize observer from difflegend up to the parent component and fix two layout issues which prevented component from calculating it's own width correctly.

webapp/javascript/components/FlameGraph/FlameGraphComponent/DiffLegendPaletteDropdown.tsx
```diff
@@ -3,45 +3,87 @@ import Dropdown, { MenuItem, MenuButton } from '@ui/Dropdown';
...
-    </Dropdown>
+          </MenuItem>
+        ))}
+      </Dropdown>
+    </div>
   );
 };
 
+/**
+ * TODO: unify this and toolbar's
```
Rperry2174, line 55: I'm fine leaving this as a todo, but as a follow up after we merge we should create an issue so we don't forget about this

PR #760, opened 2022-01-25
chore(adhoc): adds analytics

petethepig:


pkg/config/config.go
```diff
@@ -27,6 +27,8 @@ type Config struct {
 }
 
 type Adhoc struct {
+	AnalyticsOptOut bool `def:"false" desc:"disables analytics" mapstructure:"analytics-opt-out"`
```
abeaumont, line 30: https://github.com/pyroscope-io/docs/blob/main/docs/agent-configuration-adhoc.mdx should be updated to reflect this new option.

abeaumont: Done in https://github.com/pyroscope-io/docs/pull/33/commits/3e1980f6e03985e3d428780d1228eee784b6c109

PR #761, opened 2022-01-25
chore(standalone): adds logos to standalone html

petethepig:
This adds a logo to the top of the page and also add a footer with build info, it's implemented the same way as in the regular pages. 

Toolbar:
Screen Shot 2022-01-24 at 4 52 23 PM

Footer:
Screen Shot 2022-01-24 at 4 47 20 PM

### DIFF
--- a/pkg/structs/flamebearer/html.go
+++ b/pkg/structs/flamebearer/html.go
@@ -8,6 +8,8 @@ import (
 	"net/http"
 	"regexp"
 	"text/template"
+
+	"github.com/pyroscope-io/pyroscope/pkg/build"
 )
 
 // FlamebearerToStandaloneHTML converts and writes a flamebearer into HTML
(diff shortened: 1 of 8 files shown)


pkg/structs/flamebearer/html.go
```diff
@@ -26,9 +28,8 @@ func FlamebearerToStandaloneHTML(fb *FlamebearerProfile, dir http.FileSystem, w
 	scriptTpl, err := template.New("standalone").Parse(
 		`
 <script type="text/javascript">
-	try {
-		eval('window.flamegraph = {{ .Flamegraph }}');
-	} catch(e) {}
+	window.flamegraph = {{ .Flamegraph }}
```
petethepig, line 31: @eh-am I figured the eval here was not necessary so I removed it. But maybe I'm missing something, let me know.

PR #766, opened 2022-01-25
feat: experimental tracing integration

kolesnikovae:
First steps toward integrating profiling with tracing. This PR demonstrates how a Go application can be instrumented with CPU profiler in conjunction with OpenTelemetry and Honeycomb tracing. At this point, this is more of a proof of concept than a practical feature.

Before the PR is merged, you can use dev build:
```
PYROSCOPE_IMAGE=pyroscope/pyroscope-dev:git-d5004149c2357466f709c1142c756197b00a3686 HONEYCOMB_API_KEY={you-api-key} docker-compose up --build
```

TODO:
 - [ ] Make it visible on the timeline (?)
 - [ ] Switch to a separate DB(s).

We also need to decide on wether we should write profiles annotated with `profile_id` to the regular storage as well so that they would be available for regular queries; without `profile_id` label, apparently. The problem is that it would cause additional pressure on the storage: basically, we will have either to duplicate data, or to reference profiles somehow, either way this entails performance implications. On the other hand, if we drop such profiles, users may lack the information: effectively, only traced requests are profiled. I believe, we can not sacrifice that.

---

The example demonstrates how Pyroscope can be used in conjunction with tracing. In the example we will instrument a sample application with OpenTelemetry tracer and will be using [Honeycomb.io](https://www.honeycomb.io) observability platform.

To achieve that, we will be using a special label – `profile_id` that is set by the profiler dynamically. Our simple application specifies Span ID as a profile ID which establishes one-to-one relation between a trace span execution scope and the profiling scope.

By default, only the root span gets annotated (the first span created locally), this is done to circumvent the fact that the profiler records only the time spent on CPU. Otherwise, all the children profiles should be merged to get the full representation of the root span profile.

In practice, the scenario can be different: an arbitrary string can be set as a profile ID. The most important feature is that profiles with the same ID are merged by storage upon insert.

Pyroscope handles profiles with `profile_id` label in a very specific way and stores them separately from others. As a result, such profiles can not be accessed using regular queries that aggregate the data: the very idea of identifiers is to ensure request-level granularity.

There are number of limitations:
  1. Only Go CPU profiling is fully supported at the moment.
  2. Due to the very idea of the sampling profilers, spans shorter than the sample interval may not be captured. For example, Go CPU profiler probes stack traces 100 times per second, meaning that spans shorter than 10ms may not be captured.

image

image


yeya24 (commented):
This is great! Does it work with pull model as well?

kolesnikovae:
Thank you @yeya24! 

Yes, it should work regardless of the delivery mode and I'm planning to add pull mode example as well before merging – PR is still at an early stage.

Any ideas on how to make such an integration more helpful are welcome :)

yeya24:
Instead of searching traceID from profiles, this feature does the opposite thing: search profile ID from traces.
So how do users get the profile ID for each single profile? Is it visible in UI?
And this only makes sense for single profile, not merged profile? :

kolesnikovae:
> Instead of searching traceID from profiles, this feature does the opposite thing: search profile ID from traces.
> So how do users get the profile ID for each single profile? Is it visible in UI?

That's right. The idea is to enable users to query request-level profiles (actually, of the arbitrary scope). I've added a screenshot with Honeycomb UI to depict how to can look like.

> And this only makes sense for single profile, not merged profile? :

I think that we should preserve merged profiles (dropping `profile_id` label, apparently – otherwise, we quickly reach so high cardinality that storage won't be able to handle queries). It's not implemented yet, we need to make the decision.

lizthegrey:
> Instead of searching traceID from profiles, this feature does the opposite thing: search profile ID from traces. So how do users get the profile ID for each single profile? Is it visible in UI? And this only makes sense for single profile, not merged profile? :

Because usually the profile ID is one and the same as the span ID.

PR #770, opened 2022-01-27
feat(frontend): settings page with users and key management

shaleynikov:
WIP Settings page & server


webapp/javascript/services/users.ts
```diff
@@ -0,0 +1,26 @@
...
+  const response = await request('/api/users');
+
+  try {
+    if (response.isOk) {
+      console.log('Parsing', response.value);
+      return parse(response.value);
+    }
+  } catch (e) {
+    console.error(e);
+  }
```
eh-am, line 23: Sorry, why was this needed? The whole point of using `Result` is to not deal with exceptions, so why was this throwing an exception? Let me know, maybe we messed something up in the base implementation.

webapp/javascript/redux/reducers/settings.ts
```diff
@@ -0,0 +1,63 @@
...
+  // There's no 'loading'
+  users:
+    | { type: 'loaded'; data: Users }
+    | { type: 'reloading'; data: Users }
+    | { type: 'failed'; data: Users };
+}
+
+// Define the initial state using that type
+const initialState: SettingsRootState = {
+  users: { type: 'loaded', data: (window as any).initialState.users },
```
eh-am, line 18: Wait is this populated from the server?

webapp/javascript/components/Settings/Users/UserAddForm.tsx
```diff
@@ -0,0 +1,105 @@
...
+    };
+    console.log(JSON.stringify(data));
+    fetch('/api/users', {
+      method: 'POST',
+      body: JSON.stringify(data),
+      headers: [
+        ['Content-Type', 'application/json'],
+        ['Accept', 'application/json'],
+      ],
+    })
```
eh-am, line 39: I assume this is due to the PR being a WIP, but I think this should be moved to the `services/users`?

webapp/javascript/components/Settings/Users/UserAddForm.tsx
```diff
@@ -0,0 +1,105 @@
...
+      .then((r) => r.json())
+      .then((r) => {
+        console.log(r);
+        return r;
+      })
+      .then(() => {
+        setForm({});
+        dispatch(reloadUsers());
```
eh-am, line 47: And all this call to a thunk

webapp/javascript/components/Settings/Users/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+function Users() {
+  const dispatch = useAppDispatch();
+  const users = useAppSelector(selectUsers);
+  console.log('Rendering users');
+  useEffect(() => {
+    if (typeof users === 'undefined') {
+      console.log('Reloading users');
+      dispatch(reloadUsers());
+    }
+  });
```
eh-am, line 35: Is this the case where we load users once?
If so i think it can be written as 
```suggestion
  useEffect(() => {
      dispatch(reloadUsers());
  }, []);
```

webapp/javascript/components/Settings/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+      <h1>Settings</h1>
+      <nav>
+        <ul className={styles.settingsNav}>
+          <li>
+            <NavLink
+              to={`${url}`}
+              exact
+              className={(isActive) =>
+                styles.navLink + (isActive ? ` ${styles.navLinkActive}` : '')
+              }
```
eh-am, line 30: Can we simplify these with that https://www.npmjs.com/package/classnames lib?

webapp/javascript/components/Sidebar.tsx
```diff
@@ -71,6 +72,10 @@ export default function Sidebar2() {
     isRouteActive('/adhoc-single') ||
     isRouteActive('/adhoc-comparison') ||
     isRouteActive('/adhoc-comparison-diff');
+  const isSettingsActive =
+    isRouteActive('/settings') ||
+    isRouteActive('/settings/users') ||
+    isRouteActive('/settings/api-keys');
```
eh-am, line 78: Can we be lazy and use `isRouteActive(/settings*/)` that requires updating that `isRouteActive` function to support regex but it should be fine?

My worry is having to come back to update this function every time we add a new tab to the settings.

eh-am:
Also would be nice to use https://github.com/pyroscope-io/pyroscope/blob/main/stories/Notifications.stories.tsx when the requests fail!

webapp/javascript/components/Settings/Users/UserAddForm.tsx
shaleynikov: Yeah, ofc all api related code goes to `services`. I just was at high speed :-)

webapp/javascript/redux/reducers/settings.ts
shaleynikov: Actually It looks like I left something uncommitted

webapp/javascript/services/users.ts
shaleynikov: it may throw an exception once it doesn't pass `zod` validation inside `parse(response.value);`

kolesnikovae:
LGTM. Thank you @shaleynikov, nice job! I'll take care of the backend stuff.

webapp/templates/login.html
```diff
@@ -13,14 +13,66 @@
...
+          <div>
+            <div><label>Email or username</label></div>
+            <input type="text" name="user" autocapitalize="none" placeholder="email or username" id="username" />
+          </div>
+          <div>
+            <div><label>Password</label></div>
+            <input type="password" name="password" aria-label="password input field" id="password" autocomplete="current-password"  placeholder="password" />
+          </div>
+          <div>
+            <button type="button" class="sign-in-button" onClick="login()">Log In</button>
```
petethepig, line 60: I'd move `login` function to `form onSubmit`, and make the button `type=submit`. that way it works when people press enter and things like that.

shaleynikov: Sure, I'd change all other forms same way.

PR #772, opened 2022-01-28
improve: fix unsafeStrToSlice panic due to empty string

linthan:
due to empty string, the unsafeStrToSlice will result in the panic of the pyroscope server.


abeaumont:
Thanks for reporting and fixing the issue @linthan. I believe it may be better to remove `unsafeStroToSlice` completely and to have better support for empty strings. Take a look to #773

kolesnikovae (commented):
Thank you for fixing the issue @linthan!

I’m wondering if this is a hypothetic case or we do have traces with empty function/frame names. That would mean we should make sure such a profile is handled correctly along the whole workflow (ingest/query).

kolesnikovae (approved):

PR #773, opened 2022-01-28
fix: remove unsafe string conversion and support empty strings.

abeaumont:
This is a different approach to PR #772. Instead of filtering out
empty strings in one of the possible insertion paths, we handle it
properly to prevent problems both from this or other insertion paths:
- We remove the unsafe string to bytes conversion, using instead the
insertion functions that handle strings.
- We handle the corner case of empty strings properly to prevent an
infinite loop.

### DIFF
--- a/pkg/storage/tree/tree_profile.go
+++ b/pkg/storage/tree/tree_profile.go
@@ -2,8 +2,6 @@ package tree
 
 import (
 	"encoding/binary"
-	"reflect"
-	"unsafe"
 
 	"github.com/cespare/xxhash/v2"
 )
@@ -106,7 +104,7 @@ func (r *ProfileReader) readTrees(x *Profile, c labelsCache, f Finder) {
 	if len(indexes) == 0 {
 		return
 	}
-	stack := make([][]byte, 0, 16)
+	stack := make([]string, 0, 16)
 	for _, s := range x.Sample {
 		for i := len(s.LocationId) - 1; i >= 0; i-- {
 			// Resolve stack.
@@ -128,21 +126,17 @@ func (r *ProfileReader) readTrees(x *Profile, c labelsCache, f Finder) {
 				if !ok {
 					continue
 				}
-				stack = append(stack, unsafeStrToSlice(x.StringTable[fn.Name]))
+				stack = append(stack, x.StringTable[fn.Name])
 			}
 		}
 		// Insert tree nodes.
 		for i, vi := range indexes {
 			if v := uint64(s.Value[vi]); v != 0 {
-				c.getOrCreate(types[i], s.Label).InsertStack(stack, v)
+				c.getOrCreate(types[i], s.Label).InsertStackString(stack, v)
 			}
 		}
 		stack = stack[:0]
 	}
-}
-
-func unsafeStrToSlice(s string) []byte {
-	return (*[0x7fff0000]byte)(unsafe.Pointer((*reflect.StringHeader)(unsafe.Pointer(&s)).Data))[:len(s):len(s)]
 }
 
 // sample type -> labels hash -> entry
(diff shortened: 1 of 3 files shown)


pkg/storage/tree/tree_profile.go
```diff
@@ -128,23 +126,19 @@ func (r *ProfileReader) readTrees(x *Profile, c labelsCache, f Finder) {
...
 				}
-				stack = append(stack, unsafeStrToSlice(x.StringTable[fn.Name]))
+				stack = append(stack, x.StringTable[fn.Name])
 			}
 		}
 		// Insert tree nodes.
 		for i, vi := range indexes {
 			if v := uint64(s.Value[vi]); v != 0 {
-				c.getOrCreate(types[i], s.Label).InsertStack(stack, v)
+				c.getOrCreate(types[i], s.Label).InsertStackString(stack, v)
```
kolesnikovae, line 135: I don't think we should do this. I've considered the option to use [the compiler optimisation](https://github.com/golang/go/wiki/CompilerOptimizations#range-over-bytes) but refused due to the performance costs (see https://github.com/pyroscope-io/pyroscope/pull/663, Failed attempts p.2):

```
goos: darwin
goarch: amd64
pkg: github.com/pyroscope-io/pyroscope/pkg/storage/tree
cpu: Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz

Benchmark_ProfileReader_String-4   	     165	   7073024 ns/op	  409041 B/op	    9226 allocs/op
Benchmark_ProfileReader_Bytes-4   	     445	   2795861 ns/op	  409296 B/op	    9226 allocs/op
```

`Benchmark_ProfileReader_String` – this branch.
`Benchmark_ProfileReader_Bytes` – main.

We can see that both approaches ~equal in terms of allocations (the compiler does not allocate space for the byte slice in both cases) but we also can see that `Benchmark_ProfileReader_Bytes` solution is twice more efficient in terms of CPU time. The benchmark can be found here: https://github.com/pyroscope-io/pyroscope/pull/774

I'd like to emphasise that this path is a pretty hot one: such a performance degradation will reflect on the resource consumption. But it will hardly affect ingestion throughput, because of the memory bounds.

---

Instead, we could change treeName label type to string, which would allow us to eliminate any tricks:
```go
func (n *treeNode) insertStr(targetLabel string) *treeNode {
	i := sort.Search(len(n.ChildrenNodes), func(i int) bool {
		return strings.Compare(n.ChildrenNodes[i].Name, targetLabel) >= 0
	})
	if i > len(n.ChildrenNodes)-1 || n.ChildrenNodes[i].Name != targetLabel {
		// Not sure if we need to copy targetLabel.
		// At least in case of pprof.
		child := newNode(targetLabel)
		n.ChildrenNodes = append(n.ChildrenNodes, child)
		copy(n.ChildrenNodes[i+1:], n.ChildrenNodes[i:])
		n.ChildrenNodes[i] = child
	}
	return n.ChildrenNodes[i]
}
```

abeaumont: Oh, I missed you have tried this route and discarded this. Thanks for taking the effort of adding benchmarking information.

Switching label type to string OTOH would make it worse when we have a byte slice as a key, which I think happens in some other code paths... otherwise, it sounds good!

kolesnikovae: I should have added a benchmark earlier, sorry! I'm not happy that we do this obscure unsafe thing.

> Switching label type to string OTOH would make it worse when we have a byte slice as a key, which I think happens in some other code paths... otherwise, it sounds good!

Yep, therefore I postponed this change when was working on this - it requires way more careful analysis I could do at that time. Moreover, I still think we can replace this label with an offset+length mark and use a reusable byte buffer.

abeaumont:
Closing as this is not the best approach

Issue #692, opened 2022-01-10
Create github action to automatically resolve conflicts in yarn.lock or go.mod files


petethepig:
We often encounter this situation where there's a merge conflict in either `yarn.lock` or `go.mod` file. Here's an [example of a PR with such issue](https://github.com/pyroscope-io/pyroscope/pull/691)

![Screen Shot 2022-01-09 at 9 48 59 PM](https://user-images.githubusercontent.com/662636/148723114-0eca2363-8e1e-41d2-af0a-a48e2d908acf.png)

Every time this happens somebody has to manually do something along these lines:

```bash
git checkout <pr-branch>
git merge main
yarn install # or go mod tidy
git add yarn.lock # or git add go.sum
git commit -m 'resolves merge conflict'
git push origin <pr-branch>
```

This happens quite often and we waste a lot of time doing this manually. I figured this can be automated. 

Unfortunately, I could not find any off-the-shelf solutions. The closest I could find is this gist: https://gist.github.com/amacneil/60bf679f357bad9d62103cfdc86cbd74

We need to either:
* find something existing that can deal with this issue / or
* create a new github workflow that would do this

---

For commit author we can use pyroscopebot, here's a snippet we use in other places:
```
git config user.name "Pyroscope Bot"
git config user.email "dmitry+bot@pyroscope.io"
```

PR #780, opened 2022-01-31
chore: Added action to solve conflicts on yarn.lock and go.sum

juliosaraiva:
This PR will solve the Issue #692

### DIFF
--- a/.github/workflows/fix-conflicts.yml
+++ b/.github/workflows/fix-conflicts.yml
@@ -1,10 +1,12 @@
 name: Solve conflicts
 
-on:
-  pull_request:
-    types:
-      - synchronize
-      - labeled
+on: workflow_dispatch
+# TODO: when we confirm this works well we should uncomment this
+# on:
+#   pull_request:
+#     types:
+#       - synchronize
+#       - labeled
 
 jobs:
   fix-conflicts:
--- a/.github/workflows/fix-conflicts.yml
+++ b/.github/workflows/fix-conflicts.yml
@@ -1,0 +1,32 @@
+name: Solve conflicts
+
+on:
+  pull_request:
+    types:
+      - synchronize
+      - labeled
+
+jobs:
+  fix-conflicts:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Resolves merge conflicts
+        if: github.event.label.name == "has conflicts"
+        run: |
+          git config user.name "Pyroscope Bot"
+          git config user.email "dmitry+bot@pyroscope.io"
+          files=(`git diff --name-only --diff-filter=U -- yarn.lock go.sum`)
+          if [ ${#files[@]} -ge 1 ];then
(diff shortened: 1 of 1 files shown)


.github/workflows/fix-conflicts.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Resolves merge conflicts
+        if: github.event.label.name == 'has conflicts'
+        run: |
+          git config user.name "Pyroscope Bot"
+          git config user.email "dmitry+bot@pyroscope.io"
+          files=(`git diff --name-only --diff-filter=U -- yarn.lock go.mod`)
+          if [ ${#files[@]} -ge 1 ];then
+            git checkout origin/main -- "$files"
+            yarn
+            go mod vendor
```
petethepig, line 23: why do `go mod vendor` and not something like `go mod tidy`?

.github/workflows/fix-conflicts.yml
```diff
@@ -0,0 +1,27 @@
+name: Solve conflicts
+
+on:
+  pull_request:
+    types:
+      - synchronize
```
petethepig, line 6: what does this `synchronize` thing mean? I couldn't find anything in github actions docs

.github/workflows/fix-conflicts.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  fix-conflicts:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Resolves merge conflicts
+        if: github.event.label.name == 'has conflicts'
```
petethepig, line 15: so this label is also added by a github workflow / action. What if that workflow runs after this new workflow you added? is this what that `syncronize` thing above is for?

.github/workflows/fix-conflicts.yml
juliosaraiva: `syncronize` means that this action will be trigged each time that a push occurs to PR, but it will only run if is there any label `has conflicts`
Here are some references:
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request
https://frontside.com/blog/2020-05-26-github-actions-pull_request/

.github/workflows/fix-conflicts.yml
juliosaraiva: I was confused about that option. I'll update it to this one.

.github/workflows/fix-conflicts.yml
juliosaraiva: The ideia is that this workflow should run each time that a PR is update with a push request, but the step that will solve conflicts will run only if a conflict in any of these two files was found

.github/workflows/fix-conflicts.yml
juliosaraiva: @petethepig For GoLang I guess the file must be `go.sum` instead of `go.mod`, what do you think?

PR #785, opened 2022-02-01
feat: add some basic flamebearer validation.

abeaumont:


pkg/structs/flamebearer/flamebearer.go
```diff
@@ -100,3 +102,46 @@ func newTimeline(timeline *segment.Timeline) *FlamebearerTimelineV1 {
...
+	}
+	if len(fb.Flamebearer.Levels) == 0 {
+		return fmt.Errorf("a profile must have at least one profiling level")
+	}
+	var mod int
+	switch format {
+	case tree.FormatSingle:
+		mod = 4
+	case tree.FormatDouble:
+		mod = 7
```
eh-am, line 132: Maybe we also need to check `leftTicks` and `rightTicks` exist?
https://github.com/pyroscope-io/pyroscope/issues/762

Also `numTicks` and `maxSelf` ?

eh-am (approved):
Beautiful!

I honestly wonder if at this point we shouldn't call this validation from both points (input and output), just to give us some peace of mind.

Another discussion to be made is v2 having other fields such as the ones I added in the `/render` endpoint. Which in hindsight didn't seem like such a good idea.
 https://github.com/pyroscope-io/pyroscope/blob/49da33846baa66efee669506c9428777c48500d9/pkg/server/render.go#L115-L131

abeaumont:
> I honestly wonder if at this point we shouldn't call this validation from both points (input and output), just to give us some peace of mind.

That's a good point. I think generation should be already covered statically with a test suite rather than having to check this dynamically on every call.

eh-am:
> leave them out of the format?
To leave them out in the format. Now we have diferent models (`/render` vs `/adhoc` for example), which affects how the UI behaves. 

To give you a concrete example, when exporting an adhoc file via the UI, we can't easily generate a file `${appname}.json`, since the model from adhoc doesn't have the appName there.

Rperry2174:
> To give you a concrete example, when exporting an adhoc file via the UI, we can't easily generate a file ${appname}.json, since the model from adhoc doesn't have the appName there.

Not sure how much extra work it would be or if this PR covers this but would be great to support this usecase somehow:
easily generate a file `${appname}.json`

pkg/structs/flamebearer/flamebearer.go
abeaumont: I think I'll leave this out of this PR:
- Having `leftTicks` and `rightTicks` in a single profile is something that shouldn't happen, but it's not a problem either (they'll just be ignored). I think it may be better to focus on a new data format where these requirements are better defined (mandatory when needed, non-existent when they are not).
- I'd prefer to improve our test-suite at tree generation before checking that `numTicks` and `maxSelf` match the expected result.

abeaumont:
This PR doesn't cover changes in the data format, it just validates the data format that we currently support. I have created #791 for changes in the data format and I'll work on a proposal for that. I'd merge this PR as is.

eh-am (approved):

PR #798, opened 2022-02-04
feat(frontend): add package coloring for rust

eh-am:
Closes https://github.com/pyroscope-io/pyroscope/issues/789

This PR contains 2 parts:

# Some prep work, adding tests for existing coloring integrations
Couldn't add much tests for `phpspy` since our example is too simple.
Plus simplified the `default` one, since it seems to use `go` scheme by default.

# Supports rust
For that the regex used is to just get whatever is before the first `::`.

It's a bit naive, since it doesn't deal with cases such as 
```
<core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
```

That @omarabid said it should be treated as `core::future::future::Future::poll`.

But should be a good start.

image


petethepig (approved):
Looks great. Thanks for adding this @eh-am !

abeaumont:
What would you need for PHP? Opening an issue with it may be useful.

PR #799, opened 2022-02-07
feat(frontend): adds ability to export to flamegraph.com

petethepig:


webapp/javascript/components/ExportData.tsx
```diff
@@ -82,6 +112,52 @@ function ExportData(props: ExportDataProps) {
...
+          dlLink.target = '_blank';
+          dlLink.href = json.url;
+
+          document.body.appendChild(dlLink);
+          dlLink.click();
+          document.body.removeChild(dlLink);
+        })
+        .catch((e) => {
+          console.error(e);
+          // TODO: show error to user via notifications
```
petethepig, line 155: @eh-am curious how you think I should implement this. I was thinking about passing some callback (`errorHandler`) from the parent component to this component. That way we preserve its reusability.

eh-am: This is fine for now, but at some point the best of handling this is to lift this up to receive a `exportFlamegraphComFunction`, since  each environment will have a different way of handling this, eg grafana.

PR #810, opened 2022-02-09
allow custom Authorization

EricWJP:
supported: OAuth2 Proxy(Bearer )
   Added support: Basic Authentication (Basic )
change:
   pkg/agent/upstream/remote/remote.go:128
     request.Header.Set("Authorization", "Bearer "+r.cfg.AuthToken)
  to:
   pkg/agent/upstream/remote/remote.go:128
       request.Header.Set("Authorization", r.cfg.AuthToken)

### DIFF
--- a/pkg/agent/upstream/remote/remote.go
+++ b/pkg/agent/upstream/remote/remote.go
@@ -125,7 +125,7 @@ func (r *Remote) uploadProfile(j *upstream.UploadJob) error {
 	request.Header.Set("Content-Type", "binary/octet-stream+trie")
 
 	if r.cfg.AuthToken != "" {
-		request.Header.Set("Authorization", "Bearer "+r.cfg.AuthToken)
+		request.Header.Set("Authorization", r.cfg.AuthToken)
 	}
 
 	// do the request and get the response


abeaumont (changes requested):
Hi @EricWJP, thanks for your contribution!

The change is not backward compatible and it would break all the existing Pyroscope deployments with an auth token. If we want to allow generic authorization I'd suggest creating a new configuration variable (`Authorization` maybe?) instead. If we don't want to support multiple ways of doing the same thing we may need to deprecate `AuthToken`. What do you think @petethepig @kolesnikovae?

kolesnikovae:
Thank you for the contribution @EricWJP! We recently moved the Go client package into a separate module - please see https://github.com/pyroscope-io/client. This client package is still maintained but will be deprecated at some point.

I like the proposed change but as @abeaumont mentioned we can't break backward compatibility. I think we should extend [`pyroscope.Config`](https://github.com/pyroscope-io/client/blob/ee66a0b0711a6c2f83d0a67b00c9e3375610bacd/pyroscope/api.go#L11-L20) to accept arbitrary headers (and maybe cookies) that would override existing ones.

By the way, we're also planning to release a basic Identity and Access Management layer next week, so you won't need any reverse proxies to authenticate clients - that should work out of the box.

If you don't mind, I'll close the PR.

PR #812, opened 2022-02-09
fix: ignore root node when converting a flamebearer to a tree.

abeaumont:


petethepig:
@abeaumont Can you explain the issue a little bit more?

abeaumont:
The problem is related due to the flamebearer format. In a flamebearer, a root node `total` is added, which serves as the first level. This node doesn't exist in the tree, it's generated in the conversion (see for example https://github.com/pyroscope-io/pyroscope/blob/main/pkg/structs/flamebearer/flamebearer_test.go#L27-L53

When we are doing the reverse conversion, from flamebearer to tree, we need to remove this `total` node, otherwise it'll be part of the tree, and when the tree is converted back to a flamebearer, it'll have a duplicated total: the one in the tree, and the one added during the tree to flamebearer conversion (if you use the adhoc diff view, you'll see them duplicated there).

I have created a test for the Diff conversion function to showcase the problem: https://github.com/pyroscope-io/pyroscope/pull/813/commits/f7ca4a2bdd6ed41bfa02247a29a0cd578625ba2d.
The test is currently failing since that branch doesn't contain the fix: https://github.com/pyroscope-io/pyroscope/runs/5140109084?check_suite_focus=true

(I have also found out a different bug while creating that test, #820)

PR #815, opened 2022-02-09
chore(frontend): import improvements

eh-am:
Improves a bunch of things:

* Until we publish the `FlamegraphRenderer` component as a library, the webpack alias just make things more complicated than necessary for people that import that component directly. This PR removes some of the alias (particularly the ones in the path of importing `FlamegraphRenderer`)
* improve typing of components used by `FlamegraphRenderer`
* inline `true-myth` library (since it uses ES modules and we don't want to force users to use it)


eh-am:
Not required anymore after the component is published as a library

PR #829, opened 2022-02-12
chore(lib): publish FlamegraphRenderer as library

eh-am:
Publish our `FlamegraphRenderer` component as library.

It's still pretty beta:
* we don't publish the types
* API is messy
* there's really no concept of versions (we will always publish whatever is in main)

etc.

We will be using it internally and smoothing off the rough edges in the meanwhile.


webapp/lib/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "pyroscope",
+  "version": "0.0.1-%VERSION%",
+  "main": "index.lib.js",
```
petethepig, line 4: I don't think this is right, I only see lib.js and such in the package (no index.lib.js).

webapp/lib/index.ts
```diff
@@ -0,0 +1,5 @@
+import FlameGraphRenderer from '../javascript/components/FlameGraph/FlameGraphRenderer';
+import Flamegraph from '../javascript/components/FlameGraph/FlameGraphComponent/Flamegraph';
+import ExportData from '../javascript/components/ExportData';
+
+export { FlameGraphRenderer as FlamegraphRenderer, Flamegraph, ExportData };
```
petethepig, line 5: I think we should also add this one:
```
import { decodeFlamebearer } from '../node_modules/pyroscope/webapp/javascript/models/flamebearer';
```

Because my understanding is that if the data is coming from a server you need this function to convert it into something the component understands, right?

webapp/lib/package.json
```diff
@@ -0,0 +1,7 @@
...
+  "browser": "index.js",
+  "license": "MIT"
```
petethepig, line 6: Not sure but I think we should probably list dependencies here.

petethepig, line 6: Also, we use Apache 2 for license everywhere else, so we should definitely change that.

webapp/lib/index.ts
eh-am: That's correct, will do.

However just so that we are on the same page, the plan is to move that `decodeFlamebearer` function to happen within the `FlamegraphRenderer` component, so that the API is impler.

webapp/lib/package.json
eh-am: That's a great question. For the browser there's no need to install anything, since we package everything.

Now for node it's a different story, but I don't know how to separate those. Maybe we need different package.json?

PR #848, opened 2022-02-15
chore: add a php-fpm example

samoilenko:
1. cd /examples/php-fpm
2. docker-compose down -v (optional step, if it's not a first run)
3. docker-compose up --build
Screenshot 2022-02-18 at 10 13 02
4. open 127.0.0.1:4040
Screenshot 2022-02-18 at 10 13 22


examples/php-fpm/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+      - '4040:4040'
+    command:
+      - 'server'
+  nginx:
+    container_name: pyroscope_nginx
+    build:
+      context: .
+      dockerfile: nginx/Dockerfile
+    ports:
+      - 80:80
```
eh-am, line 17: Opening ports under 1024 require root IIRC, may be better to use a different one?

eh-am (commented):
Maybe I am missing something, but it looks like we need to access the application once to kick off the php script?

I think it would be nice to at least document that.

examples/php-fpm/docker-compose.yml
samoilenko: nginx was removed from the example

examples/php-fpm/php/index.php
```diff
@@ -0,0 +1,9 @@
+<?php
+$x = 1;
+
+while(true) {
+  echo "The number is: $x\n";
+  $x++;
+  sleep(1);
+}
+?>
```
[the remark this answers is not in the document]
samoilenko: replaced

samoilenko:
> I think it would be nice to at least document that.

I removed Nginx from the example and now to get metrics, containers just should be run

examples/php-fpm/php/run.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+php-fpm -D
+
+screen -d -m sh -c "SCRIPT_FILENAME=./index.php REQUEST_METHOD=GET cgi-fcgi -bind -connect 127.0.0.1:9000"
+
+pid=$(pgrep "php-fpm: master process")
+
+pyroscope connect --pid $pid --spy-name phpspy
```
abeaumont, line 9: I think this can be simplified by using `exec` instead of connect, no need to separate creating from profiling and the `pgrep` would be needed, right?

examples/php-fpm/php/run.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+php-fpm -D
+
+screen -d -m sh -c "SCRIPT_FILENAME=./index.php REQUEST_METHOD=GET cgi-fcgi -bind -connect 127.0.0.1:9000"
```
abeaumont, line 5: This should work fine without `screen`, it'd be nice to avoid the dependency when running the example :)

abeaumont:
This looks good, thanks @samoilenko! Left a couple of minor comments

eh-am:
Hey! Unfortunately it's still not working here :(

While trying to debug, I brought back nginx and used `curl` to trigger the request. Seems to be working but there may be a better of doing it.

```diff
diff --git a/examples/php-fpm/docker-compose.yml b/examples/php-fpm/docker-compose.yml
index 33bcef91..697d7efd 100644
--- a/examples/php-fpm/docker-compose.yml
+++ b/examples/php-fpm/docker-compose.yml
@@ -15,5 +15,17 @@ services:
       dockerfile: php/Dockerfile
     cap_add:
       - SYS_PTRACE
+  nginx:
+    container_name: pyroscope_nginx
+    build:
+      context: .
+      dockerfile: nginx/Dockerfile
+  client:
+    container_name: pyroscope_ubuntu
+    image: curlimages/curl
+    command:
+    - curl
+    - pyroscope_nginx
+
 
 
diff --git a/examples/php-fpm/nginx/conf/Dockerfile b/examples/php-fpm/nginx/conf/Dockerfile
new file mode 100644
index 00000000..a2a872a2
--- /dev/null
+++ b/examples/php-fpm/nginx/conf/Dockerfile
@@ -0,0 +1,3 @@
+FROM nginx:1.9-alpine
+
+COPY nginx/conf/nginx.conf /etc/nginx/conf.d/default.conf
diff --git a/examples/php-fpm/nginx/conf/nginx.conf b/examples/php-fpm/nginx/conf/nginx.conf
new file mode 100644
index 00000000..a2c80e5b
--- /dev/null
+++ b/examples/php-fpm/nginx/conf/nginx.conf
@@ -0,0 +1,17 @@
+server {
+   listen       80 default_server;
+   root         /var/www/html;
+   index        index.php;
+
+    location / {
+        try_files $uri $uri/ /index.php?$query_string;
+    }
+
+   location ~ \.php$ {
+        fastcgi_split_path_info ^(.+\.php)(/.+)$;
+        fastcgi_pass   pyroscope_php-fpm:9000;
+        fastcgi_index  index.php;
+        fastcgi_param  SCRIPT_FILENAME  $document_root$fastcgi_script_name;
+        include        fastcgi_params;
+    }
+}
```

samoilenko:
@abeaumont @eh-am let's decide which option (with Nginx or with cgi-fcgi) should be left. After that can continue. :)

eh-am:
@samoilenko up to you, my point is that I could NOT get it working with our previous commit.

examples/php-fpm/php/run.sh
samoilenko: done

examples/php-fpm/php/run.sh
samoilenko: done

abeaumont (approved):
@samoilenko This looks much better now, great work! I'd move the example inside the php/ dir (move the example already there to a subdirectory), to group the different examples by language, (see for example the python/ dir)

eh-am (approved):
Looks good!

Only question I have is about the stacktrace, shouldn't we be seeing the `work/fastFunction/slowFunction/dummy` there?

image

petethepig:
@eh-am sorry, I merged because I saw you approved it and I missed you comment. 

@samoilenko — i guess it would be nice to understand why we only see `sleep`s. I would recommend replacing a sleep with a loop. something like `while (i < 10000) { i++; }`. Feel free to add it in a separate PR.

samoilenko:
@petethepig Hi, I'll do it, but after we win against russia. Now I can't do it.

petethepig:
@samoilenko Fair enough. Stay safe, Слава Україні!

PR #868, opened 2022-02-18
feat: extract various components

shaleynikov:


webapp/javascript/components/Settings/Security/ChangePasswordForm.tsx
```diff
@@ -57,37 +58,31 @@ function ChangePasswordForm(props) {
...
+          <InputField
+            label="New password"
+            type="password"
+            placeholder="New password"
+            name="password"
+            required
+            onChange={handleChange}
+          />
+          <InputField
+            label="Conform new password"
```
eh-am, line 79:
```suggestion
            label="Confirm new password"
```

webapp/javascript/ui/StatusMessage/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+  children: string;
+}
+
+export default function StatusMessage({ type, children }: StatusMessageProps) {
+  return children ? (
+    <div
+      className={cx({
+        [styles.statusMessage]: true,
+        [styles.error]: type === 'error',
+        [styles.success]: type === 'success',
```
eh-am, line 16: This kinda implies it's possible to have both `styles.error` and `styles.success`, which is obviously untrue.

Can we rewrite as a pure function?
```js
const  getClassnameForType = () => {
  switch (type) {
    case 'error': return styles.error;
    case 'success': return styles.success;
  }
}
```

PR #871, opened 2022-02-18
docs: add an adhoc push example for rust integration.

abeaumont:
This is not yet working as it depends on
https://github.com/pyroscope-io/pyroscope-rs/issues/8

It also shows a couple of things that could be improved from the UX perspective: the need to call start and stop explicitly.


Rperry2174:
Going to close for now... we can re-add this once https://github.com/tikv/pprof-rs/pull/102 is merged

PR #873, opened 2022-02-18
ci: adds oauth mock

petethepig:
cc @eh-am @kolesnikovae @shaleynikov

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,6 +113,7 @@
     "jest-image-snapshot": "^4.5.1",
     "lint-staged": "^11.1.2",
     "monaco-editor-webpack-plugin": "^1.9.0",
+    "oauth2-mock-server": "^4.3.0",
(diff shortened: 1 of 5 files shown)


eh-am:
I think that's perfect, that's exactly what we need!

Just one thing I didn't understand: what does the gitlab flow has different from the others? From my perspective that's a generic Oauth one no?

petethepig:
> Just one thing I didn't understand: what does the gitlab flow has different from the others? From my perspective that's a generic Oauth one no?

Yeah, so the reason I'm using gitlab are:
* we don't have a dedicated "generic oauth" provider oauth
* out of all the ones we do have gitlab is the only one (I think) that has support for custom URLs (and the reason that one is set up like that is because gitlab has an open source version that you can install yourself.

Issue #592, opened 2021-12-08
Provide web endpoints to view targets discovered using k8s service discovery 


dannykopping:
As an operator, I'd like an easy way to know which pods are being discovered by pyroscope, so that I can troubleshoot why some pods are missing. A log line could work, but this would likely just log once and it'd be hard to track down. A web UI would probably be best.

Prometheus provides a view: http://demo.robustperception.io:9090/service-discovery
This allows the operator to see which pods were discovered, which labels they have and what the relabel config execution produces.

http://demo.robustperception.io:9090/targets is also pretty useful, to know which targets are actually scraped, how long ago they were scraped and how long it took.

PR #877, opened 2022-02-21
Provide web endpoints to view targets discovered using k8s service di…

gawicks:
…scovery #592

image


pkg/server/controller.go
```diff
@@ -264,11 +273,41 @@ func (ctrl *Controller) serverMux() (http.Handler, error) {
 		{"/metrics", promhttp.Handler().ServeHTTP},
 		{"/exported-metrics", ctrl.exportedMetricsHandler},
 		{"/healthz", ctrl.healthz},
+		{"/targets", ctrl.activeTargetsHandler},
```
petethepig, line 276: I think this should be in `diagnosticSecureRoutes`, we don't want everyone to have access to this.

webapp/javascript/components/ServiceDiscoveryApp.tsx
```diff
@@ -0,0 +1,206 @@
...
+  return (
+    <div
+      className="p-2"
+      style={{
+        fontSize: '12px',
+        width: '100%',
+        height: '100%',
+        overflow: 'auto',
+        margin: '0 0 0 20px',
+      }}
```
petethepig, line 51: we're trying to switch to this pattern for styles:
* you put all styles in a module.scss file https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/components/ExportData.module.scss
* then you import it from your component https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/components/ExportData.tsx#L11
* then you use classes by names https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/components/ExportData.tsx#L247

gawicks: Please ignore the inline styles. I've purposefully left it this way so I could iterate on the design faster!

eh-am (commented):
Love this, it's gonna be quite useful!

I have a couple comments:
1. Can you do the data fetching part in a `service`, and do some runtime validations on the data?

Example https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/services/appNames.ts#L22

2. Can you do the redux bit in its own store? Similar to https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/redux/reducers/settings.ts
Also if you could use `redux-toolkit`? Similar to https://github.com/pyroscope-io/pyroscope/blob/2bebc82be08641dd9c4901675555901b5781740b/webapp/javascript/components/NameSelector.tsx#L34

petethepig (changes requested):
Code looks really good. One thing I'm thinking is for consistency's sake I think we should make the output in /targets use `camelCase` (lower case for the first letter).

gawicks:
@petethepig the casing of the output should be fixed in yesterday's commits.

dannykopping:
Sweet, thanks a lot!

PR #889, opened 2022-02-23
feat: added nodejs profiler demo

shaleynikov:
It implements `(100).times("very ")` basic implementation of nodejs profiler, which uses `@datadog/pprof`

Just run `make dev` and select `node.cpu{}` or `node.wall{}` from app list


eh-am:
Superseeded by https://github.com/pyroscope-io/pyroscope-nodejs

PR #898, opened 2022-02-28
feat: add pprof profiles multiplexing

kolesnikovae:
The PR addresses https://github.com/pyroscope-io/pyroscope/issues/818.

It's important to note that this affects only pprof format (Go clients). Profiles of other formats are ingested one by one, making it almost impossible to merge scope-tagged ones (with `profile_id` label) into "baseline" profiles. This is not a problem as for now because the tracing integration support was announced only for Go; in future we should either implement this logic in the client libraries, or switch to pprof.


lizthegrey:
Oh thank fuck, thank you all. Let me know when it gets into a tagged release.

PR #909, opened 2022-03-02
chore(webapp): move to lerna

eh-am:
`make dev` should still work fine.

This is not the final work, but since the PR is already too big, it's a good idea to continue in a new Pull Request.


eh-am:
After internal conversation we decided it does not make sense moving `webapp` to `packages/webapp`, so I am closing this PR in favour of https://github.com/pyroscope-io/pyroscope/pull/919

Issue #907, opened 2022-03-02
Remove non-empty 'Full name' and 'email' requirement


eh-am:
In settings page the frontend makes full name and email to be required, but that's not really required from backend perspective. We should relax these requirements in the frontend.

![Peek 2022-03-02 11-13](https://user-images.githubusercontent.com/6951209/156378632-a85a2ee0-b005-4a0f-82fc-42ae9c27e9f8.gif)

webstradev:
Started to take a look at this, unsure why the full name is behaving as a required field because out of the three it looks like only Username and Email have a required attribute. But as I said, I will take a look.

![image](https://user-images.githubusercontent.com/82543732/156444148-f7ba2a66-5d3f-4276-8ff8-ce0140acc057.png)

eh-am:
Closed by https://github.com/pyroscope-io/pyroscope/pull/912

PR #912, opened 2022-03-03
chore: allow empty email on edit profile page

webstradev:
Solves #907 

This pull request removes the required attribute from the email field on the Edit Profile page (/settings).

The initial issue also states that the Full Name field has a non-empty requirement, however, this does not seem to be the case on the latest build. (See [this image](https://user-images.githubusercontent.com/82543732/156444148-f7ba2a66-5d3f-4276-8ff8-ce0140acc057.png) which was the 3 input fields before I started)

### DIFF
--- a/webapp/javascript/components/Settings/Preferences/index.tsx
+++ b/webapp/javascript/components/Settings/Preferences/index.tsx
@@ -72,7 +72,6 @@ function Preferences(props) {
           type="text"
           placeholder="email"
           value={form?.email}
-          required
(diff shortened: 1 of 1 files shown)


webstradev:
I'm not sure why the build has failing with branch not found. It's my first time contributing here.

eh-am:
RE: the job failing is due to https://github.com/pyroscope-io/pyroscope/issues/913

So don't worry about it

PR #918, opened 2022-03-04
Refactor: move colors to css global variables

EgorMozheiko:


eh-am:
There's something wrong with 
* packages/pyroscope-datasource-plugin
* packages/pyroscope-flamegraph
* packages/pyroscope-panel-plugin

As far as I can see they haven't been touched, so why is there a version bump for all of them?

PR #933, opened 2022-03-10
feat(frontend): add confirmation modals stories

EgorMozheiko:


eh-am (changes requested):
Hey, I am not fully understanding what's this for?

My understanding is that the `ConfirmDelete` component is not really "generic", in a sense that it can't be used in other places. To me, it's strictly scoped to the Settings page.

To me, the only thing generic is the "Modal", as presented here https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/ui/Modals/index.ts

Then the storybook would show different aspects of that Modal (a danger modal, a confirmation modal etc), similar to what grafana has here https://developers.grafana.com/ui/latest/index.html?path=/story/overlays-confirmmodal--basic

What I recomm

eh-am:
Hey @EgorMozheiko !

In that same fashion, we should move `ConfirmDelete` out of `@ui/Modals` since it's not really generic. The rule of thumb is to think: will other pyroscope applications (eg backoffice apps) use `{componentName}`?

The point is that we plan on moving the whole `@ui` into its own package. Pretty much like a company-wide component library.

PR #936, opened 2022-03-11
fix(frontend): date range picker styling

EgorMozheiko:


eh-am (changes requested):
Hey @EgorMozheiko 

For PRs like this it's kinda difficult for us to figure out what are you solving specifically. 
Can you add more details?
Eg a screenshot of previous/after

Also, I assume that style it was overwriting some other styling, that's why you made it more specific. Can confirm that's the reasoning?

EgorMozheiko:
@eh-am 
This is the fix for https://github.com/pyroscope-io/pyroscope/issues/876.
The problem was that the styles for submit button got applied to the month switch buttons. 
Before: 
image
Now:
image

PR #942, opened 2022-03-14
feat(webapp): allow comparing distinct queries/tags

eh-am:
**THIS IS FOR THE COMPARISON PAGE ONLY**

Related to https://github.com/pyroscope-io/pyroscope/issues/658

Allows to compare 2 distinct queries (and apps, for that matter):
image

# "Main" timeline uses line visualization
image

# Fallbacking to bar
When there's a single data point, we fallback to `bar`, otherwise there's nothing visible (to make a line you need 2 points after all)
image

# Skewing
When the timelines are the same (when you are querying the same application), we slightly skew the second one so that's visible. Although it becomes technically incorrect data, I hope it helps people visualize that are indeed 2 timelines.
image

THE HEURISTIC THAT FIGURES OUT IF 2 TIMELINES ARE THE SAME NEEDS TO BE IMPROVED, RIGHT NOW IT CHECKS FOR 100% EXACTNESS  (IE WE SUM THEM AND CHECK IF THE RESULT IS THE SAME). This is **bad**.

# Don't show selection for the opposite timeline

Previous:
image

After:
image
Hopefully it makes it less confusing


webapp/javascript/components/TimelineChartWrapper.tsx
```diff
@@ -242,23 +228,121 @@ class TimelineChartWrapper extends React.Component<
...
+    return false;
+  }
+
+  const add = (acc: number, a: number) => acc + a;
+
+  // TODO: actually check if they are the same
+  // this is a very poor heuristic
+  const sumA = dataA.samples.reduce(add, 0);
+  const sumB = dataB.samples.reduce(add, 0);
+  return sumA === sumB;
```
eh-am, line 324: There are many ways to improve this but I am looking for the simplest one. Any suggestions?

petethepig:
@Rperry2174 and I reviewed it and played with it a bunch — everything works great, merging. Good job!

PR #946, opened 2022-03-15
Chore/improve service error

eh-am:
Closes https://github.com/pyroscope-io/pyroscope/issues/941

# Displays more info to the user
image

We can assume most users of pyroscope are quite technical and would like to know as much as possible when something fails. Hence we display the `error.message` in a notification message

# Log full errors to the console
In the same manner, we also log the entire Error object to the console (using a redux middleware):
image

In the future we can report these kind of errors to the server, so that we can understand better what kind of errors our users have.

### DIFF
--- a/webapp/javascript/redux/store.ts
+++ b/webapp/javascript/redux/store.ts
@@ -8,6 +8,7 @@ import {
   PURGE,
   REGISTER,
 } from 'redux-persist';
+import { deserializeError } from 'serialize-error';
 
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore: Until we rewrite FlamegraphRenderer in typescript this will do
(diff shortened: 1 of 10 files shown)


webapp/javascript/redux/store.ts
```diff
@@ -41,9 +42,18 @@ const reducer = combineReducers({
...
-    // TODO: it would be nice to have an actual error here
-    //    console.error(action.error.stack);
+    // since redux-toolkit serializes errors
+    // we should deserialize them back
+    // https://github.com/reduxjs/redux-toolkit/blob/db0d7dc20939b62f8c59631cc030575b78642296/packages/toolkit/src/createAsyncThunk.ts#L94
+    try {
+      const deserialized = deserializeError(action.error);
+      console.error(deserialized);
+
+      // TODO: report error to server?
```
eh-am, line 52: @petethepig here's where the stats api would be useful

webapp/javascript/services/base.ts
```diff
@@ -107,23 +117,35 @@ export async function request(
         }
       }
```
eh-am, line 119: @abeaumont @kolesnikovae
you guys may know this better, is any other case we are missing?
eg server responds with an error message in a different field other than errors | error | message

webapp/javascript/redux/store.ts
petethepig: agreed

webapp/javascript/services/base.ts
abeaumont: I think we are not sending errors in JSON format in all the cases. Some consistency in the API would be helpful here.

PR #947, opened 2022-03-16
feat: extract go templates to react

EgorMozheiko:


webapp/javascript/services/users.ts
```diff
@@ -81,6 +81,45 @@ export async function loadCurrentUser(): Promise<
   return Result.err<User, RequestError>(response.error);
 }
 
+export async function logIn({
+  username,
+  password,
+}: {
+  username: string;
+  password: string;
+}): Promise<any> {
```
eh-am, line 90: Why any? 

If we do know about the response type, then let's actually parse it to validate it's correct.

If we don't know/don't care, then let's just type it as `unknown`.

webapp/javascript/ui/Icons/index.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+
```
eh-am, line 2: Honestly I would say to move this out of `ui/Icons`, since they have a different look and feel from the fontAwesome ones we use. So we probably don't want to have them reusable.

eh-am, line 2: Honestly I would say to move this out of `ui/Icons`, since they have a different look and feel from the fontAwesome ones we use. So we probably don't want to have them reusable.

webapp/javascript/ui/InputField/index.tsx
```diff
@@ -1,15 +1,22 @@
...
 
 /* eslint-disable react/jsx-props-no-spreading */
-function InputField({ label, ...rest }: IInputFieldProps) {
+function InputField({ label, styling, ...rest }: IInputFieldProps) {
   return (
-    <div className={styles.inputWrapper}>
+    <div
+      className={cx({
+        [styles.inputWrapper]: true,
+        [styling]: true,
```
eh-am, line 17: `styling` refers to a classname? if so let's call that instead.

eh-am, line 17: `styling` refers to a classname? if so let's call that instead.

webapp/javascript/index.jsx
```diff
@@ -45,36 +46,47 @@ function App() {
...
-        <Route path="/settings">
-          <Settings />
+        <Route exact path="/login">
+          <SignInPage />
         </Route>
-        <Route path="/service-discovery">
-          <ServiceDiscoveryApp />
+        <Route exact path="/signup">
+          <SignUpPage />
         </Route>
```
eh-am, line 55: Don't we have to add a condition here, since the pyroscope instance may not allow signup?

eh-am, line 55: Don't we have to add a condition here, since the pyroscope instance may not allow signup?

webapp/javascript/index.jsx
```diff
@@ -45,36 +46,47 @@ function App() {
...
-        <Route path="/comparison">
-          <ContinuousComparisonView />
-        </Route>
-        <Route path="/comparison-diff">
-          <ContinuousDiffView />
-        </Route>
-        <Route path="/settings">
-          <Settings />
+        <Route exact path="/login">
+          <SignInPage />
```
eh-am, line 51: Don't we have to add a condition here, since the pyroscope instance may be public?

eh-am, line 51: Don't we have to add a condition here, since the pyroscope instance may be public?

webapp/javascript/components/IntroPages/SignIn/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+    } catch (e) {
+      setForm({ ...form, errors: e.errors || [e.message] });
+    }
+  }
+
+  useEffect(() => {
+    if (currentUser) {
+      history.push('/');
+    }
+  }, [currentUser]);
```
eh-am, line 54: What is this doing specifically?

eh-am, line 54: What is this doing specifically?

webapp/javascript/services/users.ts
```diff
@@ -81,6 +81,45 @@ export async function loadCurrentUser(): Promise<
...
+  const response = await request(`/login`, {
+    method: 'POST',
+    body: JSON.stringify({ username, password: passwordEncode(password) }),
+  });
+  if (response.isOk) {
+    return Promise.resolve(response.value);
+  }
+
+  return Promise.reject(response.error);
```
eh-am, line 99: The pattern we are following is to never fail at the service level, and let that be handled at the component level. Can you return a `Result/Err` instead?

eh-am, line 99: The pattern we are following is to never fail at the service level, and let that be handled at the component level. Can you return a `Result/Err` instead?

webapp/javascript/index.jsx
```diff
@@ -22,8 +22,9 @@ import ServiceDiscoveryApp from './components/ServiceDiscoveryApp';
 import ServerNotifications from './components/ServerNotifications';
 // since this style is practically all pages
 import '@pyroscope/flamegraph/dist/index.css';
-// global css variables
-// import './variables.css';
+
```
eh-am, line 25: Can we move these to `webapp/javascript/pages`?

eh-am, line 25: Can we move these to `webapp/javascript/pages`?

eh-am (changes requested):
Looks quite good! 

I left a few messages, mostly nitpicks.

eh-am:
@EgorMozheiko can you have a look at the CI checks failing?

pkg/server/handler.go
```diff
@@ -47,6 +48,8 @@ func (ctrl *Controller) indexHandler() http.HandlerFunc {
 		} else if path == "/service-discovery" {
 			ctrl.statsInc("service-discovery")
 			ctrl.renderIndexPage(rw, r)
+		} else if path == "/login" || path == "/signup" {
```
petethepig, line 51: let's add other pages as well, particularly `/forbidden`, `/welcome` and `/redirect`

webapp/javascript/pages/IntroPages/NotFound/index.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import cx from 'classnames';
+import { Link } from 'react-router-dom';
+import styles from '../IntroPages.module.css';
+
+function NotFoundPage() {
```
petethepig, line 6: nice touch

webapp/javascript/pages/IntroPages/SignIn/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+            onChange={handleFormChange}
+            required
+          />
+        </div>
+        <button className={styles.button} type="submit">
+          Log in
+        </button>
+        <Divider />
+        <div className={cx(buttonStyles.buttonContainer)}>
+          <Link
```
petethepig, line 97: In the original implementation these are added conditionally, only if these a particular sign in method is enabled. For example: https://github.com/pyroscope-io/enterprise/blob/97a60023e3e9d5b9fa705eacf0900386ee8367c7/webapp/templates/login.html#L50

We should probably pass these variables to the frontend the same way we pass other things for example, here's how we pass "features":
* https://github.com/pyroscope-io/pyroscope/blob/ac5a1284d5d9c925093a2be54deb720eff29bac8/pkg/server/handler.go#L110-L120
* https://github.com/pyroscope-io/enterprise/blob/97a60023e3e9d5b9fa705eacf0900386ee8367c7/webapp/templates/index.html#L39-L41
then later they are accessible via `window.features`



--- 
you can test all of this with a config like this:
```

auth:
  signup-default-role: Admin
  ingestion:
    enabled: true

  # Internal authentication provider settings.
  internal:
    enabled: true

  # Google OAuth2 authentication settings.
  google:
    enabled: false

  # Github OAuth2 authentication settings.
  github:
    enabled: true

  # Gitlab OAuth2 authentication settings.
  gitlab:
    enabled: false

```

webapp/javascript/pages/IntroPages/SignIn/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+  return (
+    <div className={styles.loginWrapper}>
+      <form className={styles.form} onSubmit={handleSubmit}>
+        <div className={styles.formHeader}>
+          <div className={styles.logo} />
+          <h1>Welcome to Pyroscope</h1>
+          <h3>Log in to continue</h3>
+        </div>
+        <div>
```
petethepig, line 69: see my comment below about these fields being conditional. The regular sign in inputs are only available if you have internal authentication enabled, e.g:
```
auth:
  internal:
    enabled: true
```

petethepig (changes requested):
The new pages look great! Good job @EgorMozheiko 

I do have a few comments though:
* I added some comments directly in the code
* Let's move other pages as well, `welcome`, `forbidden` and `redirect`. (and also I guess we can make 404 work?). We should also delete the old templates.
* there's a bug I think where it tries to load a user and fails even though I'm on the login page:
* Screen Shot 2022-03-28 at 11 43 34 AM

pkg/server/handler.go
shaleynikov: Still discussing on the way it should be handled. Seems `/welcome` and `/redirect` does pretty much the same.

webapp/javascript/pages/IntroPages/NotFound/index.tsx
shaleynikov: Now it's handled with go server, but here is a switch so make it work on react side

webapp/javascript/redux/reducers/user.ts
```diff
@@ -34,6 +34,10 @@ export const loadCurrentUser = createAsyncThunk(
     if ('code' in res.error && res.error.code === 404) {
       return Promise.resolve({ id: 0, role: 'anonymous' });
     }
+    // Suppress 401 error on login screen
+    if ('code' in res.error && window?.location?.pathname === '/login') {
```
petethepig, line 38: if a user uses baseURL I think this might be something other than login, right? e.g if baseURL = `/pyroscope` this page will be at `/pyroscope/login`

webapp/javascript/services/base.ts
```diff
@@ -112,7 +112,12 @@ export async function request(
       // Check if it's 401 unauthorized error
       if (response.status === 401) {
         // TODO: Introduce some kind of interceptor (?)
-        if (window && window.location) {
+        if (
+          window &&
+          window.location &&
+          window.location.pathname !== '/login' &&
+          window.location.pathname !== '/signup'
```
petethepig, line 119: same as previous comment

webapp/templates/404.html
```diff
@@ -0,0 +1,54 @@
+<!DOCTYPE html>
```
petethepig, line 1: Could we also render this via react? e.g when server detects a 404 we render "index" page, but it shows a 404 because no route is found?

webapp/javascript/redux/reducers/user.ts
shaleynikov: Noted. 

Yeah that's bascially why we need e2e tests

webapp/javascript/services/base.ts
shaleynikov: fixed

webapp/templates/404.html
shaleynikov: done

webapp/javascript/pages/constants.ts
```diff
@@ -0,0 +1,16 @@
...
+  COMPARISON_VIEW = '/comparison',
+  COMPARISON_DIFF_VIEW = '/comparison-diff',
+  SETTINGS = '/settings',
+  LOGIN = '/login',
+  SIGNUP = '/signup',
+  SERVICE_DISCOVERY = '/service-discovery',
+  ADHOC_SINGLE = '/adhoc-single',
+  ADHOC_COMPARISON = '/adhoc-comparison',
+  ADHOC_COMPARISON_DIFF = '/adhoc-comparison-diff',
+}
```
petethepig, line 12: thanks for doing this, this will definitely make it easier to work with these in the future.

petethepig, line 12: thanks for doing this, this will definitely make it easier to work with these in the future.

petethepig:
Closed in favor of #1031

PR #948, opened 2022-03-17
feat: create a page that shows your current config

gawicks:
localhost_4040_settings_config


pkg/server/config.go
```diff
@@ -1,16 +1,41 @@
...
+	if config.Auth.Google.ClientSecret != "" {
+		config.Auth.Google.ClientSecret = sensitive
+	}
+	if config.Auth.Gitlab.ClientSecret != "" {
+		config.Auth.Gitlab.ClientSecret = sensitive
+	}
+	if config.Auth.Internal.AdminUser.Password != "" {
+		config.Auth.Internal.AdminUser.Password = sensitive
+	}
+
```
gawicks, line 31: I masked everything that looked like a secret in server configuration. Have I missed anything?

webapp/javascript/services/currentConfig.ts
```diff
@@ -0,0 +1,18 @@
...
+export async function fetchCurrentConfig(): Promise<
+  Result<CurrentConfig, RequestError | ZodError>
+> {
+  const response = await request('status/config');
+
+  if (response.isOk) {
+    return parse(response.value);
+  }
+
+  return Result.err<CurrentConfig, RequestError>(response.error);
```
eh-am, line 17: I think this can be simplified with
```
  import { request, parseResponse } from './base';

  (...)
  return parseResponse(response, CurrentConfigSchema);
```

Where `CurrentConfigSchema` is

```
export const CurrentConfigSchema = z.object({
  yaml: z.string(),
});
```

eh-am (commented):
Nice, looks quite good!

A few thoughts:

# Comparing with the previous `/config` I see couple things:
1. We are kinda breaking the API (since we renamed it), but let's assume people have NOT USED IT programatically, so I guess it's fine?
2. It returns values in JSON format. I guess we should still support that.
image

# case sensitveness
I haven't actually verified this, but I wonder if this generated config actually works. Asking because from what I recall, the config we use is separated by dashes (`-`). Example: https://github.com/pyroscope-io/pyroscope/blob/main/scripts/oauth-mock/pyroscope-config.yml

image

# using prism
I wonder if we could use [prism](https://prismjs.com/) to syntax highlight the output to make it more pleasant. 
I know we are already using it for the tags bar. So maybe we could also use it here? https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/components/TagsBar.tsx#L15

pkg/server/config.go
petethepig: Or even better, can we somehow do it with annotations? For example, to avoid these things showing up in JSON output we add `json:"-"` annotations to fields. Maybe we can do the same thing with YAML?

gawicks: The idea was to provide a hint that the parameter has been configured but not necessarily something we should display.

eh-am (approved):
Looks quite good, I like it!

RE: where/how to hide sensitive data @petethepig any thoughts here?

pkg/config/marshal.go
```diff
@@ -0,0 +1,23 @@
...
+	}
+	if config.Auth.Google.ClientSecret != "" {
+		config.Auth.Google.ClientSecret = sensitive
+	}
+	if config.Auth.Gitlab.ClientSecret != "" {
+		config.Auth.Gitlab.ClientSecret = sensitive
+	}
+	if config.Auth.Internal.AdminUser.Password != "" {
+		config.Auth.Internal.AdminUser.Password = sensitive
+	}
```
petethepig, line 21: The fact that we have to manually add these here is scary to me. I'm worried that next time we add a sensitive parameter we'll forget to add it here. How can we avoid that? Maybe with annotations? That's that thing I was taking about, adding `yaml:"-"` to annotations. Would that work?

pkg/config/config.go
```diff
@@ -97,153 +97,153 @@ type Target struct {
...
+	ClientID             string   `json:"-" deprecated:"true" def:"" desc:"client ID generated for Github API" yaml:"client-id" mapstructure:"client-id"`
+	ClientSecret         string   `json:"-" deprecated:"true" def:"" desc:"client secret generated for Github API" yaml:"client-secret" mapstructure:"client-secret"`
+	RedirectURL          string   `json:"-" deprecated:"true" def:"" desc:"url that Github will redirect to after logging in. Has to be in form <pathToPyroscopeServer/auth/github/callback>" yaml:"redirect-url" mapstructure:"redirect-url"`
+	AuthURL              string   `json:"-" deprecated:"true" def:"https://github.com/login/oauth/authorize" desc:"auth url for Github API" yaml:"auth-url" mapstructure:"auth-url"`
+	TokenURL             string   `json:"-" deprecated:"true" def:"https://github.com/login/oauth/access_token" desc:"token url for Github API" yaml:"token-url" mapstructure:"token-url"`
+	AllowedOrganizations []string `json:"-" deprecated:"true" def:"" desc:"list of organizations that are allowed to login through github" yaml:"allowed-organizations" mapstructure:"allowed-organizations"`
```
petethepig, line 246: I'm worried that next time we add a new parameter here we might forget to add these annotations (they are also prone to typos). Is there a way to write a custom linter that would go over all of these structs and ensure `yaml` annotation matches `mapstructure` annotation, and they both match the name (`AllowedOrganizations` in this case)?

petethepig:
@gawicks 

I have a couple of comments:
1. Could you change it so that service discovery and config are under tabs? Kind of like what we have on /settings page.
2. It worries me that next time we add new config parameters we might forget to update `yaml` thing and also add it to list of  sensitive parameters. We need to come up with some way automating the review process for these things.
3. RE JSON, I'm fine getting rid of it if that makes things easier.

gawicks:
@petethepig Fair point, it's easy to miss these magic strings when adding a new parameter. Perhaps a combination of linting and tests would help. I'll look into it.

gawicks:
@petethepig Can we merge this?

kolesnikovae (approved):
Looks good!

Although using yaml tags is slightly annoying and prone to human errors, I don't know a  way to handle this better.

PR #950, opened 2022-03-17
chore(webapp): app and tags selectors search position adjustment

pavelpashkovsky:
Fixed: https://github.com/pyroscope-io/pyroscope/issues/758

Demo: https://monosnap.com/file/FIVtgpV6YC2mk5jvihNQTaJojkczRa
Demo for tag selector: https://monosnap.com/file/YJ6cuKemgjpRQHkWC59SpqlFbdwUTh

### DIFF
--- a/webapp/javascript/components/NameSelector.tsx
+++ b/webapp/javascript/components/NameSelector.tsx
@@ -54,8 +54,9 @@ function NameSelector() {
     </MenuItem>
   )) as ShamefulAny;
 
-  const noApp =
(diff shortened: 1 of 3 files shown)


eh-am (commented):
LOVE IT!!!!

Since it's so easy to do, can we do the same things for tags? `TagsBar` component

pavelpashkovsky:
>Since it's so easy to do, can we do the same things for tags? TagsBar component

@eh-am sure, but can you please help me with example when I can get some amount of real tags from backend on dev mode?

PR #965, opened 2022-03-22
feat(webapp): add modal for custom export name

EgorMozheiko:
Issue: https://github.com/pyroscope-io/pyroscope/issues/872


eh-am (changes requested):
Looks guide good! I added some minor comments

Visually speaking, what if we make the components closer to what's on the rest of the page?
Like the buttons, the input box etc. Otherwise it feels a little too...disconnectex.

image

eh-am:
Looks good, the only thing I noticed is that for diff it uses "flamegraph" as the default name 
image


But merging this anyway.

PR #967, opened 2022-03-22
Feat/diff compare queries

eh-am:


pkg/structs/flamebearer/flamebearer.go
```diff
@@ -109,19 +110,47 @@ func NewProfile(name string, output *storage.GetOutput, maxNodes int) Flamebeare
...
+	}
+
+	if left.SampleRate != right.SampleRate {
+		// if one of them is empty, it still makes sense merging the profiles
+		if left.SampleRate != 0 && right.SampleRate != 0 {
+			msg := fmt.Sprintf("left sample rate (%d) does not match right sample rate (%d)", left.SampleRate, right.SampleRate)
+			return FlamebearerProfile{}, errors.New(msg)
+		}
+	}
+
```
eh-am, line 134: @abeaumont I added some crappy validations here, anything else you find missing?

pkg/adhoc/server/convert.go
```diff
@@ -118,8 +110,7 @@ func DiffV1(name string, base, diff *flamebearer.FlamebearerProfile, maxNodes in
 		name = n
 	}
 
-	fb = flamebearer.NewCombinedProfile(name, out, bOut, dOut, maxNodes)
-	return fb, nil
+	return flamebearer.NewCombinedProfile(name, bOut, dOut, maxNodes)
```
abeaumont, line 113: Note that `bOut`, `dOut` don't have metadata currently, they should be completed with metadata from `base` and `diff`.

pkg/structs/flamebearer/flamebearer.go
```diff
@@ -109,19 +110,47 @@ func NewProfile(name string, output *storage.GetOutput, maxNodes int) Flamebeare
 	}
 }
 
-func NewCombinedProfile(name string, output, left, right *storage.GetOutput, maxNodes int) FlamebearerProfile {
+func isEmpty(t *storage.GetOutput) bool {
```
abeaumont, line 113: nitpick: move all private functions below public functions.

pkg/structs/flamebearer/flamebearer.go
abeaumont: It looks good, not sure about making the fields optional though. I think we may want to make them mandatory and report an error if any of those is missing (I _think_ they are always set up when ingesting, would have to double-check). That would increase the strictness, what do you think?

eh-am: I agree, except that there's a valid case where things are optional: when you are comparing stuff like this, where one of them is empty:
image

"why would you compare an empty one"? it may just happen, for example, your app is brand new (so left half, will be selecting an empty time range)

pkg/adhoc/server/convert.go
eh-am: good catch, thanks!

PR #969, opened 2022-03-23
fix(flamegraph): clicking on anywhere on a row selects that row

eh-am:


eh-am:
is github actions down again?

PR #979, opened 2022-03-25
feat(frontend): better timeline selector

pavelpashkovsky:
Issue: https://github.com/pyroscope-io/pyroscope/issues/953
Demo: https://monosnap.com/file/OKLa31HFRxnplxOBqgyX7AK80n6fFZ

We've done requirements from task:

> - Leave the data bars themselves as pink
> - make the end bars "dashed" grey (lets make this configurable so we can play with different colors)
> - make the "shaded selected area" a slightly opaque grey
> - Add the "handle" on the left and the right side of the range to make it clear that people can change this
> - Remove the text that says "select left time range" because now hopefully it will be clearer

But if you need functionality that user should change selection by dragging side handles, it assumes rewriting Flot source code dramatically, because it doesn't support selection adjustment at all (if user selects some time range, Flot deleted old selection). I can't estimate time for such a refactor at this moment, it might take 1.5-2 weeks.


webapp/javascript/components/TimelineChartWrapper.tsx
```diff
@@ -37,6 +37,8 @@ type TimelineChartWrapperProps = {
     left?: Marking;
     right?: Marking;
   };
+
+  selectionType: 'single' | 'double';
```
eh-am, line 41: Can you comment what this field does? Preferrably using tsdoc https://tsdoc.org/pages/tags/example/

webapp/javascript/components/TimelineChartWrapper.tsx
```diff
@@ -61,6 +63,7 @@ class TimelineChartWrapper extends React.Component<
       },
       selection: {
         mode: 'x',
+        selectionType: props.selectionType || 'single',
```
eh-am, line 66: If it's a required value, then we don't need to set a default no?

webapp/javascript/components/TimelineFlotChartSelection.ts
```diff
@@ -0,0 +1,298 @@
+/* eslint-disable */
+// @ts-nocheck
```
eh-am, line 2: How difficult would it be to add types to this plugin?

webapp/javascript/components/TimelineFlotChart.ts
```diff
@@ -0,0 +1,3706 @@
+/* eslint-disable */
```
eh-am, line 1: What is going on with this file? I assume you didn't write this code from scratch, so where is it coming from?

webapp/javascript/components/TimelineChartWrapper.tsx
```diff
@@ -140,7 +143,7 @@ class TimelineChartWrapper extends React.Component<
       const to = new Date(formatAsOBject(m.to)).getTime();
 
       // We make the sides thicker to indicate the boundary
-      const boundary = { lineWidth: 3, color: m.color.rgb() };
+      const boundary = { lineWidth: 1, color: m.color.rgb() };
```
eh-am, line 146: Question: is this still required? The whole point was to have a "thicker" line. Now if they are exactly the same as normally, then it may not be needed no?

eh-am (commented):
That's cool! Can you go over a little more detail on what was actually implemented? I am a little bit confused  by these 2 new files  `TimelineFlotChart` and  `TimelineFlotChartSelection`. Feels like maintaining them is going to be hard.


RE the design I will defer to @Rperry2174 which is the one that envisioned these changes.

pavelpashkovsky:
@eh-am I'will fix all your comments a bit later, but in general it's styling of timeline selection, requirements from this task assumes editing Flot source code, so I had to extract Flot render logic to separated file and inject some logic into it. If you have more gracefull solution please let me know.

webapp/javascript/components/TimelineFlotChartSelection.ts
pavelpashkovsky: this file was deleted

webapp/javascript/components/TimelineChartWrapper.tsx
pavelpashkovsky: added comment on it

webapp/javascript/components/TimelineFlotChart.ts
pavelpashkovsky: Right, it's Flot's core rendering logic file. Changing appearance of timeline selection assumes editing code here. If you know any easier way please let me know.

eh-am: Alright, can you then
* point to the file this was copied from (and its version), probably just linking to the commit hash in github should be enough
* comment which sections you overwrote with custom logic

my perspective here is that anybody looking at this code 6 months from now will have a hard time understanding what's going on

pavelpashkovsky: added comment on that file

pavelpashkovsky:
@eh-am can you help please, what the correct way to update snapshots locally? `yarn test:ss`? is there any documentation on testing?

webapp/javascript/components/TimelineChartWrapper.tsx
eh-am: >selection looks different

How different? Consider that this is the public api of that component. Ideally one does not have to look into the implementation to figure out what it does.

pavelpashkovsky: updated comment

Rperry2174:
closing for now... we will _definitely_ revisit this soon, but just don't have enough time to play around with it yet

PR #998, opened 2022-04-04
feat: added examples for nodejs profilier

shaleynikov:


shaleynikov:
Here I also fixed race condition on flamegraph being rendered before data on timeline pops out

PR #1004, opened 2022-04-06
fix: disable selection if there is no data inside

pavelpashkovsky:
Issue: https://github.com/pyroscope-io/pyroscope/issues/934
Demo: https://monosnap.com/file/H1kCE6tgSEuxxaGKJd6Hnz1cBzMCTr


eh-am:
Again, do we need to fork the float libary to fix this? Feels like killing an ant with a sledgehammer.

pavelpashkovsky:
@eh-am completely agree with you, but I'm not sure, that there is any easier way to customize Float from outside. Because all functionality of this lib based on adding jQuery plugins. And they don't have any kind of public API. In this PR I adjusted timeline UI behaviour if selection disabled, in this case I disabled binding handlers inside of Flot plugins

eh-am:
My suggestion is to just do a NOOP when the selected value is bogus
https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/components/TimelineChartWrapper.tsx#L225

```js
const customOnSelect = (from, until) => {
  if (!isInvalidInterval(from, until)) {
    this.props.onSelect(from, until);
  }

}
```

Although not as well done as the solution you brought in (the custom icon is pretty nice), it does the job from engineering perspective.

cc @petethepig your thoughts?

pavelpashkovsky:
@eh-am , yeah, it's a bit overengineering, I've just tried to provide solution with predictable user experience. Sure, your solution make sense, I will update the PR in accordance with it a bit later

petethepig:
I agree with Eduardo that copying code from other libraries (essentially forking) should be a last resort type action.

To me it seems that maybe a deeper problem is that when there's no data we return `null` for timeline (see screenshot below). Maybe it would be better if we made it so that the backend returns a timeline full of zeros instead. I think that would partially fix the issue. I'll play with it today and report my findings.

Screen Shot 2022-04-07 at 6 17 55 PM

pavelpashkovsky:
@eh-am , looks like @petethepig's contribution fixed this issue. Please take a look at this video [https://monosnap.com/file/VqM16Y9ps5rm2ZOUA82Uf8s5rMXdwd]

Does this issue still need to have some fix from my side guys? May I close this PR?

eh-am:
Yeah Dmitry's changes seem that fixed the issue, I don't think we need anything here!

PR #1007, opened 2022-04-06
chore(webapp): get rid of initialState

eh-am:
Now we load application names (by making a request) instead of having it prepopulated by the server. This makes the frontend/backend more decoupled (althought we currently add build info and feature flags, which both can be dealt with differently)


webapp/javascript/components/Continuous.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const appNames = useAppSelector(selectAppNames);
+
+  useEffect(() => {
+    async function loadAppNames() {
+      if (appNames.length <= 0) {
+        // Load application names
+        const names = await dispatch(reloadAppNames());
+
+        // Pick the first one
+        if (names.payload.length > 0) {
```
petethepig, line 24: This should only happen if query is empty.

We should also add `{}` to the end of it so that it's clear that you can add tags.

petethepig (changes requested):
There's one issue where if you select an app and then refresh the page the app selection gets lost, I added a comment above, where I think the issue is.

webapp/javascript/components/Continuous.tsx
eh-am: Good catch, thanks!

RE adding `{}` this is tricky (when to add `{}`? when it's baked already?), so I went back and refactored the code to use nominal typing. Now the action expects a `Query` which you can then convert via `brandQuery(string) => String` or `queryFromAppName(string) => Query`

petethepig (approved):
Works great now, good job!

PR #1012, opened 2022-04-08
fix(server): always returns timeline even if there's not data

petethepig:
See my comment [here](https://github.com/pyroscope-io/pyroscope/pull/1004#issuecomment-1092346973) for more context.

I intentionally added it to the controller (presentation layer), and not storage, because there's value in distinguishing between outputs like:
* successful reads, `out != nil`
* no data reads, `out == nil`

### DIFF
--- a/pkg/server/render.go
+++ b/pkg/server/render.go
@@ -96,9 +96,11 @@ func (rh *RenderHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		WriteInternalServerError(rh.log, w, err, "failed to retrieve data")
 		return
 	}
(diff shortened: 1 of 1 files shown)


eh-am:
Closes https://github.com/pyroscope-io/pyroscope/issues/930

eh-am:
U created an issue for that [earlier](https://github.com/pyroscope-io/pyroscope/issues/930), but i believe there's a semantic difference difference between "pyroscope received NO data (empty)" vs "pyroscope did NOT receive any data (eg app is down, network failure etc)".

Are we fine with just making them the same?

PR #1024, opened 2022-04-12
fix: prevents double counting in the table #896

### LINKED ISSUE #896: Total time is wrongly calculated with recursive functions.
As it can be seen in the screenshot, the program runs for ~40s, but `FibKt.main` report 2x and `FibKt.fib` multiple times that:

kotlin

I suppose that the problem is that the `Total` is aggregating the total times of each function by name, without taking into account node connections in the tree: only the first appearance of the function in each of the tree branches should be counted.

shaleynikov:
Got same issues with nodejs profiler as described at #896 . There are many many recursive calls in nodejs so it makes numbers look really wierd.

The idea behind this fix is following: we add only first function appearance to total value and discard next.


eh-am:
This is nice!

I would love to see a unit test though (granted we currently don't have any for that part of the code).

shaleynikov:
It looks like we have recursive functions on `render.json` e2e fixtures. 
Going to update it

eh-am:
That too, but I was thinking unit tests, not only to validate it works, but also as documentation of the behaviour changes (this PR being an example of that).

petethepig:
We're not ready to merge this yet, there are some corner cases that aren't handled properly. Closing again :)

PR #1025, opened 2022-04-12
chore: reduce list of watched files by air

eh-am:
As per https://github.com/pyroscope-io/docs/pull/59

Ideally we would use a allowlist instead of denylist, but... `¯\_(ツ)_/¯`


thisisobate (approved):
Tested locally and it works as expected!

thisisobate (changes requested):
I just noticed the CPU utilization grew from 20% to almost 60%. Tagging @eh-am for visibility.

Rperry2174:
talked with @thisisobate about it live that this is fine

PR #1038, opened 2022-04-19
Add loki for logging and derived fields in tracing example

Rperry2174:
The goal is to eventually add a derived field for pyroscope in addition to jaeger:

For example this configuration:
Screen Shot 2022-04-19 at 10 52 39 AM

Allows us to cick directly on a log and create a query on the opposite side with the `trace_id` in the query box for jaeger

We'd like to be able to do the same thing eventually with Pyroscope i.e.
image

Note: This will be slightly broken at first because not all spans have corresponding profiles (only the root spans do)

cc/ @pavelpashkovsky

### DIFF
--- a/examples/tracing/jaeger/docker-compose.yml
+++ b/examples/tracing/jaeger/docker-compose.yml
@@ -7,6 +7,10 @@ services:
       - '6831:6831/udp'
       - '16686:16686'
       - '14268:14268'
+    logging:
+      driver: loki
+      options:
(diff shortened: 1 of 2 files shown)


yeya24:
Wow love this idea! Any upstream work you are doing on the Grafana side to support this?

Rperry2174:
hey @yeya24 we're going to reach out to the grafana folks and see if they might be interested in merging upstream. For now we are just using [custom image for grafana ](https://github.com/pyroscope-io/grafana) internally

PR #1042, opened 2022-04-20
Update tracing example

Rperry2174:
Update tracing example to have proper embedded flamegraph
image


examples/tracing/jaeger/pyroscope.yaml
```diff
@@ -1,6 +1,7 @@
 ---
 cors:
-  allowed-origins: ['http://localhost:3000', 'http://localhost:16686']
+  allowed-origins:
+    ['http://localhost:3000', 'http://localhost:16686', 'http://localhost:4000']
```
Rperry2174, line 4: @pavelpashkovsky jaeger-ui is served on port :4000 which is why we continued to get cors error

PR #1048, opened 2022-04-21
tests:  added e2e tests for oauth

shaleynikov:


Makefile
```diff
@@ -3,8 +3,6 @@ GOBUILD=go build -trimpath
 ARCH ?= $(shell uname -m)
 OS ?= $(shell uname)
 
-SERVERTARGETS ?= dev server
-SERVERPARAMS ?=$(filter-out $(SERVERTARGETS), $(MAKECMDGOALS))
```
petethepig, line 7: what if it's empty?

petethepig: tested it, I guess it just works :)

.github/workflows/cypress-tests-auth.yml
```diff
@@ -39,7 +39,9 @@ jobs:
         with:
           build: make e2e-build
           wait-on: http://localhost:4040
-          start: make server -- --auth.internal.enabled --auth.internal.signup-enabled
+          start: |
```
petethepig, line 42: didn't know you can do multiple lines like this, that's cool https://github.com/cypress-io/github-action#start-multiple-servers

pkg/server/controller.go
```diff
@@ -293,7 +293,7 @@ func (ctrl *Controller) serverMux() (http.Handler, error) {
 		{"/healthz", ctrl.healthz},
 	})
 
-	r.NotFoundHandler = ctrl.notfoundHandler()
```
petethepig, line 296: pls delete notfoundHandler now that we don't use it

webapp/javascript/pages/IntroPages/SignIn/index.tsx
```diff
@@ -46,7 +47,9 @@ function SignInPage() {
 
       const res = await logIn({ username, password });
       if (res.isOk) {
-        dispatch(loadCurrentUser());
+        history.replace(
+          (location.state as any)?.redir || PAGES.CONTINOUS_SINGLE_VIEW
```
petethepig, line 51: this fixes https://github.com/pyroscope-io/pyroscope/issues/1037 right?

petethepig: Will it work properly with BaseURL? For example, if redir is null, this will redirect to `/`, not `BaseURL`, right?

webapp/javascript/index.jsx
```diff
@@ -52,27 +53,37 @@ function App() {
...
+              </Continuous>
+            </Protected>
           </Route>
           <Route path={PAGES.SERVICE_DISCOVERY}>
-            <ServiceDiscoveryApp />
+            <Protected>
+              <ServiceDiscoveryApp />
+            </Protected>
           </Route>
           {isAdhocUIEnabled && (
```
petethepig, line 88: This should also be under `<Protected>` I think.

petethepig (approved):
Looks good, added a couple of comments

Makefile
shaleynikov: yeah, I used this hack until I needed to supply more complicated params, so I just run it with
```
make SERVERPARAMS="--config=./server.yml"
```

webapp/javascript/pages/IntroPages/SignIn/index.tsx
shaleynikov: No, history shouldn't be aware. Wondering why tests are passing.

.github/workflows/cypress-tests-auth.yml
shaleynikov: Neither did I, found that in that exact marvelous documentation

PR #1050, opened 2022-04-21
fix(jfr): fixes a parser regression introduced in 1.15.0

eh-am:


examples/java-jfr/run.sh
```diff
@@ -0,0 +1,4 @@
+export PYROSCOPE_APPLICATION_NAME=fibonacci-java-lock-push
+export PYROSCOPE_PROFILER_EVENT=lock
+export PYROSCOPE_SERVER_ADDRESS=http://pyroscope:4040
```
petethepig, line 3: I'd move it to docker-compose, close to `PYROSCOPE_LOG_LEVEL` variable. Also would add format explicitly:
```
PYROSCOPE_FORMAT=jfr
```

It's in the [docs](https://pyroscope.io/docs/java/)

Issue #1049, opened 2022-04-21
Replace rgb(112, 219, 237) in palette for something darker


eh-am:
![image](https://user-images.githubusercontent.com/6951209/164463093-7ee2f0da-33e9-4da9-914f-8fe0d3ef3775.png)

The contrast is not very good
As you can see the contrast is not very good
https://github.com/pyroscope-io/pyroscope/blob/main/packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/colorPalette.ts#L34

Plus it's a good idea to have a look at every single color in the palette.

eh-am:
The problem I see here is that the same color is used in both the table and the flamegraph, but the text color is white in the former and black in the latter.

![image](https://user-images.githubusercontent.com/6951209/164470555-f9672508-8612-404d-8f83-5134db7b84db.png)

We can't just make the text black in both, since you may have cases like the following, where the background is not 100% width.
![image](https://user-images.githubusercontent.com/6951209/164470734-6d5e4dc0-5668-4244-9c6e-5d232d1c8348.png)

eh-am:
Reopning this since the current solution looks weird for diff view
![Screen Shot 2022-04-25 at 3 00 38 PM](https://user-images.githubusercontent.com/6951209/165522631-9791e13a-bcce-46d1-935d-81546d7ebf5b.png)

deepto98:
Can I pick this up?

PR #1051, opened 2022-04-21
fixed low contrast issue

Tejax-v2:
fixes: #1049


eh-am:
Awesome, thank you for your contribution!

The problem I see here is that the same color is used in both the table and the flamegraph, but the text color is white in the former and black in the latter.

image

We can't just make the text black in both, since you may have cases like the following, where the background is not 100% width.
image

PR #1055, opened 2022-04-22
fix(webapp): tag selector overflow

pavelpashkovsky:
FIX: https://github.com/pyroscope-io/pyroscope/issues/1043
DEMO:
image

2 important things

1. `<input>` was changed to `<textarea>`, I use `react-textarea-autosize` for correct resizing query textarea. This npm module was already injected to Pyroscope as submodule before.
2. I had to transform query font to monospace, this is the easiest and the most accurate approach to keep the same size of symbols inside of `<textarea>` and `<code>` and keep them matching each other.


pavelpashkovsky:
@eh-am I adjusted query input overflow, but solution isn't ideal, because at the same time this UI element is `input` and `code`. I'd appreciate any ideas on it

eh-am (commented):
Nice!

Just out of curiosity, did you try solving this with CSS only?
Somehow making the `<code>`element having the same width as the input box. 
Then we don't have to do the 
```jsx
style={{ width: `calc(${inputWidth}px - 0.75em)` }}
```
part, which seems a little bit finicky.

But yeah it doesn't seem too trivial, it probably requires creating a wrapper around both the input and the code part, so that they both share the same width, then making one of them `position: absolute`.

webapp/javascript/components/TagsBar.tsx
```diff
@@ -102,6 +110,33 @@ function QueryInput({ initialQuery, onSubmit }: QueryInputProps) {
     }
   }, [query]);
 
+  useEffect(() => {
```
eh-am, line 113: In a high level can you explain what's goin on here?

eh-am (approved):
Looks quite good for me! One thing I noticed is that ctrl+enter is not submitting the query.

webapp/javascript/components/TagsBar.tsx
```diff
@@ -102,6 +110,33 @@ function QueryInput({ initialQuery, onSubmit }: QueryInputProps) {
...
+    setTextAreaSize({
+      width: textareaRef?.current?.['offsetWidth'] || 0,
+      height: textareaRef?.current?.['offsetHeight'] || 0,
+    });
+  }, [query, windowWidth, onSubmit]);
+
+  const onChange = (e: any) => {
```
eh-am, line 120: Can we type this any? :)

webapp/javascript/components/TagsBar.tsx
pavelpashkovsky: We have to keep `<code>` width and height exactly as `<textarea>` width and height. This effect allows us to recalculate them after updating any of useEffect dependencies. We get width/height from `textareaRef?.current?.['offsetWidth' | 'offsetHeight']`, this is the most accurate way to get dimensions of `<textarea>`

webapp/javascript/components/TagsBar.tsx
pavelpashkovsky: done :)

pavelpashkovsky:
@eh-am 
> I noticed is that ctrl+enter is not submitting the query.

It works in the same way as at https://demo.pyroscope.io/. UI detects if user changed query string inside of input, if so, it allows user submit by ctrl+enter. Should I add possibility to submit all requests despite if they didn't change?

webapp/javascript/components/TagsBar.tsx
eh-am: Wait i still see it there

pavelpashkovsky: some kind of github caching issue
image

webapp/javascript/components/TagsBar.tsx
pavelpashkovsky: There are 2 approaches for changing textarea height: changing height directly <textarea style={{height: "100px"}}>, and changing amount of rows <textarea rows={3}>. In this case changing rows in more convenient than changing textarea height (at least it is less demanding). onChange consists logic of incrementing or decrementing amout of textarea's rows depending on amount of symbols

eh-am: I see! I'm just not a big fan of calling `e.target` directly as opposed to going through the whole react cycle (use a state, then that using that state to set the `rows/scrollTop` / or using `refs`). I don't have any strong arguments except "it looks too imperative/I don't see many docs using that approach". Any thoughts on that?

That being said, can you explain a similar explanation you gave above above the `const onChange` line? I come from the principle I (Eduardo) got slightly confused by that, then maybe others will also will.

pavelpashkovsky: My fault, your comment made me to go to the `react-textarea-autosize` src code to check what is going on under the hood, and it looks like we can get rid of these uncontrolled operations on Pyroscope side. Actions with `event` duplicated logic there. Sorry, I haven't checked it before I changed `<textarea>` to `TextareaAutosize`.

eh-am: No worries that's what reviews are for :P

PR #1059, opened 2022-04-25
chore: bump pyroscope-models to 0.0.7

eh-am:
I've noticed npm publish job is failing (eg https://github.com/pyroscope-io/pyroscope/runs/6159562890?check_suite_focus=true) not quite sure how the tag already exists though.


eh-am:
NVM just deleted the tags and recreated the job (not quite sure what happened there in the first place)

PR #1060, opened 2022-04-26
feat: nodejs push & pull mode

shaleynikov:


pkg/scrape/target.go
```diff
@@ -100,6 +102,15 @@ func (t *Target) offset(interval time.Duration) time.Duration {
 	return time.Duration(next)
 }
 
+func (t *Target) SpyName() string {
+	for _, l := range t.discoveredLabels {
+		if l.Name == model.SpyNameLabel && l.Value != "" {
+			return l.Value
+		}
+	}
+	return spyName
```
eh-am, line 111: Maybe we should signal somehow that it's defaulting to a deprecated spyName?

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/color.ts
```diff
@@ -91,6 +91,8 @@ function spyToRegex(spyName: string) {
       return /^(?<packageName>(.*\/)*)(?<filename>.*\.py+)(?<line_info>.*)$/;
     case 'rbspy':
       return /^(?<packageName>(.*\/)*)(?<filename>.*\.rb+)(?<line_info>.*)$/;
+    case 'nodespy':
```
eh-am, line 94: Let's add a test to color.spec.ts please

examples/nodejs/express-ts/profiler.js
```diff
@@ -0,0 +1,107 @@
+/* eslint-disable */
```
eh-am, line 1: I believe this file should go away since we have the library published?

examples/nodejs/express-ts/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "type": "module",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@pyroscope/nodejs": "github:pyroscope-io/pyroscope-nodejs",
```
eh-am, line 12: Should we point to a published version?

examples/nodejs/express-ts/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
...
+  pyroscope:
+    image: 'pyroscope/pyroscope:latest'
+    ports:
+      - '4040:4040'
+    command:
+      - 'server'
+  app:
+    build: .
+    cap_add:
+      - SYS_PTRACE
```
eh-am, line 13: Is this actually needed or just bad copy paste?

examples/nodejs/express-pull/server.yml
```diff
@@ -0,0 +1,12 @@
+---
+log-level: debug
+scrape-configs:
+  - job-name: testing
+    enabled-profiles: [cpu]
```
eh-am, line 5: Shall we enable memory as well?

eh-am (commented):
Looks good :)

pkg/scrape/target.go
shaleynikov: Yeah, it's definitely is a good idea. I peeked and it seems it's a bigger change that it might seem as we need to change `SpyName()` return type to something like `(string, error)` but it doesn't supported along the code yet. Let me create separate issue for that

shaleynikov: https://github.com/pyroscope-io/pyroscope/issues/1066

examples/nodejs/express-ts/docker-compose.yml
shaleynikov: ofc copy-paste )

eh-am (approved):
Looks good!

My only concern now is RE migration steps (partially covered by 
 https://github.com/pyroscope-io/pyroscope/issues/1066, plus docs).

We can talk about this later though.

Thank you!

PR #1069, opened 2022-04-28
Fix Standalone Flamegraph styles  [WIP]

pavelpashkovsky:


webapp/javascript/redux/reducers/ui.ts
```diff
@@ -36,11 +36,17 @@ type SidebarState =
 
 export interface UiState {
   sidebar: SidebarState;
+  colorMode: ColorMode;
+}
+
+export enum ColorMode {
+  Dark = 'dark',
+  Light = 'light',
 }
```
eh-am, line 45: Let's use union types instead, since it type checks better among other things.

eh-am (commented):
Not sure about current approach. We are effectively duplicating the CSS under `flamegraph.scss` and `profile.scss`, which will make keeping them in sync complicated. I think it's relative okay to duplicate stuff like the global `font` and whatnot, but duplicating `select` and `input` may be too much. For these I believe it's better to create a component like we need for `@ui/button`, then scope its css using css variables?

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/colorPalette.ts
```diff
@@ -56,3 +65,11 @@ export const ColorBlindPalette: FlamegraphPalette = {
   neutralColor: Color.rgb(148, 142, 142),
   badColor: Color.rgb(220, 50, 32),
 };
+
+export const ColorBlindLightPalette: FlamegraphPalette = {
```
eh-am, line 69: This is cool and stuff, let's not do this in THIS pr.

webapp/sass/flamegraph.scss
```diff
@@ -0,0 +1,232 @@
...
+    .table-item-button {
+      border: none;
+      display: flex;
+      flex-flow: row nowrap;
+      align-items: baseline;
+      width: 100%;
+      cursor: pointer;
+    }
+  }
+}
```
eh-am, line 174: This whole block here can be scoped using css modules to the FlamegraphTable no?

webapp/sass/flamegraph.scss
```diff
@@ -0,0 +1,232 @@
...
+      border-left-color: var(--ps-c-green);
+      border-right-color: var(--ps-c-green);
+      border-top-color: var(--ps-c-green);
+      border-bottom-color: var(--ps-c-green);
+    }
+  }
+
+  & + .viz-switch {
+    margin-left: 10px;
+  }
```
eh-am, line 231: Is this even used?

webapp/sass/variables.css
```diff
@@ -31,3 +35,41 @@
   --ps-c-intro-bg-1: #3dc1d3cc;
   --ps-c-intro-bg-2: #d1283980;
 }
+
+[data-theme='light'] {
+  --ps-c-black: #ffffff;
```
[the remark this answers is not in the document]
pavelpashkovsky: Yeah, at this moment I don't have any solution for this,  I just copied previous naming approach. This kind of naming It's not really suitable for current circumstances. We have to keep the same name of the color for `:root` and for `[data-theme='light']` sections but naming based on it's color, you provided good example. Any ideas?

pavelpashkovsky:
> Not sure about current approach. We are effectively duplicating the CSS under `flamegraph.scss` and `profile.scss`, which will make keeping them in sync complicated. I think it's relative okay to duplicate stuff like the global `font` and whatnot, but duplicating `select` and `input` may be too much. For these I believe it's better to create a component like we need for `@ui/button`, then scope its css using css modules*?

You are right, I moved common selectors to separated file `common.scss` and imported this file from `flamegraph.scss` and from `profile.scss`, so no duplication at this moment.

webapp/sass/variables.css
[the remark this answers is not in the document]
pavelpashkovsky: maybe something like 
`--input-border: red`
`--main-background: blue`
`--text-color: green`
?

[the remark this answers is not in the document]
eh-am: That's definitely better yes!

One thing I've seen previously is  the use of `{primary,secondary,tertiary,...} color`, eg `primary-color`. That way we can change their value freely

webapp/sass/common.scss
```diff
@@ -0,0 +1,198 @@
...
+input {
+  @include outline;
+  display: inline-block;
+  position: relative;
+  border-radius: 4px;
+  background-color: var(--ps-c-gray-light-2);
+  border: 1px solid var(--ps-c-gray-light-3);
+}
+
+.flamegraph-table {
```
eh-am, line 62: Where is this one used other than here? Can we scope that to the `FlamegraphTable` component using css modules?

PR #1071, opened 2022-04-28
Update Rust examples

omarabid:


examples/rust/rideshare/server/src/main.rs
```diff
@@ -49,21 +47,41 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
     });
 
+    let (add_tag, remove_tag) = agent_running.tag_wrapper();
+    let add = Arc::new(add_tag);
+    let remove = Arc::new(remove_tag);
+
     // Bike Route
-    let bike = warp::path("bike").map(|| {
+    let bike = warp::path("bike").map(move || {
+        add("vehicule".to_string(), "bike".to_string());
```
Rperry2174, line 56: can you change all instances of `vehicule` to the proper spelling of `vehicle` (unless there is a reason you're spelling it differently)

omarabid: Good catch! (it's the French spelling)

PR #1073, opened 2022-04-29
feat(flamegraph): User should be able to adjust title visibility over the Flamegraph

pavelpashkovsky:
FIX: https://github.com/pyroscope-io/pyroscope/issues/1072


pavelpashkovsky:
@eh-am, make sense, replaced with `showToolbar`

PR #1089, opened 2022-05-06
Feat: nodejs updates

shaleynikov:
Fixed:
* All examples are shipped with dockerfiles and docker-compose files
* ESM support added for `express-ts` example
* Updated api to match 0.2.x version of `@pyroscope/nodejs`
* Sample Types are now compatible with go


pkg/storage/tree/pprof.go
```diff
@@ -67,12 +67,14 @@ var DefaultSampleTypeMapping = map[string]*SampleTypeConfig{
 		Sampled:     true,
 	},
 	"objects": {
-		Units:      "objects",
```
petethepig, line 70: Messaged in slack, but basically, I'm trying to figure out if we need `objects` and `space` at all. It seems like these are very similar to `inuse_objects` and `inuse_space`

shaleynikov: Makes sense. 

On latest `@pyroscope/nodejs` I adjust profile to meet golang sample types name

https://github.com/pyroscope-io/pyroscope-nodejs/pull/12

PR #1096, opened 2022-05-10
feat: add support for labels when used with JFR and async-profiler's contextId

korniltsev:
Right now tags are sent as protobuf along with JFR as a separate multipart form field.
All events are grouped by contextId to minimize `storage.Putter.Put` operations
Other logic should be the same as previously


pkg/convert/jfr/parser.go
```diff
@@ -14,155 +14,224 @@ import (
...
-			if fs := frames(tp.StackTrace); fs != nil {
-				lockSamples.InsertStackString(fs, 1)
-				lockDuration.InsertStackString(fs, uint64(tp.Duration))
-			}
 		case *parser.ActiveSetting:
 			as := e.(*parser.ActiveSetting)
 			switch as.Name {
 			case "event":
-				event = as.Value
+				event = as.Value // this can be passed from the agent as upload param
```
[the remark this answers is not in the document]
petethepig: Also, this will only work if there's one type per chunk. Can we guarantee that that's always true? (I don't know too much about JFR internals)

I think if we can guarantee it, then this code is totally fine, we just need to remove this comment.

pkg/convert/jfr/parser.go
```diff
@@ -14,155 +14,224 @@ import (
...
 	for _, c := range chunks {
-		if pErr := parse(ctx, c, s, pi); pErr != nil {
+		if pErr := parse(ctx, c, s, pi, tags); pErr != nil {
 			err = multierror.Append(err, pErr)
 		}
 	}
 	pi.Val = nil
 	return err
 }
+func resolveTags(contextId int64, tags *Tags) map[string]string {
```
petethepig, line 35: nit: needs a new line before the function

pkg/convert/jfr/parser.go
```diff
@@ -14,155 +14,224 @@ import (
 	"github.com/pyroscope-io/pyroscope/pkg/storage/tree"
 )
 
-func ParseJFR(ctx context.Context, r io.Reader, s storage.Putter, pi *storage.PutInput) (err error) {
+type Tags struct {
```
petethepig, line 17: I know this is annoying, and we're not doing the best job of following this convention at the moment, but we're trying to switch from calling these `tags` to calling these `labels` as this is the name most other profilers adopted. Could you rename everything called `tags` to `labels`?

pkg/convert/jfr/parser.go
```diff
@@ -14,155 +14,224 @@ import (
...
+}
+
+func groupEventsByContextId(events []parser.Parseable) map[int64][]parser.Parseable {
+	res := make(map[int64][]parser.Parseable)
+	for _, e := range events {
+		switch e.(type) {
+		case *parser.ExecutionSample:
+			es := e.(*parser.ExecutionSample)
+			contextId := es.ContextId
+			res[contextId] = append(res[contextId], e)
```
petethepig, line 215: nit: these can just be `res[es.ContextId`] = ...` (no need for an extra variable imo)

I don't have a strong opinion on this though.

petethepig (changes requested):
It looks good overall. There are a few go nits, but I think performance-wise and code-wise it all looks good.

A couple of things we need to change before merging this:
* parse http form in a more reliable manner (see comments above)
* add some tests [here](https://github.com/pyroscope-io/pyroscope/blob/946468dbf42ff4450edc94762812ddb4a5f3482d/pkg/server/ingest_test.go#L238-L270) — i think a simple new test ensuring that multiple tags get created when you upload a jfr with tags should be good enough.
* fix linting errors
* address other comments above
* rename tags to labels

pkg/convert/jfr/parser.go
korniltsev: inlined

pkg/convert/jfr/parser.go
korniltsev: renamed

pkg/parser/parser.go
```diff
@@ -124,6 +127,33 @@ func (p *Parser) Put(ctx context.Context, in *PutInput) (err error, pErr error)
 	return err, pErr
 }
 
+func parseJFRMultipart(ctx context.Context, in *PutInput, err error, p *Parser, pi *storage.PutInput) error {
+	reader := multipart.NewReader(in.Body, in.MultipartBoundary)
+	contextsPart, err := reader.NextPart()
```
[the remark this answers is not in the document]
korniltsev: I rewrote with the same approach as pprof - read whole request into memory up to 32M.

I think I can parse  `jfr` requests without reading whole reqeust in memory 
1. iterate over multipart fields, if found labels - parse labels, if found jfr - parse jfr
2. link labels to jfr & put into storage.

@petethepig let me know if I should do it or it is ok to keep as it is now.

pkg/convert/jfr/parser.go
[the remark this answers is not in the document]
korniltsev: I removed the comment and left the things as they were before the PR

korniltsev:
Oops, got test timeout on CI. 
Unexpectedly,  a JFR file produced on my machine takes 600ms to parse on my machine and >2s on CI. 
It should be much less than 600ms
I will investigate reasons.

KorniltsevCryptopay239:
Yey, CI timeout is gone after jfr-parser update

petethepig (approved):

PR #1097, opened 2022-05-11
Fix: isolating styles for standalone flamegraph

pavelpashkovsky:


webapp/sass/variables.css
```diff
@@ -9,25 +9,75 @@
...
+  --ps-c-red-danger: #23caba;
+  --ps-c-red-light: #23caba;
+  --ps-c-orange-light-1: #17b2c3;
+  --ps-c-orange-light-2: #0392d9;
+  --ps-c-orange-dark-2: #03a3f2;
+  --ps-c-orange-dark-1: #1ac7da;
+  --ps-c-orange-dark-3: #2074ac;
+  --ps-c-intro-bg-1: #c23e2ccc;
+  --ps-c-intro-bg-2: #2ed7c680;
+}
```
eh-am, line 83: not needed in this pr

.storybook/preview.js
```diff
@@ -1,4 +1,4 @@
-import '../webapp/sass/profile.scss';
+// import '../webapp/sass/profile.scss';
```
eh-am, line 1: this is good for testing, but this will mess up all other storybook pages (since they assume some global css)

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/index.tsx
```diff
@@ -36,6 +36,7 @@ interface FlamegraphProps {
   palette: FlamegraphPalette;
   setPalette: (p: FlamegraphPalette) => void;
   toolbarVisible?: boolean;
+  colorMode?: 'light' | 'dark';
```
eh-am, line 39: not needed in this pr

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/index.tsx
```diff
@@ -244,6 +245,7 @@ export default function FlameGraphComponent(props: FlamegraphProps) {
         palette={palette}
         setPalette={setPalette}
         toolbarVisible={toolbarVisible}
+        colorMode={colorMode}
```
eh-am, line 248: not needed in this pr

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphRenderer.tsx
```diff
@@ -76,6 +76,7 @@ interface FlamegraphRendererProps {
   showPyroscopeLogo?: boolean;
   renderLogo?: boolean;
   ExportData?: React.ComponentProps<typeof Graph>['ExportData'];
+  colorMode?: 'light' | 'dark';
```
eh-am, line 79: not need in this pr

stories/Flamegraph.stories.tsx
```diff
@@ -36,6 +37,21 @@ const SimpleTree = {
...
+    colorMode={args['Light Color Mode'] ? 'light' : 'dark'}
+  />
+);
+
+export const ColorMode = Template.bind({});
+
+ColorMode.args = {
+  ['Light Color Mode']: false,
+};
+
```
eh-am, line 54: not needed in this pr

eh-am (commented):
since flamegraph.scss imports `common.scss`, which contains global css (like the input here https://github.com/pyroscope-io/pyroscope/pull/1097/files#diff-762a0efb83b959f3d95fb93aa6ad0957ea6f193e39b4707ec9832d54eefdd495R53-R60)

what happens if the user's input?
eg i have my own application with `<input type="text">`, with my own styling, then I import the flamegraph

is the final input styling affected by?

pavelpashkovsky:
> since flamegraph.scss imports `common.scss`, which contains global css (like the input here https://github.com/pyroscope-io/pyroscope/pull/1097/files#diff-762a0efb83b959f3d95fb93aa6ad0957ea6f193e39b4707ec9832d54eefdd495R53-R60)
> 
> what happens if the user's input? eg i have my own application with `<input type="text">`, with my own styling, then I import the flamegraph
> 
> is the final input styling affected by?

I assigned `.flamegraph-root` to the root div of Flamegraph, then I imported common styles under this selector
```
.flamegraph-root {
  @import 'common';
 
 //...
}
```
in this case inner styles would be assigned as child styles (take a loot at this screenshot), IMO no need to have any additional isolation
image

webapp/sass/flamegraph.scss
```diff
@@ -0,0 +1,18 @@
+@use 'sass:math';
+@use 'variables' as *;
+@use 'mixins/outline' as *;
+
+@import '~sanitize.css';
+@import '~sanitize.css/forms.css';
+@import '~sanitize.css/typography.css';
```
eh-am, line 7: Won't these affect the user's global CSS as well?

pavelpashkovsky: Theoretically it might affect user's global styles, but "cascading level" of these styles would be much lower than user's global styles, so I'm not sure that it can broke/change something.

>sanitize.css is a CSS library that provides consistent, cross-browser default styling of HTML elements alongside useful defaults. [[from here](https://csstools.github.io/sanitize.css/)]

eh-am: >but "cascading level" of these styles would be much lower than user's global styles,
I don't get this part. What if the user uses this other library https://github.com/resetercss/reseter.css for example, wouldn't our css rules (coming from sanitize.css) conflict with theirs?

eh-am: Here's an example

i modified `stories/Flamegraph.stories.tsx`, imagine this is the end user code
to have an `input` with red background (don't ask)
image
image

image
then i imported the flamegraph.css
image

Now my input is not red anymore!
image

eh-am:
Looks quite good!

It works with grafana \o/
image

One thing I noticed is that the background color is different from what it usually looks like
image
image

pavelpashkovsky:
@eh-am, hmmm, I don't see the difference on my display. Thanks for your help. There is a difference in background because I set background color for `.flamegraph-root`, I left a comment on it, but then deleted this comment by mistake. Are you sure that setting some bg color on `.flamegraph-root` is a good solution, because if you inject Flamegraph into app with light color mode, it looks weird?

pavelpashkovsky:
and is this PR good enough to merge or requires any improvements?

Issue #939, opened 2022-03-12
fix request cancellations


eh-am:
After https://github.com/pyroscope-io/pyroscope/pull/927 we ended up getting rid of Request Canceling, we have to get it back:

1. Update the functions in the `services` to take an `AbortController` (Like this https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/services/appNames.ts#L9, but making it mandatory)
2. Update the data fetching call to use (eg https://github.com/pyroscope-io/pyroscope/pull/927/files#diff-d6dda51d3d4ec84407a5925f924d35e47bb367777d3ff5a928bee70ccb4d5c07R366), just follow the compiler after step 1
3. Update the components (where `dispatch` is called) to use the promise result, and call `.abort` in the callback (example https://redux-toolkit.js.org/api/createAsyncThunk#canceling-while-running)

We have to do this for most (all) requests.

shaleynikov:
Here is a proof-of-concept on request cancellation feature, please review #1099

PR #1099, opened 2022-05-11
feat: poc on request cancellation

shaleynikov:
Here a request cancellation proof-of-concept, based on #939 

Demo is here: https://www.loom.com/share/6b1d6692aa5c4997bfef7793aa903ac3

Since it's hard to catch cancellation effect on my desktop, I throttled connection to `Slow 3g` so hitting reload button makes it cancel previous requests.

Requests are actually get cancelled, but if we want to prevent server to stop waisting resources processing unneeded requests, we can debounce (or throttle) all the controls we have.

If it looks okay, I'll finish the rest of remaining services.


shaleynikov:
@eh-am LMK what do you think

webapp/javascript/redux/reducers/settings.ts
```diff
@@ -34,15 +35,29 @@ type ApiKeysState = {
...
+
     if (res.isOk) {
       return Promise.resolve(res.value);
     }
 
+    if (res.isErr && res.error instanceof RequestIncompleteError) {
+      return Promise.reject(res);
+    }
+
+    console.log(res);
```
eh-am: One problem with this is that the return type changed to be either `Promise<typeof res.error | typeof res | typeof res.value>`. From my understanding we can still `Promise.reject(res.error)` here right?

Another thing is I think we can make this more clear this by checking for `e.name == 'AbortError'` here https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/services/base.ts#L97, then we know exactly it's an `AbortError` (request has ben aborted) and not an `RequestIncompleteError` (eg user is without network)

webapp/javascript/redux/reducers/settings.ts
```diff
@@ -34,15 +35,29 @@ type ApiKeysState = {
   data?: APIKeys;
 };
 const apiKeysInitialState: ApiKeysState = { type: 'pristine', data: undefined };
+let reloadApiKeysAbortController: AbortController | undefined;
```
eh-am, line 37: The problem with having a single `AbortController` per action is that some actions are called twice (eg comparison view)

image

Only thing that I can think now is split those up (making 2 different actions), or add some custom logic per action (storing 2 abort controlelrs for example)

webapp/javascript/redux/reducers/settings.ts
```diff
@@ -34,15 +35,29 @@ type ApiKeysState = {
...
 export const reloadApiKeys = createAsyncThunk(
   'newRoot/reloadAPIKeys',
   async (_, thunkAPI) => {
-    const res = await fetchAPIKeys();
+    if (reloadApiKeysAbortController) {
+      reloadApiKeysAbortController.abort();
+    }
+    reloadApiKeysAbortController = new AbortController();
+
+    thunkAPI.signal = reloadApiKeysAbortController.signal;
```
eh-am, line 47: Do we need this? The `thunkAPI` already has a AbortSignal (via `.signal`), can't we just make it so that the component `useEffect` cleanup will call `.abort`? That way we don't have to deal with this at the redux level.
Now that I am thinking it looks like we need this for the exact case you showed: clicking on a button multiple times. Nevermind!

Ugh this would be so much easier to solve with observables!

webapp/javascript/redux/reducers/settings.ts
shaleynikov: yep, noted

webapp/javascript/components/Settings/APIKeys/index.tsx
```diff
@@ -22,7 +22,10 @@ const ApiKeys = () => {
   const history = useHistory();
 
   useEffect(() => {
-    dispatch(reloadApiKeys());
+    const reload = dispatch(reloadApiKeys());
+    return () => {
+      reload.abort();
+    };
```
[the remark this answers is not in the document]
shaleynikov: yeah, that looks better

webapp/javascript/redux/reducers/settings.ts
shaleynikov: > Now that I am thinking it looks like we need this for the exact case you showed: clicking on a button multiple times. Nevermind!
Yes, that was done to cancel both `fetch` and thunk

webapp/javascript/redux/reducers/settings.ts
shaleynikov: let's see when we get to it

Rperry2174:
closing in favor of #1143 cc @pavelpashkovsky in case there's anything you can use from here

PR #1103, opened 2022-05-13
feat: Color mode

pavelpashkovsky:
Web app and Flamegraph color mode. based on https://github.com/pyroscope-io/pyroscope/pull/1097 and (https://github.com/pyroscope-io/pyroscope/pull/1069 closed)


eh-am:
I think i am missing some context here, I am running the storybook and it looks like this

https://user-images.githubusercontent.com/6951209/170535019-6b1c717b-f626-42e1-98f7-63857c72db73.mp4

Am I missing something?

Rperry2174:
fwiw @eh-am in the actual ui it doesn't look the same: 
image

pavelpashkovsky:
sorry for confusion guys, but let me explain. Make sure you pulled all latest changes from the PR, sometimes browser caches styles from previous branch

story of Flamegraph looks weird because I removed background color from it. When user injects Flamegraph to web app, we don't know which background-color/theme the app has. If the Flamegraph has built-in background, it can broke general styling of the parent app. User can set background for Flamegraph which he likes. [I added parent app background emulation for the Flamegraph story to make it look better (video)](https://monosnap.com/file/PPlceqkLGGSEZRAYj1YIrYctdK6HM2)

Rperry2174:
Nice work thanks @pavelpashkovsky !!!

PR #1104, opened 2022-05-13
chore: decouple storage queue

kolesnikovae:
DO NOT MERGE: WIP

We used to call `storage.Put` directly in self-profiling and `/ingest` endpoint for all formats but pprof; scrape manager and pprof parser use the internal storage queue instead.

The change unifies ingestion methods: `Enqueue` method is removed and all `storage.Putter` consumers use `storage.IngestionQueue` regardless of the ingestion method. ~It's important to note that `storage.IngestionQueue` does not discard ingestion request when the queue is full, instead it blocks caller until context cancellation/successful enqueue/server stop - whichever occurs first.~

This allows to inject another processing layer - a proxy/relay that can handle storage input before it gets enqueued or written to the storage: basically, now we can build a processing pipeline from `storage.Putter`s.


pkg/server/ingest.go
```diff
@@ -54,21 +55,15 @@ func (h ingestHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-	// this method returns two errors to distinguish between parsing and ingestion errors
-	// TODO(petethepig): maybe there's a more idiomatic way to do this?
-	err, ingestErr := h.parser.Put(r.Context(), pi)
-
-	if err != nil {
-		h.httpUtils.WriteError(r, w, http.StatusUnprocessableEntity, err, "error happened while parsing request body")
-		return
-	}
-
-	if ingestErr != nil {
```
petethepig, line 66: thank you for refactoring this

petethepig (approved):
Slightly concerned about the fact that the queue blocks now (talked about it on Slack), but outside of that it looks good!

kolesnikovae:
> Slightly concerned about the fact that the queue blocks now (talked about it on Slack), but outside of that it looks good!

Reverted. Let's revisit this place later.

petethepig (approved):
LGTM, I think we can merge this now

PR #1107, opened 2022-05-19
feat: UTC timezone [WIP]

pavelpashkovsky:


eh-am:
Haven't looked at the code yet but here are my thoughts:

- would be useful to be able to pick another option other than utc/local, for example even though i work in brazil ( - 3) we often talk in pst (-7 i think?)
- is this setting global? Eg all dates within the app would be following the setting, and if it's global isn't it weird for this setting to be under the timepicker component?
- is this setting persisted? Browser only (local storage)? 
- this is more of a reflection, when sharing profiles, do people want to send them in a given format (eg utc)  or not?

pavelpashkovsky:
@eh-am 
>is this setting global? Eg all dates within the app would be following the setting, and if it's global...

This setting is global for all pages of Pyroscope web app. It's important to understand that I updated only the "front-end" of interacting with time. This PR doesn't affect functions which set/calculate time, it's only for fuctions which "render" time (time selectors/ timelines etc.). Pyroscope already stores and works with time in unix format what equals to UTC.

>...isn't it weird for this setting to be under the timepicker component?

You are right, it's just for POC. Can you describe your vision of this selector?

>is this setting persisted? Browser only (local storage)?

yes it is, I keep this param in redux state

>this is more of a reflection, when sharing profiles, do people want to send them in a given format (eg utc) or not?

no, sharing reports logic stays the same. recepient will see this report in timezone which he set on his browser

eh-am (commented):
I'm slightly confused by why we need to maintain a `isUTC`, isn't utc == gmt and therefore, offset 0?

Couldn't some of the code be simplified by simply using an offset? Then if we ever implement support for other timezones it's just a matter of taking different offsets? [0]

[0] Althought I still feel like offsets are not the best abstraction, since offset change in eg daylight savings.

pavelpashkovsky:
@eh-am agree with you. In fact we operate time offset, not UTC/local time, I refactored my solution in a bit in accordance with your ideas.

>...Then if we ever implement support for other timezones...

But in case you want to add support for all timezones in future, there might be difficulties with refactor timelines, because Flot timeline supports only 2 options - utc and local time (might need serious refactor)

Rperry2174:
Thanks @pavelpashkovsky !! Looks great :)

PR #1110, opened 2022-05-24
fix: application name without profile type

kolesnikovae:


petethepig (approved):
Good catch, thanks for fixing this!

PR #1112, opened 2022-05-26
fix: jfr ingestion issue in 0.17.0

petethepig:
The issue is that jfr parser was built slightly differently compared to other parsers. It reused the same PutInput structure for multiple inserts within one request.

Moreover, pi.Val was set to nil at the after all writes were finished. 

This flaw was exposed after we merged #1104 because PutInput structs were now stored a little bit longer in memory and inserted asynchronously.

I refactored the jfr parser code to create new PutInput structs for each input, I also DRYed the code up a bit.

---

The tests didn't catch this because in tests the controller was initialized without a queue. This was my mistake, for some reason I thought tests were using the queue as well, and that's why I approved #1104. I updated tests to use a queue.

### DIFF
--- a/pkg/convert/jfr/parser.go
+++ b/pkg/convert/jfr/parser.go
@@ -14,8 +14,8 @@ import (
 	"github.com/pyroscope-io/pyroscope/pkg/storage/tree"
 )
 
-func ParseJFR(ctx context.Context, r io.Reader, s storage.Putter, pi *storage.PutInput) (err error) {
-	chunks, err := parser.Parse(r)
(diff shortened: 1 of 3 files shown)


pkg/convert/jfr/parser.go
```diff
@@ -24,11 +24,10 @@ func ParseJFR(ctx context.Context, r io.Reader, s storage.Putter, pi *storage.Pu
 			err = multierror.Append(err, pErr)
 		}
 	}
-	pi.Val = nil
```
petethepig, line 27: This is the reason we had a panic

PR #1114, opened 2022-05-26
fix: flamegraph palette selector checkmark styles

pavelpashkovsky:
Edited by Eduardo:

before it was like this
image (43)


packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/DiffLegendPaletteDropdown.module.css
```diff
@@ -31,6 +31,7 @@
 
 .dropdownItem svg {
   margin-left: 1em;
+  fill: white;
```
Rperry2174, line 34: didn't we move all colors to a single file ? should this be a variable instead to support light / dark mode?

eh-am (approved):

PR #1122, opened 2022-06-01
feat: remote write

eh-am:
Allow writing to a remote storage (eg any server that implements the `/ingest` endpoint, including another pyroscope instance).
This allows one to run pyroscope on the edge (in a k8s cluster using pull mode) and then push that data to a more central pyroscope instance.

API is still not fully ready, but will merge for now.


pkg/remotewrite/client.go
```diff
@@ -0,0 +1,91 @@
...
+	if !(res.StatusCode >= 200 && res.StatusCode < 300) {
+		return ErrNotOkResponse
+	}
+
+	return nil
+}
+
+func (r *Client) putInputToRequest(pi *parser.PutInput) (*http.Request, error) {
+	// TODO(eh-am): copy put.Profile?
+	req, err := http.NewRequest("POST", r.config.Address, pi.Profile)
```
kolesnikovae, line 65: I've added `pi.Clone()` method - please use it instead of copying manually: it's behaviour is pretty tricky (I don't really like this but didn't find a trivial way to eliminate double parsing in pull mode).

pkg/remotewrite/client.go
```diff
@@ -0,0 +1,129 @@
...
+		return multierror.Append(err, ErrMakingRequest)
+	}
+
+	if !(res.StatusCode >= 200 && res.StatusCode < 300) {
+		// read all the response body
+		respBody, _ := ioutil.ReadAll(res.Body)
+		return multierror.Append(ErrNotOkResponse, fmt.Errorf("status code: '%d'. body: '%s'", res.StatusCode, respBody))
+	}
+
+	return nil
```
kolesnikovae, line 69: I think we need to close response body to allow Transport to reuse the connection:
```
// The http Client and Transport guarantee that Body is always
// non-nil, even on responses without a body or responses with
// a zero-length body. It is the caller's responsibility to
// close Body. The default HTTP client's Transport may not
// reuse HTTP/1.x "keep-alive" TCP connections if the Body is
// not read to completion and closed.
```
https://cs.opensource.google/go/go/+/go1.18.3:src/net/http/response.go;l=35

pkg/selfprofiling/upstream.go
```diff
@@ -0,0 +1,79 @@
...
+		return
+	}
+
+	profile := &pprof.RawProfile{
+		Profile:          j.Profile,
+		SampleTypeConfig: tree.DefaultSampleTypeMapping,
+	}
+	if len(j.PrevProfile) > 0 {
+		profile.PreviousProfile = j.PrevProfile
+	}
```
kolesnikovae, line 64: This makes pyroscope server to parse its own heap profiles twice :/

kolesnikovae (approved):

petethepig (approved):
Good to merge

PR #1125, opened 2022-06-06
fix: infinite loop when no apps are available

petethepig:
This fixes an issue where if you don't have any apps it will have an infinite loop trying to get app names.

To reproduce, run  something like:
```
rm -rf tmp/pyroscope-storage
PYROSCOPE_NO_SELF_PROFILING=true make dev
```

I think this PR fixes it.  @eh-am I'm not sure if this is the right fix, please review.


eh-am:
@petethepig I've split the loading / auto selecting an app effects. Now if you don't have an app, then hit the refresh app button, it will auto select the first one.

eh-am (approved):

PR #1131, opened 2022-06-07
docs: update java rideshare example to use artifact from maven central

korniltsev:


petethepig:
@korniltsev now that I think about it... I think we should do the same with other examples

PR #1136, opened 2022-06-08
refactor(webapp): Text Input as standalone component

pavelpashkovsky:
REF: https://github.com/pyroscope-io/pyroscope/issues/1126, https://github.com/pyroscope-io/pyroscope/pull/1127

All `<input/>` usages were updated to use `@ui/Input`, including it's styles

### DIFF
--- a/webapp/javascript/components/Settings/Security/ChangePasswordForm.tsx
+++ b/webapp/javascript/components/Settings/Security/ChangePasswordForm.tsx
@@ -66,6 +66,7 @@ function ChangePasswordForm(props: ShamefulAny) {
             name="oldPassword"
             required
             onChange={handleChange}
+            value={form.oldPassword}
           />
           <InputField
             label="New password"
@@ -74,6 +75,7 @@ function ChangePasswordForm(props: ShamefulAny) {
             name="password"
             required
             onChange={handleChange}
+            value={form.password}
           />
           <InputField
             label="Confirm new password"
@@ -82,6 +84,7 @@ function ChangePasswordForm(props: ShamefulAny) {
             name="passwordAgain"
             required
             onChange={handleChange}
+            value={form.passwordAgain}
           />
           <Button type="submit" kind="secondary">
             Save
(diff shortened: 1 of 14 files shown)


pavelpashkovsky:
@pavelpashkovsky check `NameSelector.module.scss` if https://github.com/pyroscope-io/pyroscope/pull/1120 merged first

webapp/javascript/components/Settings/Security/ChangePasswordForm.tsx
```diff
@@ -66,6 +66,7 @@ function ChangePasswordForm(props: ShamefulAny) {
             name="oldPassword"
             required
             onChange={handleChange}
+            value={form.oldPassword}
```
Rperry2174, line 69: how was this working without these before? What different functionality does adding them now do?

webapp/javascript/ui/Input.tsx
```diff
@@ -1,41 +1,52 @@
...
-      name={name}
-      placeholder={placeholder}
-      minLength={minLength}
-      debounceTimeout={debounceTimeout}
-      onChange={onChange}
-      value={value}
-    />
-  );
-}
+const Input = React.forwardRef(
```
Rperry2174, line 18: Why change this to `React.forwardRef`

webapp/javascript/components/Settings/Security/ChangePasswordForm.tsx
pavelpashkovsky: technically it's possible, text input can render it's value without passing value to it, I prefer passing value to all inputs

`<input type="text" /> == <input type="text" value={value}>`

webapp/javascript/ui/Input.tsx
pavelpashkovsky: Here is the answer: https://stackoverflow.com/questions/58578570/value-of-using-react-forwardref-vs-custom-ref-prop
I used this approach because there are inputs with forwarded ref as `<input ref={ref}>` directly. But if you use input as functional standalone component `<Input ref={ref} />` doesn't work (`<Input customRef={ref} />` works). But, no problem, if you don't like this approach I can pass refs as <Input inputRef={ref} />, just let me know

Rperry2174 (approved):
Looks 🔥 ! Thanks @pavelpashkovsky

PR #1142, opened 2022-06-09
ruby/python examples without SYS_PTRACE

omarabid:
custom pip/gem builds that can run without SYS_PTRACE

This is a testing/showcase PR, do not merge.


petethepig:
@omarabid We can close this one, right?

PR #1143, opened 2022-06-09
Request cancellation for profile data [WIP]

pavelpashkovsky:


eh-am (commented):
This looks solid!

The only concern I have is that in thunk when an action is requested again (because a new one was instantiated), a `rejected` action is sent, which in the reducer somebody could interpret that as a failure (right now we don't do anything in the reducer, but it could happen in the future).

I mean, problem is that at the reducer level we don't distinguish what's a valid error vs what's simply a rejection.

I don't have a better suggestion right now. So it should be fine :P

eh-am:
Another thing is that we should cancel requests when the component is unmounting (eg changing routes).

webapp/javascript/redux/reducers/continuous.ts
```diff
@@ -646,21 +759,39 @@ export const selectAppNames = (state: RootState) =>
 export const selectComparisonState = (state: RootState) =>
   state.continuous.comparisonView;
 
-export const selectIsLoadingData = (state: RootState) => {
+export const selectIsLoadingData = (state: RootState, pathname?: PAGES) => {
```
eh-am, line 762: Hmm, I don't quite get why this logic is necessary, can you explain bit more?

Here's what I am thinking:
Let's say I am in the singleView page, if cancellation works properly, nothing should be loading from the other pages. Therefore there's no make this distinction?

eh-am (commented):
Gonna test more thoroughly, but looks good!

One thing I was wondering is that there's no way from store perspective to distinguish some data has simply failed to request vs it was cancelled. Don't know if that's necessary also.

Just thinking out loud here:
Imagine we decide to show a X instead of a loading spinner somewhere to indicate data has FAILED to load. Then we cancel the request:
a) by changing route, then the component is unmounted and we shouldn't see the X. Ok good!
b) by making a new request, then the value in the store should be `reloading` almost immediately after the rejected, so it shouldn't show a X.

Alright I think we are good :P

webapp/javascript/redux/reducers/continuous.ts
pavelpashkovsky: >if cancellation works properly, nothing should be loading from the other pages

you are right, but image this case:
you are on Single View, you requests some data for this page and then go away to another page. At this moment "singleView data" thunk cancels with type `reloading` or `loading` in `state.continuous.singleView.type` . I wanted to reject this thunk with `failed` status, but if you go back to Single View page, you won't see any data until new request fulfilled. I've done this logic in this way just for user convenience.
image

 This explanation might be confusing, you can just take a look at `state.continuous.singleView.type` when you quit SingleView page and source code of SingleView page (there is logic based on loading/reloading/loaded status
image


but in case when data failed/cancelled this page renders nothing. I think it's reasonable, so I didn't put `failed` status there

eh-am (approved):

PR #1148, opened 2022-06-09
chore: Add javaspy coloring for flamegraphs

Rperry2174:
Adds coloring for java:
https://flamegraph.com/share/8022667e-e6d7-11ec-925c-fab0c2c9d0a0

Tested with:
https://regex101.com/r/W5U0uD/2

image

cc @korniltsev @petethepig

### DIFF
--- a/packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/color.ts
+++ b/packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/color.ts
@@ -94,7 +94,7 @@ function spyToRegex(spyName: string) {
     case 'nodespy':
       return /^(\.\/node_modules\/)?(?<packageName>[^/]*)(?<filename>.*\.?(jsx?|tsx?)?):(?<functionName>.*):(?<line_info>.*)$/;
     case 'javaspy':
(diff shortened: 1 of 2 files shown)


korniltsev:
For  `org/apache/catalina/core/ApplicationFilterChain.doFilter` package is `org/apache/catalina/core/`, class/file name is `ApplicationFilterChain` and method name is `doFilter`

Packages usually start with reversed domain name (`org/apache`) but technically it can be 1 depth package like 
`foo/Bar.baz` or  even 0 - no package `Bar.baz`

Maybe something like this `^(?<packageName>.+\/)?(?<filneame>.+\.)(?<functionName>.+)$` ?

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/color.ts
```diff
@@ -93,6 +93,8 @@ function spyToRegex(spyName: string) {
       return /^(?<packageName>(.*\/)*)(?<filename>.*\.rb+)(?<line_info>.*)$/;
     case 'nodespy':
       return /^(\.\/node_modules\/)?(?<packageName>[^/]*)(?<filename>.*\.?(jsx?|tsx?)?):(?<functionName>.*):(?<line_info>.*)$/;
+    case 'javaspy':
```
Rperry2174, line 96: @petethepig is this correct (and does the client send this already or do we need to make a change somewhere)

PR #1149, opened 2022-06-10
fix: Merge jvm generated classes in jfr at ingestion time

korniltsev:
https://github.com/pyroscope-io/pyroscope/issues/1129
- `jdk/internal/reflect/GeneratedMethodAccessor31` => `jdk/internal/reflect/GeneratedMethodAccessor_`
- `org/example/rideshare/OrderService$$Lambda$669.0x0000000800fd7318` => `org/example/rideshare/OrderService$$Lambda$_`


korniltsev:
I wrote a tiny bench on 3M a jar file
```go
func Benchmark(b *testing.B) {
	f, _ := ioutil.ReadFile("/Users/kyprah/pyroscope/java/last.jfr")
	pi := &storage.PutInput{
		Key: segment.NewKey(map[string]string{"foo": "bar"}),
	}
	for i := 0; i < b.N; i++ {
		ParseJFR(context.TODO(),
			&mockPutter{},
			bytes.NewBuffer(f),
			pi,
			&LabelsSnapshot{},
		)
	}
}
```
- Without regexes
`Benchmark-8   	       2	 896356370 ns/op`
- With regexes on each frame of each stacktrace
`Benchmark-8   	       1	2261908016 ns/op`
- With regexes in symbol processor 
`Benchmark-8   	       2	 890091210 ns/op`

I did not push the bench to the PR branch because I don't want to commit a 3M file to git repo

Performance wise we can merge it.

One thing left to decide is whether we need a flag to enable/disable this merging behaviour or not.

pkg/convert/jfr/parser.go
```diff
@@ -3,19 +3,21 @@ package jfr
...
 	"github.com/pyroscope-io/pyroscope/pkg/storage/metadata"
+	"io"
+	"regexp"
 
 	"github.com/pyroscope-io/pyroscope/pkg/storage"
 	"github.com/pyroscope-io/pyroscope/pkg/storage/segment"
 	"github.com/pyroscope-io/pyroscope/pkg/storage/tree"
 )
 
 func ParseJFR(ctx context.Context, s storage.Putter, body io.Reader, pi *storage.PutInput, jfrLabels *LabelsSnapshot) (err error) {
```
korniltsev, line 17: Do we need a flag to enable/disable merging behaviour?

pkg/convert/jfr/parser.go
```diff
@@ -203,3 +205,24 @@ func frames(st *parser.StackTrace) []string {
...
+
+// jdk/internal/reflect/GeneratedMethodAccessor31
+var generatedMethodAccessor = regexp.MustCompile("^(jdk/internal/reflect/GeneratedMethodAccessor)(\\d+)$")
+
+// org/example/rideshare/OrderService$$Lambda$669.0x0000000800fd7318.run
+var lambdaGeneratedEnclosingClass = regexp.MustCompile("^(.+\\$\\$Lambda\\$)\\d+[./](0x[\\da-f]+|\\d+)$")
+
+func mergeJVMGeneratedClasses(frame string) string {
+	frame = generatedMethodAccessor.ReplaceAllString(frame, "${1}_")
+	frame = lambdaGeneratedEnclosingClass.ReplaceAllString(frame, "${1}_")
```
korniltsev, line 217: I appended `_` to the merged frame name, I want to put there something more meaningful, like `_merged` or `_$$merged` or `_$$redacted` `_$$pyroscope_merged` 
What do you think?

pkg/convert/jfr/parser.go
petethepig: Let's merge like this and we can add it later if people ask for it.

pkg/convert/jfr/parser.go
petethepig: Same here

petethepig (approved):
Looks great, merging!

PR #1156, opened 2022-06-13
test: remove flaky cypress tests

eh-am:
Remove most cypress snapshot tests, since:
* they are flaky
* require a linux machine to (re)generate them
* we are only testing against the flamegraph canvas, which
  * a) should be very difficult to break (since we QA every single PR using the review apps)
  * b) we already test it using unit tests, which is much less flaky (https://github.com/pyroscope-io/pyroscope/blob/12cf3a975215b9cb8b5c14d75b89437140c79284/packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/__image_snapshots__/canvas-renderer-spec-tsx-canvas-renderer-group-snapshot-collapses-small-blocks-into-one-1-snap.png)


Rperry2174 (approved):
LGTM 🔥

PR #1160, opened 2022-06-16
chore(frontend)[refactored]: request cancellation for profile data

pavelpashkovsky:
based on https://github.com/pyroscope-io/pyroscope/pull/1143, logic updated in accordance with @eh-am thoughts
The difference is:
- Cancellations get back previous loaded state if it exists. 
- Deleted all newly added 'failed' state.
- Loading indicator logic reverted to initial state (every single page tracks if there is any active fetching on any page)


webapp/javascript/hooks/timeline.hook.ts
```diff
@@ -6,12 +6,15 @@ import {
...
 } from '@webapp/redux/reducers/continuous';
 import Color from 'color';
+import useCancelRequestOnUnmount from '@webapp/hooks/cancelRequestOnUnmount.hook';
 
 // Purple
 export const leftColor = Color('rgb(208, 102, 212)');
 // Blue
 export const rightColor = Color('rgb(19, 152, 246)');
 
+let timelineData: ShamefulAny;
```
eh-am: Ugh, isn't there a way to type this? Maybe with `ReturnType<dispatch>`?

pavelpashkovsky: there is no this code in last version of this file, but there are places when I left this problem. let me check

eh-am (approved):

PR #1166, opened 2022-06-20
Replaced rgb(112, 219, 237) in palette for rgb(10,63,77)

Tusharkshahi:


Rperry2174:
Hey @Tusharkshahi could you replace with something a little lighter like: rgb 78, 146, 249 ? 
image

PR #1168, opened 2022-06-21
WIP fix(jfr) exemplars baseline

korniltsev:


pkg/server/ingest_test.go
```diff
@@ -328,12 +328,14 @@ var _ = Describe("server", func() {
...
+								id    string
+								sleep time.Duration
+								key   string
 							}
 							cids := []contextID{
-								{id: "0", key: `test.app.` + t + `{foo="bar", baz="qux"}`},
-								{id: "1", key: `test.app.` + t + `{foo="bar", baz="qux", thread_name="pool-2-thread-8"}`},
+								{id: "0", sleep: 500 * time.Millisecond, key: `test.app.` + t + `{foo="bar", baz="qux"}`},
+								{id: "1", sleep: 500 * time.Millisecond, key: `test.app.` + t + `{foo="bar", baz="qux", thread_name="pool-2-thread-8"}`},
+								{id: "2", sleep: 6 * time.Second, key: `test.app.` + t + `{foo="bar", baz="qux", thread_name="pool-2-thread-8",profile_id="239239239239"}`},
```
petethepig, line 338: 6 seconds seems a bit too much? Can we do a lower value?

petethepig (approved):
Looks good. I only had one nit about the test duration.

---

Looking at this got me thinking about other formats though. If we add support for labels to other formats in the future it seems like we'll have to implement the same duplication functionality for yet another format again.

I'm thinking, could we instead move the duplication logic deeper into the storage, e.g right around [here](https://github.com/pyroscope-io/pyroscope/blob/main/pkg/storage/storage_put.go#L40-L42) — I think this would be a slightly better design. Also, if we had it implemented that way in the first place we wouldn't need this PR 🤪

What do you guys think? 

@korniltsev — Would it be hard to refactor it like that? 
@kolesnikovae — Is there any potential drawbacks to it?

pkg/server/ingest_test.go
korniltsev: It is because of `exemplarBatchDuration = time.Second * 5`, we could either lower the `exemplarBatchDuration` for the test or add some force flush logic, I am not sure which is better or maybe there are other options, I'd ask @kolesnikovae opinion.
Also the other problem here is that we do this 8 times because of the nested contexts - ones per every type of sample(cpu, objects, ...etc). It would be nice to rewrite the test to put the jfr once and query the date 8 times. Should I do it in this PR?

korniltsev:
Moving duplication logic deeper should work.
I think the reason it was not deep into storage but as close to the parsing as possible is to merge duplicated (without profile_id) stacktraces early and only put them once - save some `Sotrage#Put' invocations. I am not very familiar with storage yet and don't know how performance could change. 
But I could try.

pkg/server/ingest_test.go
kolesnikovae: > It is because of exemplarBatchDuration = time.Second * 5, we could either lower the exemplarBatchDuration for the test or add some force flush logic

I'm afraid flushing exemplars from outside is only available in https://github.com/pyroscope-io/pyroscope/pull/1169: it adds `storage.ExemplarsInternals()` which also returns `Sync()` callback. I think I'll merge it soon, so you can rebase.

I'd say that tests in this package should not depend on the actual storage implementation. It's better to use an ingester mock. As of now our tests cover multiple "layers":
- ingestion - raw HTTP request -> `ingestion.IngestInput`
- parsing: `ingestion.IngestInput` -> `storage.PutInput`
- persistent storage: `storage.PutInput` -> data in BadgerDB

I won't nag about absolutely hermetic unit tests but we could do this better.
- If the ingester handler is only responsible for processing HTTP requests to `ingestion.IngestInput`, our test should validate only this conversion.
- Details of how raw profiling data of one of the format is parsed should be covered in the parser tests.

kolesnikovae:
@korniltsev precisely described the rationale for the current implementation.

I think we could try. Actually, it's hard to estimate impact on the performance – it may turn out that the penalty can be small but this will require substantial change of `storage.Put` and I doubt we are willing to do this now: even these lines right below the `HasProfileID` split will impact performance (we will see it in `pyroscope.server.cpu`):
```go
	s.logger.WithFields(logrus.Fields{
		"startTime":       pi.StartTime.String(),
		"endTime":         pi.EndTime.String(),
		"key":             pi.Key.Normalized(),
		"samples":         pi.Val.Samples(),
		"units":           pi.Units,
		"aggregationType": pi.AggregationType,
	}).Debug("storage.Put")
```

The bigger problem is that for **_every exemplar (a raw profile can include hundreds of them)_** we need:
1. Update labels - direct call to badger DB, which is quite expensive:
```go
	for k, v := range pi.Key.Labels() {
		s.labels.Put(k, v)
	}
```

2. Update affected dimensions (cached, but still pretty expensive in terms of allocations):
```go
	for k, v := range pi.Key.Labels() {
		key := k + ":" + v
		r, err := s.dimensions.GetOrCreate(key)
		if err != nil {
			s.logger.Errorf("dimensions cache for %v: %v", key, err)
			continue
		}
		r.(*dimension.Dimension).Insert([]byte(sk))
		s.dimensions.Put(key, r)
	}
```

3. Get segment from the cache and update its metadata: default json encoder in Go is not very performant.
```go
	r, err := s.segments.GetOrCreate(sk)
	if err != nil {
		return fmt.Errorf("segments cache for %v: %v", sk, err)
	}

	st := r.(*segment.Segment)
	st.SetMetadata(pi.SpyName, pi.SampleRate, pi.Units, pi.AggregationType)
```

4. Iterate over affected segment nodes, fetch corresponding trees, merge the cloned exemplar tree - the most expensive part:
```go
	err = st.Put(pi.StartTime, pi.EndTime, samples, func(depth int, t time.Time, r *big.Rat, addons []segment.Addon) {
		tk := pi.Key.TreeKey(depth, t)
		res, err := s.trees.GetOrCreate(tk)
		if err != nil {
			s.logger.Errorf("trees cache for %v: %v", tk, err)
			return
		}
		cachedTree := res.(*tree.Tree)
		treeClone := pi.Val.Clone(r)
		for _, addon := range addons {
			if res, ok := s.trees.Lookup(pi.Key.TreeKey(addon.Depth, addon.T)); ok {
				ta := res.(*tree.Tree)
				ta.RLock()
				treeClone.Merge(ta)
				ta.RUnlock()
			}
		}
		cachedTree.Lock()
		cachedTree.Merge(treeClone)
		cachedTree.Unlock()
		s.trees.Put(tk, cachedTree)
	})
```

I can't see how this can be optimised in a trivial way.





Regarding skipping exemplars – I think it's okay if drop them after parsing. I anticipate that this is a temporary measure and we won't use this option after exemplars fully supported in the cloud storage.

kolesnikovae:
Speaking of design: I'd better think about batch insert.

pkg/server/ingest_test.go
kolesnikovae: I have just merged https://github.com/pyroscope-io/pyroscope/pull/1169 - sorry for the delay

PR #1178, opened 2022-06-25
feat: adds proper support for goroutines, block, mutex profiling

petethepig:
Corresponding PR on the client side: https://github.com/pyroscope-io/client/pull/15

To enable this in pull mode:
```yaml
scrape-configs:
  - job-name: testing
    enabled-profiles: [cpu, mem, goroutines, mutex, block]
```

Caveat is that mutex and block profiling requires an extra step:
```golang
runtime.SetMutexProfileFraction(5)
runtime.SetBlockProfileRate(5)
```


kolesnikovae (approved):
LGTM

PR #1188, opened 2022-06-29
feat(frontend): support disabling exporting to flamegraph.com

eh-am:
Closes https://github.com/pyroscope-io/pyroscope/issues/1186 and https://github.com/pyroscope-io/pyroscope/issues/1119

I kept our standard of using `--disable/--no` for backend flags, and in the frontend to use `enable`.

There's some repetition in the `<FlamegraphRenderer>` (having to pass the same flags multiple times), but it gives us flexibility to change as we want. At some point when we make all the continuous pages have feature parity (eg export to pprof doesn't work with diff) we may clean this up.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -159,6 +159,8 @@ type Server struct {
 	SelfProfilingTags map[string]string `name:"self-profiling-tag" def:"" desc:"tag in key=value form. The flag may be specified multiple times" mapstructure:"self-profiling-tags" yaml:"self-profiling-tags"`
 
 	RemoteWrite RemoteWrite `yaml:"remote-write" mapstructure:"remote-write"`
+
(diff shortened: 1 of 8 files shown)


petethepig (approved):

kavfixnel:
Does anyone have documentation for how to actually use this setting? Can't seem to find anything from a quick search.

tigerinus:
All issues on this subject point to this PR, but still clueless about how to disable sending data to flamegraph.com

Please add documentation. 
Please add documentation. 
Please add documentation.

eh-am:
Hi folks, I don't work on this project anymore, but from a quick look it seems the server doesn't pass that flag (`exportToFlamegraphDotComEnabled`) to the frontend anymore. I recommend creating a new issue.

PR #1191, opened 2022-06-29
Merge branch 'main' of https://github.com/Tusharkshahi/pyroscope

Tusharkshahi:
I replaced rgb(112,219,237) in palette for rgb(78,146,249).


Rperry2174 (approved):
Definitely an improvement over the initial color! Merging and then fixing tests right after

PR #1195, opened 2022-06-30
fix: default name when exporting diff

dogfrogfog:
examples:
diff_pyroscope.server.inuse_objects_last_3_hours
diff_pyroscope.server.inuse_objects_last_12_hours
diff_pyroscope.server.inuse_objects_2022-06-30_0209-to-2022-06-30_1034
diff_pyroscope.server.mutex_duration_last_3_hours
diff_pyroscope.server.mutex_duration_2022-06-30_1014-to-2022-06-30_1120

Снимок экрана 2022-06-30 в 12 29 02 
Снимок экрана 2022-06-30 в 12 28 48

if we leave spaces and capitalized L 

Снимок экрана 2022-06-30 в 12 29 26

### DIFF
--- a/webapp/javascript/util/formatDate.ts
+++ b/webapp/javascript/util/formatDate.ts
@@ -64,17 +64,6 @@ export function readableRange(
   return `${format(d1, dateFormat)} - ${format(d2, dateFormat)}`;
 }
 
-export function dateForExportFilename(from: string, until: string) {
-  const dateFormat = 'yyyy-MM-dd_HHmm';
-  if (/^now-/.test(from) && until === 'now') {
-    const { number, _multiplier } = convertPresetsToDate(from);
-    return `last_${number}_${_multiplier}`;
-  }
-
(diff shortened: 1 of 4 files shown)


webapp/javascript/util/formatDate.ts
```diff
@@ -68,7 +68,7 @@ export function dateForExportFilename(from: string, until: string) {
   const dateFormat = 'yyyy-MM-dd_HHmm';
   if (/^now-/.test(from) && until === 'now') {
     const { number, _multiplier } = convertPresetsToDate(from);
-    return `Last ${number} ${_multiplier}`;
+    return `last_${number}_${_multiplier}`;
```
eh-am, line 71: You said this function is only used in `ExportData`, right? If so can we move to that file (ExportData.tsx)? I feel like having it here is confusing. I was expecting it to be something generic that was used somewhere else.

eh-am (approved):

PR #1197, opened 2022-06-30
fix: colors on login pages

petethepig:


Rperry2174 (approved):
Sorry I broke this in my refactoring 😅... If we could not use opacity so much that would be ideal though because opacity ends up weird on different backgrounds (i.e. light mode)

PR #1199, opened 2022-07-01
feat(webapp): new app selector

pavelpashkovsky:


eh-am (commented):
This looks really good!

Few thoughts:

* Can we keep the icons width the same? Otherwise the app name is slightly shifted.

https://user-images.githubusercontent.com/6951209/176917512-af6bf8ae-841f-4c9f-b9b0-f14bcbcd6eea.mp4
* We need some kind of pagination based on the window height, same as we already have:
https://user-images.githubusercontent.com/6951209/176918756-2cebc67d-92f7-493c-bd0e-2909fbf19e89.mp4


https://user-images.githubusercontent.com/6951209/176918875-bd94854e-8c2b-4680-a3ef-52acf76f3345.mp4




* This is very personal, but I don't like this icon much 
image. For some reason it reminds me of Firefox (not that I dislike Firefox, it's just that it doesn't convey any meaning).


* The SELECT APPLICATION is a little underwhelming, it doesn't convey that's a title
image


* Since this component is going to grow in complexity in the feature, can we have some tests? Preferrably using react testing library. Feel free to ping me for examples!


But as I said, it looks good and it will definitely allow us to do great things with this component :)

webapp/javascript/components/AppSelector/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+    appNames.indexOf(q) !== -1 ? q : ''
+  );
+  const appNamesState = useAppSelector(selectAppNamesState);
+
+  const [selectorOpened, toggleSelector] = useState(false);
+
+  const selectAppName = (name: string) => {
+    const appNameQuery = queryFromAppName(name);
+
+    if (onSelectedName) {
```
eh-am, line 50: This was a bit of a hack from the previous implementation IIRC, can you add a small comment explaining that we allow overwriting the behaviour when a name is selected? Could be in the props.

webapp/javascript/components/AppSelector/SelectorModal.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const [selected, select] = useState<string[]>([]);
+
+  const filteredAppNames = useMemo(
+    () =>
+      appNames.filter((n: string) =>
+        n.toLowerCase().includes(filter.trim().toLowerCase())
+      ),
+    [filter, appNames]
+  );
+
```
eh-am, line 38: There's a bit of logic here in this component, I would appreciate if you could add comments explaining them in a high level

Rperry2174 (approved):
looks great thanks @pavelpashkovsky

Issue #623, opened 2021-12-17
Bug when focusing on a node and then changing time range


petethepig:
When you click on a node on the flamegraph and then change timerange the selected node does not persist, and a different node is selected. I suspect it's because we store index instead of the whole path to the node.

Here's a gif that illustrates the problem
![2021-12-16 20 13 42](https://user-images.githubusercontent.com/662636/146488305-4c25a637-fe15-4ef4-891c-36f26325fe13.gif)

I think a better behavior would be to store the whole path internally, and when flamegraph changes check if a node at that path is available, and if not, reset to the unfocused state.

PR #1200, opened 2022-07-01
fix(frontend): don't crash when flamegraph changes

eh-am:
This was raised by @kolesnikovae  

We introduced a bug in https://github.com/pyroscope-io/pyroscope/pull/1184
Basically when flamegraphs change, we may not always be able to calculate the bar contents, since the flamegraphs may have different sizes.


kolesnikovae:
Closes #623 I guess

Issue #1204, opened 2022-07-01
Add titles to charts / flamegraphs


Rperry2174:
Right now we don't have any titles that help to explain what each of the elements on the page are. Initially we did this because we wanted to fit as much content as we could on the page, but after thinking about it and talking to some new users I think it would make sense for us to spend a little real estate giving names / titles to the elements. 

### Comparison Page
- Makes it clearer which color corresponds with which graph
- Introduces the concept of "baseline" and "comparison" graph as terms (as opposed to "left" and "right" see #1181)
- Makes it clearer what the top timeline shows

Note: The top timeline will have to change depending on the application that is chosen (i.e. memory or cpu or itimer or goroutines etc) 

![image](https://user-images.githubusercontent.com/23323466/176959916-2fb86179-5128-4ff6-8584-c4d2f5c2f97f.png)

### Diff Page
I think diff page is the most confusing as it sort of "assumes" a user came from the comparison page. I think reorganizing the page along with the titles will make it clear that it is a diff between the "baseline" and "comparison" graphs (as well as which query is the baseline and which query is the comparison)

I also think that separating each element into its own <Box /> will make the page look cleaner as well. 

![image](https://user-images.githubusercontent.com/23323466/176960399-e51dd78c-4861-4b36-bffc-23206e6d2ac1.png)

Again we should also update the tooltip to say "baseline" and "comparison" as opposed to "left" and "right" as mentioned in #1181

Rperry2174:
closed in #1208

PR #1208, opened 2022-07-04
feat: add titles to charts / flamegraphs

dogfrogfog:
#1204

### DIFF
--- a/webapp/javascript/pages/ContinuousComparisonView.tsx
+++ b/webapp/javascript/pages/ContinuousComparisonView.tsx
@@ -105,6 +105,7 @@ function ComparisonApp() {
           data-testid="comparison-container"
         >
           <Box className={styles.comparisonPane}>
+            <TimelineTitle title="Baseline Flamegraph" color={leftColor} />
             <TagsBar
               query={leftQuery}
               tags={leftTags}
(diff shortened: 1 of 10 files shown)


Rperry2174:
I know its still draft, but two things I'd add in:
image
Theres some extra padding for `diff-instructions-wrapper` that makes the gap larger. Lets remove that

also let's do font size 18 instead of 22 for the titles

eh-am:
https://user-images.githubusercontent.com/6951209/177356441-70fee6b5-e827-41e1-afd6-882eeb7cecb3.mp4

Also the content shift is bit annoying, can we always have something in the header (fixed height, or a dummy element dunno)?

webapp/javascript/pages/ContinuousComparisonView.tsx
```diff
@@ -102,11 +97,8 @@ function ComparisonApp() {
               right: { from: rightFrom, to: rightUntil, color: rightColor },
             }}
             timezone={timezone}
-            title={
-              <TimelineTitle
-                titleKey={isSidesHasSameUnits ? leftSide?.metadata.units : ''}
-              />
-            }
+            // using leftSide because comparison view (for now) should match the application at the top of the screen
```
dogfrogfog, line 100: shouldn't we expand(or remove) this comment ?

dogfrogfog:
latest version demo

https://user-images.githubusercontent.com/47758224/177386299-9f1f53b1-899b-48a5-b9a1-4896e2ec6545.mov

eh-am (approved):
Looks good! Don't forget to add to Standalone export too (https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/standalone.tsx)!

PR #1209, opened 2022-07-05
feat: support for micro, milli, and nano seconds

kolesnikovae:


petethepig (approved):

pkg/util/attime/attime.go
```diff
@@ -24,8 +24,19 @@ func Parse(s string) time.Time {
 			return t
 		}
 
+		// Unix time.
 		v, _ := strconv.Atoi(s)
-		return time.Unix(int64(v), 0)
+		switch len(s) {
+		default:
+			// Seconds.
+			return time.Unix(int64(v), 0)
```
petethepig, line 32: We might need to revisit this logic around November 20, 2286.

petethepig (approved):

pkg/util/attime/attime.go
kolesnikovae: I'll set a reminder!

PR #1211, opened 2022-07-05
fix: single view app update should change comp/diff view left and right apps

dogfrogfog:
https://user-images.githubusercontent.com/47758224/177382161-31767f11-d859-4a6c-b16e-6e644493b7c1.mov


dogfrogfog:
right, thx, i will remove it

Rperry2174 (approved):
looks good thanks!

PR #1212, opened 2022-07-05
fix(flamegraph): do a deep comparison for whether the flamegraph is the same

eh-am:
Closes https://github.com/pyroscope-io/pyroscope/issues/1192

Changed the naive `numTicks` for a deep comparison. Although JSON serialization is slow, which is a not big deal (since our flamegraphs are small), but they could be in the future, hence why I left a comment as a hint.

### DIFF
--- a/packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphRenderer.tsx
+++ b/packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphRenderer.tsx
@@ -232,8 +232,9 @@ class FlameGraphRenderer extends React.Component<
   };
 
   isSameFlamebearer = (prevFlame: Flamebearer, currFlame: Flamebearer) => {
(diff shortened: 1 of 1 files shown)


packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphRenderer.tsx
```diff
@@ -232,8 +232,9 @@ class FlameGraphRenderer extends React.Component<
   };
 
   isSameFlamebearer = (prevFlame: Flamebearer, currFlame: Flamebearer) => {
-    // This is a poor heuristic, but it should work most of the times
-    return prevFlame.numTicks === currFlame.numTicks;
+    // TODO: come up with a less resource intensive operation
```
petethepig, line 235: I think a better solution would be to generate a random ID for each response on the server side.

petethepig (approved):

PR #1213, opened 2022-07-05
[DO NOT MERGE] tracing flamegraph example

petethepig:
Preview 

Screen Shot 2022-07-05 at 3 22 42 PM

To run this:

First, start dev server
```
make dev
```

Then go to http://localhost:4040/settings/tracing


Rperry2174:
Next steps: 

1. move this function https://github.com/pyroscope-io/pyroscope/pull/1213/files#diff-679fa369ac963025edba98a12e194c20406272e512ae44db3d1ecb56397c8f9bR27 to the flamegraph package
2. Create a prop for input that is tracing data (since it is a different format)
3. create logic that accepts tracing data if it exists otherwise uses flamegraph input
4. Double check coloring makes sense (should be close to trace coloring) 

cc @dogfrogfog

eh-am:
converted to draft so that we don't accidentally merge it

petethepig:
superseded  by #1219

PR #1214, opened 2022-07-05
fix: improved nodes coloring by fixing murmur math

petethepig:
I don't know how this happened but we introduced this bug at some point. I changed it to proper math. E.g see code [here](https://github.com/jsdelfino/murmurhash-32/blob/master/murmurhash3_gc.js#L44-L47)


Rperry2174 (approved):
looks good

PR #1219, opened 2022-07-06
finishing work for 'tracing flamegraph' branch

dogfrogfog:
for demo i used `jaeger-ui` app with `"@pyroscope/flamegraph": "0.9.7-1219-1fa90c4.0",(from one of build pipelines)`.
didnt manage to get more test data to illustrate bigger graphs. FlamegraphRenderer was just imported to the page with trace data, which is /trace, to show the demo

- move `traceToFlamebearer` converter to flamegraph package this function
- add typescript and trace schema
- create a prop for input that is tracing data (since it is a different format)
- create logic that accepts tracing data if it exists otherwise uses flamegraph input
- check coloring (should be close to trace coloring) ......green is close 
- change renderLogo property
- add showCredit property
- add logo with low z-index to canvas

https://user-images.githubusercontent.com/47758224/177644841-1ca4a505-fee5-4fbd-8caf-78267abd1444.mov


packages/pyroscope-flamegraph/src/convert/convert.ts
```diff
@@ -0,0 +1,92 @@
...
+  let result: Flamebearer = {
+    format: 'single',
+    numTicks: 0,
+    sampleRate: 1000000,
+    names: [],
+    levels: [],
+    units: 'samples',
+    spyName: 'tracing',
+
+    // we use state for that ?
```
dogfrogfog, line 21: ---

packages/pyroscope-flamegraph/src/convert/convert.ts
```diff
@@ -0,0 +1,93 @@
...
+    // fitMode: 'HEAD',
+
+    // this fields are only on test data. do we need that ?
+    // topLevel: 0,
+    // rangeMin: 0,
+    // rangeMax: 1,
+  };
```
dogfrogfog, line 29: can you comment here please

eh-am:
I wonder if we should support `trace` as a first class integration, and instead support a generic `hierarchical` data, so that we can display any arbitrary hierarchical data (eg webpack bundles report, file system usage).

petethepig:
@eh-am maybe not instead but in addition to it? I think this is a good one to discuss thursday.

packages/pyroscope-flamegraph/src/convert/convert.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export function traceToFlamebearer(trace: Trace): Flamebearer {
+  let result: Flamebearer = {
+    format: 'single',
+    numTicks: 0,
+    sampleRate: 1000000,
+    names: [],
+    levels: [],
+    units: 'samples',
+    spyName: 'tracing',
```
eh-am, line 20: Shall we call it `jaeger-tracing`? @petethepig

dogfrogfog:
latest updates demo

https://user-images.githubusercontent.com/47758224/177984161-8ad3f520-fc2e-47d7-87b5-65842625a1c5.mov

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/index.tsx
```diff
@@ -23,6 +24,7 @@ interface FlamegraphProps {
   fitMode: ConstructorParameters<typeof Flamegraph>[3];
   highlightQuery: ConstructorParameters<typeof Flamegraph>[4];
   zoom: ConstructorParameters<typeof Flamegraph>[5];
+  showCredit: boolean;
```
eh-am, line 27: we should make this optional, but defaulting to true

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/index.spec.tsx
```diff
@@ -22,6 +22,7 @@ describe('FlamegraphComponent', () => {
 
     render(
       <FlamegraphComponent
+        showCredit
```
eh-am, line 25: if it's optional, why did you have to add it here?

dogfrogfog: it's optional for the parent component, which is [FlameGraphRenderer](https://github.com/pyroscope-io/pyroscope/pull/1219/files#diff-6a469ade99d60ee9e8195ee6e6e0a707f7ab092dfc7d1abd3242260853e07d8f), with default `true`. passing `showCredit` here is the emulation of parent default prop

eh-am: oh that makes sense!

Rperry2174 (approved):
Looks great! Thanks @dogfrogfog

PR #1229, opened 2022-07-07
refactor(flamegraph): profiler table in typescript

eh-am:
not my proudest moment, but added types to `ProfilerTable`

this is a first step to simplify its logic later


packages/pyroscope-models/src/flamebearer.ts
```diff
@@ -16,6 +16,8 @@ export type Flamebearer = {
    */
   levels: number[][];
   numTicks: number;
+  maxSelf: number;
```
eh-am, line 19: in ideal conditions we would break the API here since this is now a required field, but since it's still < 1.0.0 I will leave as it is

Issue #1223, opened 2022-07-06
Implement delete app functionality as an HTTP endpoint


petethepig:
Currently it's only possible via the admin command.

We should add this functionality to the HTTP controller. We should make it so that it requires authentication. If there's no authentication mechanisms set up on the server it should return an error.

eh-am:
Maybe we could even use the same admin server?  https://github.com/pyroscope-io/pyroscope/blob/main/pkg/admin/server.go#L42

petethepig:
done in https://github.com/pyroscope-io/pyroscope/commit/b82f42625f5ee80fad9701d1abd2e0b059aee46c

PR #1239, opened 2022-07-11
feat(api): Implement delete app functionality as an HTTP endpoint backend #1223 

korniltsev:


pkg/server/controller.go
```diff
@@ -256,6 +256,15 @@ func (ctrl *Controller) serverMux() (http.Handler, error) {
 		ctrl.drainMiddleware,
 		ctrl.authMiddleware(nil))
 
+	deleteAppRouter := r.Path("/delete-app").Subrouter()
```
eh-am, line 259: For consistency sake I would suggest to use ´DELETE /apps/`
https://github.com/pyroscope-io/pyroscope/blob/main/pkg/admin/server.go#L42

pkg/server/controller.go
```diff
@@ -256,6 +256,15 @@ func (ctrl *Controller) serverMux() (http.Handler, error) {
...
+	deleteAppRouter.Use(
+		api.AuthMiddleware(nil, ctrl.authService, ctrl.httpUtils),
+		authz.NewAuthorizer(ctrl.log, httputils.NewDefaultHelper(ctrl.log)).RequireOneOf(
+			authz.Role(model.AdminRole),
+			authz.Role(model.AgentRole),
```
eh-am, line 264: Agent too? To me it would be only Admin, to keep in the same veins as the socket admin server.

eh-am (commented):
Looks good, I would honestly try to somehow reuse all routes from here https://github.com/pyroscope-io/pyroscope/blob/main/pkg/admin/server.go#L40-L44

Since from my perspective, everything a local admin user can do, a remote admin can do as well.

korniltsev:
@eh-am I've pulled `admin.Controller` 👍

eh-am (approved):

eh-am:
Closes https://github.com/pyroscope-io/pyroscope/issues/1223

petethepig (approved):
Looks good. I have 2 questions:

* what will happen if a user has no auth methods set up? will unauthorized users be able to delete apps? (Ideally we should make it so that they can't).

pkg/admin/controller.go
```diff
@@ -44,6 +44,13 @@ func NewController(
 	}
 }
 
+func (ctrl *Controller) RegisterHandlers(r *mux.Router) {
+	r.HandleFunc("/v1/apps", ctrl.HandleGetApps).Methods("GET")
+	r.HandleFunc("/v1/apps", ctrl.HandleDeleteApp).Methods("DELETE")
```
petethepig, line 49: this one should be /api/apps

pkg/admin/controller.go
```diff
@@ -44,6 +44,13 @@ func NewController(
...
+	r.HandleFunc("/v1/users/{username}", ctrl.UpdateUserHandler).Methods("PATCH")
+	r.HandleFunc("/v1/storage/cleanup", ctrl.StorageCleanupHandler).Methods("PUT")
```
petethepig, line 51: these two let's remove from HTTP controller for now

kolesnikovae:
Hey @korniltsev! The PR looks good to me.

One question I'd like to raise is should we return the list of applications as a list of objects, not just strings? I just created an issue https://github.com/pyroscope-io/pyroscope/issues/1263 explaining the current issue and rationale for the new API.

/cc @petethepig, @eh-am

pkg/server/controller.go
```diff
@@ -256,6 +260,17 @@ func (ctrl *Controller) serverMux() (http.Handler, error) {
...
 
+	if ctrl.adminController != nil {
+		adminRouter := apiRouter.PathPrefix("/admin/").Subrouter()
+		adminRouter.Use(
+			api.AuthMiddleware(nil, ctrl.authService, ctrl.httpUtils),
+			authz.NewAuthorizer(ctrl.log, httputils.NewDefaultHelper(ctrl.log)).RequireOneOf(
+				authz.Role(model.AdminRole),
+			))
+		adminRouter.HandleFunc("/apps", ctrl.adminController.HandleGetApps).Methods("GET")
+		adminRouter.HandleFunc("/apps", ctrl.adminController.HandleDeleteApp).Methods("DELETE")
```
kolesnikovae, line 271: Just a thought: shouldn't the route be `DELETE /apps/{app_name}`? I understand that this would require changing the handler, but it's still worth considering :)

korniltsev: In https://github.com/pyroscope-io/pyroscope/issues/1263 a potential object in the future has sampleType field 
What if we would like to delete an app for only specific sampleType? Do we also put it in url? as path or query? Right now current DeleteInput json seems extendable the same way the new proposed returned object from `GET /apps`

kolesnikovae: Good points. I'd say that we could represent an application as a collection of profile types: /apps/{app_name}/profile-types/{profile_type}

Let's maybe leave it as it is and revisit later

kolesnikovae:
I think it makes sense to extend the existing handler

pkg/server/controller.go
```diff
@@ -256,6 +260,17 @@ func (ctrl *Controller) serverMux() (http.Handler, error) {
...
+	}
+
```
kolesnikovae, line 273: It looks like `/apps` depend on the `svc.config.EnableExperimentalAdmin` option. I'm not sure we want this dependency.

pkg/admin/service.go
```diff
@@ -1,28 +1,34 @@
...
 	m := &AdminService{
 		v,
 	}
-
 	return m
 }
 
-func (m *AdminService) GetApps() (appNames []string) {
-	return m.storage.GetAppNames(context.TODO())
+func (m *AdminService) GetAppNames(ctx context.Context) (appNames []string) {
```
kolesnikovae, line 24: It seems the service does nothing but proxying requests to the storage. Please consider removing it

pkg/admin/controller.go
```diff
@@ -44,14 +44,31 @@ func NewController(
...
+	res := make([]AppInfo, 0, len(apps.Apps))
+	for _, app := range apps.Apps {
+		it := AppInfo{
+			Name: app.Name,
+		}
+		res = append(res, it)
+	}
+	w.WriteHeader(http.StatusOK)
+	ctrl.writeResponseJSON(w, res)
+}
```
kolesnikovae, line 70: Please consider refactoring app handlers to `pkg/server` package due to:
1. Consistency with the existing HTTP API implementation.
2. Proper error handling – HTTP API methods use [a special helper](https://github.com/pyroscope-io/pyroscope/blob/c6486ffbee94f0a7277aa15f40b9000a796c196b/pkg/server/render.go#L81).
3. Independence of admin controller on the initialisation: now the main controller depends on the admin one.

kolesnikovae (changes requested):

kolesnikovae (changes requested):

kolesnikovae (approved):

petethepig:
The PR is ready but we're going to wait for Anton to finish his refactoring before merging this.

PR #1241, opened 2022-07-11
chore(frontend): remove tracing page

eh-am:
It was pushed by accident in https://github.com/pyroscope-io/pyroscope/pull/1233


Rperry2174 (approved):
I did remove it.. then added it back because I thought it might be causing lint problems 😅 . Thanks for removing again though

PR #1244, opened 2022-07-11
feat: adds support for group-by queries on the backend

petethepig:
To test:
```shell
# to create data
curl -X POST --data 'foo;bar;baz 100' 'http://localhost:4040/ingest?name=0-app%7Bfoo%3D%22bar%22%7D'
curl -X POST --data 'foo;bar;baz 100' 'http://localhost:4040/ingest?name=0-app%7Bfoo%3D%22baz%22%7D'

# then to get the data
curl 'http://localhost:4040/render?from=now-1m&until=now&query=0-app&groupBy=foo&format=json' | jq .groups
```

### DIFF
--- a/pkg/storage/storage_get.go
+++ b/pkg/storage/storage_get.go
@@ -21,11 +21,14 @@ type GetInput struct {
 	EndTime   time.Time
 	Key       *segment.Key
 	Query     *flameql.Query
+	// TODO: make this a part of the query
+	GroupBy string
 }
 
 type GetOutput struct {
 	Tree            *tree.Tree
 	Timeline        *segment.Timeline
+	Groups          map[string]*segment.Timeline
 	SpyName         string
 	SampleRate      uint32
 	Count           uint64
@@ -88,6 +91,7 @@ func (s *Storage) Get(ctx context.Context, gi *GetInput) (*GetOutput, error) {
 		aggregationType = "sum"
 		timeline        = segment.GenerateTimeline(gi.StartTime, gi.EndTime)
 	)
+	timelines := make(map[string]*segment.Timeline)
 
 	for _, k := range dimensionKeys() {
 		// TODO: refactor, store `Key`s in dimensions
@@ -108,7 +112,16 @@ func (s *Storage) Get(ctx context.Context, gi *GetInput) (*GetOutput, error) {
 			aggregationType = averageAggregationType
 		}
 
+		timelineKey := "*"
+		if v, ok := parsedKey.Labels()[gi.GroupBy]; ok {
+			timelineKey = v
+		}
+		if _, ok := timelines[timelineKey]; !ok {
+			timelines[timelineKey] = segment.GenerateTimeline(gi.StartTime, gi.EndTime)
+		}
+
 		timeline.PopulateTimeline(st)
+		timelines[timelineKey].PopulateTimeline(st)
 		lastSegment = st
 
 		trace.Logf(ctx, traceCatGetCallback, "segment_key=%s", key)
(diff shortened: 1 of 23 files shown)


petethepig:
To test:
```shell
# to create data
curl -X POST --data 'foo;bar;baz 100' 'http://localhost:4040/ingest?name=0-app%7Bfoo%3D%22bar%22%7D'
curl -X POST --data 'foo;bar;baz 100' 'http://localhost:4040/ingest?name=0-app%7Bfoo%3D%22baz%22%7D'

# then to get the data
curl 'http://localhost:4040/render?from=now-1m&until=now&query=0-app&groupBy=foo&format=json' | jq .groups
```

Rperry2174:
also works with rideshare example:
```
curl 'http://localhost:4040/render?from=now-1m&until=now&query=ride-sharing-app.cpu%7B%7D&groupBy=region&format=json'  | jq .groups
```
fyi it didn't work for me without a time range not sure if thats intentional or not

pkg/storage/storage_get.go
```diff
@@ -21,11 +21,14 @@ type GetInput struct {
 	EndTime   time.Time
 	Key       *segment.Key
 	Query     *flameql.Query
+	// TODO: make this a part of the query
+	GroupBy string
```
kolesnikovae, line 25: Maybe we should make this a slice, like in the most of query languages. I anticipate that many users will find this useful. Another very handy operator is `without`, with makes query engine to ignore the specified labels. I'm not sure if we need this (and, especially, in the first version), just a noting.

Also, I wonder if we're still going to maintain _similarity_ to PromQL. I think that it may make sense to go our own way. Otherwise, we should follow [their semantics](https://prometheus.io/docs/prometheus/latest/querying/operators/#aggregation-operators) (e.g there `group by` is an aggregation operator and does slightly different things).

kolesnikovae (approved):

pkg/storage/storage_get.go
```diff
@@ -108,7 +112,16 @@ func (s *Storage) Get(ctx context.Context, gi *GetInput) (*GetOutput, error) {
 			aggregationType = averageAggregationType
 		}
 
+		timelineKey := "*"
```
eh-am, line 115: What does a `*` group mean?

petethepig: It's for the case when there's no value for a certain key:value pair. E.g if somebody ingested data at `app{foo="bar"}`, and you're grouping by `baz`. in this example `baz` is empty, so the data written to `app{foo="bar"}` becomes `*`. Data written to `app{baz="bazz"}` becomes `bazz`.

Ideally we should come up with something better than `*` I think.

pkg/storage/storage_get.go
petethepig: Good points. Will make it a slice in the future, it should be backward-compatible with the HTTP API (e.g `groupBy=foo,bar`. Same with `without`, (e.g `groupBy=!foo`).

Eventually we might add it to flameQL too. I think it's too early for these decisions tho. At this point my goal with this current design is to make it so that we *can* add all these additional features in the future.

webapp/javascript/services/render.ts
```diff
@@ -39,6 +39,7 @@ export async function renderSingle(
     z.object({ timeline: TimelineSchema })
   )
     .merge(z.object({ telemetry: z.object({}).passthrough().optional() }))
+    .merge(z.object({ groups: z.object({}) }))
```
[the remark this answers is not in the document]
eh-am: Not relevant for flamegraph.com, it only runs here/cloud.

Issue #1234, opened 2022-07-11
Improve Tooltip styling


Rperry2174:
I think our tooltip could be organized a little bitter to accomplish two goals:
1. Make it easier to read the information in the tooltip
2. Don't trim names in the toooltip see #1217
3. Let users know that there is additional functionality that is available when right clicking on a node in the flamegraph 

I propose a couple of potential options, but personally I think we should go with option C

![tooltip_01-01](https://user-images.githubusercontent.com/23323466/178175110-5ccf39cc-f83a-4a20-a447-f3c48486de85.png)

Rperry2174:
closed in #1246

PR #1246, opened 2022-07-12
feat: add new tooltip design

dogfrogfog:
[#1234](https://github.com/pyroscope-io/pyroscope/issues/1234)

### DIFF
--- a/packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/Tooltip.module.scss
+++ b/packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/Tooltip.module.scss
@@ -1,21 +1,59 @@
 .flamegraphTooltip {
-  max-width: 80%;
+  width: 320px;
 
   position: fixed;
   background: #ffffff;
-  white-space: nowrap;
   box-shadow: 1px 2px 4px 0px rgba(0, 0, 0, 0.3);
   border-radius: 4px;
-  padding: 3px 5px;
   color: #333;
   font-size: 12px;
   visibility: hidden;
-  z-index: 1;
-}
+  z-index: 2;
 
-.flamegraphTooltipName {
-  font-weight: 500;
-  white-space: nowrap;
-  overflow: hidden;
-  text-overflow: ellipsis;
+  .flamegraphTooltipName {
+    border-bottom: 1px solid #333;
+    font-weight: bold;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    padding: 8px 5px;
+  }
+
+  .functionName {
+    padding: 8px 5px 0;
+    min-height: 50px;
+  }
+
+  .contentContainer {
+    padding: 8px 5px;
+    border-bottom: 1px solid #333;
+
+    .content {
+      display: flex;
+    }
+
+    p {
+      text-align: left;
+      margin: 0;
+      width: 50%;
+    }
+  }
+
+  .rightClickInfo {
+    padding: 8px 5px;
+    display: flex;
+    justify-content: center;
+
+    svg {
+      display: block;
+      align-self: center;
+      width: 40px;
+      height: 40px;
+      margin-right: 10px;
+    }
+
+    span {
+      font-size: 14px;
+    }
+  }
 }
--- a/packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/Tooltip.module.scss
+++ b/packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/Tooltip.module.scss
@@ -23,10 +23,12 @@
     overflow: hidden;
     text-overflow: ellipsis;
     padding: 8px 5px;
+    font-family: monospace;
   }
 
   .functionName {
     padding: 8px 5px 0;
+    font-family: monospace;
   }
 
   .tooltipTable {
--- a/packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/Tooltip.module.scss
+++ b/packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/Tooltip.module.scss
@@ -1,5 +1,5 @@
 .flamegraphTooltip {
-  width: 320px;
+  width: 420px;
   position: fixed;
   background: var(--ps-neutral-2);
   box-shadow: 1px 2px 4px 0px rgba(0, 0, 0, 0.3);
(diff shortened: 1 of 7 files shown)


packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/Tooltip.module.scss
```diff
@@ -1,21 +1,59 @@
...
+    font-weight: bold;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    padding: 8px 5px;
+  }
+
+  .functionName {
+    padding: 8px 5px 0;
+    min-height: 50px;
```
dogfrogfog, line 24: 50px = ~2 lines

https://user-images.githubusercontent.com/47758224/178529516-893de6c1-8d7c-4064-9400-41e37bee5ff2.mov

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/Tooltip.tsx
```diff
@@ -11,6 +11,8 @@ import {
 
 import { DefaultPalette, FlamegraphPalette } from './colorPalette';
 import styles from './Tooltip.module.scss';
+import Icon from '@webapp/ui/Icon';
```
eh-am, line 14: Yeah this is a circular dependency, not entirely sure how it affects the bundle size.

If it affects negatively I would recommend to just inline this Icon.
But yeah in the near future we should make another package just for ui.

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/Tooltip.module.scss
eh-am: I don't quite get this part, why minimum 2 lines?

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/Tooltip.tsx
```diff
@@ -58,8 +96,12 @@ export default function Tooltip(props: TooltipProps) {
         color: '',
       },
     },
-    left: '',
-    right: '',
+    tooltipData: [] as {
+      units: Units;
+      percent: string | number;
+      samples: string;
+      formatedValue: string;
```
eh-am, line 103:
```suggestion
      formattedValue: string;
```

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/Tooltip.tsx
```diff
@@ -109,13 +151,13 @@ export default function Tooltip(props: TooltipProps) {
...
-            sampleRate,
-            numTicks,
-            units
-          );
+          const newLeftContent = {
+            percent: formatPercent(data.total / numTicks),
+            samples:
+              units === 'trace_samples' ? '' : numberWithCommas(data.total),
+            units,
+            formatedValue: formatter.format(data.total, sampleRate),
```
eh-am, line 159: Is this calculation valid for tracing? Asking because sampleRate has some bogus value here
 https://github.com/pyroscope-io/pyroscope/blob/main/packages/pyroscope-flamegraph/src/convert/convert.ts#L17

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/Tooltip.tsx
```diff
@@ -269,13 +329,19 @@ function formatDouble(
   const leftPercent = ratioToPercent(leftRatio);
   const rightPercent = ratioToPercent(rightRatio);
 
-  const left = `Baseline: ${numberWithCommas(
-    totalLeft
-  )} samples, ${formatter.format(totalLeft, sampleRate)} (${leftPercent}%)`;
+  const newLeft = {
+    percent: leftPercent,
```
eh-am, line 333: I think this is a bug that was already happening but I got a `NaN` here 

https://user-images.githubusercontent.com/6951209/178542964-38e01418-5890-4b29-b5f0-0a6505e38240.mp4

eh-am (commented):
Looks good, added a few comments.

Ah another thing is the Hand icon. I feel like it's too big and distracting. And honestly, at this point we don't have too many features to use with right click so although it's good for telling people about it, I don't think it's really useful? IMO would be much better to only add this after we have a nice feature like "share this node" or something.

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/Tooltip.tsx
eh-am: To reproduce use something like this 
image

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/Tooltip.tsx
Rperry2174: I do think we should keep the icon. I will give a different one in slack that is an SVG like the one in the image

Rperry2174 (commented):
Looks awesome so far! A couple of tweaks:
1. Adding a slight grey to the header for some contrast:
Lets use either (curious which one you like better):
- `#d8d8d8` 
- `#b5b5b5`
- `#9a9aa0`

we will likely play around with this so please add a new variable to `variables.css` for `tooltip-header-bg`

2. Let's make the "divider" lines a little bit less "heavy" (similar to what you did with the sideline dividers

3. Let's format the actual data as a table (to be consistent with what I'm recommending for No. 4
image


5. Diff view:
Let's turn diff view into a table like this instead. We should leave the "diff" column empty for CPU time and for samples because those are not "normalized" values. 

For example, someone can select 10 minutes for the baseline graph and 5 minutes for the comparison graph. When we calcualte the "percentage diff" it is calcualting the normalized percentage share of CPU time. However, samples and CPU time themselves are absolute values and thus not normalized. 

Long term we will fine a better way to do this, but for now let's just do it this way for this PR to keep parity with the current functionality and we will update more later 

image

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/Tooltip.module.scss
dogfrogfog: i wanted to ask what you do think about it ? ....ref to [this](https://github.com/pyroscope-io/pyroscope/pull/1208#issuecomment-1175155291)...or this min-height doesnt make any sense because mostly we have long function names

Rperry2174: yeah for this use case I don't think we need the minimum 2 lines , but I like the idea :)

Rperry2174 (commented):
Looks great a few more tweaks / comments:
image

- red: change `% of cpu` to `share of cpu`
- red: remove extra "samples" text
- green: Make this text the same monospaced font that is on the flamegraph for the function names

With diff view similar comments:
image

After that I think it's good to merge!

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/Tooltip.tsx
```diff
@@ -48,6 +49,51 @@ export type TooltipProps = {
...
+  },
+  trace_samples: {
+    percent: '<TBD>',
+    formattedValue: '<TBD>',
+  },
+  '': {
+    percent: '',
+    formattedValue: '',
+  },
+};
```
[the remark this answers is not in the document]
Rperry2174: Sounds good... also @petethepig can you provide some input here?

Rperry2174 (approved):
Looks awesome! Thanks @dogfrogfog

PR #1252, opened 2022-07-13
chore(flamegraph): ship raw files

eh-am:
Ship `FlamegraphRenderer.tsx` directly (without bundling/minifying). For clarification, we still ship the flamegraph bundled for the people that want it.

This allows people to use the unminified version if they want, which allows for better tree shaking.
For us, it allows not having to build the flamegraph separately while running locally, which is faster and doesn't contain race conditions (https://github.com/pyroscope-io/pyroscope/pull/1161).

Some other arguments against bundling/minification https://www.briefs.fm/3-minutes-with-kent/221

Did the same thing for `@pyroscope/models`.


Rperry2174 (approved):
LGTM

PR #1230, opened 2022-07-08
feat(webapp): Improve Tagbar tag UI/UX for removing tags [WIP]

pavelpashkovsky:
https://github.com/pyroscope-io/pyroscope/issues/1202

PR #1256, opened 2022-07-13
Removing tags [WIP]

pavelpashkovsky:
FOR: https://github.com/pyroscope-io/pyroscope/issues/1201


Rperry2174:
closing for now along with #1230 because of the pills size and text size not matching up and making this a difficult feature to add

PR #1259, opened 2022-07-13
feat: update right-click context menu

dogfrogfog:
https://github.com/pyroscope-io/pyroscope/issues/1235

https://user-images.githubusercontent.com/6951209/179028292-20ba6fe0-333f-4a09-bb65-c9cf47c4833a.mp4

### DIFF
--- a/packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/index.tsx
+++ b/packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/index.tsx
@@ -204,7 +204,7 @@ export default function FlameGraphComponent(props: FlamegraphProps) {
             : 'Highlight';
 
         return (
-          <MenuItem key="hightlight-similar" onClick={onClick}>
+          <MenuItem key="highlight-similar-nodes" onClick={onClick}>
             <FontAwesomeIcon icon={faHighlighter} />
             {actionName} similar nodes
           </MenuItem>
--- a/packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/index.tsx
+++ b/packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/index.tsx
@@ -1,6 +1,12 @@
 /* eslint-disable no-unused-expressions */
 import React, { useCallback, useRef } from 'react';
 import clsx from 'clsx';
+import Icon from '@webapp/ui/Icon';
+import {
+  faRedo,
+  faCopy,
+  faHighlighter,
+} from '@fortawesome/free-solid-svg-icons';
 import { MenuItem } from '@szhsin/react-menu';
 import useResizeObserver from '@react-hook/resize-observer';
 import { Maybe } from 'true-myth';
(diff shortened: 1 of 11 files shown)


Rperry2174:
couple of small tweaks here as well:
I think right now the "right click" message is a little to overwhelming in the tooltip. Lets:

- make the right click logo a little smaller (let me know if you need a smaller svg)
- make the text a little bit grey so it stands out less
- move the text to be on one line (hopefully it fits, if not let me know and we can reword it a little bit) or potentially increase tooltip width a little bit


image

eh-am:
Now that I am thinking, would be nice to be able to "unhighlight" at any point. Currently you only get a "unhighlight" after clicking on the exact same node.

Also we should come up with a better name, "unhighlight" is kinda weird. Maybe "Clear highlight"?
Also "similar nodes" is kinda confusing, what's a node? What if we call "highlight function"?

Another thing I noticed is a bug we already had, when the flamegraph changes we should clear the highlight. Similar to what we do here https://github.com/pyroscope-io/pyroscope/blob/main/packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphRenderer.tsx#L210

https://user-images.githubusercontent.com/6951209/179028292-20ba6fe0-333f-4a09-bb65-c9cf47c4833a.mp4

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/index.tsx
```diff
@@ -162,19 +172,56 @@ export default function FlameGraphComponent(props: FlamegraphProps) {
...
+      };
+
+      const HighlightSimilarNodesItem = () => {
+        const onClick = () => {
+          setActiveItem({ name: barName });
+        };
+
+        const actionName =
+          selectedItem.isJust && selectedItem.value === barName
+            ? 'Unhighlight'
```
Rperry2174, line 203:
```suggestion
            ? 'Clear highlight'
```

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/index.tsx
```diff
@@ -162,19 +172,56 @@ export default function FlameGraphComponent(props: FlamegraphProps) {
...
+
+      const HighlightSimilarNodesItem = () => {
+        const onClick = () => {
+          setActiveItem({ name: barName });
+        };
+
+        const actionName =
+          selectedItem.isJust && selectedItem.value === barName
+            ? 'Clear highlight'
+            : 'Highlight';
```
Rperry2174, line 204:
```suggestion
            : 'Highlight similar nodes';
```

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/index.tsx
```diff
@@ -162,19 +172,56 @@ export default function FlameGraphComponent(props: FlamegraphProps) {
...
+
+        return (
+          <MenuItem key="highlight-similar-nodes" onClick={onClick}>
+            <FontAwesomeIcon icon={faHighlighter} />
+            {actionName} similar nodes
```
Rperry2174, line 209:
```suggestion
            {actionName}
```

Rperry2174 (approved):
Looks great! Thanks @dogfrogfog

PR #1262, opened 2022-07-14
chore: adds a flamebearer diff function

petethepig:
Need to fix some TS stuff, and then it's good to go.


Rperry2174:
Moved some code from this PR to here: https://github.com/pyroscope-io/flamegraph-renderer-example/commit/7ab78fca60401fc25a3462d198993cd5438b38c0

PR #1266, opened 2022-07-14
ci: improve js tests/lint speed

eh-am:
* Set jest `max-workers` to be the number of cores (as per https://jestjs.io/docs/troubleshooting#tests-are-extremely-slow-on-docker-andor-continuous-integration-ci-server)
* Split js lint/tests into different actions (since tests don't require building the dependencies)

For comparison

Before:
image
image

Now:
image

Thought slowest tests are still the cypress.

### DIFF
--- a/.github/lint-js.yml
+++ b/.github/lint-js.yml
@@ -1,4 +1,4 @@
-name: JS Tests
+name: JS Lint
 
 on:
   push:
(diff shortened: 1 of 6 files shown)


Rperry2174 (approved):
awesome nice work! would be great if we could get flamegraphs from this :)

PR #1270, opened 2022-07-15
ci(cypress): improve cypress tests speed

eh-am:
* Don't record videos/screenshots. Although it's slightly inconvenient not having them, it allows us to speed the tests quite a bit.
* Don't run in docker, since we are not taking snapshots anymore
* When building the app (for e2e), build go/frontend concurrently. To be able to do that embed assets had to be disabled, which means the binary is not self contained and not a real reproduction of the production binary.
* I experimented with sharing the build artifact (for us the pyroscope binary), but it seems the overhead of {storing, pulling} the artifact/finding a runner is already bigger than just letting the jobs build the artifact themselves

Before:
image
image
image

After:
image


Rperry2174 (approved):
🏃 💨  !

PR #1276, opened 2022-07-15
docs: updates tracing example with the right package name

petethepig:
/cc @Rperry2174


Rperry2174 (approved):
LGTM should we add this in honeycomb example too?

PR #1278, opened 2022-07-18
chore: @pyroscope/models ship ts files too

eh-am:
Long story short, since pyroscope webapp uses `@pyroscope/models/src`, so if you use `@pyroscope/webapp` you also need access to `@pyroscope/models/src`, therefore src needs to be shipped.


Rperry2174 (approved):
LGTM

PR #1283, opened 2022-07-18
feat: implements support for environment variable substitutions in config file

petethepig:
Still need to handle $1:$2 case properly, I have an idea for it, just need to implement it.

### DIFF
--- a/pkg/cli/command.go
+++ b/pkg/cli/command.go
@@ -3,7 +3,6 @@ package cli
 import (
 	"errors"
 	"fmt"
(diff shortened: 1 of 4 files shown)


eh-am:
you may be interested in https://github.com/pyroscope-io/pyroscope/blob/main/pkg/cli/server.go#L393

petethepig:
@eh-am I am very much indeed! I should definitely add it there otherwise it won't work for remote write 🤦🏼

PR #1284, opened 2022-07-18
chore: cleanup light mode highlight color

Rperry2174:
Before:
Screen Shot 2022-07-18 at 4 18 08 PM

After:
Screen Shot 2022-07-18 at 4 16 08 PM

### DIFF
--- a/webapp/sass/common.scss
+++ b/webapp/sass/common.scss
@@ -80,7 +80,7 @@ tt {
     }
 
     span {
-      text-shadow: 0 0 1px var(--ps-neutral-1);
+      text-shadow: 0 0 1px var(--ps-table-row-text-shadow);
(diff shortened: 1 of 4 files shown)


webapp/sass/common.scss
```diff
@@ -80,7 +80,7 @@ tt {
     }
 
     span {
-      text-shadow: 0 0 1px var(--ps-neutral-1);
```
Rperry2174, line 83: @eh-am so apparently this affects the shadow on the tables... Why is this css located here instead of in `packages/pyroscope-flamegraph/src/ProfilerTable.module.scss` ? same with all other table related css code?

petethepig (approved):
LGTM

Issue #1282, opened 2022-07-18
Create tag-explorer page for analyzing tag breakdowns


Rperry2174:
### Goal: 
The idea of the explore page is to give a slightly “higher” level view of your data than why Pyroscope currently gives you.
I created a mockup of what I was thinking about for another page on the sidebar called “tag explorer”

### Current problem
For example, right now for our rideshare example, there is one particular region and one particular vehicle that have performance issues, but if we didn’t know that before looking at the example how could we figure out which vehicle and which region are the “problem” areas?

In current Pyroscope you have to click one by one through each region and then one by one through each vehicle and then possible combine region + vehicle in order to figure this out. The comparison page makes this slightly faster 2x but only because you can compare two things at once…

### Proposed solution 
![explore_page_02-01](https://user-images.githubusercontent.com/23323466/179627038-2c5a21b0-9d3b-4c43-8d75-63d0ff1ce125.png)

In this proposed “explore” page i’m proposing that we allow users to select a tag_key and see all the series associated with the corresponding tag values on one graph.

**Note: Lets NOT add the “where” dropdown for v1**

In this situation, if one region is consuming significantly more cpu, even if there are 30 regions, we would be able to see at first glance that 29 are normal and one series is high.

We could then dig further and look at the “descriptive statistics” for that series compared to the other series.

Finally, we could see the flamegraph associated with that series and if we want we can use the buttons to go to Single view, comparison view, or diff view.

### Implementation details
Dmitry added support for “group by” queries here: https://github.com/pyroscope-io/pyroscope/pull/1244

We can do group by, but not “where”. So for the first implementation we should not include the “where” dropdown.

### Descriptive statistics:
For V1 we can just stub these. 
![image](https://user-images.githubusercontent.com/23323466/179627147-de2ad4f9-ffb2-41bf-991e-9ddbf295c720.png)

For example we will eventually calculate all these column headers on the frontend. But for now we should just use something simple like `avg_samples` and just use that for all 9 columns. 

The more important functionality is to be able to click on one of the tags and have it show up in the flamegraph below. 

### Long term benefits or proposed solution
- This allows for the potential to do “alerts” based off of these statistics
- Opens the door / lays the foundation for more advanced analytics
- Once we have an endpoint for merging flamegraphs we could potentially do something like comparing a single flamegraph for <25% percentile to >75% percentile. Note: I’m considering each 10s “bar” it’s own data point…
- etc

### Reference
[...]

PR #1293, opened 2022-07-21
feat: create tag-explorer page for analyzing tag breakdowns

dogfrogfog:
#1282


dogfrogfog:
https://user-images.githubusercontent.com/47758224/180509601-06872079-c227-4fb1-9a1a-df5932ffdd06.mov

webapp/javascript/pages/ExploreView.module.scss
```diff
@@ -49,6 +49,19 @@
         background-color: var(--ps-green-highlight);
         color: var(--ps-tooltip-text);
       }
+
+      .tagName {
+        display: flex;
+        align-items: center;
+
+        .color {
```
Rperry2174, line 57: Lets call this something a little more specific like `tagColor`

webapp/javascript/pages/ExploreView.tsx
```diff
@@ -0,0 +1,304 @@
...
+  selectAppTags,
+  TagsState,
+  fetchExploreView,
+  fetchExploreViewProfile,
+} from '@webapp/redux/reducers/continuous';
+import { queryToAppName } from '@webapp/models/query';
+
+import styles from './ExploreView.module.scss';
+
+const timelineSeriesColors = [
```
Rperry2174, line 32: lets try using this array [here](https://github.com/pyroscope-io/pyroscope/blob/main/packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/colorPalette.ts#L22) (ideally we could import it somehow?)

eh-am (commented):
1. Let's call it `Explore Tags` page to make it more clear it's all about Tags!

2. Table colors are bit off
image
Compare that to the flamegraph table
image

3. If we not gonna do anything with these buttons for now, let's remove it
image

4. What should happen in this case when there's no tags?
Right now it simply doesn't work.
image

cc @Rperry2174 

5. Show a `no tags available` when there's no tags.
image

Example:
image

webapp/javascript/pages/ExploreView.tsx
```diff
@@ -0,0 +1,309 @@
...
+          <button>Export</button>
+          <button>Single</button>
+          <button>Comparison</button>
+          <button>Diff</button>
+        </div>
+      </div>
+      <table className={styles.exploreTable}>
+        <thead>
+          <tr>
+            <th>{groupsData[0]?.tagName === '*' ? 'App' : 'Tag'} name</th>
```
eh-am, line 214: What is this doing?

webapp/javascript/pages/ExploreView.tsx
```diff
@@ -0,0 +1,309 @@
...
+          </tr>
+        </thead>
+        <tbody>
+          {groupsData.map(({ tagName, color, data }) => {
+            const mean = calculateMean(data.samples);
+
+            return (
+              <tr
+                className={
+                  tagName === groupByTagValue ? styles.activeTagRow : ''
```
eh-am, line 233: Let's make this more clear

```tsx
const isTagSelected = (tag: string) => tagName === groupByTagValue
```

webapp/javascript/pages/ExploreView.tsx
```diff
@@ -0,0 +1,309 @@
...
+                }
+                onClick={() => handleGroupByTagValueChange(tagName)}
+                key={tagName}
+              >
+                <td>
+                  {tagName === '*' ? (
+                    appName
+                  ) : (
```
eh-am, line 241: Second time I see this and I don't know what `tagName === '*'` means. Let's create a helper function to deal with this case and add some comments there

webapp/javascript/pages/ExploreView.tsx
```diff
@@ -0,0 +1,309 @@
...
+  const tagKeys = Object.keys(tags.tags);
+
+  const handleClick = (e: ClickEvent) => {
+    handleTagChange(e.value);
+  };
+
+  return (
+    <div className={styles.header}>
+      <span className={styles.title}>
+        {appName.isJust ? appName.value : ''}
```
eh-am, line 289:
```suggestion
        {appName.unwrapOr('')}
```

Rperry2174:
Let's also fix spacing here so elements are evenly spaced out:
image

webapp/javascript/pages/ExploreView.tsx
```diff
@@ -0,0 +1,309 @@
...
+        <Dropdown
+          label="tags"
+          value={selectedTag ? `tag: ${selectedTag}` : 'select tag'}
+          onItemClick={handleClick}
+        >
+          {tagKeys.map((tagName) => (
+            <MenuItem key={tagName} value={tagName}>
+              {tagName}
+            </MenuItem>
+          ))}
```
eh-am, line 302: I already mentioned this, but let's handle the case where there's no tags available.

webapp/javascript/redux/reducers/continuous.ts
```diff
@@ -1,26 +1,69 @@
...
+      timeline: Timeline;
+      profile: Profile;
+    }
+  | {
+      type: 'reloading';
+      timeline: Timeline;
+      profile: Profile;
+    };
+
+type ExploreView =
```
eh-am, line 36: Why do we need a `timeline` field? From my understanding we only use data from the groups field.

webapp/javascript/services/render.ts
```diff
@@ -39,7 +45,7 @@ export async function renderSingle(
     z.object({ timeline: TimelineSchema })
   )
     .merge(z.object({ telemetry: z.object({}).passthrough().optional() }))
-    .merge(z.object({ groups: z.object({}).passthrough().optional() }))
+    .merge(z.object({ groups: GroupsSchema.optional() }))
```
eh-am, line 48: my understanding is that this case doesn't use `groups` data at all, since it's deal with by renderExplore, so do we need this?

eh-am (commented):
Added few comments code wise, but overall looks nice! I really liked how you managed to made the code consistent with what we already have style-wise :)

webapp/javascript/services/render.ts
dogfrogfog: correct, but i think its nice to have it here, because `groups` are part of the response, even though we don't use it 
https://github.com/pyroscope-io/pyroscope/pull/1244#discussion_r919350675

eh-am: I disagree exactly for those reasons, if it's not used, then it's just adding complexity for no reason. Everything that is in the code base must be there for a reason.

A prime example of this is golang and its clean interfaces (eg  [io.Writer](https://pkg.go.dev/io#Writer)), by reducing the API scope you plug whatever you want, as long as it fulfills the interface.

webapp/javascript/pages/ExploreView.tsx
dogfrogfog: added few comments, and refactored `*` comparison (not only in this component)

webapp/javascript/redux/reducers/continuous.ts
dogfrogfog: removed

Rperry2174:
one more request -- currently when you select the `group by` dropdown, nothing happens by default and it's not clear that you're supposed to be looking at a flamegraph. Let's fix that by defaulting to the first tag being selected in the table

image

Rperry2174:
Looks great... Although another idea here :)

Can we also make it so that if no tag is selected, then we highlight the "group by" dropdown with this orange border:
image

I think that will make it clearer to the user that they should stat by selecting a tag in that dropdown

Rperry2174 (approved):
lgtm!

PR #1300, opened 2022-07-22
chore(flamegraph/models): make it mandatory to handle all spyNames

eh-am:
Propagate `spyName` and narrow its type (now it's a union instead of a simple string).

This forces us to update `spyToRegex` whenever there's an update to the spyNames list.
For Example, I removed the `pyroscope-rs` and `unknown` cases, then sint started barking:

```
@pyroscope/flamegraph:   78:11  error  Switch is not exhaustive. Cases not matched: "pyroscope-rs" | "unknown"  @typescript-eslint/switch-exhaustiveness-check
```


Rperry2174 (approved):
Lgtm!

PR #1304, opened 2022-07-25
feat: concurrent storage put

kolesnikovae:
- [x] Make `storage.Put` to work concurrently
 - [x] Make sure cache `GetOrCreate` is thread-safe
 - [x] Add storage queue params
 - [ ] Document new params
 - [x] ~Build tags for long ad-hoc tests~ – It may make sense to introduce a build tag for long tests (ad-hoc). As for now I adjusted the stress test so that it would run <1m.
 - [x] ~Check if RW mutexes are good in out case~ – Experiments don't show improvements (See https://github.com/golang/go/issues/17973)
 
 The change introduces two configuration options for the storage queue:
 - `storage-queue-workers`
 - `storage-queue-size`

~On migration, especially if the deployment is large, storage throughput will potentially increase significantly, as there will be less dropped profiles, which poses some risks. We can set default parameters so that effectively writes happen serially (just one worker) to workaround this.~

We decided to disable concurrent writes by default, so that the change does not affect existing deployments on update. However, users still may tweak parameters if required.

/cc @petethepig


kolesnikovae:
@petethepig I have updated the PR so that it won't affect existing deployments: storage queue will have one worker by default, thus effectively doing writes synchronously. Please, give it a look

petethepig (approved):

PR #1314, opened 2022-07-27
feat(ebpf): pythonless, portable ebpf

korniltsev:
In it's current state this PR:
* removes python dependency
* we now compile BPF program once (CO:RE) — no longer compile BPF program each 10 seconds
* there's no longer need for linux headers to be present on the system

---

Have been running in our cluster for the last few days and everything seems to be working great — there's no memory leaks and CPU performance overhead is currently at around 0.25%.

TODOs:
* [x] make a separate command
* [x] make some parameters (like symbol cache sizes) configurable
* [x] add labels from k8s automatically
* [x] ~~use pprof as transport format (might make it more efficient once we add labels from k8s)~~ - decided to postpone, maybe reuse transporttrie, maybe something else


petethepig:
marked as ready so that I can get a docker image

pkg/agent/ebpfspy/symbols_linux.go
```diff
@@ -0,0 +1,86 @@
...
+	if res < 0 {
+		if symbol.offset > 0 {
+			return "", uint64(symbol.offset), C.GoString(symbol.module)
+		}
+		return "", addr, ""
+	}
+	if pid == 0 {
+		return C.GoString(symbol.name), uint64(symbol.offset), C.GoString(symbol.module)
+	} else {
+		defer C.bcc_symbol_free_demangle_name(symbolC)
```
kolesnikovae, line 75: I'm not sure but [the doc says](https://github.com/iovisor/bcc/blob/f6c4b4c618a1a471fdfdd8bc852578f383793b63/src/cc/bcc_syms.h#L60) that `bcc_symbol_free_demangle_name` should be called after getting results from `bcc_symcache_resolve`, however we do this only in this code path

pkg/agent/ebpfspy/symbols_linux.go
```diff
@@ -0,0 +1,86 @@
...
+	if cache, ok := sc.cachePerPid[pid]; ok {
+		return cache
+	}
+	pidC := C.int(pid)
+	if pid == 0 {
+		pidC = C.int(-1)
+	}
+	symbolOpt := C.struct_bcc_symbol_option{use_symbol_type: C.uint(1 << elf.STT_FUNC)}
+	symbolOptC := (*C.struct_bcc_symbol_option)(unsafe.Pointer(&symbolOpt))
+	cache := C.bcc_symcache_new(pidC, symbolOptC)
```
kolesnikovae, line 45: I wonder how this ownership is managed. Shouldn't we free it explicitly with [bcc_free_symcache](https://github.com/iovisor/bcc/blob/f6c4b4c618a1a471fdfdd8bc852578f383793b63/src/cc/bcc_syms.h#L57-L58) at some point and delete the entry from Go map?

korniltsev: I agree we should call `bcc_free_symcache` and delete it from the map
- when the ebpfspy is stopping
- when pid is dead
- maybe even add some lru cache for some 
I suspect this unfreed cache may be the cause of todays memory consumption spike

pkg/agent/ebpfspy/symbols_linux.go
korniltsev: yes, I agree 
I will call it right after bcc_symcache_resolve

pkg/agent/ebpfspy/symbols_linux.go
kolesnikovae: An LRU is great idea, I like it

petethepig (changes requested):
Can we get `pkg/agent/ebpfspy/bpf/vmlinux.h` from somewhere during image building process? We shouldn't commit it to the repo.

Same with `profile.bpf.o` file

pkg/agent/ebpfspy/bpf/profile.bpf.c
```diff
@@ -0,0 +1,60 @@
...
+	val = bpf_map_lookup_elem(&counts, &key);
+	if (val)
+		(*val)++;
+	else
+		bpf_map_update_elem(&counts, &key, &one, BPF_NOEXIST);
+	return 0;
+}
+
+
+char _license[] SEC("license") = "GPL"; //todo
```
petethepig, line 60: Please set your code editor to finish files with new lines, see our Dev guide: https://pyroscope.io/docs/developer-guide/#go

pkg/config/config.go
```diff
@@ -325,6 +326,28 @@ type Connect struct {
...
+	ServerAddress          string        `def:"http://localhost:4040" desc:"address of the pyroscope server" mapstructure:"server-address"`
+	AuthToken              string        `def:"" desc:"authorization token used to upload profiling data" mapstructure:"auth-token"`
+	UpstreamThreads        int           `def:"4" desc:"number of upload threads" mapstructure:"upstream-threads"`
+	UpstreamRequestTimeout time.Duration `def:"10s" desc:"profile upload timeout" mapstructure:"upstream-request-timeout"`
+
+	Tags map[string]string `name:"tag" def:"" desc:"tag in key=value form. The flag may be specified multiple times" mapstructure:"tags"`
+
+	Pid                int    `def:"-1" desc:"PID of the process you want to profile. Pass -1 to profile the whole system" mapstructure:"pid"`
+	DetectSubprocesses bool   `def:"false" desc:"makes pyroscope keep track of and profile subprocesses of the main process" mapstructure:"detect-subprocesses"`
+	SymbolCacheSize    int    `def:"256" desc:"todo" mapstructure:"symbol-cache-size"`
```
petethepig, line 347:
```suggestion
	SymbolCacheSize    int    `def:"256" desc:"symbol cache in megabytes" mapstructure:"symbol-cache-size"`
```

korniltsev: It is not megabytes it is number of processes to keep symbols parsed in memory

cmd/pyroscope/command/ebpfspy.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/pyroscope-io/pyroscope/pkg/cli"
+	"github.com/pyroscope-io/pyroscope/pkg/config"
+	"github.com/pyroscope-io/pyroscope/pkg/exec"
+)
+
+func newEBPFSpyCmd(cfg *config.EBPFSpy) *cobra.Command {
+	vpr := newViper()
+	connectCmd := &cobra.Command{
+		Use:   "ebpfspy [flags]",
+		Short: "todo",
```
petethepig, line 17: pls add a description

pkg/config/config.go
```diff
@@ -325,6 +326,28 @@ type Connect struct {
 	Pid int `def:"0" desc:"PID of the process you want to profile. Pass -1 to profile the whole system (only supported by ebpfspy)" mapstructure:"pid"`
 }
 
+type EBPFSpy struct {
```
petethepig, line 329:
```suggestion
type EBPF struct {
```

pkg/config/config.go
petethepig:
```suggestion
	SymbolCacheSize    int    `def:"256" desc:"max size of symbols cache (1 entry per process)" mapstructure:"symbol-cache-size"`
```

petethepig (changes requested):
we should make the command be `pyroscope ebpf`. ebpfspy is more obscure, and long term we're trying to get rid of this whole "spy" suffix.

petethepig (approved):
Looks good!

PR #1318, opened 2022-07-28
feat: configurable disk alert threshold

kolesnikovae:
Currently, we have a hardcoded threshold 512MB for the low free disk space alert. The change makes this value configurable (`storage-disk-threshold` option), and instead of an absolute value it takes a percent of available disk space (disk size), which defaults to 5%.

Essentially, the value also serves as a space quota size, that is reserved and not used by storage: once server hits the threshold, storage writes start being dropped. This should allow us to prevent cases when pyroscope server can't start because of BadgerDB failing to acquire the lock file due to the lack of free space.

Although it is not entirely clear how we get into this situation with a hardcoded value, my best guess is that this might be related to how BadgerDB deletes data: effectively, garbage collection rewrites value log files temorary consuming more disk space.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -112,6 +112,8 @@ type Server struct {
 	CacheEvictThreshold float64 `def:"0.25" desc:"percentage of memory at which cache evictions start" mapstructure:"cache-evict-threshold"`
 	CacheEvictVolume    float64 `def:"0.33" desc:"percentage of cache that is evicted per eviction run" mapstructure:"cache-evict-volume"`
 
+	DiskThreshold float64 `def:"5" desc:"percentage of available disk space at which ingestion requests are discarded. Defaults to 5%. Set 0 to disable" mapstructure:"storage-disk-threshold"`
+
 	Database Database `mapstructure:"database"`
 
 	// TODO: I don't think a lot of people will change these values.
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -104,15 +104,15 @@ type Server struct {
 	LogLevel       string `def:"info" desc:"log level: debug|info|warn|error" mapstructure:"log-level"`
 	BadgerLogLevel string `def:"error" desc:"log level: debug|info|warn|error" mapstructure:"badger-log-level"`
 
-	StoragePath     string `def:"<installPrefix>/var/lib/pyroscope" desc:"directory where pyroscope stores profiling data" mapstructure:"storage-path"`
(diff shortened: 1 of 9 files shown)


pkg/util/disk/usage_test.go
```diff
@@ -9,15 +9,29 @@ import (
...
+			It("returns non-zero Total", func() {
+				Expect(u.Total).To(BeNumerically(">", 0))
+			})
+
+			It("returns non-zero Available", func() {
+				Expect(u.Available).To(BeNumerically(">", 0))
+			})
+
+			It("returns Available < Total", func() {
+				Expect(u.Available).To(BeNumerically("<", u.Total))
```
kolesnikovae, line 34: Actually, I'm not 100% if these are reasonable assertions

pkg/config/config.go
```diff
@@ -104,7 +104,9 @@ type Server struct {
 	LogLevel       string `def:"info" desc:"log level: debug|info|warn|error" mapstructure:"log-level"`
 	BadgerLogLevel string `def:"error" desc:"log level: debug|info|warn|error" mapstructure:"badger-log-level"`
 
-	StoragePath     string `def:"<installPrefix>/var/lib/pyroscope" desc:"directory where pyroscope stores profiling data" mapstructure:"storage-path"`
+	StoragePath          string  `def:"<installPrefix>/var/lib/pyroscope" desc:"directory where pyroscope stores profiling data" mapstructure:"storage-path"`
+	StorageDiskThreshold float64 `def:"5" desc:"percentage of available disk space at which ingestion requests are discarded. Defaults to 5%. Set 0 to disable" mapstructure:"storage-disk-threshold"`
```
petethepig, line 108: I would call it something a little more explicit. Right now it's unclear from the name what this parameter does.

Maybe `MinFreeSpacePercentage`?

petethepig (changes requested):
I have two comments:
* if you have less than 10GB of space the minimum space is going to be set pretty low, e.g if you have 5GB disk it's going to be set at 256MB — this is way to low for the system to operate. So I propose that we add another hard limit (e.g say 1GB), and then we take something like `max(1GB, 5%*total space)`
* the naming of the parameter could be better imo, mentioned it in the comments

kolesnikovae:
> I have two comments:
>
> if you have less than 10GB of space the minimum space is going to be set pretty low, e.g if you have 5GB disk it's going to be set at 256MB — this is way to low for the system to operate. So I propose that we add another hard limit (e.g say 1GB), and then we take something like max(1GB, 5%*total space)
> the naming of the parameter could be better imo, mentioned it in the comments

Fixed

petethepig (approved):

PR #1321, opened 2022-07-29
chore: allows empty names in names array

petethepig:
I do see empty names in our profiles every once in a while. While those might be caused by bugs, I could also see people uploading stacktraces with empty names, and in that case that wouldn't be a bug and we should probably still render them.

Therefore I'm relaxing this rule. @eh-am what do you think?

### DIFF
--- a/packages/pyroscope-models/src/profile.ts
+++ b/packages/pyroscope-models/src/profile.ts
@@ -1,7 +1,7 @@
 import { z } from 'zod';
 
 export const FlamebearerSchema = z.object({
-  names: z.array(z.string().nonempty()),
+  names: z.array(z.string()),
   levels: z.array(z.array(z.number())),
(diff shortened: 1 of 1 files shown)


eh-am:
I think there are certain places that do logic based on the stack name (eg the select thing, or the heuristic to keep a node in focus after timeline changes). Instead of tracking this down I would prefer to have a default value, like 'unknown'? 

> see people uploading stacktraces with empty names
What would this mean from a use case perspective? Like anonymous functions?

petethepig:
> What would this mean from a use case perspective? Like anonymous functions?

Honestly, it's likely bugs. But then having this check or not is the difference between seeing a slightly buggy flamegraph vs not seeing anything at all and I'd rather see something.

Gonna merge this for now and later maybe we can come up with a better long term solution

PR #1323, opened 2022-07-29
feat(webapp): update timeline appearance and refactor flot plugins

pavelpashkovsky:
Short summary on this PR
We don't edit any code in Flot anymore. But there are cases when new functionality based on some built-in plugins. In case of upgrading time selection functionality I extracted `react-flot/flot/jquery.flot.selection.min` into a file `./TimelineChartSelection`, where I customised existing logic in accordance with task requirements. It's important to note, that there are 2 types of timeline selection: `double` and `single`. All customisation was applied only for `single` type of selection. `Double` works in old fashion way (using built-in Flot selection, not plugin-built one).

image

image
- - -
I also extended in a bit TimelineChartWrapper API. Now you can set colors for selection boundaries and overlay. Adding colors depending on color mode is in progress

image


pavelpashkovsky:
looks like selection boundaries and overlay are too much pale.

pavelpashkovsky:
@Rperry2174, I can be wrong, but I've just remembered that you wanted to remove these captions several months ago? Should I do this?

image

webapp/javascript/pages/ContinuousComparisonView.tsx
```diff
@@ -95,15 +99,26 @@ function ComparisonApp() {
               dispatch(actions.setFromAndUntil({ from, until }));
             }}
             markings={{
-              left: { from: leftFrom, to: leftUntil, color: leftColor },
-              right: { from: rightFrom, to: rightUntil, color: rightColor },
+              left: {
+                from: leftFrom,
+                to: leftUntil,
+                color: leftColor,
+                overlayColor: leftColor.alpha(0.3),
```
Rperry2174, line 106: What does `overLayColor` correspond to? is it this 
image

Rperry2174:
> @Rperry2174, I can be wrong, but I've just remembered that you wanted to remove these captions several months ago? Should I do this?

Yeah we can remove these now. 

Given the current way that the code is structured what would be the steps to be able to control the selection like this?
time_selector_mockup_00_3

webapp/javascript/pages/ContinuousComparisonView.tsx
pavelpashkovsky: @Rperry2174 you are right, it's visible semi transparent area between selection boundaries, you can set it's color by changing `overlayColor ` prop

Rperry2174:
almost there! I definitely like these selectors better.. Can we make this icon an "open" hand:
image

pavelpashkovsky:
@Rperry2174, I changed resize cursor to open hand, please re/create-server

Rperry2174 (approved):
LGTM! Nice work @pavelpashkovsky

PR #1324, opened 2022-07-29
feat: tag explorer page v2

dogfrogfog:
https://github.com/pyroscope-io/pyroscope/issues/1319

change base to `main` after [feature/app-explore-page](https://github.com/pyroscope-io/pyroscope/tree/feature/app-explore-page) merge

todo:
- [x] when a tag is selected in the table, the other series on the graph should "dim" themselves
- [x] We should add a "where" clause after group by which filters both series and results in the table: ride-sharing-app.cpu grouped by tag: region where region=*.north
- [x] we should add buttons which allow you to hop to comparison or diff page
- [x] we should make the "export to flamegraph.com" include tag information in the data sent there
- [ ] we could add a button to export the table (maybe)
- [ ] we should potentially consider other / addition / better descriptive statistics to add


dogfrogfog:
server responds with `group: {}`, is that expected behavior ? cc @Rperry2174 @eh-am

PR #1325, opened 2022-07-29
chore(frontend): cache bust css

eh-am:
Since I run with dev tools open and cache disabled by default, I never had this issue. I only came to notice while seeing @pavelpashkovsky's screen.

Basically when running locally we generate files with constant file names, so that we don't calculate a hash, which is slow. The downside is that the browser will cache that file.

I added a random query string at the end of css files (only in local dev), so that it invalidates the browser's cache.


eh-am:
Actually I can't reproduce the original issue anymore

PR #1327, opened 2022-07-29
feat: upload any arbitrary data (collapsed/pprof/json) via adhoc ui

eh-am:
It uses the same api as adhoc (and therefore flamegraph.com)

still a WIP


petethepig (approved):
Looks good!

Rperry2174:
Not sure if this is intentional or not, but only two export options here:
image

Rperry2174:
Also we definitely gotta fix the styling of the diff page, but I think that's out of scope for this PR:

https://user-images.githubusercontent.com/23323466/182502978-479705e9-5663-439d-900c-df32acc81c87.mov

Rperry2174 (approved):
Nothing stuck out as problematic in the code and the code look much cleaner now 👍  

Tested it and everything looked good there

eh-am:
>Not sure if this is intentional or not, but only two export options here:
I don't think this ever worked (this is from demo-dev before merging this PR):
image

Created an issue though, can work on it next
https://github.com/pyroscope-io/pyroscope/issues/1342

PR #1329, opened 2022-08-01
feat: tag explorer view v2

dogfrogfog:
https://github.com/pyroscope-io/pyroscope/issues/1319

!`groupBy` feature should be added to the cloud server! to make page work

todo:
- [x] when a tag is selected in the table, the other series on the graph should "dim" themselves
- [x] we should add a "where" clause after group by which filters both series and results in the table: ride-sharing-app.cpu grouped by tag: region where region=*.north
- [x] we should add buttons which allow you to hop to comparison or diff page
- [x] we should make the "export to flamegraph.com" include tag information in the data sent there
we could add a button to export the table (maybe)
we should potentially consider other / addition / better descriptive statistics to add


Rperry2174:
when I select all and then hit refresh it jumps back to the previously selected group by (seems like "All" isn't getting put into the browser)

https://user-images.githubusercontent.com/23323466/182467016-4bf96052-b613-4c9e-9457-ccb56fedb94c.mov

Rperry2174:
also lets make the color square here of fixed size so that when the screen is small it doesn't shrink like this:
image

Rperry2174:
lets make the page default to "all" when a user initially selects a groupby value (notice how the page jumps currently from all to ap-south) 

https://user-images.githubusercontent.com/23323466/182647193-ff8cd040-8f93-4bc9-88b8-ea22b6004ce0.mov

also lets remove the buttons again.. We'll add them back in a separate Pr after I have time to mock up what it should look like

Rperry2174:
For the buttons lets do the following:
- Single: Goes to the single view page except with the current tag selected
- comparison: creates a modal that looks like this which then goes to the comparison page with the two selected tags preselcted on that page
- Diff: same as comparison page except with diff

We should make this new modal a new component:
comparison_button_00-01

Rperry2174:
Looks pretty good a couple of comments:

1. "Baseline" and "comparison" should be in the center, but lets left align the text in the selectable options (the tags)
2. "BASELINE" -> "Baseline" 
3. "COMPARISON" -> "Comparison"
4. Lets change the title to be Normal cased instead of all capitalized: SELECT TAG FOR COMPARISON --> "Select Tags for Comparison" (note make tagS plural)

Screen Shot 2022-08-04 at 8 20 55 PM

For the dropdowns themselves, the dropdowns that we added in this PR (and the button) for "Single", "comparison", and "diff" don't match the styling of the rest of the dropdowns on the page. For example the size is different and there is no hover effect like theother dropdown in this video. 

Can you make the styling of the new dropdowns and the new button look the same as the other dropdown?

https://user-images.githubusercontent.com/23323466/182995463-9dc64697-6ba9-437b-a13b-c2920df9393c.mov

Rperry2174 (approved):
LGTM!

Issue #573, opened 2021-11-30
Make Trie decoding more efficient


kolesnikovae:
The current `trasporttrie.IterateRaw` implementation uses the `io.CopyN` method which allocates 32K buffers by default for every tree leaf (a __stack__). This should be either revised to not use copying at all, or reuse the buffers.

It’s hard to estimate how helpful this change can be, but it is clear that it does not require any effort and will improve ingestion HTTP handler performance to some degree.

dleviminzi:
Hello, I was looking at this issue and I am not sure that it is an issue. This issue is pretty old so I imagine that what I've learned was not the case when it was opened. 

Anyway, the `io.CopyN` creates a `LimitReader` from `nameLen` and `br` and then passes that along with `b` down until `copyBuffer`. At that point `copyBuffer` uses the `ReadFrom` method instead of allocating a buffer. 

So, unless I misunderstood something, this ```io.CopyN(b, br, int64(nameLen))``` essentially ends up calling ```b.ReadFrom(io.LimitReader(br, int64(nameLen)))```.  It could be changed to the latter on line 73. It might be more direct and/or clear, but idk if you'd consider that a worthwhile change.

kolesnikovae:
Hello @dleviminzi, thank you for looking at this! I think that your understanding is correct – the buffer is not being allocated: I missed the fact that `Copy` has a type-specific behaviour. I like your idea replacing the `CopyN` call with `b.ReadFrom` as this way is more explicit.

What caught my attention is that the function expects the caller to provide a byte slice as a temporary buffer. This is not bad per se, but it doesn't seem we use it: the only [caller](https://github.com/pyroscope-io/pyroscope/blob/main/pkg/convert/profile/profile.go#L41) I see creates a new slice every time. Initial imlementation was using a [buffer pool](https://github.com/valyala/bytebufferpool), IIRC. I think that we should either allocate the slice within the `IterateRaw` method, or return the buffer pool back. However, this is a separate issue.

Honestly, from the performance perspective I can hardly say these changes can have significant impact, but they can definitely improve readability and maintainability. Your contribution would be very appreciated!

dleviminzi:
Ok, I will create a pr with the change. It will make the call stack easier to read!

kolesnikovae:
Great! Thank you so much!

PR #1335, opened 2022-08-02
call ReadFrom directly instead of CopyN

dleviminzi:
This change stems from the conversation on #573


kolesnikovae (approved):
@dleviminzi thank you for the contribution!

PR #1339, opened 2022-08-03
chore(flamegraph): reduce barrier for usage

pavelpashkovsky:
Edited by Eduardo:

* remove `true-myth` frmo dependencies, since we publish the flamegraph bundled, there should be no need to pull that dependency
* downgrade typescript usage, [`import type` was introduced in typescript 3.8](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html), which prohibits usage from users in older versions

### DIFF
--- a/packages/pyroscope-flamegraph/package.json
+++ b/packages/pyroscope-flamegraph/package.json
@@ -23,12 +23,10 @@
   },
   "peerDependencies": {
     "react": ">=16.14.0",
-    "react-dom": ">=16.14.0"
+    "react-dom": ">=16.14.0",
+    "true-myth": "^5.1.2"
   },
   "devDependencies": {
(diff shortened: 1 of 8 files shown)


eh-am:
I wanna test if we can't just tell people to somehow overwrite the type files, since I don't really like downgrading our usage of typescript features.

eh-am:
There's this https://github.com/sandersn/downlevel-dts

Rperry2174 (approved):
LGTM added another issue to follow up at some point: 
https://github.com/pyroscope-io/pyroscope/issues/1346

pavelpashkovsky:
CLOSED, MOVED to https://github.com/pyroscope-io/pyroscope/pull/1348

PR #1343, opened 2022-08-03
chore: add a script to obfuscate json files

eh-am:
Sometimes we need to post flamegraph json files (for reproducibility, example https://github.com/pyroscope-io/pyroscope/pull/1341#issuecomment-1204048760), but we don't want to expose function names etc. This simple scripts just replaces these somewhat confidential data (app name and node names) for random strings.


Rperry2174 (approved):
💫

PR #1344, opened 2022-08-03
chore(flamegraph): don't show <TBD> for unknown case in tooltip footer

eh-am:
Don't know if this is a case that can ever happen in the wild, but nevertheless it's better to not show anything instead of `TBD

https://user-images.githubusercontent.com/6951209/182663231-197c3f9a-86d5-4c66-99b3-301eade7a153.mp4

`


Rperry2174 (approved):
Agreed makes sense

PR #1345, opened 2022-08-03
chore(models): default spyName to 'unknown' when not detected

eh-am:
When a non officially supported `spyName` is passed, default to `unknown`.


Rperry2174 (approved):
💪

Issue #1346, opened 2022-08-03
Add warning / mechanism for supporting older Typescript versions for Flamegraph


Rperry2174:
In #1339 we made some changes to be able to support older typescript version with our flamegraph component (particularly for [Jaeger](https://github.com/jaegertracing/jaeger-ui/issues/390) usage in this case).. 

We should come up with a way to either alert us in CI if we use a Typescript feature that is not supported in typescript 3 or a way to convert typescript 4 to typescript 3 

This may be an option: https://github.com/sandersn/downlevel-dts

eh-am:
Implemented in https://github.com/pyroscope-io/pyroscope/pull/1348

PR #1347, opened 2022-08-03
chore(flamegraph): downlevel ts types to 3.4

eh-am:
This is so that we ship flamegraph with typings for typescript 3.4 (to make it easier for end users with non updated ts version), but still allows us to use newer features internally.

Related https://github.com/pyroscope-io/pyroscope/pull/1339 also #1346

UPD by @pavelpashkovsky : https://github.com/pyroscope-io/pyroscope/pull/1339 closed and moved to https://github.com/pyroscope-io/pyroscope/pull/1348 (All changes from this PR implemented in it)


eh-am:
Superseeded by https://github.com/pyroscope-io/pyroscope/pull/1348

PR #1351, opened 2022-08-04
feat(webapp): persist uploaded data via adhoc ui

eh-am:
Closes https://github.com/pyroscope-io/pyroscope/issues/1317

https://user-images.githubusercontent.com/6951209/182932596-653c8400-74b4-415f-be9e-5dbcedb1c301.mp4

The frontend can be simplified/improved:
* have a single path of loading a profile (from file list)
* auto select profile after uploading

### DIFF
--- a/pkg/adhoc/cli.go
+++ b/pkg/adhoc/cli.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/fatih/color"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/pyroscope-io/pyroscope/pkg/adhoc/writer"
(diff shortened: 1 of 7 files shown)


Rperry2174 (approved):
Looks good to me.. Only thing thats a little weird is that you upload a `.pb.gz` and then the persisted data is a `.json`, but I think thats probably fine

eh-am:
@Rperry2174 that's bc we convert that data before storing in adhoc.
We could however store as the original format and do the conversion at retrieval time.

The disadvantage of converting the file is that we may lose some data, example is how only pick up the first profile (in a pprof with multiple profiles)

What do you think?

PR #1352, opened 2022-08-04
Feat/adhoc UI

eh-am:
Depends on https://github.com/pyroscope-io/pyroscope/pull/1351

Keep in mind that the look and few here is horrible, this is strictly about functionality.

https://user-images.githubusercontent.com/6951209/182949514-56da59f6-a721-442e-b58e-5e8a8cc024d4.mp4
* After uploading, select the file automatically
* After uploading, change tabs automatically
* Allow uploading in the adhoc diff page

### DIFF
--- a/package.json
+++ b/package.json
@@ -188,6 +188,7 @@
     "d3-array": "^3.1.1",
     "d3-scale": "^4.0.2",
     "d3-time": "^3.0.0",
+    "downlevel-dts": "^0.10.0",
     "date-fns": "^2.27.0",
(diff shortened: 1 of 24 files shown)


Rperry2174:
I feel like maybe instead of the button saying "upload" it should say "save"... To me that would make it clearer that you're really kind of "save"ing it locally rather than sending it somewhere. 

Also with the styling I would be ok merging this as is and doing styling separately

Rperry2174 (approved):
🔥

PR #1356, opened 2022-08-06
feat: add gzip content-encoding support

korniltsev:
- add gzip decoding at ingest.go based on `Content-Encoding` header
- add gzip encoding to Remote

### DIFF
--- a/pkg/agent/upstream/remote/remote.go
+++ b/pkg/agent/upstream/remote/remote.go
@@ -133,7 +133,7 @@ func (r *Remote) uploadProfile(j *upstream.UploadJob) error {
 	if r.cfg.AuthToken != "" {
 		request.Header.Set("Authorization", "Bearer "+r.cfg.AuthToken)
 	}
-	if r.cfg.UpstreamRequestCompression == "gzip" {
+	if ce != "" {
 		request.Header.Set("Content-Encoding", ce)
 	}
 
--- a/pkg/agent/upstream/remote/remote.go
+++ b/pkg/agent/upstream/remote/remote.go
@@ -2,6 +2,7 @@ package remote
 
 import (
 	"bytes"
+	"compress/gzip"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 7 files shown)


petethepig:
Closing for now because:
* on the client side it's not super clear if gzipping transporttrie trees comes with much of an advantage
* on the server side I merged this PR https://github.com/pyroscope-io/pyroscope/pull/1357 and i think it's a slightly better solution

I think we should experiment with gzip at some point and maybe then we can reopen this.

pkg/agent/upstream/remote/remote.go
```diff
@@ -118,7 +120,11 @@ func (r *Remote) uploadProfile(j *upstream.UploadJob) error {
 
 	r.Logger.Debugf("uploading at %s", u.String())
 	// new a request for the job
-	request, err := http.NewRequest("POST", u.String(), bytes.NewReader(j.Trie.Bytes()))
+	ce, body, err := r.compressBody(j.Trie.Bytes())
```
petethepig, line 123: General go comment:

I think something like [this](https://gist.github.com/tomcatzh/cf8040820962e0f8c04700eb3b2f26be) would be slightly better because it would use less memory due to streaming, but I'm not 100% sure maybe there are downsides / other considerations I'm missing.

@kolesnikovae what do you think?

korniltsev:
It is gzipping transportrie really good on client side
For example with  [this profiles](https://flamegraph.com/share/72cc769f-16ad-11ed-8e1e-8eb7188980dc) & ebpspy I get ~9x
```
gzipping 317116 => 36456   
gzipping 384663 => 42508   
gzipping 426227 => 48235   
...
```
Let's play with it in cloudstorage?

pkg/agent/upstream/remote/remote.go
kolesnikovae: A gzip reader fits perfectly, I think. Also, can't find any downsides, but the example snippet may need some better error handling 

BTW, if `Trie` satisfied `io.WriterTo`, we'd be able to make the streaming even more efficient:
```go
func NewGzipReader(src io.WriteTo) io.Reader
```

PR #1373, opened 2022-08-09
fix(webapp): yarn storybook can't resolve module

pavelpashkovsky:


eh-am:
I solved in a different way here https://github.com/pyroscope-io/pyroscope/pull/1375
But your is good too, merging it

eh-am (approved):

PR #1374, opened 2022-08-09
fix(webapp): Fit mode styles

pavelpashkovsky:
- Fixed FitMode select height in web app
image

- Removed `webapp/javascript/components/ProfilerHeader.module.css` as unused

before:
image
after:
image


eh-am:
Should be superseeded by https://github.com/pyroscope-io/pyroscope/pull/1388

PR #1375, opened 2022-08-09
chore: fix storybook/flamegraph:dev command

eh-am:
As raised by @pavelpashkovsky 

* Fixes `dev:flamegraph` command. I thought it was not necessary anymore (since the webapp now builds the flamegraph component directly), but it's still needed if you want to develop the flamegraph component as a standalone thing, most likely consuming via the storybook.
* Also fixes storybook in general, since some components import `@webapp/ui`, which is not aliased. At some point we should move all UI components to its own component library and use that instead, but as of now I just added the alias.


Rperry2174 (approved):
👍

PR #1384, opened 2022-08-11
feat: support high number of series in explore view timeline

dogfrogfog:
https://github.com/pyroscope-io/pyroscope/issues/1364

### DIFF
--- a/webapp/javascript/pages/TagExplorerView.tsx
+++ b/webapp/javascript/pages/TagExplorerView.tsx
@@ -39,6 +39,13 @@ import { PAGES } from './constants';
 
 import styles from './TagExplorerView.module.scss';
 
+const getTimelineColorFromIndex = (index: number): Color => {
+  const maxIndex = DefaultPalette.colors.length - 1;
+  const colorPaletteIndex = index > maxIndex ? index % maxIndex : index;
+
+  return Color(DefaultPalette.colors[colorPaletteIndex]);
+};
+
 function TagExplorerView() {
   const { offset } = useTimeZone();
   const { colorMode } = useColorMode();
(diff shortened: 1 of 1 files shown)


webapp/javascript/pages/TagExplorerView.tsx
```diff
@@ -39,6 +39,13 @@ import { PAGES } from './constants';
 
 import styles from './TagExplorerView.module.scss';
 
+const getTimelineColorFromIndex = (index: number): Color => {
+  const maxIndex = DefaultPalette.colors.length - 1;
+  const colorPaletteIndex = index > maxIndex ? index % maxIndex : index;
```
eh-am, line 44:
```suggestion
  const colorPaletteIndex = index % maxIndex;
```

webapp/javascript/pages/TagExplorerView.tsx
```diff
@@ -39,6 +39,13 @@ import { PAGES } from './constants';
 
 import styles from './TagExplorerView.module.scss';
 
+const getTimelineColorFromIndex = (index: number): Color => {
```
eh-am, line 42:
```suggestion
const getTimelineColor = (index: number): Color => {
```
Since a) we use typescript so Intellisense can be assumed and b) there's already a parameter named index

eh-am: also let's receive a Palette declaratively

eh-am (changes requested):

webapp/javascript/pages/TagExplorerView.tsx
dogfrogfog: im i got it right or what exactly do you mean by that ?

Rperry2174: I think he means lets pass the Palette into the function

Rperry2174: and also I think he means to remove `Index` from the funciton name since it's implied by the fact that you pass an index into the function

Rperry2174 (approved):
LGTM 🔥

PR #1387, opened 2022-08-11
fix: Make trace units be time based

Rperry2174:
Fixes units for traces which should always be time as mentioned in [Jaeger repo](https://github.com/jaegertracing/jaeger-ui/pull/977#discussion_r943615134)

Before:
image

After:
image


packages/pyroscope-flamegraph/src/format/format.spec.ts
```diff
@@ -37,6 +37,17 @@ describe('format', () => {
         expect(df.format(8000, 100)).toBe('80.00 seconds');
       });
 
+      it('correctly formats trace samples', () => {
+        const df = getFormatter(80, 2, 'trace_samples');
+
+        expect(df.format(0, 100)).toBe('0.00 seconds');
+        expect(df.format(0.001, 100)).toBe('< 0.01 seconds');
```
yurishkuro, line 44: Does this support sub-second units? Always keeping seconds as the unit can lose a lot of precision, especially for some short traces. Jaeger UI automatically changes the units between seconds, ms, μs.

https://github.com/jaegertracing/jaeger-ui/blob/e2ae62af03e820f261713cb32ba0290c734e34f8/packages/jaeger-ui/src/utils/date.tsx#L109

yurishkuro: image

Rperry2174: makes sense. Created an issue for this:
https://github.com/pyroscope-io/pyroscope/issues/1393

PR #1388, opened 2022-08-11
fix(flamegraph): fix its styling

eh-am:
* Makes `@pyroscope/flamegraph` and `webapp` have the same styling rules (mostly the normalization ones)
* Publish storybook under github action artifacts, so that we can visually inspect the components
* Inline `sanitize.css`, for some reason I could make it so that a nested `import ~sanitize.css` would be inlined (tried with postcss plugin and nothing)

---

The original problem is that the webapp and flamegraph (standalone) shipped different sets of CSS. Which means it would look just fine in webapp (which is what most people would be running), then in the standalone version (imported via `@pyroscope/flamegraph` package), it would look slightly off.

Turns out the difference is that the flamegraph was not shipping the styles from normalize.css.
However, we can't just ship those styles, since they are global and they may affect end user's style (imagine I embed the flamegraph into my component, and then somehow it overrides my app's button!).

Therefore we need to scope those rules.

However, if we scope those rules, then they start to override the ones from css modules. The reason is simple.
Consider a sanitize.css rule for `table`:
```css
table {
  border-collapse: collapse;
}
```
If we scope under `.pyroscope` it will become:
```css
.pyroscope table {
  border-collapse: collapse;
}
```

Then let's say we have a css modules rule, which normally would override the one from sanitize css (since class has higher precedence over a simple tag)

```css
.oiJX5Zp1+W8aJOcWSU-baQ== {
  border-collapse: separate;
}
```
In our case, `.pyroscope table` is higher than `.oiJX5Zp1+W8aJOcWSU-baQ==`, therefore our css module style would never be applied.

To "solve" this I ended up using a "custom" element (not confuse with [real custom elements](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements)), here called `pyro-flamegraph`.

That way, `pyro-flamegraph table` still has a lower specificity than `.oiJX5Zp1+W8aJOcWSU-baQ==`, restoring the original behavior of having classes generated by CSS modules to overwrite the styles from the reset.

---
# Disclaimer
This may break for people that assume the flamegraph component has a certain shape, eg in snapshot tests.
Since we publish the component as < 1.0.0, I won't publish as a breaking change (even though it is).


.github/workflows/storybook.yml
```diff
@@ -0,0 +1,38 @@
+name: Storybook
```
petethepig, line 1: storybook is back!

petethepig (approved):
Interesting solution! LGTM

PR #1392, opened 2022-08-12
chore(webapp): Improve UI of Adhoc drag and drop [WIP]

pavelpashkovsky:
Refactor ADHOC d-n-d file uploader UI described here: https://github.com/pyroscope-io/pyroscope/issues/1334
demo coming soon...


Rperry2174:
For the drag-and-drop background colors lets:

1. For default color lets make  `--ps-upload-profile-area-bg` use the same color as `--ps-dropdown-background`
2. After a file has been added lets make the button use these disabled styles similar to reset here:
image

pavelpashkovsky:
@Rperry2174, @eh-am 
Hey guys, can you please test the [demo](https://pr-1392.pyroscope.io/adhoc-single?query=rideshare-app-golang.alloc_objects%7B%7D) of this feature and share your thought about couple of points:
1) Is there any issues with UI design on this feature (proper colors, elements size, appearance in light mode etc.)? 
2) When you put non-json file for d-n-d area, there should by spyname and units selectors. I pass these params to `upload` function, similar to reference logic of `Upload2()` function. Seems like server function which works on `/api/adhoc/v1/upload` doesn't handle these params. You can easily reproduce it on [demo](https://pr-1392.pyroscope.io/adhoc-single?query=rideshare-app-golang.alloc_objects%7B%7D). Just put random non-json file, set some params for spyname and units, click Save. Then go to browser devtools and review `upload` request body. How can I go further with this logic?

webapp/javascript/pages/adhoc/components/FileUploader.tsx
```diff
@@ -1,21 +1,54 @@
...
+  'pyroscope-rs': 'Rust',
+  rbspy: 'Ruby',
+  javaspy: 'Java',
+  dotnetspy: '.NET',
+  nodespy: 'NodeJS',
+  ebpfspy: 'eBPF',
+  other: 'Other',
+};
+
+const availableUnits = {
```
eh-am, line 32: There are other units available. Let's pull that and `spynames` from https://github.com/pyroscope-io/pyroscope/blob/main/packages/pyroscope-models/src/spyName.ts and https://github.com/pyroscope-io/pyroscope/blob/main/packages/pyroscope-models/src/profile.ts#L14-L22

webapp/javascript/pages/adhoc/components/utils.ts
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable default-case, @typescript-eslint/switch-exhaustiveness-check, consistent-return */
```
eh-am, line 1: `Utils` is a footgun, if we can let's call it with a more descriptive name.

`labels.ts` or `humanize.ts`

eh-am: Also you don't want to disable `@typescript-eslint/switch-exhaustiveness-check`, since that's what will force us to handle all use cases.

eh-am: Eg
Screen Shot 2022-08-19 at 12 16 14

eh-am: Also from what I see we are using `SpyNameFirstClass`, which is a narrower type than `SpyName`. We could export SpyNameFirstClass as an enum type and only deal with the cases we actually care about:

Screen Shot 2022-08-19 at 12 22 14

eh-am (changes requested):

webapp/javascript/pages/adhoc/components/utils.ts
pavelpashkovsky: fixed

Rperry2174 (approved):
lgtm thanks @pavelpashkovsky !

PR #1396, opened 2022-08-12
feat: reuse modal component

dogfrogfog:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1354
    - we have two modal selectors using same UI
    - escape duplication
## Changes
- modal UI part from AppSelector component and explorer page table from ViewTagsSelectLinkModal was moved to new reusable component
- refactor app selector and tags selector to use common UI component

https://user-images.githubusercontent.com/47758224/184434971-20dfe748-e5f5-4bab-90fd-f3b46d39a48b.mov

## Concerns
- different "no data" messages
-  max height for app selector modal is too big


eh-am:
I will have a more in-depth look, in the meanwhile, can you add this component to storybook?

eh-am (approved):

PR #1397, opened 2022-08-12
feat(webapp): set title automatically

eh-am:
Adds a `PageTitleWithAppName` component (aliased to `PageTitle` for succinctness) that uses [`react-helmet`](https://github.com/nfl/react-helmet) under the hood to set a title. Also updates all pages to add a title.

---
I used the react-helmet library since it supports other fields, and has some nice logic to merge/overwrite different (so let's say you can a generic title, and then down the component tree you overwrite that title something more specific).

Also with that `PageTitle` component, adds a suffix with the app name (default to `Pyroscope`), this is so that we can reuse this same code (eg pyroscope cloud), and just modify the app name.

For the simple pages (settings, adhoc) I let the `<PageTitle/>` component next to the route. I found that's easier to reason about. However, for the more complex ones (eg the ones with queries), I added to the page component itself (eg `ContinuousSingleView.tsx`), since they have more info on how to build the title (ie they know about the current queries).

There were a few quirks (non exhaustive list):
* diffing the same app, in that case you don't want to have redundant information like `diff between MYAPP and MYAPP`, so I just left `diff A`
* sometimes the query is not defined (still have to investigate that in a diff PR, I think it has to do with how we set up a query dynamically), so we have to account for these cases, so that we don't accidentally have titles with `undefined`, hence a `formatTitle` function that abstracts all this.

https://user-images.githubusercontent.com/6951209/184716553-5bfc7104-50bb-4ac2-b748-1762d22f5579.mp4


Rperry2174 (approved):
I don't feel too strongly but a minor inconsitency is that for adhoc we don't add the "view" part but for continuous pages we add "view"... 

Personally I don't think "view" adds much and I would remove it if they're all gonna be one way, but again fine if you merge as is as well:

image

PR #1402, opened 2022-08-15
fix: add trace_id span_id fields to java jaeger example

korniltsev:


Rperry2174:
Looks great.. Can we tweak the log message to also have something unique that will allow us to identify it as java? Maybe add rideshare-java to the message somewhere:

The query bar is populated via this regexp:
image

so in order for us to link directly to the _java_ app we need to have a slightly different regexp that is specific to java:

image

examples/tracing/jaeger/java/rideshare/src/main/resources/application.properties
```diff
@@ -1 +1,3 @@
 server.port=5000
+logging.pattern.level=%clr(%5p) %clr([ trace_id=%X{traceId:-} span_id=%X{spanId:-} ])
```
Rperry2174, line 2:
```suggestion
logging.pattern.level=%clr(%5p) %clr([ trace_id=%X{traceId:-} java_span_id=%X{spanId:-} ])
```

korniltsev:
looks good

Rperry2174 (approved):
LGTM -- it would be nice if we only logged this for the _root_ span instead of all spans, but not sure if thats possible here. Fwiw here is how we did it in Go:
https://github.com/pyroscope-io/pyroscope/pull/1124/files

PR #1409, opened 2022-08-16
chore: implements convertPprofToProfile function

petethepig:


packages/pyroscope-flamegraph/src/convert/convertPprofToProfile.ts
```diff
@@ -0,0 +1,141 @@
...
+    resultFlamebearer.levels
+  );
+
+  return {
+    version: 1,
+    flamebearer: resultFlamebearer,
+    metadata: {
+      format: 'single',
+      units: 'samples',
+      spyName: 'gospy',
```
eh-am, line 137: Can we make these the defaults, but allowing to be overwritten?

webapp/javascript/components/Settings/index.tsx
```diff
@@ -21,6 +21,26 @@ import styles from './Settings.module.css';
...
+  var binaryString = window.atob(base64);
+  var len = binaryString.length;
+  var bytes = new Uint8Array(len);
+  for (var i = 0; i < len; i++) {
+    bytes[i] = binaryString.charCodeAt(i);
+  }
+  return bytes;
+}
+
+const pprofBuf = base64ToArrayBuffer(`
```
eh-am, line 39: Let's use this as a test for `convertPprofToProfile`

packages/pyroscope-flamegraph/src/convert/convertPprofToProfile.ts
```diff
@@ -0,0 +1,141 @@
+/* eslint-disable import/prefer-default-export */
+import type { Profile } from '@pyroscope/models/src';
+
+// import {Profile as PProfProfile} from 'pprof/proto';
+
+import { perftools } from 'pprof/proto/profile';
+import { string } from 'zod';
```
eh-am, line 7: I think you imported this accidentally?

packages/pyroscope-flamegraph/src/convert/convertPprofToProfile.ts
```diff
@@ -0,0 +1,141 @@
...
+      pprofProfile.stringTable[pprofProfile.sampleType[i].type] ===
+      selectedSampleType
+    ) {
+      sampleTypeIndex = i;
+      break;
+    }
+  }
+
+  if (sampleTypeIndex === -1) {
+    throw 'sample type is not found in the profile';
```
eh-am, line 42: Not a big fan of throwing errors, we could return undefined so that forces people using typescript to handle `Profile | undefined`. 

Or if we are going to go with the error throwing version, to document that vehaviour with [tsdoc](https://tsdoc.org/pages/tags/example/).

eh-am (commented):
Looks good!

I am a little bit afraid of the `pprof` dependency though. I had issues installing it before, am I am afraid that may pose a challenge for regular users of the flamegraph component, which do not care about converting from pprof.

packages/pyroscope-flamegraph/src/convert/convertPprofToProfile.ts
petethepig: probably

PR #1410, opened 2022-08-16
fix(webapp): refresh data when (re)submitting query

eh-am:
Closes https://github.com/pyroscope-io/pyroscope/issues/1406

Hitting execute/resubmitting the same query now refreshes all data (timelines and flamegraphs).

---

RE `QueryInput`:
* Move to its own file
* Add tests
* Move to css modules

Update comparison/diff to, when resubmitting, to reload EVERYTHING, including the opposite's timeline's and flamegraph. For example, when resubmitting the left side, it will also reload: 
* the main timeline
* the left AND right timeline
* left and right flamegraphs

Which has the side effect of having your selection be "lost". Here's an illustration:

https://user-images.githubusercontent.com/6951209/184947854-6cc60266-e549-4bf9-8656-4d61363004c9.mp4

We discussed internally and it should be fine for now.


Rperry2174 (approved):
LGTM

PR #1417, opened 2022-08-17
feat(adhoc): support passing custom spyName/unit

eh-am:
Just noticed @pavelpashkovsky needs this for https://github.com/pyroscope-io/pyroscope/pull/1392. This is just the backend part.

Basically allow passing custom spyName/units when uploading adhoc files. It **does not work** with diff, since diff takes 2 profile json files (which we assume are already in the correct shape).

RE implementation:
Added to the model's `Convert` so that we can reuse same functionality in other places transparently.

Modeled fields after @dogfrogfog's implementation in flamegraph.com

### DIFF
--- a/pkg/adhoc/server/model.go
+++ b/pkg/adhoc/server/model.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"unicode"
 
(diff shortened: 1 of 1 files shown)


Rperry2174 (approved):
lgtm

petethepig (approved):

PR #1418, opened 2022-08-17
feat(flamegraph): added sub-second units support for trace visualization

StasDachinsky:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1393
## Changes
 - added ms and μs support for trace_samples formatter 
 - added tests for μs, ms, hour, day, month and year 
## Concerns
-

### DIFF
--- a/packages/pyroscope-flamegraph/src/format/format.spec.ts
+++ b/packages/pyroscope-flamegraph/src/format/format.spec.ts
@@ -41,11 +41,15 @@ describe('format', () => {
         const df = getFormatter(80, 2, 'trace_samples');
 
         expect(df.format(0, 100)).toBe('0.00 seconds');
-        expect(df.format(0.001, 100)).toBe('< 0.01 seconds');
+        expect(df.format(0.001, 100)).toBe('10.00 μs');
+        expect(df.format(0.01555, 100)).toBe('155.50 μs');
+        expect(df.format(0.1, 100)).toBe('1.00 ms');
+        expect(df.format(9.3, 100)).toBe('93.00 ms');
         expect(df.format(100, 100)).toBe('1.00 second');
         expect(df.format(2000, 100)).toBe('20.00 seconds');
-        expect(df.format(2012.3, 100)).toBe('20.12 seconds');
+        expect(df.format(2012.3, 100)).toBe('20123.00 ms');
         expect(df.format(8000, 100)).toBe('80.00 seconds');
+        expect(df.format(374.12, 100)).toBe('3741200.00 μs');
       });
 
       it('correctly formats duration when maxdur = 80', () => {
--- a/packages/pyroscope-flamegraph/src/format/format.spec.ts
+++ b/packages/pyroscope-flamegraph/src/format/format.spec.ts
@@ -40,16 +40,11 @@ describe('format', () => {
       it('correctly formats trace samples', () => {
         const df = getFormatter(80, 2, 'trace_samples');
 
-        expect(df.format(0, 100)).toBe('0.00 seconds');
-        expect(df.format(0.001, 100)).toBe('10.00 μs');
-        expect(df.format(0.01555, 100)).toBe('155.50 μs');
-        expect(df.format(0.1, 100)).toBe('1.00 ms');
-        expect(df.format(9.3, 100)).toBe('93.00 ms');
+        expect(df.format(0.001, 100)).toBe('< 0.01 seconds');
         expect(df.format(100, 100)).toBe('1.00 second');
         expect(df.format(2000, 100)).toBe('20.00 seconds');
-        expect(df.format(2012.3, 100)).toBe('20123.00 ms');
+        expect(df.format(2012.3, 100)).toBe('20.12 seconds');
         expect(df.format(8000, 100)).toBe('80.00 seconds');
-        expect(df.format(374.12, 100)).toBe('3741200.00 μs');
       });
 
       it('correctly formats duration when maxdur = 80', () => {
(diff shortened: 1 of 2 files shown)


packages/pyroscope-flamegraph/src/format/format.spec.ts
```diff
@@ -41,11 +41,15 @@ describe('format', () => {
...
         expect(df.format(0, 100)).toBe('0.00 seconds');
-        expect(df.format(0.001, 100)).toBe('< 0.01 seconds');
+        expect(df.format(0.001, 100)).toBe('10.00 μs');
+        expect(df.format(0.01555, 100)).toBe('155.50 μs');
+        expect(df.format(0.1, 100)).toBe('1.00 ms');
+        expect(df.format(9.3, 100)).toBe('93.00 ms');
         expect(df.format(100, 100)).toBe('1.00 second');
         expect(df.format(2000, 100)).toBe('20.00 seconds');
-        expect(df.format(2012.3, 100)).toBe('20.12 seconds');
+        expect(df.format(2012.3, 100)).toBe('20123.00 ms');
```
eh-am, line 50: I don't quite get this, why do we use `seconds` for `2000`, but `ms` for `2012.3`?

packages/pyroscope-flamegraph/src/format/format.ts
```diff
@@ -90,6 +91,73 @@ class DurationFormatter {
...
+  divider = 1;
+
+  suffix = 'second';
+
+  durations: [number, string][] = [
+    [60, 'minute'],
+    [60, 'hour'],
+    [24, 'day'],
+    [30, 'month'],
+    [12, 'year'],
```
eh-am, line 106: for completion it's worth adding tests for these

packages/pyroscope-flamegraph/src/format/format.ts
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable max-classes-per-file */
 import { Units } from '@pyroscope/models/src';
+import _last from 'lodash/last';
```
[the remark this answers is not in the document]
petethepig: Agreed, I'm also not a big fan of these — there's value in keeping the list of dependencies as low as possible.

packages/pyroscope-flamegraph/src/format/format.spec.ts
```diff
@@ -41,11 +41,15 @@ describe('format', () => {
...
         expect(df.format(8000, 100)).toBe('80.00 seconds');
+        expect(df.format(374.12, 100)).toBe('3741200.00 μs');
```
petethepig, line 52: this one also doesn't seem right

packages/pyroscope-flamegraph/src/format/format.ts
```diff
@@ -90,6 +91,73 @@ class DurationFormatter {
...
+  ];
+
+  subSecondDurations: [number, string][] = [
+    [1000, 'ms'],
+    [1000, 'μs'],
+  ];
+
+  units = '';
+
+  constructor(maxDur: number, units?: string) {
```
petethepig, line 116: `units` isn't really used by any of the users, right? Therefore I'd remove it altogether.

packages/pyroscope-flamegraph/src/format/format.ts
```diff
@@ -90,6 +91,73 @@ class DurationFormatter {
...
+      } else {
+        break;
+      }
+    }
+  }
+
+  format(samples: number, sampleRate: number): string {
+    let n = samples / sampleRate / this.divider;
+
+    if (n && !Number.isInteger(n) && this.divider === 1) {
```
petethepig, line 140: This logic is hard to understand, and it is faulty (see test cases above). I would not invest any more time into it.

I think you could use existing `DurationFormatter`, but modify it in two ways:
* Add milliseconds and microseconds to `durations` array, with `1000` multipliers each.
* Modify line 138 to be `let n = samples / sampleRate * 1000000 / this.divider` to compensate for extra precision.

And the rest of it should work just fine. 

@StasDachinsky let me know what you think.

petethepig: And by the way, if that's the case (and these are the only changes we're making to this formatter) we probably don't need a separate class and instead we could have a constructor argument in the existing `DurationFormatter`.

/cc @Rperry2174 — let us know if there's more changes that need to happen. E.g jaeger does [double numbers](https://github.com/jaegertracing/jaeger-ui/blob/e2ae62af03e820f261713cb32ba0290c734e34f8/packages/jaeger-ui/src/utils/date.tsx#L109-L112)

packages/pyroscope-flamegraph/src/format/format.ts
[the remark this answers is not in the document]
StasDachinsky: Agreed, this will be removed

packages/pyroscope-flamegraph/src/format/format.ts
StasDachinsky: Ok, I will update the code according to your recommendations

StasDachinsky: @petethepig Can you please explain what you mean by the constructor argument for "DurationFormatter"? Should I keep the current logic when the suffix is selected by the constructor argument "maxDur", or do I need to update it to allow "DurationFormatter" to select the suffix based on the value passed in the format function (like in the "jaeger"), but only for "trase_samples"?

petethepig: @StasDachinsky by that I meant that we'd add a new argument to `DurationFormatter` constructor, something like `enableSubsecondPrecision`, and then if you want to include sub-second precision you'd call it with that extra boolean, e.g:

`new DurationFormatter(123, '', true)`

This way we can make it so that in all other cases `DurationFormatter` is still working the same way it used to (with no sub-second precision) and it only does work with sub-second precision in tracing mode.

petethepig (changes requested):
It all looks good, I think the last thing we need to do is make it so that sub second duration only show up for `trace_samples`. That will also fix the cypress test.

petethepig (approved):

PR #1419, opened 2022-08-17
fix(webapp): fixes crash when hitting back button

eh-am:
Original issue:

When selecting a time range and hitting the back button
https://user-images.githubusercontent.com/6951209/185229101-47444c01-7644-460e-b194-404750e681c8.mp4

This happens due to `redux-sync`, which syncs the URL with the redux state both ways.
The url changes all at once (both `from` and `until` parameters), while in redux only a single action is dispatched at a time:
Screen Shot 2022-08-17 at 16 37 44

Which leads to a somewhat inconsistent state, where we are mixing relative values (like `from`) with absolute values (unix `timestamp`)
Screen Shot 2022-08-17 at 16 37 27

Which then generates a wrong date, crashing `date-fns` (https://github.com/pyroscope-io/pyroscope/blob/e2128b9e7fdde0fc823985c098b1b86ddcf710f9/webapp/javascript/util/formatDate.ts#L56-L58)

So my lazy solution is to just check whether it's a valid date, and if not, to just return an empty string. Since this is used as a label only, it should be somewhat fine.

Ahh how I would love using `Maybe` here :)

BTW the problem also happens when loading data:
Screen Shot 2022-08-17 at 16 44 47

However since it happens so fast, we end up cancelling the previous request.

### DIFF
--- a/webapp/javascript/util/formatDate.spec.ts
+++ b/webapp/javascript/util/formatDate.spec.ts
@@ -33,6 +33,10 @@ describe('FormatDate', () => {
         '1640090089000000000',
         '2021-06-21 12:34 PM - 2021-12-21 12:34 PM',
       ],
+
+      // Return nothing when mixing absolute/relative
+      ['1624278889000000000', 'now-1h', ''],
+      ['now-1h', '1624278889000000000', ''],
     ];
 
     test.each(cases)(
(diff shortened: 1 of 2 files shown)


Rperry2174 (approved):
lgtm 🔥 !

PR #1422, opened 2022-08-18
feat(webapp): Add tooltip in explore timeline

pavelpashkovsky:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1401

## Changes
- added tooltip in Explore timeline
- [DEMO](https://pr-1422.pyroscope.io/explore?query=rideshare-app-golang.alloc_objects%7B%7D&groupBy=region&groupByValue=All)

https://user-images.githubusercontent.com/7372044/186119129-afdef336-4e47-4136-a2c1-045a0c4201d0.mp4

## Concerns
-


webapp/javascript/components/TimelineChart/ExploreTooltip.tsx
```diff
@@ -0,0 +1,75 @@
...
+        ? values.map((v) => {
+            return (
+              <div key={v?.tagName} className={styles.valueWrapper}>
+                <div
+                  className={styles.valueColor}
+                  style={{
+                    backgroundColor: Color.rgb(v.color).toString(),
+                  }}
+                />
+                <div>{v?.tagName}:</div>
```
eh-am, line 65: If v is undefined we would render just `:`. Is this correct?

webapp/javascript/components/TimelineChart/TimelineChartPlugin.ts
```diff
@@ -180,6 +162,12 @@ import { getUTCdate } from '@webapp/util/formatDate';
     };
 
     function bindEvents(plot, eventHolder) {
+      const o = plot.getOptions();
+      // don't initiate logic if there is exploreTooltip in props
```
eh-am, line 166: I can see we would return early if `exploreTooltip` is set, so the comment is a bit redundant. The real question here is why do we need to return early. Why does the logic below should NOT be initialized?

webapp/javascript/components/TimelineChart/TimelineChartWrapper.tsx
```diff
@@ -71,6 +71,7 @@ type TimelineChartWrapperProps = TimelineDataProps & {
 
   /** selection type 'single' => gray selection, 'double' => color selection */
   selectionType: 'single' | 'double';
+  exploreTooltip?: FC<ShamefulAny>;
```
eh-am, line 74: Can we type it better? Either `unknown` if its props are not really used, or to the real thing?

webapp/javascript/components/TimelineChart/TimelineTooltip.tsx
```diff
@@ -0,0 +1,189 @@
+/* eslint-disable
```
eh-am, line 1: Did you write this file from scratch? Is it forked from somewhere else?

eh-am (changes requested):
Looks good, added a few comments:

1. Just curious, did you notice that was slow then added `useMemo` to speed it up, or was it done as a defensive strategy? 


2. >webapp/javascript/components/TimelineChart/utils.ts

This file need a test
Plus [calling it `utils` is a footgun](https://github.com/pyroscope-io/pyroscope/pull/1392#discussion_r950298976).


3. How to change the content?
Let's say I have a different timeline (in a completely different page), which I want a tooltip with different format. How would we go about it?
I see we can inject a custom `ExploreTooltip`

```jsx
exploreTooltip={ExploreTooltip}
```
but a) not the best name b) we would have to reimplement positioning (https://github.com/pyroscope-io/pyroscope/pull/1422/files#diff-f89f5e18f9da46438916de88c1a5a9905c1422a530ea86288caf5e5d5e919290R32-R43) which kinda sucks


I think from all these, 3rd point is the most important one. Since we will **definitely** reuse the timeline tooltip somewhere else.

webapp/javascript/components/TimelineChart/TimelineTooltip.tsx
pavelpashkovsky: I wrote this file from scratch, here is main logic of new tooltip.

webapp/javascript/components/TimelineChart/ExploreTooltip.tsx
pavelpashkovsky: fixed, tagName is mandatory field, so optional chaining was removed

webapp/javascript/components/TimelineChart/TimelineChartWrapper.tsx
pavelpashkovsky: fixed

pavelpashkovsky:
> Just curious, did you notice that was slow then added useMemo to speed it up, or was it done as a defensive strategy?

This component re-renders really frequently, I felt the performance difference on big data sets. I think it might depend much on local environment eg. computer performance, local precesses etc.

>webapp/javascript/components/TimelineChart/utils.ts
> This file need a test
> Plus https://github.com/pyroscope-io/pyroscope/pull/1392#discussion_r950298976.

splitted this file by functions and added tests

>exploreTooltip={ExploreTooltip}
> but a) not the best name

I'd appreciate any ideas how to name this component. BTW I used word Explore not because it works on Explore Tags page, because it provides better way for data exploration on timeline :)

>we would have to reimplement positioning which kinda sucks

With which exact side of positioning implementation you are not happy?

>I want a tooltip with different format

can you please explain more about it. I'm on this issue right now. Data on tooltip must have the same format as on flamegraph, right?
image

>Since we will definitely reuse the timeline tooltip somewhere else

This tooltip fully based on Flot plugin. You can't use it outside of Flot environment. But can you please describe your vision  of possible API for using this tooltip. I mean something like this:
`
<TimelineChartWrapper
    {...someProps}
    exploreTooltip={ExploreTooltip} // <== is this incorrect?
/>
`

webapp/javascript/pages/TagExplorerView.tsx
```diff
@@ -224,6 +227,17 @@ function TagExplorerView() {
...
+                onHoverDisplayTooltip={(data) =>
+                  (
+                    <ExploreTooltip
+                      pageX={data.pageX}
+                      pageY={data.pageY}
+                      align={data.align}
+                      values={data.values}
+                      timeLabel={data.timeLabel}
+                    />
+                  ) as ShamefulAny
```
pavelpashkovsky, line 239: I typed `ExploreTooltip` as `React.FC<ExploreTooltipProps>` everywhere, but for some reason it's "transformed" to Element. Maybe @eh-am you know the reason and how to fix it correctly?
image

webapp/javascript/components/TimelineChart/TimelineTooltip.tsx
```diff
@@ -0,0 +1,195 @@
...
+      selection.to = selectionData?.xaxis?.to || -1;
+    };
+
+    plot.hooks.bindEvents.push((p: PlotType, eventHolder: EventHolderType) => {
+      eventHolder.mousemove(onMouseMove);
+      eventHolder.mouseleave(onMouseLeave);
+      eventHolder.mouseup(onMouseUp);
+      eventHolder.mousedown(onMouseDown);
+      plot.getPlaceholder().bind('plothover', onPlotHover);
+      // detect plotselecting event from ./TimelineChartSelection.ts
```
eh-am, line 165: not following, why does this tooltip plugin need to know about selection?

webapp/javascript/components/TimelineChart/clamp.ts
```diff
@@ -0,0 +1,12 @@
+function clamp(min: number, value: number, max: number) {
```
eh-am, line 1: nit but having the actual `value` in the middle is weird, either put it in the front or at the end

putting at the end makes it easier to do partial application

```js
const myClamp = clamp.bind(null, floor, celling);
myClamp(myValue)
```
then when the pipeline operator is finally implemented (https://tc39.es/proposal-pipeline-operator/)
```f#
 value 
  |> myClamp
```

but let's not get too ahead of ourselves haha

webapp/javascript/components/TimelineChart/TimelineTooltip.tsx
pavelpashkovsky: when we're selecting time range it's showing from-to:
image

webapp/javascript/components/TimelineChart/ExploreTooltip.tsx
```diff
@@ -0,0 +1,75 @@
...
+export interface ExploreTooltipProps {
+  pageX: number;
+  pageY: number;
+  align: 'left' | 'right';
+  timeLabel?: string;
+  values?: Array<{
+    closest: number[];
+    color: number[];
+    tagName: string;
+  }>;
```
eh-am, line 20: I don't fully understand this, why do we receive the `color/tagName`? what is a `tagName` in a timeline?
for example, in this timeline what's a `tagName`?

image


https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/flot/index.d.ts#L157-L164

webapp/javascript/components/TimelineChart/TimelineTooltip.tsx
eh-am: honestly i feel like showing that interval is not really useful ATM

webapp/javascript/pages/TagExplorerView.tsx
eh-am: what fields are missing?

webapp/javascript/components/TimelineChart/types.ts
```diff
@@ -0,0 +1,49 @@
+export type PlotType = {
```
eh-am, line 1: I wonder if we can't leverage typings from `@types/flot` -> https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/flot/index.d.ts#L217

webapp/javascript/components/TimelineChart/ExploreTooltip.tsx
pavelpashkovsky: since we splitted tooltip into 2 parts, `color/tagName` belong only part related to Explore page, on another pages we get time and nearest data so we can customize it as we want

webapp/javascript/components/TimelineChart/TimelineTooltip.tsx
pavelpashkovsky: removed this part

eh-am (approved):

Rperry2174:
Great work @pavelpashkovsky !

PR #1423, opened 2022-08-18
chore(webapp): cleanup

eh-am:
* Move `index.jsx` to `index.tsx`
* remove react-fps-stats dependency, since its only value is when running locally, which similar functionality can be done with chrome dev tools https://developer.chrome.com/blog/new-in-devtools-61/#fps-meter
* remove unused files
* move all other js files to typescript
* fix any other typings issues (basically there was a whole problem with storybook importing react-router-dom 6, and that was overriding our types for `react-router-dom` 5, which is what we use)


Rperry2174 (approved):
Looks great, lets check everything extra close before we release to make sure deleting this didn't break anything

PR #1424, opened 2022-08-18
feat(webapp): Add settings/apps page

StasDachinsky:


eh-am:
Will review more in depth later, but right now:
1 - https://github.com/pyroscope-io/pyroscope/issues/1416#issuecomment-1218239637
2 - Delete should only be available for pyroscope instances behind auth

webapp/javascript/components/Settings/Apps/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+      <div className={appsStyles.searchContainer}>
+        <Input
+          type="text"
+          placeholder="Search app"
+          value={search}
+          onChange={(v) => setSearchField(v.target.value)}
+          name="Search app input"
+        />
+      </div>
+      <table
```
eh-am, line 60: This doesn't seem to be using the table component https://github.com/pyroscope-io/pyroscope/pull/1403

webapp/javascript/redux/reducers/settings.ts
```diff
@@ -231,20 +265,44 @@ export const deleteAPIKey = createAsyncThunk(
   }
 );
 
+export const deleteApp = createAsyncThunk(
+  'newRoot/deleteApp',
+  async (app: App, thunkAPI) => {
+    const res = await deleteAppAPI({ name: app.name });
+
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
eh-am, line 273: That's true we do have this rule, but I wonder why it doesn't trigger in other places like here https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/redux/reducers/continuous.ts#L230

webapp/javascript/components/Settings/Apps/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+              <AppTableItem
+                app={app}
+                key={`appTableItem${app.name}`}
+                onDelete={handleDeleteApp}
+              />
+            ))
+          ) : (
+            <tr>
+              <td className={appsStyles.appsTableEmptyMessage} colSpan={7}>
+                The list is empty
```
eh-am, line 82: Are distinguishing between an empty list (ie no apps) vs there's no apps bc it's a loading state?

webapp/javascript/models/app.ts
```diff
@@ -0,0 +1,16 @@
...
+});
+
+export const appsModel = z.array(appModel);
+
+export type Apps = z.infer<typeof appsModel>;
+export type App = z.infer<typeof appModel>;
+
+export function parse(a: unknown): Result<Apps, ZodError> {
+  return modelToResult(appsModel, a);
+}
```
eh-am, line 16: I don't think this is necessary anymore, see https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/services/tags.ts#L16 and https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/models/tags.ts for a reference

webapp/javascript/components/Settings/Apps/index.tsx
StasDachinsky: my bad, fixing

webapp/javascript/redux/reducers/settings.ts
StasDachinsky: it does, but warnings only

webapp/javascript/services/apps.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export async function deleteApp(data: {
+  name: string;
+}): Promise<Result<boolean, RequestError | ZodError>> {
+  const { name } = data;
+  const response = await request(`/api/apps`, {
+    method: 'DELETE',
+    body: JSON.stringify({ name }),
+  });
```
eh-am, line 30: This can take a long time (there are instances with 15+ minutes). How much is the timeout currently?

StasDachinsky: We are using `fetch` for request, default timeouts is 300s in Chrome and 90s in Firefox and looks like there is no way to increase this values. Possible fix may be switching to axios or similar lib

webapp/javascript/components/Settings/Apps/index.tsx
StasDachinsky: Rn it's just empty list while data is loading, do we have some kinda spinner component for it? If so, we need to update user list too

webapp/javascript/components/Settings/Apps/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+];
+
+function Apps() {
+  const dispatch = useAppDispatch();
+  const apps = useAppSelector(selectApps);
+  const [search, setSearchField] = useState('');
+
+  useEffect(() => {
+    dispatch(reloadApps());
+  });
```
petethepig, line 31: I tried running this and currently it gets into an infinite loop. I think maybe there needs to be an empty array here as the second argument or something like that (not a big expert in react so I'm not sure)

Screen Shot 2022-08-19 at 2 58 19 AM

webapp/javascript/components/Settings/Apps/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+import Input from '@webapp/ui/Input';
+import TableUI from '@webapp/ui/Table';
+import cl from 'classnames';
+
+import appsStyles from './Apps.module.css';
+import tableStyles from '../SettingsTable.module.scss';
+import { getAppTableRows } from './getAppTableRows';
+
+const headRow = [
+  { name: '', label: '', sortable: 0 },
```
petethepig, line 19: I think this one is not needed. E.g look at this screenshot, "Name" is too far to the right.
Screen Shot 2022-08-19 at 3 00 57 AM

It looks like there are 2 `td` elements per `tr`, so it would make sense to have 2 for `th` as well.

webapp/javascript/components/Settings/Apps/index.tsx
petethepig: Addressed in https://github.com/pyroscope-io/pyroscope/pull/1424/commits/66cc28df6dc973c3295e501e65807ebd888b5195

webapp/javascript/components/Settings/Apps/index.tsx
petethepig: Addressed in https://github.com/pyroscope-io/pyroscope/pull/1424/commits/66cc28df6dc973c3295e501e65807ebd888b5195

eh-am:
Looks good! Have a couple thoughts:

1. There's hover effect, but no actual action when clicking, which is kinda misleading IMO.

https://user-images.githubusercontent.com/6951209/185654804-d18a171a-1274-423c-974f-58d3ec6c150b.mp4

2. It it possible to follow what we do in the CLI (https://github.com/pyroscope-io/pyroscope/issues/1416#issuecomment-1218239637), and a) add more info on what's going to happen when deleting (ie it will take lots of time) and b) add a confirmation that requires retyping the app name (not super sold on this if it's too difficult)

For a) That message needs to be overwriten, since behaviour will change in the cloud

3. Which is similar to 2, but it's important to in the popup to show the app name you are deleting, since this is a destructive action. 
image

Rperry2174:
image

Lets add the actual app name that they're deleting to this message

StasDachinsky:
Last version in action
https://user-images.githubusercontent.com/23450818/185906665-69b7acd6-fec3-4fc5-ad71-781b031c567e.mov

webapp/javascript/ui/Modals/Modal.module.css
```diff
@@ -2,10 +2,6 @@
   background-color: var(--ps-ui-foreground);
 }
 
-.popup label {
-  color: var(--ps-neutral-2);
-}
-
```
StasDachinsky, line 8: I removed this style because it hide inputLabel in modal window and looks like we are not using inputLabel anywhere else

StasDachinsky:
https://user-images.githubusercontent.com/23450818/185907966-0d27b8fc-42df-412d-b4b0-84d494b9195c.mov

webapp/javascript/components/Settings/Apps/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  const handleDeleteApp = (app: App) => {
+    setAppsInProcessing([...appsInProcessing, app.name]);
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    dispatch(deleteApp(app))
+      .unwrap()
+      .then(() => {
+        setAppsInProcessing(appsInProcessing.filter((x) => x !== app.name));
+        setDeletedApps([...deletedApps, app.name]);
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
pavelpashkovsky, line 54: Maybe it's better to move same rules to top of the file (line 48, 54 etc.)?

`/* eslint-disable @typescript-eslint/no-floating-promises */
`

webapp/javascript/models/app.ts
```diff
@@ -0,0 +1,10 @@
+import { z } from 'zod';
+
+export const appModel = z.object({
```
pavelpashkovsky, line 3: @eh-am should we use uppercase in zod entities?

webapp/javascript/components/Settings/Apps/getAppTableRows.tsx
```diff
@@ -0,0 +1,73 @@
...
+import Button from '@webapp/ui/Button';
+import Icon from '@webapp/ui/Icon';
+import { App, Apps } from '@webapp/models/app';
+import type { BodyRow } from '@webapp/ui/Table';
+import confirmDelete from '@webapp/components/Modals/ConfirmDelete';
+import LoadingSpinner from '@webapp/ui/LoadingSpinner';
+
+import styles from './AppTableItem.module.css';
+
+interface IDeleteButtorProps {
```
dogfrogfog, line 13: we dont use `I` prefix

webapp/javascript/redux/reducers/settings.ts
```diff
@@ -15,25 +16,36 @@ import {
...
+import { fetchApps, deleteApp as deleteAppAPI } from '@webapp/services/apps';
 import type { RootState } from '@webapp/redux/store';
 import { addNotification } from './notifications';
 import { createAsyncThunk } from '../async-thunk';
 
-type UsersState = {
-  type: 'pristine' | 'loading' | 'loaded' | 'failed';
-  data?: Users;
-};
+enum FetchStatus {
```
pavelpashkovsky, line 24: enum keys and values are the same, maybe it's better to use Numeric enums, instead of String ones?

webapp/javascript/components/Modals/ConfirmDelete/index.tsx
```diff
@@ -1,11 +1,43 @@
...
+  object?: string;
+  warningMsg?: string;
+  onConfirm: () => void;
+  withConfirmationInput?: boolean;
+}
+
+function confirmDelete({
+  objectName,
+  objectType,
+  object,
```
dogfrogfog, line 15: can we use  `objectName` prop instead of `object` ? since they have same `string` type

webapp/javascript/ui/Table.tsx
```diff
@@ -91,9 +93,14 @@ function Table({
   table,
   tableBodyRef,
   className,
+  isLoading,
```
pavelpashkovsky, line 96: You've added loading state for all tables in the app, @eh-am does it work for web app?
image

webapp/javascript/models/app.ts
pavelpashkovsky: like `appModel  => AppModel `

webapp/javascript/redux/reducers/settings.ts
StasDachinsky: Good point, but we are using the same strings statuses in other reducers and I think we should refactor them all in the other task

webapp/javascript/components/Settings/Apps/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+  { name: '', label: '', sortable: 0 },
+];
+
+function Apps() {
+  const dispatch = useAppDispatch();
+  const apps = useAppSelector(selectApps);
+  const isLoading = useAppSelector(selectIsLoadingApps);
+  const [search, setSearchField] = useState('');
+  const [appsInProcessing, setAppsInProcessing] = useState([] as string[]);
+  const [deletedApps, setDeletedApps] = useState([] as string[]);
```
dogfrogfog, line 31:
```suggestion
  const [deletedApps, setDeletedApps] = useState<string[]>([]);
```

webapp/javascript/models/app.ts
dogfrogfog: or `AppSchema`

webapp/javascript/components/Settings/Apps/index.tsx
StasDachinsky: fixed

webapp/javascript/components/Modals/ConfirmDelete/index.tsx
StasDachinsky: fixed

webapp/javascript/components/Settings/Apps/getAppTableRows.tsx
StasDachinsky: fixed

webapp/javascript/components/Settings/Apps/index.tsx
StasDachinsky: fixed

webapp/javascript/models/app.ts
StasDachinsky: I made names here like in this [users.ts](https://github.com/pyroscope-io/pyroscope/blob/525ba0c693080a2e0a48469a91d6c70ece1d1ab3/webapp/javascript/models/users.ts#L19) file

petethepig (approved):
LGTM

StasDachinsky:
Last version with changes requested by @Rperry2174

https://user-images.githubusercontent.com/23450818/186453124-2b87ff5a-527a-4292-a8b3-410015cefafd.mov

PR #1425, opened 2022-08-18
chore: Table ui component tests warn

dogfrogfog:
## Brief
- fix Test ui component tests warnings
- ensure that test render and real one are as close to each other as possible
## Changes
- wrap hook export function call in `act`
- to make your test run closer to how React works in the browser, every component update function call should be wrapped in `act` (https://reactjs.org/docs/test-utils.html#act)
## Concerns
-


Rperry2174:
@dogfrogfog can you please add some more details to the description describing what the issue was here and why this solution fixes it? 

I'm not familiar with this error and would be nice to have some more context for the future!

petethepig (approved):

PR #1429, opened 2022-08-19
fix: do not use sampling ticker for resettable spies

korniltsev:


pkg/util/alignedticker/aligned_ticker.go
```diff
@@ -0,0 +1,57 @@
+package alignedticker
```
petethepig, line 1: thanks for implementing this, we needed this!

petethepig (approved):

PR #1430, opened 2022-08-19
chore(models): default flamebearer names to 'unknown' when it's empty

eh-am:
Follow up from https://github.com/pyroscope-io/pyroscope/pull/1321

Previous to this PR we would just render nodes with empy names.
image

However down the road code assumes it's non empty, eg:
Screen Shot 2022-08-19 at 11 07 51

So to maintain the invariant and simplify all this, decided to default to `unknown` instead.

Keep in mind values on the table still don't make full sense though.
image


Rperry2174 (approved):
LGTM 🥇

PR #1435, opened 2022-08-23
feat(webapp): dropdown component for head-first dropdown

StasDachinsky:
## Brief
 - https://github.com/pyroscope-io/pyroscope/issues/1395
## Changes
 - updated "head first" dropdown to use dropdown component
## Concerns
 -
 
https://user-images.githubusercontent.com/23450818/186137204-a2a6d71e-8096-4d49-ab68-2ca36671ba32.mov

-


packages/pyroscope-flamegraph/src/Toolbar.tsx
```diff
@@ -14,11 +14,15 @@ import useResizeObserver from '@react-hook/resize-observer';
 import Button from '@webapp/ui/Button';
 // eslint-disable-next-line import/no-extraneous-dependencies
 import Select from '@webapp/ui/Select';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import Dropdown, { MenuItem } from '@webapp/ui/Dropdown';
 import { FitModes, HeadMode, TailMode } from './fitMode/fitMode';
-import { ViewTypes } from './FlameGraph/FlameGraphComponent/viewTypes';
 import SharedQueryInput from './SharedQueryInput';
-import styles from './Toolbar.module.css';
+import { ViewTypes } from './FlameGraph/FlameGraphComponent/viewTypes';
```
dogfrogfog, line 21: type

StasDachinsky: fixed

eh-am (changes requested):
Hey!

I don't see the way this refactor was done as an improvement. Here's why:

# 1. Iterate over keys dynamically.
Although it's cool to have things dynamic like this, it doesn't really represent the domain. Here there are only 2 values, and they are pretty much static:

```jsx
      <option value={HeadMode}>{texts.head}</option>
      <option value={TailMode}>{texts.tail}</option>
````
is way simpler than
```jsx
      {options.map(({ key, value }) => (
        <MenuItem key={key} value={value}>
          <div className={styles.fitModeDropdownMenuItem}>
            {key}
            {fitMode === value ? <CheckIcon /> : null}
          </div>
        </MenuItem>
      ))}
```


# 2. `showMode` switch case
Before we had a switch case, that depending on the `showMode` (either "small" or "large"),
would return different text.

That is verbose, yes.

The refactor now uses a single variable `isSmall`, and make the code shorter.

However now the logic is spread is 3 different places:
https://github.com/pyroscope-io/pyroscope/pull/1435/files#diff-6b79ab4172b8306a9aee4afbf2af29ab608c43c22d75bc220d81f48714465cdeR245-R252

and
https://github.com/pyroscope-io/pyroscope/pull/1435/files#diff-6b79ab4172b8306a9aee4afbf2af29ab608c43c22d75bc220d81f48714465cdeR256
and
https://github.com/pyroscope-io/pyroscope/pull/1435/files#diff-6b79ab4172b8306a9aee4afbf2af29ab608c43c22d75bc220d81f48714465cdeR260


Which makes maintenance more burdensome.

Plus, from a data model POV it's not a good idea. 
Currently, `showMode` only has 2 values: either `small` or `large`.

But that doesn't necessarily have to hold true **forever**.
On example is the moment we add more items to the toolbar, it may present a `medium` state.

And the moment we update that variable (https://github.com/pyroscope-io/pyroscope/blob/main/packages/pyroscope-flamegraph/src/Toolbar.tsx#L36) to support a `medium` value, the abstraction will be broken. `isSmall` won't make sense anymore, since `!isSmall` now means `large | medium`, as opposed to just `large`.

A switch case is the perfect case to handle this. 

Another argument is that this refactor disregards the rest of codebase, which follows a similar approach:

https://github.com/pyroscope-io/pyroscope/blob/main/packages/pyroscope-flamegraph/src/Toolbar.tsx#L162-L174
https://github.com/pyroscope-io/pyroscope/blob/main/packages/pyroscope-flamegraph/src/Toolbar.tsx#L204-L212
https://github.com/pyroscope-io/pyroscope/blob/main/packages/pyroscope-flamegraph/src/Toolbar.tsx#L348-L354
https://github.com/pyroscope-io/pyroscope/blob/main/packages/pyroscope-flamegraph/src/Toolbar.tsx#L430-L440


---
All that being said, there's a visual bug here, you may have noticed:
image

>window.innerWidth

1258

StasDachinsky:
Hey @eh-am! First I want to say big thanks for the insightful comment, it really helps to understand how everything works)

I fully agree with point 2 and have updated my code accordingly. Regarding the first point with the mappings, I agree that when we have only 2 options it's easier to read without mappings, but with the MenuItem component it's a lot of duplicated code and I do not think it's good to leave it like that. Can you please check the updated version and tell me what you thought about it?
Screenshot 2022-08-24 at 12 25 18


Also I fixed this display bug and find kinda bug? with that window size

packages/pyroscope-flamegraph/src/Toolbar.tsx
```diff
@@ -14,11 +14,15 @@ import useResizeObserver from '@react-hook/resize-observer';
 import Button from '@webapp/ui/Button';
 // eslint-disable-next-line import/no-extraneous-dependencies
 import Select from '@webapp/ui/Select';
+// eslint-disable-next-line import/no-extraneous-dependencies
```
Rperry2174, line 17: why do we need this line?

packages/pyroscope-flamegraph/src/Toolbar.spec.tsx
```diff
@@ -88,7 +88,7 @@ describe('ProfileHeader', () => {
         handleSearchChange={() => {}}
         reset={() => {}}
         updateFitMode={() => {}}
-        fitMode={HeadMode}
+        fitMode={TailMode}
```
Rperry2174, line 91: why did you switch this from `HeadMode` to `TailMode` ?

packages/pyroscope-flamegraph/src/Toolbar.tsx
eh-am: this is my fault actually, the eslint rule thinks `@webapp` is an external dependency

this would be solved after we move `ui` to its own package, which is a good idea bc right now we have a circular dependency: `@webapp` uses flamegraph `flamegraph` which uses `@webapp/ui`

Rperry2174:
Can we make all the buttons/dropdowns evenly spaced out? There's an extra gap between these: 
image

packages/pyroscope-flamegraph/src/Toolbar.spec.tsx
```diff
@@ -266,18 +266,23 @@ describe('ProfileHeader', () => {
     });
 
     it('updates to HEAD first', () => {
-      userEvent.selectOptions(
-        screen.getByRole('combobox', { name: /fit-mode/ }),
-        screen.getByRole('option', { name: /Head/ })
-      );
+      const toolbar = screen
+        .getByRole('toolbar')
+        .querySelector('.dropdownMenuButton') as HTMLElement;
```
eh-am, line 271: If we had to update these it means semantics have changed, which breaks accessibility https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role

eh-am: update the tests here https://github.com/pyroscope-io/pyroscope/pull/1435/commits/1469882669170e5443211a7e7db4496a9a23f60d to reflect the roles better

packages/pyroscope-flamegraph/src/Toolbar.spec.tsx
StasDachinsky: I thought we can test both cases here

Rperry2174 (approved):
LGTM thanks!

PR #1438, opened 2022-08-24
fix(adhoc): support malformed profiles

eh-am:
Temporary fix to support malformed profiles generated by other tools.

A better solution is suggested by @kolesnikovae here https://github.com/pyroscope-io/pyroscope/issues/1439


Rperry2174 (approved):
LGTM

PR #1446, opened 2022-08-26
chore(webapp): timeline chart, removed circles on dataPoints, added crosshair cursor

StasDachinsky:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1400
## Changes
 - added crosshair cursor
 - removed circles in data points on line charts
 - changed lineWidth to line chart from 1 to 0.8
## Concerns
- I think that simply reducing the line width for line charts is not the best option, and we need to do this dynamically based on the lengths of dataPoints and Apps arrays. Reducing the width looks good for large time periods (and big data), but almost useless for small

https://user-images.githubusercontent.com/23450818/186914561-e0d707d9-6cb9-4f89-baec-28110d336d7b.mov

### DIFF
--- a/webapp/javascript/components/TimelineChart/TimelineChartWrapper.tsx
+++ b/webapp/javascript/components/TimelineChart/TimelineChartWrapper.tsx
@@ -126,6 +126,7 @@ class TimelineChartWrapper extends React.Component<
       },
       points: {
         show: false,
+        symbol: () => {}, // function that draw points on the chart
       },
       lines: {
         show: false,
@@ -147,6 +148,7 @@ class TimelineChartWrapper extends React.Component<
             ...flotOptions,
             lines: {
               show: true,
+              lineWidth: 0.8,
             },
             bars: {
               show: false,
--- a/webapp/sass/profile.scss
+++ b/webapp/sass/profile.scss
@@ -110,7 +110,9 @@ body {
 #timeline-chart-single,
 #timeline-chart-double,
 #timeline-chart-left,
-#timeline-chart-right {
+#timeline-chart-right,
+#timeline-chart-diff,
+#timeline-chart-explore-page {
   overflow: hidden;
   cursor: crosshair;
 }


webapp/javascript/components/TimelineChart/TimelineChartSelection.ts
```diff
@@ -116,8 +119,14 @@ const handleHeight = 22;
       return isLeftSelecting ? 'left' : isRightSelecting ? 'right' : null;
     }
 
-    function setCursor(type: string) {
-      $('canvas.flot-overlay').css('cursor', type);
+    function cursor() {
+      let current = '';
+      return function setCursor(type: string) {
```
dogfrogfog, line 124: can we use anonymous function here ?

webapp/javascript/components/TimelineChart/TimelineChartSelection.ts
```diff
@@ -116,8 +117,14 @@ const handleHeight = 22;
       return isLeftSelecting ? 'left' : isRightSelecting ? 'right' : null;
     }
 
-    function setCursor(type: string) {
-      $('canvas.flot-overlay').css('cursor', type);
+    function cursor() {
```
pavelpashkovsky, line 120: function `cursor`? please add what does this function do to the name. setCursor? changeCursor?

webapp/javascript/components/TimelineChart/TimelineChartSelection.ts
```diff
@@ -116,8 +117,14 @@ const handleHeight = 22;
...
+      let current = '';
+      return function (type: string) {
```
pavelpashkovsky, line 122: what the point of returning function? can you simplify it?

webapp/javascript/components/TimelineChart/TimelineChartSelection.ts
```diff
@@ -389,6 +396,7 @@ const handleHeight = 22;
     plot.clearSelection = clearSelection;
     plot.setSelection = setSelection;
     plot.getSelection = getSelection;
+    plot.setCursor = cursor();
```
pavelpashkovsky, line 399: can you please make name of this function more informative cursor();

plot.setCursor = setCursor(); ?

pavelpashkovsky (changes requested):

pavelpashkovsky:
@StasDachinsky can you please take a look at 
pyroscope/webapp/javascript/components/TimelineChart/TimelineChartPlugin.ts

at `function bindEvents(plot, eventHolder)` you can add any handlers (or use existing) to it. might be helpful

eh-am (commented):
I don't quite understand why do we need all these changes, if we already support it here https://github.com/pyroscope-io/pyroscope/blob/5788cf9943a579acd35de224537016488a98808f/webapp/sass/profile.scss#L110-L116

I feel like this implementation does more than the trivial impl above, can you just explain it what it does?

Also, if we are going with this route, we should remove that css I linked above then.

webapp/javascript/components/TimelineChart/TimelineChartPlugin.ts
```diff
@@ -57,7 +58,17 @@ import { getUTCdate } from '@webapp/util/formatDate';
       }
     }
 
+    const setHoverCursor = (isHover: boolean) => {
+      if (this.isHover !== isHover) {
```
eh-am, line 62:
```suggestion
      const hasChanged = this.isHover !== isHover;
      if (hasChanged) {
```

a nit, but I found these are clearer to read, any chance to reduce mental overhead is a good idea

webapp/javascript/components/TimelineChart/TimelineChartSelection.ts
```diff
@@ -20,6 +20,7 @@ type PlotType = {
   clearSelection: (preventEvent: boolean) => void;
   setSelection: (ranges: any, preventEvent: any) => void;
   getSelection: () => {} | null;
+  setCursor: (cursor: string) => void;
```
eh-am, line 23:
```suggestion
  setCursor: (cursor: 'grab' | 'crosshair') => void;
```

If possible let's narrow the types as much as possible. By having just `string` thereotically a `setCursor("FOOOOOOO")` is valid, although it's obviously not true.

webapp/javascript/components/TimelineChart/TimelineChartSelection.ts
```diff
@@ -116,8 +117,14 @@ const handleHeight = 22;
       return isLeftSelecting ? 'left' : isRightSelecting ? 'right' : null;
     }
 
-    function setCursor(type: string) {
-      $('canvas.flot-overlay').css('cursor', type);
+    function cursor() {
+      let current = '';
+      return function (type: string) {
+        if (type && current !== type) {
```
eh-am, line 123: Why do you have to check `type` is truthy? If it's always truthy we don't need to check it, and if it's not truthy then we need to indicate that in the types (ie `(type: string)` becomes `(type?: string)`)

webapp/javascript/components/TimelineChart/TimelineChartSelection.ts
```diff
@@ -116,8 +117,14 @@ const handleHeight = 22;
...
+          current = type;
+          $('canvas.flot-overlay').css('cursor', type);
```
eh-am, line 125: What if there are 2 canvas in the same page?

webapp/javascript/components/TimelineChart/TimelineChartWrapper.tsx
```diff
@@ -126,6 +127,7 @@ class TimelineChartWrapper extends React.Component<
       },
       points: {
         show: false,
+        symbol: () => {}, // function that draw points on the graph
```
eh-am, line 130: Nice!

webapp/javascript/components/TimelineChart/TimelineChartSelection.ts
eh-am: I wonder if there isn't a way to get the actual element we care about cc @pavelpashkovsky

Rperry2174 (approved):
LGTM

PR #1452, opened 2022-08-31
chore: extend API with support for exemplar queries

kolesnikovae:


pkg/storage/storage_exemplars_query.go
```diff
@@ -0,0 +1,190 @@
...
+	if in.MaxValue > 0 {
+		h.MinValue = 19
+		h.MaxValue = 50
+		h.MinDepth = 1
+		h.MaxDepth = 21
+		h.Values = datasetB
+	}
+
+	return QueryExemplarsOutput{Tree: tree.New(), Heatmap: h}, nil
+}
```
kolesnikovae, line 64: @dogfrogfog, please take a look – I added two data stubs for heatmap: 
- Both samples refer to a static 10 minutes range.
- By default, the dataset A is returned: this one is interesting in the way that its nature is super close to the real data. I imagine that colouring would be pretty tricky, I'd suggest considering logarithmic scale (worth discussing).
- In case if `maxValue` parameter is provided, the mock returns dataset B, that was crafted with MinValue 18 and MaxValue 50 (on the same data set as in the first case).

Below are reference visualisations (don't mind styling – I use this for debugging only):

Dataset A:
image

Dataset B:
image


I hope you will find it helpful.

This stub will be removed before merge.

/cc @Rperry2174

pkg/structs/flamebearer/flamebearer.go
```diff
@@ -101,24 +102,80 @@ type FlamebearerTimelineV1 struct {
...
+	StartTime int64 `json:"startTime"`
+	EndTime   int64 `json:"endTime"`
+	// MinValue and MaxValue indicate boundaries of Y axis: (MinValue:MaxValue].
+	MinValue uint64 `json:"minValue"`
+	MaxValue uint64 `json:"maxValue"`
+	// MinDepth and MaxDepth indicate boundaries of Z axis: [MinDepth:MaxDepth].
+	// MinDepth is the minimal non-zero value that can be found in Values.
+	MinDepth uint64 `json:"minDepth"`
+	MaxDepth uint64 `json:"maxDepth"`
+}
```
kolesnikovae, line 128: @dogfrogfog you can use this as a reference to the actual data models (hopefully, no more changes will be made in the spec). Also, I modified the way the grid is populated as per our discussion: now maximum Y value is at 0, and the minimum is the last element of a column

pkg/storage/storage_exemplars_query.go
kolesnikovae: This request returns dataset A:
```
http://localhost:4040/api/exemplars:query?query=app{}
```
One for dataset B:
```
http://localhost:4040/api/exemplars:query?query=app{}&maxValue=0
```

Let me know if you need just a response JSON

pkg/structs/flamebearer/flamebearer.go
```diff
@@ -101,24 +102,80 @@ type FlamebearerTimelineV1 struct {
 	Watermarks    map[int]int64 `json:"watermarks"`
 }
 
+type FlamebearerHeatmapV1 struct {
```
petethepig, line 105: Nit: I would not call it flamebearer heatmap — it has nothing to do with flamebearer at this point

pkg/server/controller.go
```diff
@@ -268,20 +269,24 @@ func (ctrl *Controller) serverMux() (http.Handler, error) {
...
 		routes = append(routes, []route{
 			{"/render", ctrl.renderHandler()},
 			{"/render-diff", ctrl.renderDiffHandler()},
-			{"/merge", ctrl.mergeHandler()},
 			{"/labels", ctrl.labelsHandler()},
 			{"/label-values", ctrl.labelValuesHandler()},
 			{"/export", ctrl.exportHandler()},
+			{"/merge", ctrl.exemplarsHandler().MergeExemplars},
+			{"/api/exemplars:merge", ctrl.exemplarsHandler().MergeExemplars},
+			{"/api/exemplars:query", ctrl.exemplarsHandler().QueryExemplars},
```
petethepig, line 289: @kolesnikovae Forgot to ask you about this today, but why use colons in URL paths? I feel like it's pretty uncommon. I'd rather do something like:
```
			{"/api/exemplars/merge", ctrl.exemplarsHandler().MergeExemplars},
			{"/api/exemplars", ctrl.exemplarsHandler().QueryExemplars},
```

kolesnikovae: This is a pretty popular approach to denote custom action/method: https://cloud.google.com/apis/design/custom_methods. `/` is fine but I want to make `/api/exemplars` a collection of exemplars. `/api/exemplars/merge` would contradict to the planned endpoints:

**GET** `/api/exemplars` – Lists exemplars' properties (e.g profile_id).
**GET** `/api/exemplars/{profile_id}` – Accesses specific exemplar (an associated profile, meta, timestamps, and labels)

Also, I was thinking about renaming `query` to `search`, to avoid confusion with the URL parameter.

petethepig (approved):

PR #1454, opened 2022-08-31
fix: pprof parser formatting for rbspy

korniltsev:


pkg/convert/pprof/ctx.go
```diff
@@ -0,0 +1,16 @@
...
+const ctxPprofLineNumbers ctxPprofLineNumbersKey = "ctxPprofLineNumbers"
+
+func WithLineNumbersEnabled(ctx context.Context, v bool) context.Context {
+	return context.WithValue(ctx, ctxPprofLineNumbers, v)
+}
+
+func LineNumbersEnabledFromContext(ctx context.Context) bool {
+	v, ok := ctx.Value(ctxPprofLineNumbers).(bool)
+	return ok && v
+}
```
kolesnikovae, line 16: Passing configuration parameters via context might not be desirable:
 - https://www.ardanlabs.com/blog/2019/09/context-package-semantics-in-go.html.
 - https://dave.cheney.net/2017/01/26/context-is-for-cancelation

Please consider the following implementation:

```go
type Option func(*options)

type options struct {
  includeLineNumber bool
}

func WithLineNumber() func(*options) {
   return func(o *options) {
      o.includeLineNumber = true
   }
}

func (p *Parser) Convert(ctx context.Context, startTime, endTime time.Time, profile *tree.Profile, options ...Option) error
```

pkg/convert/pprof/parser.go
```diff
@@ -221,6 +222,16 @@ func (p *Parser) readTrees(x *tree.Profile, c tree.LabelsCache, f tree.Finder) {
 	}
 }
 
+func (p *Parser) formatStackFrame(x *tree.Profile, fn *tree.Function, line *tree.Line, withLineNumbers bool) []byte {
+	if p.spyName == "rbspy" || withLineNumbers {
+		return []byte(fmt.Sprintf("%s:%d - %s",
+			x.StringTable[fn.Filename],
+			line.Line,
+			x.StringTable[fn.Name]))
+	}
```
kolesnikovae, line 231:
```go
p.spyName == "rbspy" || withLineNumbers
```
Should we maybe pass formatter explicitly to parser constructor (as an option)?

pkg/convert/pprof/ctx.go
korniltsev: I removed this option for non rbspy

PR #1455, opened 2022-08-31
feat: annotations (WIP)

eh-am:
still a WIP

```
❯ curl -X POST -d '{ "appName": "pyroscope.server.cpu", "content": "mycontent" }' localhost:4040/api/annotations
```

Plots a red line indicating an annotation
image


pkg/sqlstore/migrations/migrations.go
```diff
@@ -110,3 +111,24 @@ func createAPIKeyTableMigration() *gormigrate.Migration {
 		},
 	}
 }
+
+func createAnnotations() *gormigrate.Migration {
+	type annotation struct {
+		ID        uint   `gorm:"primarykey"`
+		AppName   string `gorm:"type:varchar(255);not null;default:null;index:"`
```
eh-am, line 118: `default:null` whoops

PR #1456, opened 2022-09-01
chore(webapp): cleanup timeline

eh-am:
While I was reading the code for the Annotations implementation, decided to clean the code up a little bit:

* rename `Markings` to `Selection` to better indicate what it really refers to (as opposed to a barebones `marking`), this will come into play once we implement Annotations, which also uses `marking` under the hood
* add defaults for `height`, `timezone` as defaultProps, so that we don't have to check if they exist every time...


Rperry2174 (approved):
LGTM

webapp/javascript/components/TimelineChart/TimelineChartWrapper.tsx
```diff
@@ -193,29 +202,28 @@ class TimelineChartWrapper extends React.Component<
...
       };
 
       return [
         {
           xaxis: { from, to },
           color:
-            this.props.selectionType === 'double'
-              ? m.overlayColor
-              : 'tranparent',
+            this.props.selectionType === 'double' ? m.overlayColor : 'NOOP',
```
pavelpashkovsky, line 212: what the purpose of setting NOOP here?

webapp/javascript/components/TimelineChart/TimelineChartWrapper.tsx
```diff
@@ -193,29 +202,28 @@ class TimelineChartWrapper extends React.Component<
       // to use custom apperance and color for it
       const boundary = {
         lineWidth: 1,
-        color:
-          this.props.selectionType === 'double' ? m.color.rgb() : 'transparent',
+        color: m.color.rgb(),
```
pavelpashkovsky, line 205: this line breaks selection appearance in a bit
image
previous one works as expected:
image

eh-am: ahh you are absolutely right

webapp/javascript/components/TimelineChart/TimelineChartWrapper.tsx
eh-am: `noop` stands for "No operation", I am doing it here since `tranparent` is not a valid color, hence `NOOP` to make it clear it's something NOT valid

PR #1458, opened 2022-09-05
fix(flamegraph): table, buttons colors for light mode

pavelpashkovsky:
## Brief

Fix incorrect colors for UI elements on Flamegraph for light mode.

### before:
before

### after:
image

## Concerns
- imported `sanitize.css` re-writes assigned styles

### DIFF
--- a/packages/pyroscope-flamegraph/src/Toolbar.module.css
+++ b/packages/pyroscope-flamegraph/src/Toolbar.module.css
@@ -5,6 +5,10 @@
   align-items: center;
 }
 
+.navbar button {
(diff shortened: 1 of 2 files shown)


packages/pyroscope-flamegraph/src/Toolbar.module.css
```diff
@@ -5,6 +5,10 @@
   align-items: center;
 }
 
+.navbar button {
+  color: var(--ps-neutral-2);
```
eh-am, line 9: Would be nice to define that somewhat globally no? Since it's clear this color applies app-wide.

PR #1463, opened 2022-09-06
fix(flamegraph): table width

eh-am:
If users of the `flamegraph` component have `table { display: block; }`, the tooltip table won't expand to full width:

Screen Shot 2022-09-06 at 10 21 56

Screen Shot 2022-09-06 at 10 22 05

Forcing the default fixes the issue.

Also updated docs to use newest API (`profile` instead of `flamebearer`)


Rperry2174 (approved):
Lgtm thx!

PR #1465, opened 2022-09-06
docs(flamegraph): fix a typo

eh-am:


Rperry2174:
while you're editing these:

```
We recommend wrapping your component around a `Box` to give it some padding.
```
should probably be
```
We recommend wrapping  a `Box` around your component to give it some padding.
```

PR #1469, opened 2022-09-07
fix(webapp): Don't show tooltip when there's no data

pavelpashkovsky:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1467
## Changes
- Hide Explore tooltip if there is no profiling data
image
## Concerns
- I hid the tooltip, but we still have this red line which follows the cursor. If you want to hide it as well I have to extract `react-flot/flot/jquery.flot.crosshair.min` into the project and make up it there. `{hoverable: boolean}` doesn't influence on this red line.

### DIFF
--- a/webapp/javascript/components/TimelineChart/Tooltip.plugin.tsx
+++ b/webapp/javascript/components/TimelineChart/Tooltip.plugin.tsx
@@ -115,6 +115,10 @@ const TOOLTIP_WRAPPER_ID = 'explore_tooltip_parent';
           }
         );
 
+        if (!values?.length) {
+          return;
+        }
+
         const Tooltip: React.ReactElement<
           ExploreTooltipProps,
           string | React.JSXElementConstructor<ExploreTooltipProps>


webapp/javascript/components/TimelineChart/Tooltip.plugin.tsx
```diff
@@ -115,6 +115,10 @@ const TOOLTIP_WRAPPER_ID = 'explore_tooltip_parent';
           }
         );
 
+        if (!values?.length) {
```
eh-am, line 118: Bit of a sidebar, but there's another edge case we don't fully handle:
It happens when we pass a custom tooltip (on that onHoverDisplayTooltip callback), which in turn returns `null`.

Here's an illustration of the problem:
Screen Shot 2022-09-07 at 10 46 08

eh-am (commented):
I think this line is not relevant anymore right?
https://github.com/pyroscope-io/pyroscope/blob/b5ce89a680256a7d6ff76e45ec38abc721df9a89/webapp/javascript/components/TimelineChart/getFormatLabel.ts#L39

eh-am (approved):

pavelpashkovsky:
>I think this line is not relevant anymore right?

this function is used not only in this tooltip so it still relevant

PR #1471, opened 2022-09-07
chore(webapp): Don't show empty series in timeline or table

pavelpashkovsky:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1464

## Changes
- filter empty series just after they were received from backend and before they get to the redux store
image

## Concerns
-

### DIFF
--- a/webapp/javascript/redux/reducers/continuous.ts
+++ b/webapp/javascript/redux/reducers/continuous.ts
@@ -926,28 +926,9 @@ export const continuousSlice = createSlice({
     });
 
     builder.addCase(fetchTagExplorerView.fulfilled, (state, action) => {
-      let data = action.payload;
-      const { groups } = action.payload;
-      const groupNames = Object.keys(groups);
-
-      if (groupNames?.length) {
-        data = {
-          ...data,
-          groups: groupNames
-            .filter((g) => !!g.trim())
-            .reduce(
-              (acc, current) => ({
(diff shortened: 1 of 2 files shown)


pavelpashkovsky:
@eh-am does it make sense to extract this part to the function and add tests?

eh-am (changes requested):
As you can see, this can be implemented in multiple places, so it's important to understand from a conceptual POV where it belongs to:

* is it a decision on purely the **view** side? if so, let's do it at the component layer/redux selection layer (https://redux.js.org/usage/deriving-data-selectors)
* is it something backend is returning kinda wrong and it should never enter the `frontend layer`? then let's do this processing at the **data service** layer

I personally prefer to leave reducers as thin as possible. Ideally it should just get whatever data the data service layer returned and add to the store.

pavelpashkovsky:
from task:
>I think we ideally should remove this as close to the api response as possible

If I understand user should see such a data at all, so maybe it needs to be moved on backend side, right? 
cc @eh-am , @Rperry2174

eh-am:
Or, as a stopgap, to implement it here https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/services/render.ts#L198 using https://github.com/colinhacks/zod#preprocess

Rperry2174 (approved):
LGTM!

PR #1472, opened 2022-09-07
fix(webapp): hide timeline tooltip if there's no data

eh-am:
Initially raised https://github.com/pyroscope-io/pyroscope/pull/1469#discussion_r964867470

We always render a tooltip, even if there's no body:
image

It's a bit tricky to solve this, since it would require the parent to know whether the children is empty. Instead of manually rendering I decided to just use CSS to hide the tooltip if it's empty (ie its children is null).


pavelpashkovsky (approved):
LGTM!

PR #1473, opened 2022-09-07
feat(backend): annotations

eh-am:
After the annotations POC (https://github.com/pyroscope-io/pyroscope/pull/1455) I decided to create smaller PRs with more focused code.

This is the backend implementation.

* Adds a `/api/annotations` POST endpoint that creates an annotation. For reference here's a payload:
```json
{
  "appName": "myApp",
  "content": "mycontent",
  "timestamp": 1662729099
}
```
if `timestamp` is absent, we default to `time.Now()`. This is so that one can easily create an annotation to mark an event that just happened (eg a deploy).
* enhances `/render` response with a list of annotations for that given app/range, therefore there's no `GET /api/annotations` support.


pkg/sqlstore/migrations/migrations.go
```diff
@@ -110,3 +111,24 @@ func createAPIKeyTableMigration() *gormigrate.Migration {
 		},
 	}
 }
+
+func createAnnotationsTableMigration() *gormigrate.Migration {
+	type annotation struct {
+		ID        uint   `gorm:"primarykey"`
+		AppName   []byte `gorm:"type:varchar(255);not null;default:null"`
```
eh-am, line 118: I thought this would not allow creating empty fields, but it doesn't seem to work:

```
❯ curl -X POST -d '{ "appName": "", "content": "" }' localhost:4040/api/annotations
{"appName":"","content":"","timestamp":1662585565}
```

pkg/api/annotation_test.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/pyroscope-io/pyroscope/pkg/model"
+	"github.com/pyroscope-io/pyroscope/pkg/service"
+)
+
+type mockService struct {
+	createAnnotationResponse func() (*model.Annotation, error)
+}
+
+func (m *mockService) CreateAnnotation(ctx context.Context, params service.CreateAnnotationParams) (*model.Annotation, error) {
+	return m.createAnnotationResponse()
```
eh-am, line 17: @kolesnikovae  we seem to be using `gomock` for the other tests in this package, should we keep using it moving forward?

pkg/server/render.go
```diff
@@ -159,9 +177,19 @@ func (*RenderHandler) mountRenderResponse(flame flamebearer.FlamebearerProfile,
 		Query:                 gi.Query.String(),
 		MaxNodes:              maxNodes,
 	}
+
```
eh-am, line 180: Did this just to reformat in such a way it makes sense for `/render`. Example:  it's redundant to have an `appName` repeated for each annotation.

pkg/service/annotation.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	u.AppName = params.AppName
+	u.Content = params.Content
+	u.Timestamp = params.Timestamp
+
+	tx := svc.db.WithContext(ctx)
+
+	// Upsert
+	if err := tx.Where(model.CreateAnnotation{
```
eh-am, line 33: Although it's simple enough, it already has a certain logic (upserting). 

@kolesnikovae What do you think about adding yet another indirection layer with the `Repository` pattern?

pkg/api/annotation_test.go
kolesnikovae: It would be nice. However, I think it's ok to use a manually crafted mock if it is simpler and more convenient in a particular case. I was just too lazy to manually mock a dozen of methods

pkg/api/annotation.go
```diff
@@ -0,0 +1,73 @@
...
+type CreateParams struct {
+	AppName   string `json:"appName"`
+	Timestamp int64  `json:"timestamp"`
+	Content   string `json:"content"`
+}
+
+func (ctrl *AnnotationsCtrl) CreateHandler(w http.ResponseWriter, r *http.Request) {
+	var params CreateParams
+	if err := json.NewDecoder(r.Body).Decode(&params); err != nil {
+		ctrl.httpUtils.WriteInternalServerError(r, w, err, "failed to unmarshal JSON")
```
kolesnikovae, line 42: Please use `HandleError` instead, it sets proper HTTP code and creates proper response body:
https://github.com/pyroscope-io/pyroscope/blob/90f21bf9d254c442968662b4ee2699cf172537f5/pkg/server/httputils/default.go#L37-L46

pkg/model/annotation.go
```diff
@@ -0,0 +1,46 @@
...
+	"time"
+
+	"github.com/hashicorp/go-multierror"
+)
+
+var (
+	ErrAnnotationInvalidAppName   = errors.New("invalid app name")
+	ErrAnnotationInvalidTimestamp = errors.New("invalid timestamp")
+	ErrAnnotationInvalidContent   = errors.New("invalid content")
+)
```
kolesnikovae, line 14: There are some custom error types that understood by httputils helper, e.g `ValidationError` – this will automatically instruct it to return HTTP code 400 with a properly formatted JSON body listing all the errors encountered

pkg/sqlstore/migrations/migrations.go
kolesnikovae: The `annotation` type definition is private and can't be used outside the migration function (this is done so in order to avoid side effects caused by model type changes; this is not mandatory but considered a good practice).

The reason why API allows creating an annotation with empty fields is that it uses `model.Annotation` which does not have constraints:
```go
type Annotation struct {
 	AppName   string
 	Content   string
 	Timestamp time.Time
 	CreatedAt time.Time
 	UpdatedAt time.Time
 }
```

pkg/service/annotation.go
eh-am: >I'd better re-implement services for that alternative database/storage
Alright!

Although we would need to figure out where to put that common business logici independent of storage

petethepig (approved):
Looks great, I like how everything is organized and I love that everything has tests :)

PR #1474, opened 2022-09-08
feat(webapp): Make <CollapseBox /> component

pavelpashkovsky:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1421

## Changes
- Added collapsible Box
- DEMO:
https://user-images.githubusercontent.com/7372044/189065862-f43c2f4e-bc53-488f-98ff-fb8259c3e2f9.mp4
- https://pr-1474.pyroscope.io/explore?query=rideshare-app-golang.alloc_objects%7B%7D


eh-am (changes requested):
Looks good!

Few thoughts:

1. Would be nice to have a storybook example since this is effectively a new, reusable component. This is so that new devs can refer to this "library of components"

2. Behaviour is weird when title is empty:
https://user-images.githubusercontent.com/6951209/189142078-7e9999dc-9881-4aa1-9dfa-7742c83b3e39.mp4

* if a title MUST not be empty, then we should add least document it
* chevron should always be on the right

3. Another thing I have strong opinions is to ALWAYS show the chevron. Reason is for discoverability. I want to be able to figure out what the component does immediately, I shouldn't need to hover. Specially because not all people interact with a hover. (yeah i know, we also have tooltips, but still)

pavelpashkovsky:
updated! @eh-am

eh-am (approved):

PR #1477, opened 2022-09-08
feat(webapp): Add Explore Page piechart

pavelpashkovsky:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1420

## Changes
- added PieChart based on tags
- DEMO: https://pr-1477.pyroscope.io/
- 
image


webapp/javascript/pages/TagExplorerView.module.scss
```diff
@@ -108,3 +108,13 @@
     }
   }
 }
+
+.statiscticsBox {
```
eh-am, line 112:
```suggestion
.statisticsBox {
```

webapp/javascript/pages/tagExplorer/components/PieChart/index.tsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import ReactFlot from 'react-flot';
+import { TimelineGroupData } from '@webapp/components/TimelineChart/TimelineChartWrapper';
+import styles from './styles.module.scss';
+import { calculateMean } from '../../../math';
+
+const PieChart = ({ data }: { data: TimelineGroupData[] }) => {
```
eh-am, line 7: I think this can be a generic/reusable `PieChart` component, and the transformations are done outside it.

eh-am (changes requested):

Rperry2174:
Great start!
image

two thoughts:
1. We should change the columns in the table to be:

~Tag name | Event count | avg samples | std deviation samples~
Tag name | avg samples | std deviation samples | total samples |

(removing Event count)

2. We should find a way to deal with when the pie chart has a lot of series or very small slices. Maybe we can look into how other popular libraries do it like [grafana](https://grafana.com/docs/grafana/latest/visualizations/pie-chart-panel/), [D3](https://d3-graph-gallery.com/pie)

One idea is maybe for the pie chart we only show the top 5 series and group everything else into "other" 

wdyt? cc @eh-am @petethepig

pavelpashkovsky:
In my opinion grouping small slices would be the best solution.

What I found about that:
https://support.microsoft.com/en-us/office/explode-or-expand-a-pie-chart-63284b67-22ea-4960-ab1e-0a3895af68ce

https://chandoo.org/wp/group-small-slices-in-pie-charts/

https://www.addtwodigital.com/add-two-blog/2021/2/13/rule-3-not-too-many-wedges-not-too-few

I think we can show top N, and put the rest of slices under the "Other". But what to do with the "Other"? open on click? How many levels of opening it might have? I would go with 2 levels for first time

pavelpashkovsky:
image

I pushed POC on condition if we show top 5 results (hide the rest under Other). It would be nice if you `/create-server` and give some feedback. 

I'm thinking that there is another option, we can show first N slices if they take place, for example, 70%, the rest of slices should be hidden under the "Other". We can reach the best result if we can combine these 2 methods, but honestly I can't provide any condition for it :)

Rperry2174:
Yeah this one is tricky... Let's try going this route:
image

And then only show the top X labels (default to something like 3 or 5)

pavelpashkovsky:
I set threshold value as 0.17, it means, if slice bigger than 17% it will render label, not ideal but my testing shows this is kinda "optimal" value. You can play with this value in 
`webapp/javascript/pages/tagExplorer/components/TotalSamplesChart/PieChart/index.tsx`
image

Rperry2174 (approved):
LGTM! I might play around with the sizing of the pie chart but I'll do that separately

PR #1489, opened 2022-09-12
feat(webapp): annotations UI

eh-am:
Render annotations created via the backend (creating annotation via UI will come in a new PR):
Kapture 2022-09-14 at 17 28 48

There's still some visual glitches that I would like to address (in separate PRs) before considering this feature done:
* https://github.com/pyroscope-io/pyroscope/pull/1508
* https://github.com/pyroscope-io/pyroscope/pull/1504

To create an annotation
```
curl -X POST -d '{ "appName": "pyroscope.server.cpu", "content": "hi i am an annotation", "timestamp": 1663187023 }' localhost:4040/api/annotations
```

You can get the current timestamp with `date +%s`


Rperry2174:
Long term I think it would be cool to make these annotations look something like this:
mockup_annotations_00

Where there are multiple "types" depending on the type of annotations

webapp/javascript/pages/ContinuousSingleView.tsx
```diff
@@ -117,7 +127,13 @@ function ContinuousSingleView() {
             title={
               <TimelineTitle titleKey={singleView?.profile?.metadata.units} />
             }
+            annotations={annotations}
             selectionType="single"
+            onHoverDisplayTooltip={(props) => {
+              if (props.values[0].closest[1]) {
+                return <div>{props.timeLabel}</div>;
+              }
```
pavelpashkovsky, line 135: you probably need some default value here (outside this if statement)

webapp/javascript/components/TimelineChart/TimelineChartWrapper.tsx
```diff
@@ -187,72 +191,67 @@ class TimelineChartWrapper extends React.Component<
...
+      prevProps.selection !== this.props.selection ||
+      prevProps.annotations !== this.props.annotations
+    ) {
       const newFlotOptions = this.state.flotOptions;
       newFlotOptions.grid.markings = this.plotMarkings();
       this.setState({ flotOptions: newFlotOptions });
     }
   }
 
   plotMarkings = () => {
```
pavelpashkovsky, line 204: nice simplification of `plotMarkings`

webapp/javascript/components/TimelineChart/Annotation.tsx
```diff
@@ -0,0 +1,70 @@
...
+  if (!values) {
+    return Maybe.nothing();
+  }
+  if (!values[0]) {
+    return Maybe.nothing();
+  }
+  if (!values[0].closest) {
+    return Maybe.nothing();
+  }
+  if (!values[0].closest[0]) {
```
pavelpashkovsky, line 55: maybe it makes sense to check positive condition instead of checking 4 negative ones?

webapp/javascript/components/TimelineChart/Annotation.tsx
```diff
@@ -0,0 +1,70 @@
...
+  }
+  return Maybe.of(values[0].closest[0]);
+}
+
+function getClosestAnnotation(
+  annotations: { timestamp: number; content: string }[],
+  timestamp: number
+) {
+  const f = annotations.find(
+    (a) => Math.abs(a.timestamp - timestamp) < THRESHOLD
```
pavelpashkovsky, line 66: as I understand you've added `THRESHOLD` to have kind of distance of watching markings, but does it the same for all markings? how does it work you have 3 annotations near each other? for example, on timestamps 15:30:25,  15:30:27, 15:30:28?

webapp/javascript/models/annotation.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const AnnotationSchema = z.object({
+  content: z.string(),
+  // TODO(eh-am): validate it's a valid unix timestamp
+  timestamp: z.number(),
+});
+
+export type Annotation = z.infer<typeof AnnotationSchema>;
+
+export const AnnotationsSchema = z.array(AnnotationSchema);
```
dogfrogfog, line 11: seems like you dont use it elsewhere, `export` can be removed

webapp/javascript/components/TimelineChart/Annotation.tsx
eh-am: used a solution, although not perfect, hopefully will fix in https://github.com/pyroscope-io/pyroscope/issues/1505

pavelpashkovsky: @eh-am, this problem might be a bit deeper, I have no idea why, but these annotation marks (https://github.com/pyroscope-io/pyroscope/issues/1505) looks like vertical selection lines in left/right timelines on ComparisonView page. Just take look at them more attentively. Dotted line + black draggable area. I added this appearance only for left/right single timeline selection lines, is there chance you pass some incorrect props? any kind of props conflict? But no idea what they are doing in annotations PR.
image

Rperry2174 (approved):
Lgtm

PR #1494, opened 2022-09-13
chore: Update tag explorer colors

Rperry2174:
Our current colors seem a little hard to distinguish so trying a different color palette

New (left) and Old (right)

image

image

image


petethepig (approved):
LGTM

PR #1499, opened 2022-09-14
feat: add 'perf script' format upload

korniltsev:
The goal is to upload perf data to pyroscope/flamegraph:
- First idea was to try to parse perf data at server side with [perf_data_converter](https://github.com/google/perf_data_converter). But the perf and converted pprof has no symbols so we can't do it on serverside
- Second idea was to parse `perf script` output (what this pr does) like `perf script | curl  --data-binary @- http://flamegraph.com`, the problem with it is that it may be huge (like 65M for 16 cores profiling for 10 seconds).
- Third idea is to use embeded stackcollapse script `perf script report stackcollapse | curl --data-binary @- flamegraph.com`  . stackcollapse is not huge so it is better. We might decline this PR and go with third solution. Only flamegraph.com changes are required, to allow to post profiles to `/`.

I doubt someone would like to upload huge profiles when they can upload short(collapsed/) profiles

This PR allows to upload `perf script` output to adhoc/flamegraph.com

To try:
1. record
```
perf record -a -g
```
2.a upload directly to flamegraph.com
```
perf script | curl  --data-binary @- http://flamegraph.com
```
2.b or save to file and upload to  adhoc/flamegraph.com
```
perf script > events
```

### DIFF
--- a/pkg/adhoc/server/convert.go
+++ b/pkg/adhoc/server/convert.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/pyroscope-io/pyroscope/pkg/convert/perf"
 	"sort"
 	"strconv"
 	"time"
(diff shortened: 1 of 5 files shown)


petethepig (approved):
LGTM

PR #1504, opened 2022-09-14
fix(backend): don't set an annotation timestamp in controller

eh-am:
Closes https://github.com/pyroscope-io/pyroscope/issues/1503


kolesnikovae (approved):
LGTM

PR #1508, opened 2022-09-14
fix(backend): upsert annotations

eh-am:
Closes https://github.com/pyroscope-io/pyroscope/issues/1507

### DIFF
--- a/pkg/service/annotation.go
+++ b/pkg/service/annotation.go
@@ -30,10 +30,11 @@ func (svc AnnotationsService) CreateAnnotation(ctx context.Context, params model
 	tx := svc.db.WithContext(ctx)
 
 	// Upsert
-	if err := tx.Where(model.CreateAnnotation{
-		AppName: params.AppName, Timestamp: params.Timestamp,
-	}).Attrs(u).FirstOrCreate(&u).Error; err != nil {
-		return nil, err
+	if tx.Where(model.CreateAnnotation{
+		AppName:   params.AppName,
+		Timestamp: params.Timestamp,
+	}).Updates(&u).RowsAffected == 0 {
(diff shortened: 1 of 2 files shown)


pkg/service/annotation.go
```diff
@@ -30,10 +30,11 @@ func (svc AnnotationsService) CreateAnnotation(ctx context.Context, params model
...
 	// Upsert
-	if err := tx.Where(model.CreateAnnotation{
-		AppName: params.AppName, Timestamp: params.Timestamp,
-	}).Attrs(u).FirstOrCreate(&u).Error; err != nil {
-		return nil, err
+	if tx.Where(model.CreateAnnotation{
+		AppName:   params.AppName,
+		Timestamp: params.Timestamp,
+	}).Updates(&u).RowsAffected == 0 {
+		return &u, tx.Create(&u).Error
```
kolesnikovae, line 37: This looks like a working solution, though I used slightly different approach in past to avoid conflicts – you may find it useful: https://gorm.io/docs/create.html#Upsert-x2F-On-Conflict

kolesnikovae (approved):
Looks good

pkg/service/annotation.go
eh-am: I believe no conflict will be generated since I haven't added a unique constraint on the columns /facepalm https://github.com/pyroscope-io/pyroscope/blob/main/pkg/sqlstore/migrations/migrations.go#L118-L119

Hopefully it will be solved via https://github.com/pyroscope-io/pyroscope/issues/1509

kolesnikovae: Ah, I see – thank you for the explanation!

PR #1510, opened 2022-09-15
fix(webapp): show annotations tooltip only when hovering close to the marker

eh-am:
Fixes https://github.com/pyroscope-io/pyroscope/issues/1505

Before:
Kapture 2022-09-14 at 16 27 27

After:
Kapture 2022-09-15 at 12 28 09

I use flot's `p2c` function (which stands to `point to canvas`) to figure out where a timestamp (here the annotation one) would be in the canvas. (Maybe there's a different way to do it accessing markers directly, but that is too tied to a specific implementation, since we want to move the impl to a plugin, i don't think it makes sense depending on a `marker`)

Then I just check if they are within a threshold (hardcoded to 3) of the cursor. If yes, display the tooltip.


pavelpashkovsky (approved):

pavelpashkovsky:
LGTM, have you checked that this logic breaks nothing on comparison, diff and explore pages?

eh-am:
> LGTM, have you checked that this logic breaks nothing on comparison, diff and explore pages?

I don't see how they are relevant, since this logic is only for annotations.

PR #1512, opened 2022-09-15
fix(webapp): weird marking when adding annotation

eh-am:
Closes https://github.com/pyroscope-io/pyroscope/issues/1497

Before:
image

After:
image

tl;dr
Problem is that we mix `markings` with `selection`, as pointed by this commented https://github.com/pyroscope-io/pyroscope/blob/d8e0d21bde92d7c5f6a07c9cb728607d1c2cd2c0/webapp/javascript/components/TimelineChart/TimelineChartWrapper.tsx#L205-L208

But anyway, the `selection` plugin ALWAYS 
https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/components/TimelineChart/Selection.plugin.ts#L425 draws a selection, using markings https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/components/TimelineChart/Selection.plugin.ts#L429 

But in the annotations case, we draw a marking that's not an annotation!

This will be solved after we create our own plugin (https://github.com/pyroscope-io/pyroscope/issues/1495), so that an annotation is a first class concept, as opposed to a marking as it currently is.

But right now this is a hotfix that should do the job.

### DIFF
--- a/webapp/javascript/components/TimelineChart/Selection.plugin.ts
+++ b/webapp/javascript/components/TimelineChart/Selection.plugin.ts
@@ -422,7 +422,10 @@ const handleHeight = 22;
     plot.hooks.draw.push(function (plot: PlotType, ctx: CtxType) {
       const opts = plot.getOptions();
 
-      if (opts?.selection?.selectionType === 'single') {
+      if (
+        opts?.selection?.selectionType === 'single' &&
+        opts?.selection?.selectionWithHandler
+      ) {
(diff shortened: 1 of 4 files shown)


Rperry2174 (approved):
LGTM

pavelpashkovsky (approved):
LGTM, but I will take closer look on rendering annotations plugin

PR #1513, opened 2022-09-15
chore: make pair (timestamp, appName) unique

eh-am:
Closes https://github.com/pyroscope-io/pyroscope/issues/1509

Plus refactors the `upsert` as per @kolesnikovae's here https://github.com/pyroscope-io/pyroscope/pull/1508#discussion_r972034413

Here's the sql schema (generated with `.schema annotations`):
```
CREATE TABLE `annotations` (`id` integer,`app_name` text NOT NULL DEFAULT null,`timestamp` datetime,`content` text NOT NULL DEFAULT null,`created_at` datetime,`updated_at` datetime,PRIMARY KEY (`id`));
CREATE UNIQUE INDEX `idx_appname_timestamp` ON `annotations`(`app_name`,`timestamp`);
```

---
offtopic/self documenting: had to turn on sql log to make sense out of what the orm is doing

```diff
diff --git a/pkg/sqlstore/sqlstore.go b/pkg/sqlstore/sqlstore.go
index 642ad7ca..06c7e429 100644
--- a/pkg/sqlstore/sqlstore.go
+++ b/pkg/sqlstore/sqlstore.go
@@ -57,7 +57,8 @@ func (s *SQLStore) openSQLiteDB() (err error) {
 		path = s.config.Database.URL
 	}
 	s.orm, err = gorm.Open(sqlite.Open(path), &gorm.Config{
-		Logger: logger.Discard,
+		//		Logger: logger.Discard,
+		Logger: logger.Default.LogMode(logger.Info),
 	})
 	s.db, err = s.orm.DB()
 	return err
```

### DIFF
--- a/pkg/service/annotation.go
+++ b/pkg/service/annotation.go
@@ -36,6 +36,9 @@ func (svc AnnotationsService) CreateAnnotation(ctx context.Context, params model
 			{Name: "app_name"},
 			{Name: "timestamp"},
 		},
+		// Update fields we care about
+		DoUpdates: clause.AssignmentColumns([]string{"app_name", "timestamp", "content"}),
+		// Update updateAt fields
 		UpdateAll: true,
 	}).Create(&u).Error; err != nil {
 		return nil, err
--- a/pkg/service/annotation.go
+++ b/pkg/service/annotation.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/pyroscope-io/pyroscope/pkg/model"
 	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
 )
 
 type AnnotationsService struct{ db *gorm.DB }
@@ -29,12 +30,14 @@ func (svc AnnotationsService) CreateAnnotation(ctx context.Context, params model
 
 	tx := svc.db.WithContext(ctx)
 
-	// Upsert
-	if tx.Where(model.CreateAnnotation{
-		AppName:   params.AppName,
-		Timestamp: params.Timestamp,
-	}).Updates(&u).RowsAffected == 0 {
-		return &u, tx.Create(&u).Error
+	if err := tx.Clauses(clause.OnConflict{
+		Columns: []clause.Column{
+			{Name: "app_name"},
+			{Name: "timestamp"},
+		},
+		UpdateAll: true,
+	}).Create(&u).Error; err != nil {
+		return nil, err
 	}
 
 	return &u, nil
--- a/pkg/service/annotation.go
+++ b/pkg/service/annotation.go
@@ -30,6 +30,7 @@ func (svc AnnotationsService) CreateAnnotation(ctx context.Context, params model
 
 	tx := svc.db.WithContext(ctx)
 
+	// Upsert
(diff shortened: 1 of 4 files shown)


pkg/service/annotation.go
```diff
@@ -30,11 +31,17 @@ func (svc AnnotationsService) CreateAnnotation(ctx context.Context, params model
...
+			{Name: "app_name"},
+			{Name: "timestamp"},
+		},
+		// Update fields we care about
+		DoUpdates: clause.AssignmentColumns([]string{"app_name", "timestamp", "content"}),
+		// Update updateAt fields
+		UpdateAll: true,
+	}).Create(&u).Error; err != nil {
+		return nil, err
 	}
```
eh-am, line 45: Not sure this is actually better.

There are a few things happening here:
1. We need to tell which fields we are interested in conflicting (`app_name`, `timestamp`), otherwise it defaults to the primary key. I decided to not use a natural primary key here, since we may want to for eg modify an existing annotation (change `appName` for example)
2. `DoUpdates` since we need to tell which columns we want to update. Otherwise it doesn't do anything.
3. `UpdateAll` to tell it to update `updateAt` field. Here's where it gets confusing: I thought `UpdateAll` would update everything (duh), but it seems to ignore the other fields.

petethepig (approved):

kolesnikovae (approved):

pkg/service/annotation.go
kolesnikovae: I guess we only need to update `content`:
```go
DoUpdates: clause.AssignmentColumns([]string{"content"}),
```

kolesnikovae:
Regarding SQL logging – I'd say that we should not enable it by default

kolesnikovae:
Honestly, I don't quite understand the restriction: why we can't have multiple annotations for the same application and the same moment of time?

pkg/sqlstore/migrations/migrations.go
```diff
@@ -132,3 +133,20 @@ func createAnnotationsTableMigration() *gormigrate.Migration {
 		},
 	}
 }
+
+func addIndexesUniqueTableMigration() *gormigrate.Migration {
+	type annotation struct {
+		AppName   string    `gorm:"index:idx_appname_timestamp,unique;not null;default:null"`
+		Timestamp time.Time `gorm:"index:idx_appname_timestamp,unique;not null;default:null"`
```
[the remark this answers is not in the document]
eh-am: I noticed, but not in migration thought right?

eh-am:
> Regarding SQL logging – I'd say that we should not enable it by default
Yeah, I didn't even commit it. I was just documenting for the next person.

>Honestly, I don't quite understand the restriction: why can't we have multiple annotations for the same application and the same moment of time?
From a technical perspective? None.

From a practical perspective? Couple reasons (at least):
* UI is just not ready for it (instead of a simple tooltip we would need to have a more robust component)
* we may want to implement some kind of threading. eg if i add an annotation on top of an existing annotation, there's some ordering implicit where the original one needs to be displayed first

PR #1524, opened 2022-09-19
feat(webapp): create annotations via ui

eh-am:
# What does it do
Allows creating an annotation via the UI in **SINGLE VIEW only**

https://user-images.githubusercontent.com/6951209/191333752-a0eb3b69-2ed7-4d65-b588-5ae6c67e06f5.mp4

# How it works
Whatever is passed to `TimelineChart`'s `ContextMenu` props is rendered when left-clicking the timeline. The `ContextMenu` and its `MenuItems` are created dynamically (see `AddAnnotationMenuItem` for more info). This is so that we can do different things when clicking on different points (eg [allow deleting/editing an existing annotation]( https://github.com/pyroscope-io/pyroscope/issues/1526)).

# Other
* Added a `popover` element which is essentially the same as [`ModalWithToggle`](https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/ui/Modals/ModalWithToggle.tsx), except it's not tied to a button. We should add a `anchorEl` functionality to `Popover` and get rid of `ModalWithToggle`.
* Added form integration with `react-form-hook` and `zod`.  **This should be the way we do form handling moving forward**. See example in stories/Form.stories.tsx
* There's a frontend feature flag called `isAnnotationsEnabled`. The idea is to disable it in the cloud until we implement the backend there.


Rperry2174:
A couple of questions:
1. Should the annotation persist in this PR on the PR server? (I assume maybe not since that is in the cloud ?) 
2. Are annotations scoped to the entire app (i.e. is there any context of what tags were present when the annotations are created)

webapp/javascript/pages/continuous/contextMenu/AddAnnotation.menuitem.tsx
```diff
@@ -0,0 +1,120 @@
...
+            <form
+              id="annotation-form"
+              name="annotation-form"
+              onSubmit={handleSubmit((d) => {
+                onCreateAnnotation(d.content);
+              })}
+            >
+              <TextField
+                {...register('content')}
+                label="Content"
```
Rperry2174, line 86: I think this might be better to say description but don't feel too strongly about that

eh-am:
@Rperry2174 


> 1. Should the annotation persist in this PR on the PR server? (I assume maybe not since that is in the cloud ?)
It uses local storage which in this case is sqlite, AFAIK remote write only works for `/ingest` data (https://github.com/pyroscope-io/pyroscope/blob/main/pkg/cli/server.go#L152) cc @petethepig 

 >2. Are annotations scoped to the entire app (i.e. is there any context of what tags were present when the annotations are created)
Yes. it's scoped to the entire app. Tags are stripped here https://github.com/pyroscope-io/pyroscope/blob/main/pkg/model/annotation.go#L63

webapp/javascript/pages/continuous/contextMenu/AddAnnotation.menuitem.tsx
eh-am: done!

Rperry2174 (approved):
LGTM!

PR #1540, opened 2022-09-22
chore: refactor adhoc server

kolesnikovae:
Note that all the components that depend on adhoc in other repos will be updated as well


Rperry2174:
> that all the components that depend on adhoc in other repos will be updated as well

Do you mean components or things like slackbot, lambda extension, etc?

kolesnikovae:
> Do you mean components or things like slackbot, lambda extension, etc?

Only flamegraph.com and cloudstorage as far as I can see. I should say that the changes in the dependent projects caused by this PR are pretty minor and do not affect the logic

pkg/api/router/router.go
```diff
@@ -78,8 +79,19 @@ func (r *Router) RegisterAPIKeyHandlers() {
...
+
+func (r *Router) RegisterAdhocHandlers() {
+	h := api.NewAdhocHandler(r.AdhocService, httputils.NewDefaultHelper(r.Logger))
+
+	x := r.PathPrefix("/adhoc/v1").Subrouter()
+	x.Methods(http.MethodGet).PathPrefix("/profiles").HandlerFunc(h.GetProfiles)
+	x.Methods(http.MethodGet).PathPrefix("/profile/{id:[0-9a-f]+}").HandlerFunc(h.GetProfile)
+	x.Methods(http.MethodGet).PathPrefix("/diff/{left:[0-9a-f]+}/{right:[0-9a-f]+}").HandlerFunc(h.GetProfileDiff)
+	x.Methods(http.MethodPost).PathPrefix("/upload").HandlerFunc(h.Upload)
+	x.Methods(http.MethodPost).PathPrefix("/upload-diff/").HandlerFunc(h.UploadDiff)
```
kolesnikovae, line 96: I didn't find any actual usage of the endpoint in the frontend. Do we need it?

/cc @eh-am

eh-am: that is correct, i guess it's a left over

pkg/adhoc/writer/external_writer.go
```diff
@@ -97,7 +98,20 @@ func (w *externalWriter) write(name string, out *storage.GetOutput) error {
...
 		}
 	case "html":
-		fb := flamebearer.NewProfile(filename, out, w.maxNodesRender)
+		fb := flamebearer.NewProfile(flamebearer.ProfileConfig{
+			Name:      filename,
+			MaxNodes:  w.maxNodesRender,
+			Tree:      out.Tree,
+			Timeline:  out.Timeline,
+			Groups:    out.Groups,
+			Telemetry: out.Telemetry,
```
eh-am, line 107: That's funny, none of these seem to be used/relevant when writing externally (ie when exporting). I don't think they should be part of `FlamebearerProfile` struct

pkg/server/controller.go
```diff
@@ -207,6 +201,9 @@ func (ctrl *Controller) serverMux() (http.Handler, error) {
 		AuthService:        ctrl.authService,
 		UserService:        ctrl.userService,
 		AnnotationsService: ctrl.annotationsService,
+		AdhocService: service.NewAdhocService(
+			ctrl.config.MaxNodesSerialization, //
```
eh-am, line 205:
```suggestion
			ctrl.config.MaxNodesSerialization,
```

eh-am (approved):
Looks good, have a minor comment about `FlamebearerProfile`, but it's kinda unrelated to this PR.

pkg/adhoc/writer/external_writer.go
kolesnikovae: I agree, `FlamebearerProfile` is super bloated and hard to maintain. We need to revise this type

pkg/server/controller.go
kolesnikovae: Oh, right – thanks!

kolesnikovae:
@petethepig would you mind if merge this?

PR #1544, opened 2022-09-23
chore: add a dialog modal

eh-am:
Add a generic dialog, based off https://mui.com/base/api/modal-unstyled/

* breaks down the api into 4 elements: `<Dialog><DialogBody><DialogHeader><DialogFooter>`
* supports scrolling `<DialogBody>` (see video below)
* automatically set a backdrop

Kapture 2022-09-23 at 16 34 38

Related https://github.com/pyroscope-io/pyroscope/issues/1542

After this we should refactor the multiple dialogs we have to use this one.


Rperry2174 (approved):
Nice! One small nit is I feel like it looks weird when buttons are extra small so for the "OK" button I would give it a minimum width, but don't feel super strongly about that.

PR #1546, opened 2022-09-26
chore(webapp): rename /service to /api

eh-am:
Rename `/services` to `/api`
This should make it more clear that it's strictly about the api communication layer (ie IO)
Plus, it allows us to scope more generic code under `services`, as opposed to `utils` as we have currently.


petethepig (approved):
Looks good, can you make sure to update everything on the cloud side too?

eh-am:
will deal with this later

PR #1562, opened 2022-09-27
feat(annotations): allow creating multiple annotations for different apps

eh-am:
support passing an `appNames` array


korniltsev (approved):
nit: I am not sure having both fields is a good idea, maybe just keep AppNames?

eh-am:
>nit: I am not sure having both fields is a good idea, maybe just keep one array field AppNames?

Haha @petethepig and I discussed quite a lot about it, I wasn't so fond of it at the beginnig, but I guess the decision is that `appName` works well when you are creating a single annotation (eg via UI, which returns a single annotation), and `appNames` is just a utility for when creating via the API.

PR #1563, opened 2022-09-27
chore: set a transparent border on buttons

eh-am:
existing buttons with variation (secondary/primary) have `border: none`, which when paired with normal buttons show this lack of border, making them having different sizes:

Screen Shot 2022-09-27 at 16 04 33

This PR just sets a border with background color, so that's counted as the button size but not really visible.

Before:
image

After:
image


petethepig (approved):

Rperry2174 (approved):
Nice!

PR #1565, opened 2022-09-28
feat: Heatmap should show error message if no data is returned

dogfrogfog:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1551

## Changes
- add common NoData component
- change "no data" message
- handle heatmap null value
- hide flamegraph when heatmap is null

https://user-images.githubusercontent.com/47758224/192804104-2b496db3-76a1-4b3e-a81d-fd03e48505c6.mov

## Concerns
-


packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphRenderer.tsx
```diff
@@ -467,7 +467,11 @@ class FlameGraphRenderer extends React.Component<
               styles.panesWrapper
             )}`}
           >
-            {dataUnavailable ? <NoProfilingData /> : panes.map((pane) => pane)}
+            {dataUnavailable ? (
+              <NoData dataName="profiling" />
+            ) : (
+              panes.map((pane) => pane)
+            )}
```
eh-am, line 474: I think this can be simplified to simply `No data available`, without specifying a qualifier.

webapp/javascript/components/Heatmap/utils.ts
```diff
@@ -89,42 +89,42 @@ export const getSelectionData = (
...
+  (min: number, max: number, timezone: string) => (v: number) => {
+    const d = getUTCdate(
+      new Date(v / 1000000),
+      timezone === 'utc' ? 0 : new Date().getTimezoneOffset()
+    );
+    // nanoseconds -> hours
+    const hours = (max - min) / 60 / 60 / 1000 / 1000 / 1000;
+
+    return getTimelineFormatDate(d, hours);
+  };
```
eh-am, line 102: I not a big fan of having this logic repeat in multiple places (https://github.com/pyroscope-io/pyroscope/issues/1521), can we extract common logic to Can we move that here https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/util/formatDate.ts
And here only have the part that's >strictly< related to Heatmap?

dogfrogfog: ye, i will refactor it

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphRenderer.tsx
dogfrogfog: `No data available` instead of `No profiling data available for this application / time range.` ?

packages/pyroscope-flamegraph/src/format/format.ts
```diff
@@ -101,10 +101,12 @@ class DurationFormatter {
...
-      this.units ||
-      `${this.suffix}${n === 1 || this.suffix.length === 2 ? '' : 's'}`
-    }`;
+    return withUnits
+      ? `${nStr} ${
+          this.units ||
+          `${this.suffix}${n === 1 || this.suffix.length === 2 ? '' : 's'}`
+        }`
+      : nStr;
   }
```
eh-am, line 110: What's going on here? Can you add a test to demonstrate this logic? https://github.com/pyroscope-io/pyroscope/blob/59d34aaf79d7a8a873d29543cbc722d8fffd4896/packages/pyroscope-flamegraph/src/format/format.spec.ts

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphRenderer.tsx
eh-am: I think so, then we don't need all this new code. cc @Rperry2174

packages/pyroscope-flamegraph/src/format/format.ts
dogfrogfog: sure, but that's the part of different PR

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphRenderer.tsx
dogfrogfog: there was a `NoProfilingData` component which is div + span + styles to add paddings and center text, i moved the whole component to the _webapp/ui_ folder....so its not that new code

eh-am: Ah I meant allowing custom title (`profile` | `heatmap`), that's a level of customization I don't think we really need it.

Rperry2174 (approved):
lgtm!

PR #1569, opened 2022-09-29
fix: merge zstd lib generated so names for jfr

korniltsev:
changes 
`libzstd-jni-1.5.1-16931311898282279136.so.Java_com_github_luben_zstd_ZstdInputStreamNoFinalizer_decompressStream`
to 
`libzstd-jni-1.5.1-_.so.Java_com_github_luben_zstd_ZstdInputStreamNoFinalizer_decompressStream`

### DIFF
--- a/pkg/convert/jfr/parser.go
+++ b/pkg/convert/jfr/parser.go
@@ -270,10 +270,14 @@ var lambdaGeneratedEnclosingClass = regexp.MustCompile("^(.+\\$\\$Lambda\\$)\\d+
 // libzstd-jni-1.5.1-16931311898282279136.so.Java_com_github_luben_zstd_ZstdInputStreamNoFinalizer_decompressStream
 var zstdJniSoLibName = regexp.MustCompile("^(libzstd-jni-\\d+\\.\\d+\\.\\d+-)(\\d+)(\\.so)$")
 
+// ./tmp/libamazonCorrettoCryptoProvider109b39cf33c563eb.so
+var amazonCorrettoCryptoProvider = regexp.MustCompile("^\\./tmp/(libamazonCorrettoCryptoProvider)([0-9a-f]{16})(\\.so)$")
+
 func mergeJVMGeneratedClasses(frame string) string {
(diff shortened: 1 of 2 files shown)


pkg/convert/jfr/parser.go
```diff
@@ -268,9 +267,13 @@ var generatedMethodAccessor = regexp.MustCompile("^(jdk/internal/reflect/Generat
 // org/example/rideshare/OrderService$$Lambda$669.0x0000000800fd7318.run
 var lambdaGeneratedEnclosingClass = regexp.MustCompile("^(.+\\$\\$Lambda\\$)\\d+[./](0x[\\da-f]+|\\d+)$")
 
+// libzstd-jni-1.5.1-16931311898282279136.so.Java_com_github_luben_zstd_ZstdInputStreamNoFinalizer_decompressStream
+var zstdJniSoLibName = regexp.MustCompile("^(libzstd-jni-\\d+\\.\\d+\\.\\d+-)(\\d+)(\\.so)$")
```
petethepig, line 271: we should definitely expand this to more jni libraries, not just libzstd

petethepig (changes requested):

pkg/convert/jfr/parser.go
korniltsev: ./tmp/libamazonCorrettoCryptoProvider109b39cf33c563eb.so
./tmp/libamazonCorrettoCryptoProvider492aaa424fafd7e0.so

petethepig: I mean more like make it so that it would work with any arbitrary library

korniltsev: I don't get it
This naming logic is somewhat app/lib specific.
For example pyroscope adds a sha1 to a library name, amazon 8byte hex, zstd some long integer.

petethepig: ah, I didn't realize that, I thought this was something standardized

petethepig (approved):

PR #1584, opened 2022-10-01
chore: update deps in dockerfile

korniltsev:
- removed rbspy, phpspy, rustdeps from building inside docker
- update alpine to 3.16, go to 1.19, php to 7.4, node to 18
- add `apk upgrade` to pickup latest patches

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,3 @@
-#        _
-#       | |
-#  _ __ | |__  _ __  ___ _ __  _   _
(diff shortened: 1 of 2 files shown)


petethepig (changes requested):
Looks great. We should remove phpspy too

korniltsev:
Remove php.
I am not sure about removing php though. People might still use it and we have no replacement similar to ruby&python

petethepig (approved):

PR #1587, opened 2022-10-04
refactor: tree serialization improvements

petethepig:
This PR improves tree truncation algorithm. This results in less nodes being dropped (and replaced with `other` nodes).

This also introduces ability to set `max-nodes-serialization` to `-1` which effectively disables truncation.

I ran some benchmarks with a ruby app with a very leafy flamegraph. I ingested 5000 profiles and measured storage size, wall time and cpu time for the ingestions and put together a table that represents the difference between old version, new version with defaults and new version with `max-nodes-serialization` set to -1.

  | storage size | profile time | flamegraph
-- | -- | --| --
old version | 220 MB | 47s | https://flamegraph.com/share/747cf1c2-4ab2-11ed-81e2-52cfe0251908
new version defaults | 228 MB | 52s | https://flamegraph.com/share/261eae6a-4aac-11ed-81e2-52cfe0251908
new version -1 | 293 MB | 52s | https://flamegraph.com/share/31edb3ac-4ab0-11ed-81e2-52cfe0251908

### DIFF
--- a/.air.conf
+++ b/.air.conf
@@ -25,7 +25,7 @@ delay = 1000 # ms
 
 send_interrupt = true
 
-kill_delay = 100 # ms
+kill_delay = 2000000000 # ns
 
 # Stop to run old binary when build errors occur.
 stop_on_error = true
--- a/cmd/pyroscope/command/convert.go
+++ b/cmd/pyroscope/command/convert.go
@@ -39,7 +39,7 @@ func parse(input io.Reader, format string) error {
 		parser(input, func(name []byte, val int) {
 			t.Insert(name, uint64(val))
 		})
-		return t.SerializeNoDict(4096, os.Stdout)
+		return t.SerializeTruncateNoDict(4096, os.Stdout)
 	case "trie":
 		t := transporttrie.New()
 		parser(input, func(name []byte, val int) {
--- a/pkg/storage/tree/flamebearer.go
+++ b/pkg/storage/tree/flamebearer.go
@@ -21,7 +21,7 @@ type Flamebearer struct {
 	Format     Format `json:"format"`
 }
 
-var otherJsonableSlice = jsonableSlice("other")
+var lostDuringRenderingName = jsonableSlice("other")
 
 func (t *Tree) FlamebearerStruct(maxNodes int) *Flamebearer {
 	t.RLock()
@@ -81,7 +81,7 @@ func (t *Tree) FlamebearerStruct(maxNodes int) *Flamebearer {
 			otherTotal := uint64(0)
 			var otherNode *treeNode
 			for _, n := range tn.ChildrenNodes {
-				if bytes.Equal(n.Name, otherName) {
+				if bytes.Equal(n.Name, lostDuringRenderingName) {
 					otherTotal += n.Total
 					continue
 				}
@@ -96,7 +96,7 @@ func (t *Tree) FlamebearerStruct(maxNodes int) *Flamebearer {
 			}
 			if otherTotal != 0 {
 				otherNode = &treeNode{
-					Name:  otherJsonableSlice,
+					Name:  lostDuringRenderingName,
 					Total: otherTotal,
 					Self:  otherTotal,
 				}
--- a/pkg/storage/tree/flamebearer.go
+++ b/pkg/storage/tree/flamebearer.go
@@ -1,5 +1,7 @@
 package tree
 
+import "bytes"
+
 type Format string
 
 const (
@@ -18,6 +20,8 @@ type Flamebearer struct {
 	Units      string `json:"units"`
 	Format     Format `json:"format"`
 }
+
+var otherJsonableSlice = jsonableSlice("other")
 
 func (t *Tree) FlamebearerStruct(maxNodes int) *Flamebearer {
 	t.RLock()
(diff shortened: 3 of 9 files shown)


.air.conf
```diff
@@ -25,7 +25,7 @@ delay = 1000 # ms
 
 send_interrupt = true
 
-kill_delay = 100 # ms
+kill_delay = 2000000000 # ns
```
petethepig, line 28: turns out that values is in nanoseconds, hence the change — this is timeout that `air` will use before doing a `kill -9` on the process

cmd/pyroscope/command/convert.go
```diff
@@ -39,7 +39,7 @@ func parse(input io.Reader, format string) error {
 		parser(input, func(name []byte, val int) {
 			t.Insert(name, uint64(val))
 		})
-		return t.SerializeNoDict(4096, os.Stdout)
+		return t.SerializeTruncateNoDict(4096, os.Stdout)
```
petethepig, line 42: I changed the name to be consistent

pkg/storage/tree/flamebearer.go
```diff
@@ -19,6 +21,8 @@ type Flamebearer struct {
 	Format     Format `json:"format"`
 }
 
+var lostDuringRenderingName = jsonableSlice("other")
```
petethepig, line 24: Maybe we could change this to something else in the future, e.g "lost during rendering"

pkg/storage/tree/flamebearer.go
```diff
@@ -75,7 +79,12 @@ func (t *Tree) FlamebearerStruct(maxNodes int) *Flamebearer {
 
 			xOffset += int(tn.Self)
 			otherTotal := uint64(0)
+			var otherNode *treeNode
 			for _, n := range tn.ChildrenNodes {
+				if bytes.Equal(n.Name, lostDuringRenderingName) {
+					otherTotal += n.Total
+					continue
+				}
```
petethepig, line 87: it's possible that there's already an "other" node from serialization. In this case we're storing the value and combine it with the values lost during rendering

pkg/storage/tree/minval.go
```diff
@@ -1,8 +1,13 @@
 package tree
 
-import "github.com/pyroscope-io/pyroscope/pkg/structs/cappedarr"
+import (
+	"github.com/pyroscope-io/pyroscope/pkg/structs/cappedarr"
+)
 
 func (t *Tree) minValue(maxNodes int) uint64 {
+	if maxNodes == -1 {
```
petethepig, line 8: This allows for `max-nodes-serialization=-1` values in configuration

pkg/storage/tree/serialize.go
```diff
@@ -13,48 +13,7 @@ import (
 // serialization format version. it's not very useful right now, but it will be in the future
 const currentVersion = 1
 
-func (t *Tree) Serialize(d *dict.Dict, maxNodes int, w io.Writer) error {
```
petethepig, line 16: not used anywhere, superseded by `SerializeTruncate`

pkg/storage/tree/serialize.go
```diff
@@ -13,48 +13,7 @@ import (
...
-			nodes = append(tn.ChildrenNodes, nodes...)
-		}
-		_, err = varint.Write(w, cnl)
-		if err != nil {
-			return err
-		}
-	}
-	return nil
-}
+var lostDuringSerializationName = []byte("other")
```
petethepig, line 16: similar idea here, maybe we'll change this name at some point

pkg/storage/tree/serialize.go
```diff
@@ -79,64 +38,38 @@ func (t *Tree) SerializeTruncate(d *dict.Dict, maxNodes int, w io.Writer) error
...
+			}
 			if cn.Total >= minVal {
-				tn.ChildrenNodes = append(tn.ChildrenNodes, cn)
+				largeEnoughNodes = append(largeEnoughNodes, cn)
+			} else {
+				// Truncated children accounted as parent self.
+				other += cn.Total
 			}
 		}
-		if len(tn.ChildrenNodes) > 0 {
```
petethepig, line 94: this diff is unfortunate. I'd recommend just viewing the file

pkg/storage/tree/serialize.go
```diff
@@ -273,55 +207,8 @@ func DeserializeNoDict(r io.Reader) (*Tree, error) {
...
-		_, err = varint.Write(w, cnl)
-		if err != nil {
-			return err
-		}
-	}
-	return nil
-}
-
+// used in the cloud
+// this function modifies the tree, so should only be used once!
```
petethepig, line 211: @kolesnikovae I hope this is intentional

pkg/storage/tree/serialize.go
```diff
@@ -340,19 +227,34 @@ func (t *Tree) SerializeTruncateNoDict(maxNodes int, w io.Writer) error {
 			return err
 		}
 
-		val := tn.Self
+		if _, err = vw.Write(w, tn.Self); err != nil {
+			return err
+		}
 		cNodes := tn.ChildrenNodes
 		tn.ChildrenNodes = tn.ChildrenNodes[:0]
```
petethepig, line 234: I don't fully get what we're gaining by doing this, but I left it untouched. I did not copy this to the other implementation (`SerializeTruncate`) because we don't want to modify the tree there

pkg/storage/tree/truncation_test.go
```diff
@@ -0,0 +1,58 @@
...
+				buf := &bytes.Buffer{}
+				treeA.SerializeTruncateNoDict(3, buf)
+				treeB, err := DeserializeNoDict(buf)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(treeB.StringWithEmpty()).To(Equal(treeStr(`"b" 2|"c" 3|"ls" 1|`)))
+			})
+		})
+	})
+
+	// TODO(petethepig): add more tests
```
petethepig, line 57: I want to add tests with big trees but I'm afraid they're going to be too complex to comprehend, so not sure it's worth it

PR #1589, opened 2022-10-04
Support speedscope format

### LINKED ISSUE #172: Add support for speedscope format
We already have support for [a few different formats](https://github.com/pyroscope-io/pyroscope/blob/dotnet/pkg/convert/parser.go#L15-L70), including pprof.

It would be great if we added support for [speedscope format](https://gist.github.com/petethepig/c6041b26002e218ef54e67c5c6f3909a). It's somewhat similar to pprof, but it's json, not protobuf.

### TODO

* [ ] add a new parser method to [parser.go](https://github.com/pyroscope-io/pyroscope/blob/dotnet/pkg/convert/parser.go#L27)
* [ ] add a test for this new parser

vasi-stripe:
This supports both variants (evented and sampled) of Speedscope's [native format](https://github.com/jlfwong/speedscope/blob/main/src/lib/file-format-spec.ts). This would allow supporting more agents, eg [pyinstrument](https://github.com/joerick/pyinstrument).

Ingestion using the API appears to work with some sample profiles I have lying arounds. I added a couple of fixture tests [from upstream](https://github.com/jlfwong/speedscope/tree/main/sample/profiles/speedscope).

Possible issues:
* Technically, the times in Speedscope format are allowed to be doubles. It looks like they really are non-integers for pyinstrument! But the tree API only allows uint64 weights, and there's no simple way to decide what resolution the client intends.
* It's awkward to figure out the right units. The Speedscope format mixes up the concepts of time-units (s/ms/µs) and measurement-units (bytes/allocations/time). I guess at least for time-based units, we base sampleRate on them?
* It's not obvious to me how we should build keys when there are multiple profiles present of the same type
* It's not obvious how we should tell when to use metadata from the HTTP params vs. the profile format's metadata. Would be nice if we could distinguish between "HTTP param default" vs. "HTTP client explicitly chose this".
* It looks like adhoc mode uses a whole different set of converters, so I didn't build that for Speedscope (yet?)
* Evented mode is often used for flamegraphs, but Pyroscope doesn't really do those, we prefer to aggregate. Seems fine, as long as users expect this.

Closes #172

### DIFF
--- a/pkg/convert/speedscope/json.go
+++ b/pkg/convert/speedscope/json.go
@@ -1,0 +1,65 @@
+package speedscope
+
+// Description of Speedscope JSON
+// See spec: https://github.com/jlfwong/speedscope/blob/main/src/lib/file-format-spec.ts
+
+const (
+	schema = "https://www.speedscope.app/file-format-schema.json"
+
+	profileEvented = "evented"
+	profileSampled = "sampled"
+
+	unitNone         = "none"
+	unitNanoseconds  = "nanoseconds"
+	unitMicroseconds = "microseconds"
+	unitMilliseconds = "milliseconds"
+	unitSeconds      = "seconds"
+	unitBytes        = "bytes"
+
+	eventOpen  = "O"
+	eventClose = "C"
+)
+
+type speedscopeFile struct {
+	Schema             string `json:"$schema"`
+	Shared             shared
+	Profiles           []profile
+	Name               string
+	ActiveProfileIndex int
+	Exporter           string
+}
+
+type shared struct {
+	Frames []frame
+}
+
+type frame struct {
+	Name string
+	File string
+	Line int
+	Col  int
+}
+
+type profile struct {
+	Type       string
+	Name       string
+	Unit       string
+	StartValue int64
+	EndValue   int64
+
+	// Evented profile
+	Events []event
+
+	// Sample profile
+	Samples []sample
+	Weights []int64
+}
+
+type event struct {
+	Type  string
+	At    int64
+	Frame int
+}
+
+// Indexes into Frames
+type sample []int
--- a/pkg/convert/speedscope/json.go
+++ b/pkg/convert/speedscope/json.go
@@ -25,7 +25,7 @@ type speedscopeFile struct {
 	Shared             shared
 	Profiles           []profile
 	Name               string
-	ActiveProfileIndex int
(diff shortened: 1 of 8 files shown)


vasi-stripe:
@petethepig I think the npmjs test failure is spurious?

petethepig:
@vasi-stripe Yeah, that job shoudn't run for forks. You can ignore it for now. We've also disabled it in [this PR](https://github.com/pyroscope-io/pyroscope/pull/1594)

vasi-stripe:
Great! I think the only blockers then are:

* Dealing with doubles
* Anything else serious you find on review ;-)

Any thoughts on how to manage non-integer weights? Does Pyroscope assume that weights are in multiples of sampleRate?

vasi-stripe:
Filed an issue on speedscope about confusing units: https://github.com/jlfwong/speedscope/issues/406

petethepig:
@vasi-stripe It's looking good! Let me think a little more about the double thing and time units / sample rate. I'll review tonight.

petethepig (approved):
Everything looks and works great — I tried a few speedscope jsons I found online and they were getting ingested without any issues.

The only things I noticed that I figured we should improve before merging are:

* as @vasi-stripe mentioned, dealing with `double` values (we only support integers in the storage engine)
* dealing with time units such as seconds or milliseconds (most of our integrations deal with samples, which are then combined with `sampleRate` to render time units, e.g seconds). `pyinstrument` uses seconds and without sample rate adjustments the numbers on the frontend don't match values in speedscope.

I figured this is getting deep into the intricacies of our storage engine, so I put together a [PR that addresses these](https://github.com/vasi-stripe/pyroscope/pull/1). @vasi-stripe I can't commit this to your repo, but I think you should be able to merge that PR in your repo and it will become a part of this PR in `pyroscope-io/pyroscope`.

@vasi-stripe & @kolesnikovae — would love your all's feedback on it and if it's looking good let's merge it and do a release :)

kolesnikovae:
Looks really good. Nice job @vasi-stripe!

@petethepig I left a [comment](https://github.com/vasi-stripe/pyroscope/pull/1#issuecomment-1268340513) in the patch PR as it mostly concerns the storage internals and the ingestion path on the whole

vasi-stripe:
Ok, merged in the units PR. Thanks for that!

Could you re-run actions?

PR #1591, opened 2022-10-04
fix: make adhoc table fit width

dogfrogfog:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1581

## Changes
- make adhoc table fit width
- change adhoc table time format to _MMM d, yyyy_ at _H:mm_ 
- add "tooltip" (`title` span attribute) to Filename cells
- ~~add suffix column, the `suffix` value should be part of `/api/adhoc/v1/profiles` reponse~~

Screenshot 2022-10-05 at 12 28 33
Screenshot 2022-10-05 at 12 25 59

## Concerns
- ~~FileNames column values are hidden when  screen width is small (thats not a problem since we are adding suffix column?)~~
Screenshot 2022-10-05 at 12 27 16

### DIFF
--- a/webapp/javascript/components/FileList.tsx
+++ b/webapp/javascript/components/FileList.tsx
@@ -30,6 +30,12 @@ const getBodyRows = (
       (profId) => profId === profile.id
     );
 
+    const date = parseISO(profile.updatedAt);
+    const timeString = `${format(date, 'MMM d, yyyy')} at ${format(
+      date,
+      'H:MM'
+    )}`;
+
     acc.push({
       cells: [
         {
@@ -40,7 +46,7 @@ const getBodyRows = (
             </div>
           ),
         },
-        { value: format(parseISO(profile.updatedAt), 'yyyy-MM-dd') },
+        { value: timeString },
       ],
       onClick: () => {
         // Optimize to not reload the same one
--- a/webapp/javascript/components/FileList.tsx
+++ b/webapp/javascript/components/FileList.tsx
@@ -9,8 +9,10 @@ import styles from './FileList.module.scss';
 
 const dateModifiedColName = 'updatedAt';
 const fileNameColName = 'name';
+const suffixColName = 'suffix';
 const headRow = [
   { name: fileNameColName, label: 'Filename', sortable: 1 },
+  { name: suffixColName, label: 'Suffix' },
   {
     name: dateModifiedColName,
     label: 'Date Modified',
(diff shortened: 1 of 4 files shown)


webapp/javascript/components/FileList.tsx
```diff
@@ -33,13 +34,13 @@ const getBodyRows = (
...
-              {profile.name}
+            <div className={styles.profileName}>
+              <span title={profile.name}>{profile.name}</span>
               {isRowSelected && <CheckIcon className={styles.checkIcon} />}
-            </>
+            </div>
           ),
         },
-        { value: profile.updatedAt },
+        { value: format(parseISO(profile.updatedAt), 'yyyy-MM-dd') },
```
eh-am, line 43: @Rperry2174 Not sure I understand the reasoning behind this. What if I am uploading multiple files (via `adhoc` command for example) in a short interval (minutes) and I am basing off that info to navigate through my files.

eh-am (commented):
I like having the names fully visible, specially if you have a bunch of profiles (`.cpu`, `.mutex`, `.inuse_objects`) that would be all trimmed to `rideshare-app-golang...`, but from design purposes I guess it makes sense making it "prettier".

webapp/javascript/components/FileList.tsx
Rperry2174: sounds good for time lets do this format:
image

Lets add one more column with the suffix:
`cpu`, `mutex`, `inuse_objects`, etc

dogfrogfog:
updated PR description

dogfrogfog:
https://user-images.githubusercontent.com/47758224/194075862-e798ed9b-631f-4252-819c-b1a73e854d83.mov

webapp/javascript/components/FileList.tsx
```diff
@@ -29,17 +30,23 @@ const getBodyRows = (
       (profId) => profId === profile.id
     );
 
+    const date = parseISO(profile.updatedAt);
+    const timeString = `${format(date, 'MMM d, yyyy')} at ${format(
+      date,
+      'H:mm'
+    )}`;
```
eh-am, line 37: I think we are missing `PM/AM` there
image

eh-am: Yeah, I like the 24h but it's definitely confusing, specially when it's alone like this. Plus we can say we have a large US audience.

Rperry2174: I like 24h format I think this is fine

eh-am:
Closes https://github.com/pyroscope-io/pyroscope/issues/1338

Rperry2174 (approved):
lgtm

PR #1593, opened 2022-10-04
chore: quick refactor to reuse query extracting logic

eh-am:
Small refactor.

While reviewing the code I noticed one can't compare/diff 2 distinct apps, this is the reason why.
Plus used the `queryToAppName` utility instead of parsing it manually, just to keep things tidy.


webapp/javascript/hooks/populateLeftRightQuery.hook.ts
```diff
@@ -1,18 +1,24 @@
...
-function isQueriesHasSameApp(queries: string[]): boolean {
-  const appName = queries[0].split('{')[0];
+function isQueriesHasSameApp(queries: Query[]): boolean {
+  const appName = queryToAppName(queries[0]);
+  if (appName.isNothing) {
+    return false;
+  }
 
-  return queries.every((query) => query.match(appName));
+  return queries.every((query) => query.match(appName.value));
```
petethepig, line 12: Would this be better?
```suggestion
  return queries.every((query) => query.startsWith(appName.value));
```

petethepig (approved):
added a comment, other than that looks good

PR #1594, opened 2022-10-04
build: skip publish npm if user not part of pyroscope org

eh-am:
Previously when PRs were created by a non member of the pyroscope org, this `checkout` step would fail, due to the lack of access to the secrets:

image

This PR basically skips the entire pipeline if that secret is not present.


Rperry2174 (commented):
LGTM

Rperry2174 (approved):

PR #1603, opened 2022-10-06
chore: display loading indicator

eh-am:
One common complaint is not being able to visually tell data is stale or not.

This PR adds a visual indicator when data is `loading | reloading`. It's important to understand the difference: `reloading` just means we already have data, so can easily display an overlay on top of existing component. Now `loading` means initial load, which we should use [skeletons](https://mui.com/material-ui/react-skeleton/) instead in the future. To be fair, it already works well with the timelines. It's just the flamegraph that's missing an empty state.

Did only for Continuous{Single,Comparison,Diff} pages. Tried to do for Tag Explorer but it's not that straightforward. May do in an isolated PR.

Also adds a delay of 250ms before showing the overlay. Reasoning is that when requests are fast enough we don't need to display a feedback, so it feels snappy.

image

---
Github messed up the diff. Only thing I am doing with the pages is nesting all components within a `<LoadingOverlay>`
When reviewing the code


petethepig:
Looks great!

webapp/javascript/ui/Box.module.scss
```diff
@@ -3,6 +3,7 @@
   border: 1px solid var(--ps-ui-border);
   border-radius: 4px;
 
+  position: relative;
   //  display: inline-block;
```
pavelpashkovsky, line 7: can you please remove commented code

webapp/javascript/redux/reducers/continuous.ts
```diff
@@ -845,8 +841,14 @@ export const continuousSlice = createSlice({
...
     builder.addCase(fetchSideTimelines.pending, (state) => {
-      state.leftTimeline = { ...state.leftTimeline, type: 'loading' };
-      state.rightTimeline = { ...state.rightTimeline, type: 'loading' };
+      state.leftTimeline = {
+        ...state.leftTimeline,
+        type: getNextStateFromPending(state.leftTimeline.type),
+      };
+      state.rightTimeline = {
+        ...state.rightTimeline,
+        type: getNextStateFromPending(state.leftTimeline.type),
```
pavelpashkovsky, line 850: does `type` mean loading type? does it make sense to name this prop something like `loadingType`?

what the point to track next state from pending?

webapp/javascript/pages/loading.ts
```diff
@@ -0,0 +1,7 @@
+export function isLoadingOrReloading(
+  states: ('pristine' | 'loading' | 'reloading' | 'loaded')[]
+) {
+  return states.every((state) => {
```
pavelpashkovsky, line 4: [].every means `true` if all array items return `true`, just a question, didn't you consider returning `true` from this function using [].some? in my point of view, if component requires data from 2 sources, and one of them is loading, it means there is no enough data for rendering this component, so this component is in `loading` state

webapp/javascript/ui/LoadingOverlay.tsx
```diff
@@ -0,0 +1,58 @@
...
+  children?: React.ReactNode;
+  /** delay in ms before showing the overlay. this evicts a flick */
+  delay?: number;
+}) {
+  const [isVisible, setVisible] = useState(false);
+
+  // Wait for `delay` ms before showing the overlay
+  // So that it feels snappy when server is fast
+  // https://www.nngroup.com/articles/progress-indicators/
+  useEffect(() => {
```
pavelpashkovsky, line 28: delay is not required, so maybe it makes sense to not to use setTimeout at all if you don't pass `delay` prop? actually I'm not a fan of hardcoding any kind of static value of loading time :)

webapp/javascript/ui/LoadingOverlay.module.css
```diff
@@ -0,0 +1,20 @@
...
+  top: 0;
+  bottom: 0;
+  display: flex;
+  justify-content: center;
+
+  /* ignore all pointer events */
+  pointer-events: all;
+  z-index: 99;
+  transition: all 0.5s ease-in-out;
+  background: rgba(0, 0, 0, 0.7);
```
dogfrogfog, line 14: maybe background-color ?

webapp/javascript/ui/LoadingOverlay.tsx
```diff
@@ -0,0 +1,58 @@
...
+export function LoadingOverlay({
+  active = true,
+  spinnerPosition = 'center',
+  children,
+  delay = 250,
+}: {
+  /** where to position the spinner. use baseline when the component's vertical center is outside the viewport */
+  spinnerPosition?: 'center' | 'baseline';
+  active: boolean;
+  children?: React.ReactNode;
```
dogfrogfog, line 19: use without `React.` prefix ?

webapp/javascript/ui/LoadingOverlay.tsx
```diff
@@ -0,0 +1,58 @@
...
+          styles.loadingOverlay,
+          !isVisible ? styles.unactive : null
+        )}
+        style={{
+          alignItems: spinnerPosition,
+        }}
+      >
+        <LoadingSpinner size="46px" />
+      </div>
+
```
dogfrogfog, line 54: we need this empty line ?

webapp/javascript/pages/ContinuousSingleView.tsx
```diff
@@ -134,29 +136,40 @@ function ContinuousSingleView() {
       </ContextMenu>
     );
   };
-
   return (
     <div>
       <PageTitle title={formatTitle('Single', query)} />
       <div className="main-wrapper">
         <Toolbar />
+
```
dogfrogfog, line 144: empty line

webapp/javascript/ui/LoadingOverlay.tsx
eh-am: Actually the default is 250ms. This is intentional and a design choice. For fast operations (<250ms we don't want to show any loading indication).

From Nielson Norman group:
image

Same principle, but I found 1 second too broad so changed for `250ms`.

webapp/javascript/pages/loading.ts
eh-am: Good point! Updated!

webapp/javascript/redux/reducers/continuous.ts
eh-am: `type` means 
Screen Shot 2022-10-13 at 13 40 25

> what the point to track next state from pending?
It's just an utility. Imagine a state machine where given current state + action is `pending`, it returns what the next state should be. Its logic basically is: do we already have data? If so we are `reloading`, otherwise we are just `loading`.

PR #1605, opened 2022-10-06
feat(webapp): Annotations flot plugin

pavelpashkovsky:
## Brief
- add annotations plugin for timelines

## Changes
- demo screencast:

https://user-images.githubusercontent.com/7372044/195332521-958413e6-03ca-4e25-bf0a-d060dd58b519.mp4

- demo:
https://pr-1605.pyroscope.io/?query=rideshare-app-golang.alloc_objects%7B%7D

## Concerns
-


webapp/javascript/components/TimelineChart/Annotations.plugin.tsx
```diff
@@ -0,0 +1,255 @@
+import React from 'react';
+import * as ReactDOM from 'react-dom';
+import { randomId } from '@webapp/util/randomId';
+import { PlotType, CtxType, EventHolderType } from './types';
```
eh-am, line 4: is it possible to use flotjs types?

webapp/javascript/components/TimelineChart/TimelineChartWrapper.tsx
```diff
@@ -148,6 +148,7 @@ class TimelineChartWrapper extends React.Component<
         //   a position and a nearby data item object as parameters.
         clickable: true,
       },
+      annotations: null,
```
eh-am, line 151: i think a default value is simply an empty array no?

pavelpashkovsky: we have to separate cases 
- `annotations: []` (empty array) - we allow using annotation for this timeline, but at this moment annotations list is empty, user can add some
- `annotations: [Annotation, Annotation]` (non-empty array) - the same here, user can see and add annotations
- `annotations: null` - disable annotations functionality

Rperry2174:
Looks great so far. Let's have the annoatation be a little bit smaller (we can focus on UI / design of it at the end if you'd like) 
mockup_annotations_00

webapp/javascript/components/TimelineChart/TimelineChartWrapper.tsx
eh-am: what's the difference between annotations functionality being "disabled" and simply no annotation existing?

webapp/javascript/components/TimelineChart/Annotations.plugin.tsx
```diff
@@ -0,0 +1,285 @@
...
+  return {
+    x: Math.floor(extractedX.axis.p2c(positionInTimestamp.x)) + plotOffset.left,
+    y: Math.floor(extractedY.axis.p2c(positionInTimestamp.y)) + plotOffset.top,
+  };
+};
+
+const findAnnotationByCursorPosition = (
+  x: number,
+  y: number,
+  list: AnnotationPosition[] = []
```
eh-am, line 92: I get your reasoning but I find weird setting a default value >here<. This is waaay down the tree. If we have to set a default, it should be high up the tree. From maintainability it's much easier to just assume it's always present than making us second guess ourselves on every step.

webapp/javascript/components/TimelineChart/Annotations.plugin.tsx
```diff
@@ -0,0 +1,285 @@
...
+) => {
+  return list?.find((an) => {
+    return x >= an.fromX && x <= an.toX && y >= an.fromY && y <= an.toY;
+  });
+};
+
+function drawRoundRect(
```
eh-am, line 99: Although I like to keep things close to where they are used, this is too much of an implementation detail. I don't want to know about canvas drawing (when reading about annotations). This and the code that draws an icon on the canvas can be moved to its own file I think.

webapp/javascript/components/TimelineChart/Annotations.plugin.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+        ReactDOM.render(
+          <ContextMenu
+            click={{ ...pos }}
+            containerEl={containerEl}
+            timestamp={timestamp}
+            value={value}
+          />,
+          containerEl
+        );
```
eh-am, line 187: I think we can organize this better...Annotations plugin should not be responsible for rendering a Context Menu... Annotations plugin should only be responsible for rendering Annotations.

webapp/javascript/components/TimelineChart/Annotations.plugin.tsx
```diff
@@ -0,0 +1,285 @@
...
+        const { x, y } = getCursorPositionInPx(plot, pos);
+
+        const annotation = findAnnotationByCursorPosition(
+          x,
+          y,
+          annotationsPositions
+        );
+
+        if (annotation) {
+          placeholder.trigger('hoveringOnAnnotation', [{ hovering: true }]);
```
eh-am, line 143: Dunno about this... message passing like this to a completely different plugin feels very hacky...

webapp/javascript/components/TimelineChart/Annotations.plugin.tsx
```diff
@@ -0,0 +1,285 @@
...
+    plot.hooks!.draw!.push((plot, ctx: CtxType) => {
+      const o: IFlotOptions = plot.getOptions();
+
+      if (o.annotations?.length) {
+        const axes = plot.getAxes();
+        const plotOffset: { top: number; left: number } = plot.getPlotOffset();
+        const extractedX = extractRange(plot, axes, 'x');
+        const extractedY = extractRange(plot, axes, 'y');
+
+        o.annotations.forEach((a: AnnotationType) => {
```
eh-am, line 200: As I said in some other comment, when reading the functionality one doesn't really want to be concerned about canvas drawing. Move this to its own function please.

webapp/javascript/components/TimelineChart/Annotations.plugin.tsx
```diff
@@ -0,0 +1,285 @@
...
+          const left: number =
+            Math.floor(extractedX.axis.p2c(a.timestamp * 1000)) +
+            plotOffset.left;
+          const yMax =
+            Math.floor(extractedY.axis.p2c(extractedY.axis.min)) +
+            plotOffset.top;
+          const yMin = 0 + plotOffset.top;
+          const lineWidth = 1;
+          const subPixel = lineWidth / 2 || 0;
```
eh-am, line 209: what is a subpixel? what is this necessary?

webapp/javascript/components/TimelineChart/Selection.plugin.ts
```diff
@@ -346,6 +315,18 @@ const handleHeight = 22;
     plot.setSelection = setSelection;
     plot.getSelection = getSelection;
 
+    function onHoveringOnAnnotation(
+      _: EventType,
+      value?: { hovering: boolean }
+    ) {
+      // during further complication this logic should be moved to separated plugin
+      // which only handles cursor state over Flot
+      // because cursor setters conflict between each other
```
eh-am, line 324: Yeah not a fan of this at all,

If we implement icons as react elements this won't be a problem correct?

webapp/javascript/components/TimelineChart/TimelineChartWrapper.tsx
```diff
@@ -212,27 +213,36 @@ class TimelineChartWrapper extends React.Component<
...
+  composeAnnotationsList = () => {
+    return Array.isArray(this.props.annotations)
+      ? this.props.annotations?.map((a) => ({
+          timestamp: a.timestamp,
+          content: a.content,
+          type: 'message',
+          color: ANNOTATION_COLOR,
+        }))
+      : null;
+  };
```
eh-am, line 230: Already said somewhere else, but please don't add a third state to an array. An array is either empty or non-empty. If you need to have another state make it clear with another variable.

webapp/javascript/components/TimelineChart/extractRange.ts
```diff
@@ -0,0 +1,43 @@
+/* eslint-disable */
+// function taken from markings support in Flot
+export default function extractRange(
```
eh-am, line 3: I don't understand what this does, can you add a tsdoc comment?

webapp/javascript/pages/ContinuousSingleView.tsx
```diff
@@ -114,6 +115,19 @@ function ContinuousSingleView() {
     if (!isAnnotationsEnabled) {
       return null;
     }
+
+    if (props.value) {
```
eh-am, line 119: I don't understand this, when `props.value` exist we show annotation info? What is `props.value`? What does `ContextMenu` have to do with `AnnotationInfo`?

eh-am (changes requested):
Curious to hear your thoughts on why you preferred to draw on the canvas VS rendering Icons using React.

---
I think `shouldStartAnnotationsFunctionality` is unnecessary. If it's empty we don't render annotations, if it's not empty we render them. An array either has items or does not has items. Adding a third `null` state is unnecessary. If we need to control whether the entire plugin is activated or not, then we can create another variable `enableAnnotationsPlugin` or something, although I don't think this is necessary.

---
I know you have your own preferences, but for consistency let's use `function` instead of lambda functions.

webapp/javascript/components/TimelineChart/TimelineChartWrapper.tsx
pavelpashkovsky: >"disabled"

user can't see and create annotations, whole annotations functionality is disabled

> simply no annotation existing

annotations list is empty, but user can add some ones

pavelpashkovsky:
@eh-am annotation marks (square + icon) moved outside of Flot, now it's React Component. I found a way how to position them over the timeline correctly. it used to be a problem they lost correct positioning when user changes browser window width

webapp/javascript/pages/ContinuousSingleView.tsx
pavelpashkovsky: not relevant, I removed this stuff

webapp/javascript/components/TimelineChart/Selection.plugin.ts
pavelpashkovsky: removed

webapp/javascript/components/TimelineChart/Annotations.plugin.tsx
pavelpashkovsky: Honestly I got this in built-in flot rendering engine. As I understand flot uses this approach for accurate positioning some elements on X axis (+/-0.5px).

webapp/javascript/components/TimelineChart/Annotations.plugin.tsx
pavelpashkovsky: removed

webapp/javascript/components/TimelineChart/Annotations.plugin.tsx
```diff
@@ -0,0 +1,285 @@
...
+    left: number;
+  };
+
+  return {
+    x: Math.floor(extractedX.axis.p2c(positionInTimestamp.x)) + plotOffset.left,
+    y: Math.floor(extractedY.axis.p2c(positionInTimestamp.y)) + plotOffset.top,
+  };
+};
+
+const findAnnotationByCursorPosition = (
```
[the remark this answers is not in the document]
pavelpashkovsky: removed

webapp/javascript/components/TimelineChart/Annotations.plugin.tsx
```diff
@@ -0,0 +1,285 @@
...
+    return $(`#${WRAPPER_ID}`);
+  }
+
+  const body = $('body');
+  return $(`<div id="${WRAPPER_ID}" />`).appendTo(body);
+};
+
+const getCursorPositionInPx = (
+  plot: jquery.flot.plot,
+  positionInTimestamp: { x: number; y: number }
```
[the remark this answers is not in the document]
pavelpashkovsky: removed

webapp/javascript/components/TimelineChart/Annotations.plugin.tsx
pavelpashkovsky: removed

webapp/javascript/components/TimelineChart/Annotations.plugin.tsx
pavelpashkovsky: removed

webapp/javascript/components/TimelineChart/Annotations.plugin.tsx
pavelpashkovsky: can you please explain a bit more. This function opens context menu on click on canvas. Maybe I have to move this outside of flot's plugin as function?

eh-am:
It's looking good!! Couple thoughts:

image
Right now it's possible that the icon covers the title.
---
There's no visual difference between a Creation/Edit modal and a "Read" Modal

Read Modal
image

Create Modal
image

I think, for Read modal, we could make it readonly so that it has a dark background.
Honestly we should in the near future have a diff modal. Right now it's not very optimized for reading the data (the tooltip was a bit better in that regard, since you could read most of the content at once).

Rperry2174:
Yeah I agree with @eh-am that the "read" modal fields should both have dark backgrounds 

https://user-images.githubusercontent.com/6951209/195377738-74f88832-a8ca-4b98-b4c7-a95de64c04e2.png

Maybe eventually we'll add an edit button but that will be separate

pavelpashkovsky:
@eh-am , @Rperry2174 , made this input dark
image

eh-am:
RE: Icon covering the title:
image

Can you increase the margin a little bit? `20px` would work I think
image

eh-am:
Also is it possible to have visual feedback when hovering over an icon? In cases like the following it's difficult to tell which one I am hovering on.
image

webapp/javascript/pages/continuous/contextMenu/AnnotationInfo.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+interface AnnotationInfo {
+  container: AddAnnotationProps['container'];
+  popoverAnchorPoint: AddAnnotationProps['popoverAnchorPoint'];
+  timestamp: AddAnnotationProps['timestamp'];
+  timezone: AddAnnotationProps['timezone'];
+  value: { content: string; timestamp: number };
+  isOpen: boolean;
+  setIsOpen: Dispatch<SetStateAction<boolean>>;
+  popoverClassname?: string;
```
eh-am, line 19: Can you add back the tsdoc comments we have?
https://github.com/pyroscope-io/pyroscope/blob/53836ce6b2f4a5eb2debc6a2e03c0b39812bf488/webapp/javascript/pages/continuous/contextMenu/AddAnnotation.menuitem.tsx#L21-L28

webapp/javascript/pages/continuous/contextMenu/AnnotationInfo.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { useAnnotationForm } from './useAnnotationForm';
+
+interface AnnotationInfo {
+  container: AddAnnotationProps['container'];
+  popoverAnchorPoint: AddAnnotationProps['popoverAnchorPoint'];
+  timestamp: AddAnnotationProps['timestamp'];
+  timezone: AddAnnotationProps['timezone'];
+  value: { content: string; timestamp: number };
+  isOpen: boolean;
+  setIsOpen: Dispatch<SetStateAction<boolean>>;
```
eh-am, line 18:
```suggestion
  onClose: () => void;
```

From my understanding, the annotation itself can only **close** itself.

webapp/javascript/pages/continuous/contextMenu/useAnnotationForm.ts
```diff
@@ -0,0 +1,48 @@
+import * as z from 'zod';
+import { zodResolver } from '@hookform/resolvers/zod';
+import { useForm } from 'react-hook-form';
+import { format } from 'date-fns';
+import { getUTCdate, timezoneToOffset } from '@webapp/util/formatDate';
+
+interface IseAnnotationFormProps {
```
eh-am, line 7: We don't prefix our interfaces with `I`

webapp/javascript/pages/continuous/contextMenu/AddAnnotation.menuitem.tsx
```diff
@@ -69,43 +57,43 @@ function AddAnnotation(props: AddAnnotationProps) {
...
-                )}
-              />
-            </form>
-          </PopoverBody>
-          <PopoverFooter>
-            <Button type="submit" kind="secondary" form="annotation-form">
-              Save
-            </Button>
-          </PopoverFooter>
+          {isPopoverOpen ? (
```
eh-am, line 63: Instead of doing ternary conditionals, prefer to fail early.
Example

```tsx
if (!isPopoverOpen) {
  return null;
}
```

webapp/javascript/components/TimelineChart/Annotations.plugin.tsx
```diff
@@ -0,0 +1,119 @@
...
+};
+
+interface IFlotOptions extends jquery.flot.plotOptions {
+  annotations?: AnnotationType[];
+  wrapperId?: string;
+}
+
+(function ($) {
+  function init(plot: jquery.flot.plot & jquery.flot.plotOptions) {
+    plot.hooks!.draw!.push((plot, ctx: CtxType) => {
```
eh-am, line 24:
```suggestion
    plot.hooks!.draw!.push((plot, ctx: CanvasRenderingContext2D) => {
```
I don't think we need custom types here.

webapp/javascript/components/TimelineChart/Annotations.plugin.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+          drawAnnotationLine({
+            ctx,
+            yMin: plotOffset.top,
+            yMax:
+              Math.floor(extractedY.axis.p2c(extractedY.axis.min)) +
+              plotOffset.top,
+            left,
+            color: a.color,
+          });
```
eh-am, line 51: this is better, but from a reader pov i prefer to see a `renderAnnotationInTimeline` or something, which then can call these other 2 functions as you currently have

webapp/javascript/components/TimelineChart/Annotations.plugin.tsx
```diff
@@ -0,0 +1,119 @@
...
+})(jQuery);
+
+function drawAnnotationLine({
+  ctx,
+  color,
+  left,
+  yMax,
+  yMin,
+}: {
+  ctx: CtxType;
```
eh-am, line 72:
```suggestion
  ctx: CanvasRenderingContext2D;
```

webapp/javascript/components/TimelineChart/Annotations.plugin.tsx
```diff
@@ -0,0 +1,119 @@
...
+  options,
+  left,
+}: {
+  annotation: AnnotationType;
+  options: { wrapperId?: string };
+  left: number;
+}) {
+  const annotationMarkElementId = 'annotation_mark_'.concat(
+    String(annotation.timestamp)
+  );
```
eh-am, line 97: This doesn't seem to account for multiple timelines with their own annotations.

webapp/javascript/components/TimelineChart/AnnotationMark/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import AnnotationInfo from '@webapp/pages/continuous/contextMenu/AnnotationInfo';
+import useTimeZone from '@webapp/hooks/timeZone.hook';
+import { faCommentDots } from '@fortawesome/free-regular-svg-icons/faCommentDots';
+
+import styles from './styles.module.scss';
+
+interface IAnnotationMarkProps {
+  type: 'message';
+  color: Color;
```
eh-am, line 12: Pretty sure color can be inferred from the `type` (which we currently only have `message`). Each type is gonna be an item/bg color, which is static. We won't allow people to mix and match these. At least not in the foreseeable future.

webapp/javascript/components/TimelineChart/AnnotationMark/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <div
+        data-testid="annotation_mark_wrapper"
+        onClick={onClick}
+        style={{ background: color.hex() }}
+        className={styles.wrapper}
+        role="none"
+      >
+        <FontAwesomeIcon className={styles.icon} icon={getIcon(type)} />
+      </div>
+      {target ? (
```
eh-am, line 49: Early return please!

webapp/javascript/components/TimelineChart/AnnotationMark/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+  switch (type) {
+    case 'message':
+    default:
+      return faCommentDots;
+  }
+};
+
+const AnnotationMark = ({ type, color, value }: IAnnotationMarkProps) => {
+  const { offset } = useTimeZone();
+  const [visible, toggle] = useState(false);
```
eh-am, line 29:
```suggestion
  const [visible, setVisible] = useState(false);
```

just to make things more explicit

cypress/integration/webapp/annotations.ts
```diff
@@ -0,0 +1,38 @@
+describe('Annotations', () => {
+  it('add annotation flow works as expected', () => {
+    cy.visit('/').wait(3000);
```
eh-am, line 3: Also don't wait an arbitrary amount of time (here 3000 ms). You most likely are waiting on a network request, which you can mock.

Eg https://github.com/pyroscope-io/pyroscope/blob/main/cypress/integration/webapp/serviceDiscovery.ts#L4-L9

pavelpashkovsky:
@eh-am sorry, probably I deleted comments accidentally during dev process and then forgot to revert them

webapp/javascript/components/TimelineChart/Annotations.plugin.tsx
pavelpashkovsky: nice, removed my custom `CtxType`

webapp/javascript/components/TimelineChart/ContextMenu.plugin.tsx
```diff
@@ -21,63 +17,48 @@ export interface ContextMenuProps {
...
-              click={{ ...pos }}
-              containerEl={containerEl}
-              timestamp={timestamp}
-            />
-          </Provider>,
+          <ContextMenu
+            click={{ ...pos }}
+            containerEl={containerEl}
+            timestamp={Math.round(pos.x / 1000)}
+          />,
```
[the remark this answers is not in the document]
pavelpashkovsky: you need provider if you want to import any kind of redux logic inside of the body of context menu (for example, redux selectors, useDispatch etc.). but in fact it executes functions which are already were initialized in components connected to redux. You can just take a look at `webapp/javascript/pages/ContinuousSingleView.tsx`. function `dispatch` was initialized in `ContinuousSingleView`, you just execute it in ContextMenu "env"

webapp/javascript/components/TimelineChart/ContextMenu.plugin.tsx
```diff
@@ -21,63 +17,48 @@ export interface ContextMenuProps {
...
     }
 
-    const flotEl = plot.getPlaceholder();
-
-    // Register events and shutdown
-    // It's important to bind/unbind to the SAME element
-    // Since a plugin may be register/unregistered multiple times due to react re-rendering
-
-    // TODO: not entirely sure when these are disabled
     if (plot.hooks?.bindEvents) {
```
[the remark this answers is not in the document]
pavelpashkovsky: removed by mistake, reverted

webapp/javascript/components/TimelineChart/Annotations.plugin.tsx
pavelpashkovsky: >this is better, but from a reader pov i prefer to see a renderAnnotationInTimeline or something, which then can call these other 2 functions as you currently have

reworked this part, but can you please double check it to confirm that I understood you correctly

>Plus I found mark confusing. In flotjs a marking is a simple line/rectangular areas. Now you are calling mark the actual icon. That's confusing.
I renamed it to `renderAnnotationIcon`, but I'm not really happy with this decision because it's not just icon. but I can't come up anything better :) any suggestion?

webapp/javascript/components/TimelineChart/AnnotationMark/index.tsx
pavelpashkovsky: can you please explain more what you want to see here

cypress/integration/webapp/annotations.ts
pavelpashkovsky: not sure it would work in this case because after getting response this page takes some time for rendering

pavelpashkovsky: but let me research it better :)

Rperry2174 (approved):
lgtm!

PR #1610, opened 2022-10-10
perf: optimize allocations in dictionaries

kolesnikovae:
Currently we observe high CPU time consumption of Go GC in certain cases:
image

Significant part of the objects are allocated at tree serialisation when node labels are encoded using dictionaries: notice that the function is called twice, on TTL interval and at memory-pressure-based eviction – in practice, it can reach ~40% of objects:
image

It is interesting that the second biggest consumer is `varint.NewWriter` which allocates a tiny buffer.


petethepig (approved):

pkg/storage/dict/dict.go
```diff
@@ -69,10 +70,15 @@ func (t *Dict) readValue(key Key, w io.Writer) bool {
...
 func (t *Dict) Put(val Value) Key {
 	t.m.Lock()
 	defer t.m.Unlock()
-	var buf bytes.Buffer
-	t.root.findNodeAt(val, &buf)
-	return buf.Bytes()
+	b := bufferPool.Get()
+	defer bufferPool.Put(b)
+	t.root.findNodeAt(val, b)
+	k := make([]byte, b.Len())
```
petethepig, line 81: technically we're still going to make this allocation here

I wonder if we could avoid doing that somehow too

kolesnikovae: Yeah, but we don't need to re-allocate it in most cases. I think it should be a caller option to provide a buffer

kolesnikovae: Made it possible to use a caller-provided buffer. The call is only used at tree serialisation which you work on: I think we can slightly improve performance by doing something like:

```go
var b bytes.Buffer // Can be taken from a pool.
for _, n := range nodes {
   b.Reset()
   // This cast does not cause an allocation: n.Name is of jsonableSlice type which is []byte. 
   d.PutTo(&b, []byte(n.Name))
   vw.Write(w, uint64(b.Len())
   b.WriteTo(w)
   // ...
}
```

I think this should me done after tree serialisation refactoring

PR #1613, opened 2022-10-11
feat: add sandwich view for table/flamegraph

dogfrogfog:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1586

## Changes
- add additional option `sandwich` to Flamegraph component
- add sandwich view to Flamegraph component

demo link : https://pr-1613.pyroscope.io/?query=rideshare-app-golang.alloc_objects%7B%7D

https://user-images.githubusercontent.com/47758224/198092939-83a6701f-a514-4474-a7f3-477cc06bb6a9.mov

## Concerns
- not sure yet how to combine bottom nodes in callers flamegraph since flamegraph only knows how to grow from top to bottom(maybe you have any ideas?)
- we should have empty top lvl node to accumulate values and to "grow" flamegraph from top to bottom (or it's okay ?)
Screenshot 2022-10-26 at 13 03 18


dogfrogfog:
https://user-images.githubusercontent.com/47758224/195132696-8ab85d4d-4f17-4670-b4a2-390275158a42.mov

dogfrogfog:
https://user-images.githubusercontent.com/47758224/195645542-f646753d-28c5-49b3-96d8-450d433afd6c.mov

dogfrogfog:
https://user-images.githubusercontent.com/47758224/195649911-d7802c8b-a846-4e4f-9a8b-085183310924.mov

packages/pyroscope-flamegraph/src/convert/sandwichViewProfiles.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+        'total',
+        'name-2-2',
+        'name-3-1',
+        'specific-function-name',
+        'name-5-1',
+        'name-5-2',
+      ],
+      numTicks: 400,
+      sampleRate: 100,
+    });
```
eh-am, line 36: I think this is a bad idea. With a test i should able to inspect its intent by reading the code/artifacts. Here the `flamebearer` format is for all effects, gibberish.

Can you convert it back to another representation? I suggest the `tree` one which is what most people are used to.

eh-am: So what I am suggesting is:

Create a tree by hand.
Convert to flamebearer using one our functions.
Run calleesFlamebearer
Convert its output to a tree
Compare if the output tree is the same as EXPECTED_TREE (something you create manually)

This should be a test made of pure functions (input -> output), no need to mock anything.

eh-am (commented):
I don't think we should call it `sandwich view` in the Flamegraph's component internals. That name is purely for the end user. Internally we should call it what it is: `callees` and `callers`

packages/pyroscope-flamegraph/src/convert/sandwichViewProfiles.ts
```diff
@@ -0,0 +1,107 @@
...
+  const copy = JSON.parse(JSON.stringify(p));
+  copy.flamebearer.levels = deltaDiffWrapper('single', copy.flamebearer.levels);
+  const calleesResultFlamebearer = calleesFlamebearer(
+    copy.flamebearer,
+    nodeName
+  );
+  calleesResultFlamebearer.levels = deltaDiffWrapperReverse(
+    'single',
+    calleesResultFlamebearer.levels
+  );
```
eh-am, line 95: I don't get why this is needed. We do `deltaDiffWrapper` and then `deltaDiffWrapperReverse`, is it because `calleesFlamebearer` expect things in delta format? If so, shouldn't all this be part of `calleesFlamebearer`?

Rperry2174: I believe @petethepig is the one to answer this and not sure if he knows either 😅

packages/pyroscope-flamegraph/src/convert/sandwichViewProfiles.spec.ts
Rperry2174: I don't think it's a "bad" idea -- curious why you did choose this format instead of a tree @dogfrogfog ? Did you consider the tree too?

packages/pyroscope-flamegraph/src/convert/sandwichViewProfiles.ts
eh-am: the function we export

dogfrogfog:
> I don't think we should call it `sandwich view` in the Flamegraph's component internals. That name is purely for the end user. Internally we should call it what it is: `callees` and `callers`

agree, will rename it

packages/pyroscope-flamegraph/src/convert/sandwichViewProfiles.spec.ts
dogfrogfog: so i chose flamebearer because this is the function's return value...but ye I think the way Eduardo suggested is better and easier to understand... I will refactor this part

dogfrogfog:
https://user-images.githubusercontent.com/47758224/197020592-6b790aab-eec2-4745-9655-eb5c3a872cd0.mov

petethepig:
Played with the sandwich view some, it looks awesome! Great job @dogfrogfog !

I have some feedback:
* I think we should make it so that when you first open Sandwich view it shows the right pane without a flamegraph and tells the user to click an item from the table on the left, something like "⬅️ Select a function on the left to get started (cc @Rperry2174  for the exact phrasing, arrow). When I first opened the sandwich tab it was not clear to me that I needed to click an item in the table.
* I think we should add a button to the right click menu that would allow a user to go into sandwich view from a flamegraph
Screen Shot 2022-10-26 at 10 48 19 PM

Rperry2174:
Let's also use this for the sandwich logo: 

sandwich_logo_02

packages/pyroscope-flamegraph/src/convert/sandwichViewProfiles.spec.ts
```diff
@@ -0,0 +1,27 @@
+import { treeToFlamebearer, calleesFlamebearer } from './sandwichViewProfiles';
+import { flamebearersToTree } from './flamebearersToTree';
+
+import { tree, name22FunctionTreeWithTotal } from './testData';
+
+describe('Sandwich view profiles', () => {
+  it('should return correct callees flamebearer for single function appearance', () => {
```
eh-am, line 7: I think we are missing couple other cases:

1. No data (eg no callee), not sure it will be possible (after all, how is it possible to not have a single callee?) in code but still
2. Single callee (it seems you already have this one)
3. Multiple callees

Also where are the tests for callers?

packages/pyroscope-flamegraph/src/convert/sandwichViewProfiles.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+    const f = treeToFlamebearer(tree);
+
+    const resultCalleesFlamebearer = calleesFlamebearer(
+      // @ts-ignore
```
eh-am, line 11: AFAIK this disable typescript for the entire file, why do we need this?

eh-am: I think it's better to do a type assertion than simply disabling ts.

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphRenderer.tsx
```diff
@@ -431,9 +448,96 @@ class FlameGraphRenderer extends React.Component<
...
+      const callersFlamebearer = callersProfile(
+        this.state.flamebearer,
+        activeItemName
+      );
+      const calleesFlamebearer = calleesProfile(
+        this.state.flamebearer,
+        activeItemName
+      );
+
+      return (
```
eh-am, line 473: Both `Graph` are pretty similar right? Can we use a function to simplify this?

```jsx
const sandwitchGraph = (myCustomParams) => {
 return (<Graph ... {...mycustomParams} />);
}
```

eh-am (commented):
This is pretty cool, well done!

Found a couple quirks though:

1. Managed to get a consistent crash:

Click on a node, click on sandwich view, select an item in the table.

2. If I select an item in sandwich view then go to `flamegraph`, that item is still selected, alright. But when I go from a selected item in `flamegraph` to `sandwich` view the selection is cleared, which is confusing.

eh-am:
Also I don't see why we couldn't zoom a node in sandwich view. Seems useful since you can't always see the full name  eg
image

Rperry2174:
> Also I don't see why we couldn't zoom a node in sandwich view. Seems useful since you can't always see the full name  eg
> 
> image
> 
> 

I agree but let's save that for a separate pr

Rperry2174:
> 
> 
> This is pretty cool, well done!
> 
> 
> 
> Found a couple quirks though:
> 
> 
> 
> 1. Managed to get a consistent crash:
> 
> 
> 
> Click on a node, click on sandwich view, select an item in the table.
> 
> 
> 
> 2. If I select an item in sandwich view then go to `flamegraph`, that item is still selected, alright. But when I go from a selected item in `flamegraph` to `sandwich` view the selection is cleared, which is confusing.


When you say "item" do you mean flamegraph node or table row? Same with selection

packages/pyroscope-flamegraph/src/convert/sandwichViewProfiles.spec.ts
dogfrogfog: will add more cases

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphRenderer.tsx
```diff
@@ -341,6 +351,12 @@ class FlameGraphRenderer extends React.Component<
   };
 
   updateView = (newView: ViewTypes) => {
+    if (newView === 'sandwich') {
+      this.setState({
+        selectedItem: Maybe.nothing(),
```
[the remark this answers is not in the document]
dogfrogfog: good point, additionaly reseting `focusNode` and `zoom` values fixes the crush
> 1. Managed to get a consistent crash:

packages/pyroscope-flamegraph/src/convert/sandwichViewProfiles.spec.ts
dogfrogfog: talking about `callers`......i tried to test it the same way i test `callees` and seems like [flamebearersToTree](https://github.com/pyroscope-io/pyroscope/blob/e1a7860572c944408bb92b141d3b7ce5ed06ad1d/packages/pyroscope-flamegraph/src/convert/flamebearersToTree.ts) doesnt not parse the tree node if it's total is equal to zero..i implemented callers flamegraph using empty (with `total: [0]` prop) top lvl node to accumulate values.... so `flamebearersToTree` cant return correct to test

packages/pyroscope-flamegraph/src/FlameGraph/FlamegraphRenderer.module.scss
```diff
@@ -33,6 +33,48 @@
...
+    display: inline-block;
+    position: relative;
+    margin: 30px 0 0 10px;
+    padding: 8px;
+    border-radius: 5px;
+    background-color: var(--ps-sandwich-pane-info-bg);
+    color: var(--ps-neutral-1);
+
+    &::before {
+      content: '◀';
```
petethepig, line 52: Nit: I'd maybe use the [border trick](https://css-tricks.com/snippets/css/css-triangle/) instead here simply because this arrow might render differently on different OSs

petethepig (approved):
The demo looks awesome!

PR #1615, opened 2022-10-14
feat(webapp): Issue when comparison / diff timelines are out of range [WIP]

pavelpashkovsky:


eh-am:
I think we can solve this at a different level.

[Given https://demo-dev.pyroscope.io/comparison?query=rideshare-app-golang.alloc_objects%7B%7D&rightQuery=rideshare-app-golang.alloc_objects%7B%7D&leftQuery=rideshare-app-golang.alloc_objects%7B%7D&from=1665749249&until=1665749701](https://demo-dev.pyroscope.io/comparison?query=rideshare-app-golang.alloc_objects%7B%7D&rightQuery=rideshare-app-golang.alloc_objects%7B%7D&leftQuery=rideshare-app-golang.alloc_objects%7B%7D&from=1665749249&until=1665749701&leftFrom=1665749323&leftUntil=1665749397&rightFrom=1665749512&rightUntil=1665749601)

if the interval `leftFrom` and `leftUntil` is outside the [`from`, `until`] range, display a notification, maybe on top of the timeline? We have an errorMessage ocmponent somewhere, I think it could be something like that, which is similar to bootstrap's alert: https://getbootstrap.com/docs/4.0/components/alerts/
image

pavelpashkovsky:
@eh-am I think your variant is really good, interesting what @Rperry2174 thinks about it. and what is possible appearance and position of this alert?

pavelpashkovsky:
I can add tests after your review

Rperry2174:
Lets do the following:
1. Move the notification to below the main timeline
2. When clicking the button it should _extend_ the main timeline to include the entire blue range and also the entire pink range

The logic behind #2 is that this issue most frequently occurs when the main timeline is dynamic, but the comparison sub-timeline's are fixed and so we should adjust the main timeline to fit those

pavelpashkovsky:
hey guys, demo shows previous commit version. caching issue?

Rperry2174:
Looking at the commit I think its the correct version... what makes you think its the wrong version?

Rperry2174:
Looks great a couple of tweaks here:
- Let's add an "ignore" button which makes the alert disappear

For the colors lets do (and let's make sure these are variables as well):
- background: #f2cd51
- text: --ps-immutable-placeholder-text
- border: --ps-ui-border

Rperry2174:
There's currently an issue where when you select a time range after the default time ranges load, you get the error message because the pink timerange is almost immediately "out of range".

In order to prevent this, let's change the logic so that there is a buffer on the "minumum" side. For example:

Should **not** show a warning message:
Baseline (pink) timerange: 00:56 to 1:30
Comparison (blue) timerange: 1:30 to 2:00
**Main** timerange: 1:00 to 2:00

**Should** show a warning message:
Baseline (pink) timerange: 00:54 to 1:30
Comparison (blue) timerange: 1:30 to 2:00
**Main** timerange: 1:00 to 2:00

In this case the buffer is 5 minutes. 

Let's for now put this "buffer" in a variable and we may adjust this buffer later, but it should fix the problem for now

eh-am (commented):
Looks good! I have few thoughts though:


# RE: reusability

1. There's already a `StatusMessage` component, which I think can be reused here.

https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/ui/StatusMessage/index.tsx

2. Instead of creating custom buttons, we should reuse the existing button we have, and if necessary extend it in such a way that's reusable.

# RE: organization
Right now it works for comparison/diff since both use left/right, but not necessarily. Let's decouple input/output, plus it becomes testable.

The API would be something like
```tsx
<SyncTimelines onSync={onSync} main={mainTimeline} left={leftTimeline} right={rightTimeline} />
```

Since code will be pretty similar between comparison/diff, then you can add another interaction layer to reuse the same logic if that helps. But biggest point is that the `syncTimelines` functionality should NOT be aware of the store/anything else. It just takes inputs and calls outputs. 

# Tests
Given my previous point, it will be easy to test this component. And it makes it as good documentation.

webapp/javascript/components/TimelineChart/SyncTimelines/useSync.ts
```diff
@@ -0,0 +1,126 @@
...
+  }, [leftFrom, rightFrom, leftUntil, rightUntil, from, until]);
+
+  const leftSelectionMarkings = markingsFromSelection('single', {
+    from: leftFrom,
+    to: leftUntil,
+  } as Selection);
+  const rightSelectionMarkings = markingsFromSelection('single', {
+    from: rightFrom,
+    to: rightUntil,
+  } as Selection);
```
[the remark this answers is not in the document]
pavelpashkovsky: hmmm, interesting. I use this function to extract absolute time from relative one.
something like `{from: 'now', to: 'now-1h'} => {from: '1666689381792', to: '1666689991792'}`

then I use these values for comparison if selection is in/out of range

is there any more accurate way?

pavelpashkovsky:
@eh-am 
> ## RE: organization
>
I updated API in this way, now this component doesn't have any connection to redux store

> ## RE: reusability
I use `StatusMessage `, but initial UI design of this component differs from what we need for this warning. StatusMessage API allows styling it in way we need, but I did nothing with `error` and `success` appearance, what looks a bit weird from my perspective. Can you please take a look at this component and way of using it, maybe you have some better solution

image

webapp/javascript/components/TimelineChart/SyncTimelines/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+          >
+            Ignore
+          </Button>
+          <Button
+            data-testid="sync-button"
+            onClick={onSyncClick}
+            className={styles.syncButton}
+          >
+            Sync Timelines
+          </Button>
```
eh-am, line 57: What if we add an `outline` variant to our button?
Eg https://mui.com/material-ui/react-button/#outlined-button

eh-am: Then like any other variant of our buttons, we would add different styling hover/focus state, it would also look good next to a "normal" button etc.

webapp/javascript/ui/StatusMessage/index.tsx
```diff
@@ -26,7 +33,8 @@ export default function StatusMessage({ type, message }: StatusMessageProps) {
         [getClassnameForType()]: true,
       })}
     >
-      {message}
+      <div>{message}</div>
+      <div className={styles.rightSideWrapper}>{rightSide}</div>
```
eh-am, line 37: Nit, but as you pointed b4, let's hide this div if `rightSide` is not available. Less stuff in the DOM is always a good idea :P

pavelpashkovsky: is `:empty` rule suitable for this case?
`
.rightSideWrapper {
  &:empty {
    display: none;
  }
}
`

pavelpashkovsky:
image

webapp/javascript/components/TimelineChart/SyncTimelines/SyncTimelines.spec.tsx
```diff
@@ -0,0 +1,114 @@
...
+        {...propsWhenActive}
+        onSync={(from, to) => {
+          result = { from, to };
+        }}
+      />
+    );
+
+    fireEvent.click(screen.getByTestId('sync-button'));
+
+    console.log(result);
```
pavelpashkovsky, line 82: no cl

webapp/javascript/components/TimelineChart/SyncTimelines/useSync.ts
```diff
@@ -0,0 +1,105 @@
...
+    const selectionMax = Math.max(...selectionsLimits);
+
+    const offset = [
+      leftSelection.from,
+      rightSelection.from,
+      leftSelection.to,
+      rightSelection.to,
+    ].some((p) => String(p).startsWith('now'))
+      ? selectionOffset
+      : 1;
```
eh-am, line 91: you explained in slack, but can you explain it here too since this is an edge case?

petethepig (approved):
Haven't looked at the code but the demo looks great!

PR #1618, opened 2022-10-14
fix(webapp): make app selector/timerange dropdowns to be above loading overlay

eh-am:
Before:
Screen Shot 2022-10-14 at 09 30 12

After:
Screenshot 2022-10-14 at 09-24-27 Single pyroscope server cpu{} Pyroscope

---
Implementation: values may seem arbitrary, but the intention is to create different layers (0-9, 10-99, 100-999 etc). Didn't think too much through about this so feel free to change if a better approach comes up.


Rperry2174 (approved):
LGTM!

PR #1638, opened 2022-10-25
feat(webapp): Add relative time period dropdown to comparison / diff view

pavelpashkovsky:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1633

## Changes
- DEMO https://pr-1638.pyroscope.io/


pavelpashkovsky:
and can you guys explain a bit more about UI elements which is pointed on screenshot
image

there is kind of extra space on the left of mini timeline preview. can you please explain a bit more what it is? should I add some extra space to the timelines in similar way?

Rperry2174:
image

For this, the "baseline" and "comparions" timelines should always show the same time as the "main" timeline. Instead, what we should adjust is `leftFrom` and `leftUntil`. So:

Main timeline: Tuesday, 0:00 to 12:00
baseline timeline: Tuesday, 4:00 to 5:00
comparison timeline: Tuesday, 7:00 to 8:00

when you click "set baseline to 24 hours prior" it should _extend_ the main timeline to include the new baseline
Main timeline: Monday 7:00 to Tuesday 8:00 (it extends to include the new baseline)
baseline timeline: Monday, 7:00 to 8:00 (Moves from tuesday comparison timeline to 24 hours before that) 
comparison timeline: Tuesday, 7:00 to 8:00


we can likely use some of the logic from #1615 to sync the timelines when one of the selections becomes out of range

Rperry2174:
This looks great and as far as I can tell is correct (fair warning that some of this stuff will definitely change as we play around wit hit more) 

I wanted to just bring up what should happen in the case where you select a "select 24 hours prior" but the main timeline does not include that. We should basically extend the main timeline to make sure it includes that time period. Here is an example: 

For this, the "baseline" and "comparions" timelines should always show the same time as the "main" timeline. Instead, what we should adjust is `leftFrom` and `leftUntil`. So:

Main timeline: Tuesday, 0:00 to 12:00
baseline timeline: Tuesday, 4:00 to 5:00
comparison timeline: Tuesday, 7:00 to 8:00

when you click "set baseline to 24 hours prior" it should *extend* the main timeline to include the new baseline
Main timeline: Monday 7:00 to Tuesday 8:00 (it extends to include the new baseline)
baseline timeline: Monday, 7:00 to 8:00 (Moves from tuesday comparison timeline to 24 hours before that)
comparison timeline: Tuesday, 7:00 to 8:00

we can likely use some of the logic from #1615 to sync the timelines when one of the selections becomes out of range

Rperry2174:
image

The two checkboxes are just to show what "checked" and "unchecked" states look like there should really just be one checkbox 

the extra space in the preview was just because I was creating a fast mockup.. I don't feel strongly either way about it remaining or not. The only thing I feel strongly about is the label _text_ being correct (in this case 24 hours)

Rperry2174:
> are these selections were placed correctly on main and side timelines?
Yes they look correct

> should I maybe cut space at the left of pink selection. I made it in this way, because it looks more predictable, obvious
I do like the extra "buffer" in front of the pink selection

> I'm "sticking" selections to the right side of timeline, because time flows from the left to the right. is it ok?
yes this is fine 

> Should we store "N time prior" value in the global store instead of component state? if so, how we can keep this actual? User can go outside this page and make another selection.
let's not worry about this for now. I think the state technically gets saved in the url when `leftFrom` `leftUntil` `rightFrom` `rightUntil` changes 

> there is an option 1 MONTH. but how much time should it consist, 28-31d? maybe it make sense to transform this option to something like 30 DAYS?
Good point yes lets make it 30 days

petethepig:
It's looking great!

webapp/javascript/components/SideTimelineComparator/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+const cx = classNames.bind(styles);
+
+const buttons = [
+  [
+    {
+      label: '1 hour prior',
+      ms: 3600 * 1000,
+    },
+    {
+      label: '12 hour prior',
```
petethepig, line 38:
```suggestion
      label: '12 hours prior',
```

webapp/javascript/components/SideTimelineComparator/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+      ms: 43200 * 1000,
+    },
+    {
+      label: '24 hour prior',
```
petethepig, line 42:
```suggestion
      label: '24 hours prior',
```

webapp/javascript/components/SideTimelineComparator/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+    {
+      label: '2 weeks prior',
+      ms: 1209600 * 1000,
+    },
+    {
+      label: '30 days prior',
+      ms: 2592000 * 1000,
+    },
+  ],
+];
```
petethepig, line 60: Just thinking out loud, maybe it would be more useful if times are something shorter like:
| 1 hour | 12 hours |
| 2 hours | 24 hours |
| 6 hours | 1 week |


@kolesnikovae @eh-am @korniltsev  What do you guys think?

Rperry2174:
After playing with it some, it's a little weird that a dropdown on the "comparison" side affects the "baseline" side. Instead let's move it like this:

image

Rperry2174:
Let's bring in the checkbox and remove part of the text:
image

- By default the checkbox should be empty
- By default the dropdown should be set to 24 hours
- When someone clicks the checkbox it should "enforce" the time range in the dropdown

Rperry2174:
lookes great @pavelpashkovsky one tiny piece of feedback is I think the dropdown would look better if we plugged in the actual value at the top:

image

This will require us to make the dropdown slightly wider, but I think it's worth it

Rperry2174 (approved):
Lgtm!

PR #1639, opened 2022-10-25
chore: load appNames from /api/apps

eh-am:
As prep work for https://github.com/pyroscope-io/cloudstorage/issues/435
Load appNames from `/api/apps`


pkg/server/controller.go
```diff
@@ -303,13 +303,15 @@ func (ctrl *Controller) serverMux() (http.Handler, error) {
...
+
+	protectedAppsRouter := apiRouter.PathPrefix("/").Subrouter()
+	protectedAppsRouter.Use(
 		api.AuthMiddleware(nil, ctrl.authService, ctrl.httpUtils),
 		authz.NewAuthorizer(ctrl.log, httputils.NewDefaultHelper(ctrl.log)).RequireOneOf(
 			authz.Role(model.AdminRole),
 		))
-	appsRouter.HandleFunc("/apps", ctrl.getAppsHandler()).Methods("GET")
-	appsRouter.HandleFunc("/apps", ctrl.deleteAppsHandler()).Methods("DELETE")
+	protectedAppsRouter.HandleFunc("/apps", ctrl.deleteAppsHandler()).Methods("DELETE")
```
eh-am, line 314: @kolesnikovae @korniltsev @petethepig suggestions on how to handle this in a more scalable manner?

kolesnikovae: I would move this to line 220:
```go
	authorizer := authz.NewAuthorizer(ctrl.log, httputils.NewDefaultHelper(ctrl.log))
	requireAdminRole := authorizer.Require(authz.Role(model.AdminRole))

	appsRouter := apiRouter.PathPrefix("/apps").Subrouter()
	appsRouter.Path("/").
		Methods(http.MethodGet).HandlerFunc(ctrl.getAppsHandler()).
		Methods(http.MethodDelete).Handler(requireAdminRole(ctrl.deleteAppsHandler()))
```

Definitely out of scope for this PR but:
 - I would love to refactor the handler and service (interface only) to `/api` package
 - Noticed that `authz.NewAuthorizer(ctrl.log, httputils.NewDefaultHelper(ctrl.log))` takes logger twice but never uses the first one. Must be a leftover

pkg/server/controller.go
```diff
@@ -219,6 +219,16 @@ func (ctrl *Controller) serverMux() (http.Handler, error) {
 		apiRouter.RegisterAdhocHandlers()
 	}
 
+	// FIXME: Honor ctrl.config.RemoteRead.Enabled.
+	authorizer := authz.NewAuthorizer(ctrl.log, httputils.NewDefaultHelper(ctrl.log))
+	appsRouter := apiRouter.PathPrefix("/apps").Subrouter()
+	appsRouter.Use(ctrl.authMiddleware(nil))
```
kolesnikovae, line 225: `apiRouter` already has auth middleware, I'm not sure we need registering it one more time for a sub-router

pkg/server/controller.go
```diff
@@ -219,6 +219,16 @@ func (ctrl *Controller) serverMux() (http.Handler, error) {
...
+
+	appsRouter.Path("").
+		Methods(http.MethodGet).Handler(ctrl.getAppsHandler())
+	appsRouter.Path("").
+		Methods(http.MethodDelete).Handler(authorizer.RequireAdminRole()(ctrl.deleteAppsHandler()))
```
kolesnikovae: I think it can be simplified, taking into account that we don't need to specify empty path:
```go
appsRouter.Methods(http.MethodGet).Handler(ctrl.getAppsHandler())
appsRouter.Methods(http.MethodDelete).Handler(authorizer.RequireAdminRole()(ctrl.deleteAppsHandler()))
```

eh-am: :P leftover from my debugging

pkg/server/controller.go
kolesnikovae: > Ugh this is confusing. 
>
> That being said still using requireAdminRole, which (I think) means even in a public instance it's non possible to delete an app, which is fair but confusing when running locally.

That's true. Currently, if authentication is disabled, no admin actions can be performed, corresponding UI components are not shown. Now that we way want to change this, we can make an option telling what permissions to grant to an anonymous user when authentication is not enabled.

Also, not directly related but we may also consider adding `Member` role which allows more actions, for example: adding/deleting annotations, uploading adhoc profiles and so on.

> Speaking of requireAdminRole, I saw that there's a https://github.com/pyroscope-io/pyroscope/blob/main/pkg/api/authz/authz.go#L24-L26 but its use seems to be used as a middleware, I guess we should unify that.

In some cases it is more convenient to use it as a middleware (the code you reference), in others - wrap a specific handler. Typically, a collection like `/apps` has a set of CRUD methods, and each of them may require different permissions. On the other hand, if an entire collection like `/users` requires the same set of authorisation rules, it is way more convenient to use a middleware.

eh-am:
@korniltsev did I do something with admin server tests? It's failing consistent here, but from my perspective I didn't even touch that code.

pkg/server/controller.go
kolesnikovae: BTW, as discussed, using `authorizer.RequireAdminRole()` instead of `requireAdminRole := authorizer.Require(authz.Role(model.AdminRole))` is definitely preferable, thanks for pointing out this!

kolesnikovae:
> @korniltsev did I do something with admin server tests? It's failing consistent here, but from my perspective I didn't even touch that code.

I guess we used to have authentication middleware enabled regardless of the configuration. And /apps GET required Admin.

Honestly, I don't think the failing test is valid: with Auth disabled `/apps/delete` should always return 403 I believe.
```go
			Context("with Auth disabled", func() {
				It("should be accessible to only admin token", func() {
					(*cfg).Server.Auth.Internal.Enabled = false
					(*cfg).Server.EnableExperimentalAdmin = true
					runServer(cfg, func(s testServices) {
						checkRoleAccess(s, model.AdminRole, true, http.StatusOK)
						checkRoleAccess(s, model.AgentRole, false, http.StatusForbidden)
						checkRoleAccess(s, model.ReadOnlyRole, false, http.StatusForbidden)
						checkRoleAccess(s, model.InvalidRole, false, http.StatusUnauthorized)
					})
				})
			})
```

Although I hardly can find any good in allowing admin actions for unauthenticated users, we may want to make it possible to configure Role for anonymous users when authentication is disabled. We already have `SignupDefaultRole` and can add `AnonymousRole` or just `DefaultRole` that defaults to `ReadOnly`

eh-am:
Yeah just made the tests pass. Ideally we should have granular tests per route.

petethepig (approved):

PR #1643, opened 2022-10-26
feat: add a generic Tooltip component

eh-am:
Replace our `Tooltip` for material ui's one:
Screen Shot 2022-10-26 at 17 54 30
Changed it to have the same look and feeling as the previous one.

Also add a `TooltipInfoIcon` that we can reuse anywhere.
Screen Shot 2022-10-26 at 17 47 09

---
Api is
```jsx
    <Tooltip title="my title">
      <MyComponentThatWillTriggerTheTooltipOnHover />
    </Tooltip>

```
Some features:
* It triggers by default on hover (although that can be overriden https://github.com/mui/material-ui/blob/master/packages/mui-joy/src/Tooltip/Tooltip.tsx#L601)
* It auto positions correctly depending on the viewport (eg it never goes off the screen, also if there's not enough space let's say below the component, it renders on the top)
* `title` can take any react component, not only strings!
* accessibility (eg trigger on keyboard focus)
* It supports 12 different positions
* For more features see https://mui.com/material-ui/react-tooltip/#customization


eh-am:
One thing here is that I don't think the contrast against the background is good. IMO we should flip dark/light so that it has a "light" aspect.

Rperry2174:
> One thing here is that I don't think the contrast against the background is good. IMO we should flip dark/light so that it has a "light" aspect.

what if we add a slight background shadow to it?

PR #1649, opened 2022-10-31
feat: store application metadata

eh-am:
Stores `ApplicationMetadata` in the database (sqlite):

```go
type ApplicationMetadata struct {
	// Fully Qualified Name. Eg app.cpu ({__name__}.{profile_type})
	FQName string `gorm:"index,unique;not null;default:null"`

	SpyName         string
	SampleRate      uint32
	Units           metadata.Units
	AggregationType metadata.AggregationType
}
```

Every time we ingest data, we also update its `ApplicationMetadata`. There's a cache that only calls the underlying storage (db) when **the value is different from what's in memory**.
Since there's a ttl and a cache size, we may do unnecessary writes for clients with lots of applications or applications that are not ingested frequently.

All reads are done directly from the database.

Also add a migration script that runs on startup. It basically lists all apps and adds them to the database.

Also make `GET /api/apps` calls in the webapp to use this service.


pkg/cli/appMetadataMigration.go
```diff
@@ -0,0 +1,64 @@
+package cli
```
kolesnikovae, line 1: snake_case is preferred for file names

pkg/cli/appMetadataMigration.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	// If they don't exist already
+	for _, a := range appNamesFromOrigin {
+		if _, ok := appMap[a]; !ok {
+			logrus.Info("Migrating app: ", a)
+			// Write to MetadataSaver
+			m.appMetadataSaver.CreateOrUpdate(ctx, storage.Application{
+				FullyQualifiedName: a,
+			})
```
kolesnikovae, line 59: I wonder how we should deal with other properties - specifically SampleRate. Should we care about other properties at all when migrating?

A nit: the error should be logged

pkg/model/application.go
```diff
@@ -0,0 +1,17 @@
...
+	ErrApplicationNameEmpty = ValidationError{errors.New("application name can't be empty")}
+)
+
+func ValidateAppName(appName string) error {
+	// TODO(eh-am): check for invalid characters, min/max names?
+	// problem is that it needs to be in sync with existing data
+	if appName == "" {
+		return ErrApplicationNameEmpty
+	}
+	return nil
```
kolesnikovae, line 16: There is a function: https://github.com/pyroscope-io/pyroscope/blob/508466ea3e2c440bb4e5ee6803f5cfd748d32b7b/pkg/flameql/flameql.go#L87-L97

However, we don't have the max size defined. I'd say, 1024 bytes should be sufficient (for label keys as well). Regarding backward compatibility – if we don't want to issue a major release, we should either accept the risk, or allow arbitrary length. I think the risk is acceptable

pkg/service/application.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func (ApplicationService) handleError(err error) error {
+	switch {
+	case errors.Is(err, gorm.ErrRecordNotFound):
+		return model.ErrApplicationNotFound
+	default:
+		return err
+	}
+}
```
kolesnikovae, line 68: I might be mistaken but it seems that only `Get` can return `model.ErrApplicationNotFound`, so the wrapper is kinda redundant and makes error handling less explicit. Please consider handling `model.ErrApplicationNotFound` errors directly in the place they are expected

pkg/storage/types.go
```diff
@@ -131,3 +132,19 @@ type BadgerDBWithCache interface {
 	CacheInstance() *cache.Cache
 	Name() string
 }
+
+// TODO: I don't know why but I find ApplicationService name clumsy in this context.
+
+type ApplicationService interface {
+	CreateOrUpdate(context.Context, Application) error
+}
```
kolesnikovae, line 140: Seems to be a leftover

pkg/service/application_cache.go
```diff
@@ -0,0 +1,70 @@
...
+			return svc.writeToBoth(ctx, application)
+		}
+		return nil
+	}
+
+	// Not in cache
+	// Could be due to TTL
+	// Or could it be that's a new app
+	return svc.writeToBoth(ctx, application)
+}
```
kolesnikovae, line 54: I would maybe ensure that updates are throttled, e.g. once per second / 10 seconds. There is one problem that agents _may_ send different data :/ `SampleRate` causes most of concerns. I would maybe consider ignoring it and eventually removed it from the application metadata

pkg/service/auth.go
```diff
@@ -32,7 +32,8 @@ func NewAuthService(db *gorm.DB, jwtTokenService JWTTokenService) AuthService {
 //
 // External users are not allowed to use password authentication.
 // TODO(kolesnikovae): It's true for "some" authentication providers.
-//  Others may need us to pass through the credentials (e.g. LDAP).
+//
+//	Others may need us to pass through the credentials (e.g. LDAP).
```
kolesnikovae, line 36: "Others may need us to pass through the credentials (e.g. LDAP)." is part of the TODO note

kolesnikovae (commented):
Looks good! Left a couple of minor notes

pkg/cli/appMetadataMigration.go
eh-am: >Should we care about other properties at all when migrating?
No, since they will be populated at ingestion time https://github.com/pyroscope-io/pyroscope/blob/131e718cdf5f62666358c2ef17ded45f9a14caf7/pkg/storage/storage_put.go#L39

pkg/service/auth.go
eh-am: there was a tab that made the formatter go nuts, fixed

pkg/cli/app_metadata_migration.go
```diff
@@ -0,0 +1,68 @@
...
+		appMap[a.FQName] = a
+	}
+
+	// If they don't exist already
+	for _, a := range appNamesFromOrigin {
+		if _, ok := appMap[a]; !ok {
+			logrus.Info("Migrating app: ", a)
+			// Write to MetadataSaver
+			saveErr := m.appMetadataSaver.CreateOrUpdate(ctx, storage.ApplicationMetadata{
+				FQName: a,
```
petethepig, line 59: does this mean that we won't populate any of the other metadata, e.g spy name, and it would only be added on the next ingest request?

and if so, how hard would it be to make it populate other fields as well?

pkg/service/application_cache.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"context"
+	"reflect"
+	"time"
+
+	"github.com/pyroscope-io/pyroscope/pkg/storage"
+)
+
+type ApplicationWriter interface {
+	CreateOrUpdate(ctx context.Context, application storage.ApplicationMetadata) error
```
petethepig, line 12: Nit: This is the same interface as `storage.ApplicationMetadataSaver`. It may make sense to use the same one... not sure.

pkg/storage/types.go
```diff
@@ -131,3 +132,13 @@ type BadgerDBWithCache interface {
...
 }
+
+type ApplicationMetadata struct {
+	// Fully Qualified Name. Eg app.cpu ({__name__}.{profile_type})
+	FQName string `gorm:"index,unique;not null;default:null" json:"name"`
+
+	SpyName         string                   `json:"spyName,omitempty"`
+	SampleRate      uint32                   `json:"sampleRate,omitempty"`
+	Units           metadata.Units           `json:"units,omitempty"`
+	AggregationType metadata.AggregationType `json:"-"`
```
petethepig, line 143: any particular reason for not including this in JSON?

petethepig: Oh, we're not using this in the API yet, so I guess this is fine..

Yeah, I'd say, if we do ever use this in the HTTP apis we should probably expose AggregationType and also we should not `omitempty` because zod will complain about it if they are empty.

pkg/service/application_cache.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+type ApplicationCacheServiceConfig struct {
+	Size int
+	TTL  time.Duration
+}
+
+func NewApplicationCacheService(config ApplicationCacheServiceConfig, appSvc ApplicationWriter) *ApplicationCacheService {
+	if config.Size <= 0 {
+		config.Size = 100
```
petethepig, line 27: What happens if we go over 100? I would maybe set it to something higher than that, e.g 1000.

petethepig (approved):
It all looks good. I have a slight concern about the migration — it would be nice if it could add other things like spy name and units, but I don't know how hard it is to do that.

pkg/service/application_metadata.go
```diff
@@ -0,0 +1,63 @@
...
+	switch {
+	case errors.Is(res.Error, gorm.ErrRecordNotFound):
+		return app, model.ErrApplicationNotFound
+	default:
+		return app, res.Error
+	}
+}
+
+func (svc ApplicationMetadataService) CreateOrUpdate(ctx context.Context, application storage.ApplicationMetadata) error {
+	if err := model.ValidateAppName(application.FQName); err != nil {
```
petethepig, line 44: I think if the app name made it all the way here is a bit too late to do validations. I think they should either happen much earlier (in the controller), or not happen at all.

pkg/cli/app_metadata_migration.go
eh-am: Requires querying latest profile, which I don't know a good way to do so, so I would query from EPOCH until now to be certain. I bet @kolesnikovae has a better idea.

But as I said previously, it's possible that there are applications without any other metadata (eg ones consumed via collapsed), so we have to deal with the case where they are absent anyway.

pkg/service/application_metadata.go
eh-am: Well that's the approach we use in other places ¯\\\_(ツ)\_\/¯

https://github.com/pyroscope-io/pyroscope/blob/34edfddebbbd33c6dbdb7925c5d95fddef8f5200/pkg/service/user.go#L18
https://github.com/pyroscope-io/pyroscope/blob/34edfddebbbd33c6dbdb7925c5d95fddef8f5200/pkg/service/api_key.go#L17
https://github.com/pyroscope-io/pyroscope/blob/34edfddebbbd33c6dbdb7925c5d95fddef8f5200/pkg/service/annotation.go#L23

pkg/storage/types.go
eh-am: Reasoning is simple: since this is an endpoint tailored for the frontend, and since that information is not used in the frontend, there's no need to expose it.

I previously talked with @kolesnikovae and aggregationType is a storage thing, I don't think the frontend user has any use for this info. I may be mistaken.

```
❯ grep -ri 'aggregationType' webapp/ | wc -l
0
```

pkg/cli/app_metadata_migration.go
kolesnikovae: I agree, it may be tricky. One approach is to pick the latest tree of that app (via reverse badger DB iterator with app name prefix) and then load the corresponding segment

pkg/service/application_metadata.go
kolesnikovae: Also, I can remember one issue when invalid (empty) app names were generating due to some flaws in parsing. It was hard to fix. This validation would allow us to identify the error quickly.

pkg/storage/types.go
petethepig: yeah that's fair

pkg/service/application_metadata.go
petethepig: Fair enough, let's validate everywhere then :)

petethepig (approved):

PR #1650, opened 2022-10-31
chore(webapp): Format tag explorer table to proper units

pavelpashkovsky:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1629

## Changes
DEMO: https://pr-1650.pyroscope.io/

### DIFF
--- a/webapp/javascript/pages/TagExplorerView.tsx
+++ b/webapp/javascript/pages/TagExplorerView.tsx
@@ -5,6 +5,7 @@ import type { ClickEvent } from '@szhsin/react-menu';
 import Color from 'color';
 import TotalSamplesChart from '@webapp/pages/tagExplorer/components/TotalSamplesChart';
 import type { Profile, Units } from '@pyroscope/models/src';
+import { unitsDescription } from '@pyroscope/models/src';
 import Box, { CollapseBox } from '@webapp/ui/Box';
 import Toolbar from '@webapp/components/Toolbar';
 import ExportData from '@webapp/components/ExportData';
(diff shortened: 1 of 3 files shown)


Rperry2174:
Lets leave the headers the same as they were: 

- Average
- Standard deviation
- Total

But lets add units to the actual values in the table to match the values in the table below (you'll need to format them based on size -- i.e. seconds, minutes, hours, days, etc)

webapp/javascript/pages/TagExplorerView.tsx
```diff
@@ -396,9 +413,9 @@ function Table({
       label: `${groupByTag === '' ? 'App' : 'Tag'} name`,
```
Rperry2174, line 413: Can we make this be "Application" instead of "App name" so that it's consistent with what we call it at the top:
image

Rperry2174:
Even though in the flamegraphs table we do <0.01 days... since this is more "exact" is there any way we can easily do the actual number here (in any units) 

image

cc @eh-am

eh-am:
@Rperry2174 https://github.com/pyroscope-io/pyroscope/blob/main/packages/pyroscope-flamegraph/src/format/format.ts#L90 it's only enabled for `trace_samples` https://github.com/pyroscope-io/pyroscope/blob/04ab88a1d6c7c8fabefe5a46bf69c3137de25d06/packages/pyroscope-flamegraph/src/format/format.ts#L32

we need to move that flag up to`getFormatter` https://github.com/pyroscope-io/pyroscope/blob/04ab88a1d6c7c8fabefe5a46bf69c3137de25d06/packages/pyroscope-flamegraph/src/format/format.ts#L17

Rperry2174:
> @Rperry2174 @dogfrogfog https://github.com/pyroscope-io/pyroscope/blob/main/packages/pyroscope-flamegraph/src/format/format.ts#L90 it's only enabled for `trace_samples`
> 
> https://github.com/pyroscope-io/pyroscope/blob/04ab88a1d6c7c8fabefe5a46bf69c3137de25d06/packages/pyroscope-flamegraph/src/format/format.ts#L32
> 
> we need to move that flag up to`getFormatter`
> 
> https://github.com/pyroscope-io/pyroscope/blob/04ab88a1d6c7c8fabefe5a46bf69c3137de25d06/packages/pyroscope-flamegraph/src/format/format.ts#L17

lets do this separately then

Rperry2174 (approved):
lgtm

PR #1656, opened 2022-11-01
feat(webapp): [notifications] support 'warning' status and arbitrary jsx element

eh-am:
Support a `warning` level:
Screen Shot 2022-11-01 at 09 21 24

Also support an arbitrary JSX element to be passed as message, as opposed to only strings:
Screen Shot 2022-11-01 at 09 21 35


korniltsev (approved):

Rperry2174 (approved):
LGTM

PR #1657, opened 2022-11-01
refactor(webapp): Update time selector dropdown styling

pavelpashkovsky:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1632
## Changes
image

### DIFF
--- a/webapp/javascript/components/CustomDatePicker.module.scss
+++ b/webapp/javascript/components/CustomDatePicker.module.scss
@@ -2,4 +2,9 @@
   outline: none;
   border-radius: 4px;
   border: 1px solid var(--ps-ui-border);
+  background-color: var(--ps-ui-element-bg-highlight);
+}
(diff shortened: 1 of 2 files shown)


eh-am:
I think background color should be straight up white, similar to annotations here 
image

Then it's clear it something you can actually interact with.

eh-am:
Also we should reuse the `Input` element from `@ui`

pavelpashkovsky:
@eh-am 

> I think background color should be straight up white, similar to annotations here image
> 
> Then it's clear it something you can actually interact with.

am I on a right way (according to colors)?
image

eh-am:
@pavelpashkovsky i was thinking for inputs (since that's what we already do), but dropdowns are definitely more trickier so i wouldn't touch that yet :(

eh-am (approved):

PR #1658, opened 2022-11-01
feat: show gif when heatmap has no selection

dogfrogfog:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1596

## Changes
- add gif and heatmap action description
- show gif when heatmap has no selected area

demo link: https://pr-1658.pyroscope.io/exemplars/single?query=rideshare-app-golang.cpu%7B%7D

https://user-images.githubusercontent.com/47758224/199260086-4c51eb68-9ae9-4809-852b-681772dd7c6a.mov

## Concerns
-


eh-am:
I feel like we should show the gif only when the heatmap has not been interacted with. From the moment you interact with it, you shouldn't see the helper anymore. I also think we shouldn't lose existing state when starting selecting, only after the selection is finished (mouseout event i think)

Rperry2174:
lets at least add a box background to the area behind the gif... @eh-am / @dogfrogfog up to you guys if you think we need a button or not -- if we did add a button it could look like this:

image

We could also maybe make a button select a random area for the user, but I feel like that may be too much...

Rperry2174:
here's another option we could go with -- this is actually my vote:
image

wdyt @dogfrogfog @eh-am ?

eh-am:
The status message makes sense, I would use an info one though (blue bg iirc)

dogfrogfog:
> The status message makes sense, I would use an info one though (blue bg iirc)

could you explain a bit what you do mean by that..add yellow block from Ryan's ss and paint it blue ?

Rperry2174:
> > The status message makes sense, I would use an info one though (blue bg iirc)
> 
> could you explain a bit what you do mean by that..add yellow block from Ryan's ss and paint it blue or i'm missing something?

yeah I think thats what he means lets go with that @dogfrogfog

dogfrogfog:
https://user-images.githubusercontent.com/47758224/199523238-b1e069cd-7b2b-4c50-bee1-a211f865b0fc.mov

Rperry2174:
one more request... lets move the status message above to be in between the two heatmaps (and remove the text) like this 
image

Rperry2174:
Let's also only show the gif if there is heatmap data available... For example here we should not show the gif:
image

eh-am (approved):

webapp/javascript/components/Heatmap/index.tsx
```diff
@@ -25,15 +25,15 @@ interface HeatmapProps {
     startT: number,
     endT: number
   ) => void;
-  timezone: string;
   sampleRate: number;
+  timezone: string;
```
eh-am, line 29: where's this used?

dogfrogfog: https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/components/Heatmap/utils.ts#L127

dogfrogfog: i just reordered props in asc order

Issue #1659, opened 2022-11-01
Merge root nodes of callers in sandwich view


Rperry2174:
![image](https://user-images.githubusercontent.com/23323466/199266792-fd04ea0f-af89-46d6-bc12-606d47b8aad3.png)

We need to merge the bottom node here, but we need a function to do so and a tree to write that function for

petethepig:
There's a similar issue in callees too
![Screen Shot 2022-11-01 at 8 13 57 AM](https://user-images.githubusercontent.com/662636/199267921-de173476-1ebb-48c1-8ade-c2b4f9b4c7fd.png)

Rperry2174:
also what do the "total" nodes mean at the top of each of the "callers" stacks? should those be removed?

petethepig:
> also what do the "total" nodes mean at the top of each of the "callers" stacks? should those be removed

I'd keep it, I think that one is fine

dogfrogfog:
> also what do the "total" nodes mean at the top of each of the "callers" stacks? should those be removed?

The "total" node is a tree root node. To get stacks im going from root node to target function. Since root node is "total" we see it on top of each stack

dogfrogfog:
> There's a similar issue in callees too ![Screen Shot 2022-11-01 at 8 13 57 AM](https://user-images.githubusercontent.com/662636/199267921-de173476-1ebb-48c1-8ade-c2b4f9b4c7fd.png)

ye...i think we should have ability to merge _same name + same tree level_ nodes..... i think the case with callees and the callers are pretty similar... in both cases we have same [tree](https://github.com/pyroscope-io/pyroscope/blob/main/packages/pyroscope-flamegraph/src/convert/sandwichViewProfiles.ts#L8) and we want to merge nodes with the same paramethers

dogfrogfog:
this is example of `time.Since` tree with unmerged functions + corresponding flamebearer of this tree

dogfrogfog:
here is the example of `rideshare-app-golang.cpu{}` `time.Since` callers [tree](https://pastebin.com/AV6xhZbZ) and [flamebearer](https://pastebin.com/CPsR7CDT).... not exact same as in 1st ss but still pretty close
<img width="1365" alt="Screenshot 2022-11-02 at 13 37 47" src="https://user-images.githubusercontent.com/47758224/199491429-724a3db5-f61f-43a7-93a5-5035a3ca3399.png">

dogfrogfog:
[...]

PR #1670, opened 2022-11-02
chore: sandwich improvements

petethepig:
Made it do deduplication in callees and callers. Addressed #1659 

Screen Shot 2022-11-02 at 3 21 01 PM


packages/pyroscope-flamegraph/src/convert/sandwichViewProfiles.ts
```diff
@@ -53,37 +53,40 @@ export const treeToFlamebearer = (tree: TreeNode): FlamebearerData => {
...
     const nextNode = { name, ...rest, total: [total] };
-
     nestedObj.children = [nextNode];
     nestedObj = nextNode;
   });
 
   return result.children[0];
 };
 
+function dedupTree(node) {
```
Rperry2174, line 70: can you add some comments about whats happening here?

petethepig: It's a recursive function that takes a node as input and then for that node walks over all children and performs deduplication meaning that it finds children with the same name and when two children have the same name it unifies them into 1 child node.

It then recursively calls itself on each new child node.

packages/pyroscope-flamegraph/src/convert/sandwichViewProfiles.ts
```diff
@@ -170,12 +168,15 @@ export function callersFlamebearer(
   processTree(tree);
 
   subtrees.forEach((v, i) => {
-    totalNode.children.push(
-      arrayToTree(v, maxSubtreeLvl, targetFunctionTotals[i])
-    );
+    totalNode.children.push(arrayToTree(v.reverse(), targetFunctionTotals[i]));
```
petethepig, line 171: Major change I made here is that I first make an inverse (upside down) tree.
This allows me to call `dedupTree` function on the whole structure.
And then I reverse it back on line 177.

petethepig: This allows us to get rid of some other code, e.g all that `maxSubtreeLvl` logic

dogfrogfog (approved):
LGTM !! :fire:

packages/pyroscope-flamegraph/src/convert/sandwichViewProfiles.ts
eh-am: can we type `node`?

eh-am (commented):
Cool! How would one go about adding a test to cover this case? cc @dogfrogfog I think it would be useful.

packages/pyroscope-flamegraph/src/convert/sandwichViewProfiles.ts
eh-am: i don't think that's obvious, a comment in the code would be appreciated

PR #1673, opened 2022-11-03
fix: preserve query param when redirecting when baseUrl is set

eh-am:
# Problem
It was raised in our [slack channel](https://pyroscope.slack.com/archives/C01FJRYENPQ/p1666858222507879) that baseUrl + oauth did not work properly.

# The Fix
This path is only reached when both auth and baseUrl are enabled.

The way it was implemented, certain redirects would be fully encoded, including the path, producing URLs such as:
```
/pyroscope/auth/gitlab/redirect%3Fcode=d79c3ebf-4367-4c64-862e-c11cc89c58fa&scope=read_api&state=5db3bd88832f9a2900c71ed14d58957d&tls=false
```
Where `%3F` should be the `?` which represents the start of the query string.

With the change, we simply concatenate as strings, not escaping anything.

> **Warning**
Notice that a better solution should involve using more clear data structures, such as `net/url.URL`, which can hold host/path/query string separately.

# Others
* Add explanation on how to run baseURL + auth locally
* Fix an issue with our oauth2-mock-server impl only working for the very first time
* Add cypress suite to test the combination of baseUrl + oauth

### DIFF
--- a/.github/workflows/cypress-tests.yml
+++ b/.github/workflows/cypress-tests.yml
@@ -200,7 +200,7 @@ jobs:
         uses: cypress-io/github-action@v2
         with:
           wait-on: http://localhost:8080/pyroscope
-          start: make server
+          start: make server SERVERPARAMS=--config=scripts/oauth-mock/pyroscope-config-base-url.yml
           config-file: cypress/integration/auth/cypress.json
         env:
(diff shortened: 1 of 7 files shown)


petethepig (approved):
LGTM

PR #1674, opened 2022-11-03
feat(flamegraph): Redesign flamegraph toolbar to allow for more interactions

pavelpashkovsky:
## Brief
 - https://github.com/pyroscope-io/pyroscope/issues/1636
 
 ## Changes
 - DEMO: https://pr-1674.pyroscope.io/


pavelpashkovsky:
1) Let me know if some sizes, shapes or colors are incorrect.
2) There is no enough space to place everything on Toolbars on Comparison page. Any ideas?

https://user-images.githubusercontent.com/7372044/199745888-2ce05b15-81dd-488c-8792-adb38b887b8f.mp4

3) What to do with SELF | TOTAL | DIFF buttons on DiffView page. Does it make sense to transform these button in the same way as TABLE | FLAMEGRAPH+TABLE | FLAMEGRAPH (on the right side)?

4) This selector looks like font-size selector, right?
image

5) @eh-am After some updates live reload for storybook doesn't work anymore. Is there easy way to fix it?

6) Actual Export button icon differs from required icon in this task. Guys told that you have arranged to leave current button. Just to confirm.
image

7) On the mockup toolbar looks kind of divided.
image

8) @eh-am can you please help w/ code review. It's not a final version, but we have to be on the same page

Rperry2174:
image

> 1. Let me know if some sizes, shapes or colors are incorrect.

see picture above for color adjustments / styling placement adjustments 

> 2. There is no enough space to place everything on Toolbars on Comparison page. Any ideas?

When there's not enough space these should turn to dropdowns:
- head first / tail first tabs
- table / both / flamegraph / sandwich tabs

For now, let's just make everything else  _just_ be an icon with no words. Eventually we will use the [tooltip](https://github.com/pyroscope-io/pyroscope/pull/1643) that @eh-am added and things will work similar to google sheets. 

https://user-images.githubusercontent.com/23323466/199918081-f84bdd58-98b7-4cf7-9471-091b5e36afd0.mov

So this means that the following should be changed to just be their respective icons 100% of the time (we may change this later):
- Collapse nodes above
- reset view

> 3. What to do with SELF | TOTAL | DIFF buttons on DiffView page. Does it make sense to transform these button in the same way as TABLE | FLAMEGRAPH+TABLE | FLAMEGRAPH (on the right side)?

Let's worry about this later... 

> 4. This selector looks like font-size selector, right?
>    image

yes this is just for the mockup don't worry about it for now we will add font size later maybe

> 6. Actual Export button icon differs from required icon in this task. Guys told that you have arranged to leave current button. Just to confirm.
>    image

Yes we decided on the new one no need to change the export button again

> 7. On the mockup toolbar looks kind of divided.
>    image

yes toolbar is divided see picture at top of this message

> 8. @eh-am can you please help w/ code review. It's not a final version, but we have to be on the same page
> 9. Can you please explain a bit more about about this set of colors+buttons
>    image

Just for reference in case we don't like the colors chosen

Rperry2174 (approved):
LGTM! Nice work :)

PR #1676, opened 2022-11-04
fix a few function names on comments

cuishuang:


pkg/scrape/config/config_http.go
```diff
@@ -35,7 +35,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/mwitkow/go-conntrack"
```
kolesnikovae, line 38: Must've been removed unintentionally. This import is required

pkg/util/process/process_unix.go
```diff
@@ -24,7 +24,7 @@ func Exists(pid int) bool {
 	return s != "Z"
 }
 
-// SendProcess send signal s to the process p.
+// SendSignal send signal s to the process p.
```
kolesnikovae, line 27:
```suggestion
// SendSignal sends signal s to the process p.
```

kolesnikovae (changes requested):
Thank you for the contribution @cuishuang! There is one issue with a removed import – could you check it please?

Rperry2174 (approved):
LGTM! Thanks @cuishuang !

PR #1677, opened 2022-11-04
fix(webapp): pass from,until when calling /label{-values}

eh-am:
[The backend expects `from` and `until` query parameters in the `/labels` and `/label-values` endpoints](https://github.com/pyroscope-io/pyroscope/blob/508466ea3e2c440bb4e5ee6803f5cfd748d32b7b/pkg/server/labelvalues.go#L22-L23).

Here we pass them.

The catch is that to not have to have different set of tags for different time ranges, we just take the biggest time range between left/right/regular. This implies that we may pass a bigger time range than necessary, but it's fine.

~Also this has another downside: every time the range changes, we also have to load labels again, which is redundant most of the times. In the future we may do some optimization like storing the {from, until} interval and only load when a different range outside the cached range is passed.~
Also Store the `{from, until}` interval which an app loaded labels refers to. That way we don't have to reload again if loading from a subrange.

A small refactor, but removed `TagsBar` from `Toolbar` in SingleView to keep it consistent with the other views.


webapp/javascript/redux/reducers/continuous.ts
```diff
@@ -580,9 +642,13 @@ export const fetchTagValues = createAsyncThunk(
...
     }
 
+    const state = thunkAPI.getState().continuous;
+    const timerange = biggestTimeRangeInUnix(state);
     const res = await tagsService.fetchLabelValues(
       payload.label,
-      payload.query
+      payload.query,
+      timerange.from,
+      timerange.until
```
eh-am, line 651: Actually we should use the same `from, until` we used when we loaded the tag.

eh-am: fixed

Issue #1671, opened 2022-11-02
Incorrect output of `convertJaegerTraceToProfile`


bobrik:
I filed an issue with more details and repro steps with an example trace in jaeger-ui repo:

* https://github.com/jaegertracing/jaeger-ui/issues/1012

Maybe this repo is the right place to report it.

Rperry2174:
Thanks for reporting and letting us know about the other issue @bobrik.. just to make sure everyone is on the same page let's keep the communication in the issue in jaeger-ui repo

petethepig:
@bobrik I submitted [a PR](https://github.com/pyroscope-io/pyroscope/pull/1679) — would love your feedback!

aleks-p:
The function was removed from this repo along with the legacy UI. Jaeger UI reimplemented its flame graph on d3-flame-graph.

PR #1679, opened 2022-11-04
chore: fixes the logic in convertJaegerTraceToProfile function

petethepig:
Addresses #1671 / https://github.com/jaegertracing/jaeger-ui/issues/1012

## Trace 
Screen Shot 2022-11-04 at 3 15 57 PM

## Flamegraph Before

before

## Flamegraph After

after

## Notes

One of the characteristics of the trace from the original issue was that it had parallel executions, particularly in `logic-proxy: upstream` node. That's why it looked so strange before.

This PR makes it so that in such cases when the sum of children nodes' duration is above parent duration, parent duration is set to the sum of children nodes' duration.

One unfortunate consequence of this is that now durations of the top nodes are inflated and don't always match the durations in span, e.g look at the top `eyeball-facing-service: request`. While this is correct in a sense that that's how much wall time was spent processing a request, it might be confusing to some.

@bobrik What do you think about this?

### DIFF
--- a/packages/pyroscope-flamegraph/src/convert/convertJaegerTraceToProfile.ts
+++ b/packages/pyroscope-flamegraph/src/convert/convertJaegerTraceToProfile.ts
@@ -46,12 +46,12 @@ export function convertJaegerTraceToProfile(trace: Trace): Profile {
     span.children = map(groups, (group) => {
       const res = group[0];
       for (let i = 1; i < group.length; i += 1) {
-        res.duration += group[i].duration;
+        res.total += group[i].total;
       }
-      childrenDur += res.duration;
+      childrenDur += res.total;
       return res;
     });
-    span.total = span.duration || childrenDur;
+    span.total = Math.max(span.duration, childrenDur);
     span.self = Math.max(0, span.total - childrenDur);
   }
   groupSpans(root, 0);


bobrik:
> While this is correct in a sense that that's how much wall time was spent processing a request

In my example the whole request took 162ms, but your top flamegraph slab is 268ms. That doesn't seem to match your description. That doesn't seem right.

Rperry2174:
@bobrik did you see the note at the bottom? 

> One of the characteristics of the trace from the original issue was that it had parallel executions, particularly in logic-proxy: upstream node. That's why it looked so strange before.
This PR makes it so that in such cases when the sum of children nodes' duration is above parent duration, parent duration is set to the sum of children nodes' duration.
One unfortunate consequence of this is that now durations of the top nodes are inflated and don't always match the durations in span, e.g look at the top eyeball-facing-service: request. While this is correct in a sense that that's how much wall time was spent processing a request, it might be confusing to some.

This is why there is a discrepancy between the request length and what the flamegraph shows... Curious what you think the expected behavior should be here

bobrik:
> did you see the note at the bottom?

Yes, I quoted from it. I don't agree with the "that's how much wall time was spent processing a request" framing.

> While the full request only took 162ms, the flamegraph shows 268ms because there was 268ms of "work" being done (just some of it was happening in parallel).

Tracing is generally used to gauge latency (wall time), not work (which is usually a proxy for cpu time) as flamegraphs.

I don't know if there's a good way to translate concurrent spans into flamegraphs. It might be acceptable to represent them as you do with your change visually, but show the actual timing values on the slabs. As a concrete example for my trace:

```
[ logic-proxy: upstream (70.1%, 133.91ms)                                                         ]
[ logic-proxy: upstream_ttfb (60.39%, 112.75ms) ][ customer-code: http_request (60.39%, 112.69ms) ]
```

Here we scale width of `upstream_ttfb` and `http_request` to make them fit into `upstream`, but keep them proportional.

If you can add some sort of warning to `upstream` span showing that it has children adding up to more than their parent (or maybe even just parallel running children), that would be even better to avoid confusion.

Let me know if this makes sense.

Rperry2174:
hey @bobrik yes this makes sense and might be a good compromise we'll look into seeing what changes we'd need to make in order to allow the flamegraph component to make them "overwrite" the true values, but this may take a little while to dig into

cescoferraro:
@Rperry2174 how can I try this PR on my project?

cescoferraro:
I published this PR as `cescoferraro-flamegraph`

eh-am:
Hey @cescoferraro, even though you already published a PR, we have a CI step that publishes the flamegraph (https://github.com/pyroscope-io/pyroscope/actions/runs/3397437411/jobs/5649595841). You can point to `@pyroscope/flamegraph@0.25.2-1679-ab8cd3f.0` 

image

PR #1687, opened 2022-11-08
refactor(webapp): cleanup store

eh-am:
The continuous reducer file was too big (>1k lines), so I moved the thunks to their own files, also moved selectors to a single `selectors.ts` file.


Rperry2174 (approved):
LGTM

PR #1689, opened 2022-11-08
chore(webapp): rename 'latency' -> 'duration' in heatmap

eh-am:


Rperry2174 (approved):
agreed this makes more sense

PR #1697, opened 2022-11-10
feat: show percentages for diff table instead of absolute values

dogfrogfog:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1481

## Changes
- change diff table values from _absolute_ to _percent_

demo link: https://pr-1697.pyroscope.io/comparison-diff?query=rideshare-app-golang.alloc_objects%7B%7D&rightQuery=rideshare-app-golang.alloc_objects%7B%7D&leftQuery=rideshare-app-golang.alloc_objects%7B%7D

https://user-images.githubusercontent.com/47758224/201096384-0e17789a-f533-4d0b-9598-f81ad5fc8b50.mov

## Concerns
- im using .toFixed(2) and for some values that looks strange

image

i think we can either show smth like < 0.01 % like we do with absolute values or increase number of digits after comma

image


dogfrogfog:
https://user-images.githubusercontent.com/47758224/201370229-75680381-4e93-4b0d-a3ec-800bf9b387e7.mov

Rperry2174:
I think this looks great... I think we should just make the table have the exact same 3 columns as the tooltip has :

image

Let's remove the 3 button options for now 
The table should have:
- ~total (left)~ -> Baseline
- ~total (right)~ -> Comparison
- Add a column for diff

wdyt @petethepig ?

petethepig:
@Rperry2174 yeah that makes sense

dogfrogfog:
Screenshot 2022-11-14 at 16 54 41

Rperry2174:
Looks great.. lets also show the same data that shows on the flamegraph over the table as well for the tooltip:
image

dogfrogfog:
https://user-images.githubusercontent.com/47758224/201921620-1b72a1d3-6ca6-4cf4-9ab9-170b216e11dc.mov

packages/pyroscope-models/src/flamebearer.ts
```diff
@@ -40,7 +40,8 @@ export type Flamebearer = {
 
 export type addTicks =
   | { format: 'double'; leftTicks: number; rightTicks: number }
-  | { format: 'single' };
+  // todoL: remove left/right ticks
+  | { format: 'single'; leftTicks?: number; rightTicks?: number };
```
eh-am, line 44: What's going on here? My understanding is that `single` doesn't have `{left,right}Ticks`

packages/pyroscope-flamegraph/src/format/format.ts
```diff
@@ -14,6 +14,12 @@ export function ratioToPercent(ratio: number) {
   return Math.round(10000 * ratio) / 100;
 }
 
+export function percentDiff(leftPercent: number, rightPercent: number): number {
+  // difference between 2 percents
+  // https://en.wikipedia.org/wiki/Relative_change_and_difference
+  return ((rightPercent - leftPercent) / leftPercent) * 100;
+}
```
eh-am, line 21: Since we are at here, can you unify that with https://github.com/pyroscope-io/pyroscope/blob/c56a6970cfc576c9eb572f4e6d7130576ab9f926/packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/color.ts#L26

packages/pyroscope-flamegraph/src/Tooltip/TableTooltip.tsx
```diff
@@ -44,62 +48,88 @@ export default function TableTooltip({
...
-      const selfFormatted = formatter.format(
-        parseInt(selfValue, 10),
-        sampleRate
-      );
-      const totalFormated = formatter.format(
-        parseInt(totalValue, 10),
-        sampleRate
-      );
+      switch (format) {
+        case 'single': {
```
eh-am, line 55: We gonna need to add a test case for this.

eh-am (commented):
Just to be sure, are we getting rid of these?
image

cc @Rperry2174

dogfrogfog:
yes, i removed code for this buttons

Rperry2174:
> Just to be sure, are we getting rid of these? image
> 
> cc @Rperry2174

yes I think we should get rid of them (tbh many don't know what they actually mean anyway)

eh-am (approved):

PR #1698, opened 2022-11-10
feat: add Fit Mode to Context Menu

dogfrogfog:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/606

## Changes
- add Fit Mode to Context Menu (flamegraph right click menu)

demo link: https://pr-1698.pyroscope.io/?name=hotrod.python.frontend%7B%7D&query=rideshare-app-golang.cpu%7B%7D

https://user-images.githubusercontent.com/47758224/201111129-15a923c1-0655-403d-979b-0ced3563fa62.mov

## Concerns
- need to add icons


packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/index.tsx
```diff
@@ -238,6 +241,22 @@ export default function FlameGraphComponent(props: FlamegraphProps) {
...
+
+        const handleClick = () => {
+          const newValues = isHeadFirst ? TailMode : HeadMode;
+          updateFitMode(newValues);
+        };
+
+        return (
+          <MenuItem key="fit-mode" onClick={handleClick}>
+            {/* todo: add icons */}
+            Fit mode: {isHeadFirst ? 'tail first' : 'heat first'}
```
eh-am, line 255:
```suggestion
            Fit mode: {isHeadFirst ? 'tail first' : 'head first'}
```

Although since we have enough space, I think we can be more explicit `Fit {head, tail} first` or or some better wording cc @Rperry2174

Rperry2174: I think the wording is fine, but for consistency we should add the icon the same way the other options have an icon

dogfrogfog: absolutely..which one should i use ?

Rperry2174: Lets use the same icons from the toolbar (depending on whether it says "head" or "tail" first. As far as the text for the other ones we kind of explain what happens with the option:
image


What do you guys think about:

- "[ Icon ] Show text head first" 
- "[ Icon] Show text tail first"

dogfrogfog: looks good i think

https://user-images.githubusercontent.com/47758224/201632259-37090c6c-1bed-447f-86f8-da2597501bf8.mov

eh-am (approved):

eh-am:
Looks good! Thanks

PR #1707, opened 2022-11-14
chore: pprof add more sample types

korniltsev:


pkg/storage/tree/pprof.go
```diff
@@ -78,6 +78,34 @@ var DefaultSampleTypeMapping = map[string]*SampleTypeConfig{
 		Units:       metadata.LockNanosecondsUnits,
 		Cumulative:  true,
 	},
+	"cpu": {
```
petethepig, line 81: golang cpu profiles usually have `samples` and `cpu` types built-in. We need to make sure this doesn't cause issues where it would write the cpu data from go profiles twice.

petethepig: Yeah, looking at the code it looks like this will cause issues. E.g look at this code that we use to figure out if profile type should be ingested https://github.com/pyroscope-io/pyroscope/blob/23c649d586c138f6c5d598a9225aa0e8b4644c88/pkg/convert/pprof/parser.go#L173-L178

korniltsev:
I think I don't even need this, I will just pass sample type config during ingestion

PR #1714, opened 2022-11-14
chore: update light mode tooltip color

Rperry2174:
Old
image

New
image


eh-am (approved):
Looks good, even though this won't work properly in grafana. See https://github.com/pyroscope-io/pyroscope/issues/1717 for more info

PR #1722, opened 2022-11-16
feat(webapp): filter out apps that are not cpu in exemplars page

eh-am:
We only filter out the ones we are sure it's not cpu (eg the ones with a suffix that's not `.cpu`).
For the ones that we cannot identify, we still allow them.

This function will be improved once the backend starts returning more useful data (specially `units`).

Before:
image

After:
image


Rperry2174:
nice.. what happens if you come from another page where you're looking at a non-cpu app ?

PR #1728, opened 2022-11-17
refactor(flamegraph): Hide ExportData and ChangeView dividers when it necessary

pavelpashkovsky:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1710

## Changes
- DEMO: https://pr-1728.pyroscope.io/
- display only flamegraph + **NO** ExportData
image

- display only flamegraph + ExportData
image

- "display only" disabled + ExportData
image

### DIFF
--- a/packages/pyroscope-flamegraph/src/Toolbar.tsx
+++ b/packages/pyroscope-flamegraph/src/Toolbar.tsx
@@ -88,8 +88,7 @@ export interface ProfileHeaderProps {
   sharedQuery?: FlamegraphRendererProps['sharedQuery'];
 }
 
-const Divider = ({ visible = true }: { visible?: boolean }) =>
-  visible ? <div className={styles.divider} /> : null;
+const Divider = () => <div className={styles.divider} />;
 
 const Toolbar = React.memo(
   ({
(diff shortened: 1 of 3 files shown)


packages/pyroscope-flamegraph/src/Toolbar.tsx
```diff
@@ -88,7 +88,8 @@ export interface ProfileHeaderProps {
   sharedQuery?: FlamegraphRendererProps['sharedQuery'];
 }
 
-const Divider = () => <div className={styles.divider} />;
+const Divider = ({ visible = true }: { visible?: boolean }) =>
```
dogfrogfog, line 91: how about ?
```suggestion
const Divider = ({ visible = true }) =>
```

pavelpashkovsky: I think it's better to remove all unnecessary props from Divider at all

dogfrogfog (approved):

packages/pyroscope-flamegraph/src/Toolbar.tsx
```diff
@@ -140,17 +140,23 @@ const Toolbar = React.memo(
...
-            <ViewSection
-              flamegraphType={flamegraphType}
-              showMode={showMode}
-              view={view}
-              updateView={updateView}
-            />
-          )}
-          <Divider />
-          {ExportData}
+          {!disableChangingDisplay ? (
```
eh-am, line 143: Nit, but I recommend writing these boolean flags in the positive form.

Eg `enableChangingDisplay`

I found it's easier to read the different cases
`enableChangingDisplay` -> enable
`!enableCgangingDisplay` -> disable

Now compared to 
`disableChangingDisplay` -> is changing display enabled? no
`!disableChangingDisplay` -> double negative, we are disabling disabling changing display, which here means we are Enabling changing display

pavelpashkovsky: updated

eh-am (approved):

PR #1731, opened 2022-11-17
fix: make reset view available in sandwich mode context menu

dogfrogfog:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1727

## Changes
- fix ability to reset flamegraph from ContextMenu components

https://user-images.githubusercontent.com/47758224/202425277-b60efba4-ee6a-4501-aa8d-8bfc72f5a63e.mov

## Concerns
-

### DIFF
--- a/packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/index.tsx
+++ b/packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/index.tsx
@@ -265,19 +265,11 @@ export default function FlameGraphComponent(props: FlamegraphProps) {
         );
       };
 
-      const ResetItem = () => {
-        const isResetDisabled = selectedItem.isNothing && !dirty;
-
-        return (
-          <MenuItem key="reset" disabled={isResetDisabled} onClick={onReset}>
-            <FontAwesomeIcon icon={faRedo} />
-            Reset View
-          </MenuItem>
-        );
-      };
-
       return [
-        ResetItem(),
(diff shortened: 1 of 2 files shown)


packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphComponent/index.tsx
```diff
@@ -265,11 +265,20 @@ export default function FlameGraphComponent(props: FlamegraphProps) {
         );
       };
 
+      const ResetItem = () => {
+        const isResetDisabled =
+          !(selectedItem.isJust && selectedItem.value) && !dirty;
```
eh-am, line 270: why do we need to check `selectedItem.value` exists? if it's not a `Nothing` then it should have a value

dogfrogfog: correct, i will fix it

packages/pyroscope-flamegraph/src/FlameGraph/FlameGraphRenderer.tsx
```diff
@@ -377,8 +376,9 @@ class FlameGraphRenderer extends Component<
   // so that the flamegraph doesn't rerender unnecessarily
   isDirty = () => {
     return (
+      !this.state.selectedItem.isNothing ||
```
eh-am, line 379:
```suggestion
      this.state.selectedItem.isJust ||
```

eh-am (approved):
👍 :)

PR #1733, opened 2022-11-17
feat: make tag explorer modal adapt to content

dogfrogfog:
## Brief 
- https://github.com/pyroscope-io/pyroscope/issues/1517

## Changes
- remove fixed width

Screenshot 2022-11-17 at 16 00 52
Screenshot 2022-11-17 at 16 01 21

### DIFF
--- a/webapp/javascript/pages/tagExplorer/components/TagsSelector.module.scss
+++ b/webapp/javascript/pages/tagExplorer/components/TagsSelector.module.scss
@@ -1,42 +1,44 @@
-.title {
-  display: inline-block;
-  text-transform: uppercase;
-  padding: 5px;
-  color: var(--ps-select-modal-title);
-  font-weight: 700;
-}
+.tagSelector {
+  width: auto;
 
-.tags {
-  padding: 0;
-  list-style: none;
-  margin: 0;
-  text-align: left;
-  min-height: 150px;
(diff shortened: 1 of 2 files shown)


dogfrogfog:
https://user-images.githubusercontent.com/47758224/202668342-52ec4556-bf2b-499a-9a02-d1424d570468.mov

webapp/javascript/pages/tagExplorer/components/TagsSelector.module.scss
```diff
@@ -1,40 +1,45 @@
...
 
-    input {
-      border: none;
-      padding: 10px 30px;
-      width: 100%;
-      text-align: left;
-      white-space: nowrap;
-      overflow: hidden;
-      text-overflow: ellipsis;
+    li {
```
eh-am, line 21: I know this code was already like this, but I am going to say it just for us to sync about these things.

I don't see the value in nesting selectors using sass here, since we use css modules anyway it's already pretty scoped.

When using nesting it's easier to make things confusing. So I would suggest for next to just make them more or less flat.

dogfrogfog: copy 👍

Rperry2174 (approved):
lgtm

PR #1741, opened 2022-11-18
feat(flamegraph): Max height and scroll for flamegraph table 

pavelpashkovsky:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1215

## Changes
- DEMO: https://pr-1741.pyroscope.io/

### DIFF
--- a/packages/pyroscope-flamegraph/src/FlameGraph/FlamegraphRenderer.module.scss
+++ b/packages/pyroscope-flamegraph/src/FlameGraph/FlamegraphRenderer.module.scss
@@ -38,6 +38,12 @@
   flex: 1;
 }
 
+.tablePane {
+  &.vertical {
+    margin-right: 0;
+  }
+}
+
 .sandwichPane {
   .sandwichPaneInfo {
     display: inline-block;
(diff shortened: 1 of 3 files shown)


eh-am:
Hmmm I personally am not a big fan of [inline scrollbars](
https://baymard.com/blog/inline-scroll-areas).

Although we if we are going with this approach, I think would be nice to have at least 2 things:
* Sticky header
* Display an arrow or something to indicate there's more data

Another idea is to truncate the data and have a "More items" button?
image

Or, a completely different approach would be to add pagination. Although that breaks the in browser search.

pavelpashkovsky:
>Sticky header

despite having "display: sticky" implementation of this feature using only CSS might be cranky

eh-am:
>I like these ideas! But don't you think that whole table data should be always available for user?
I don't think it's suuuuuch a big deal since we have the search functionality.

But here I am just throwing out ideas, the inner scrolling is fine as it is.

Rperry2174:
going to close this for now as it's not clear if we should go the scroll route or the full page route or something in between

PR #1749, opened 2022-11-21
fix(flamegraph): Make table tooltip invisible when user not hovering on table

pavelpashkovsky:
## Brief
- Closes https://github.com/pyroscope-io/pyroscope/issues/1743

## Changes
- Fixed common situation when user moves cursor from table to flamegraph but table tooltip stays visible.


pavelpashkovsky:
Postponed this task in a while, cypress can't detect hover state on flamegraph table, needs additional investigation

eh-am:
I understand the solution, but don't understand the problem: why does it keep visible even after hovering out the table?

pavelpashkovsky:
another option is add additional `mouseover` event handler for tooltip which hides tooltip when user hovering on it. but `pointer-events: none` is simpler :)
```
const onOver = () => {
    setStyle({
      display: 'none',
    });
  };

  useEffect(() => {
    const tootipEl = tooltipRef.current;
    tootipEl.addEventListener('mouseover', onOver);

    return () => {
      tootipEl.removeEventListener('mouseover', onOver);
    };
  }, [dataSourceRef.current, memoizedOnMouseMove]);
```

eh-am (approved):

PR #1758, opened 2022-11-23
perf: cache optimization

kolesnikovae:
I'm restoring the work done in https://github.com/pyroscope-io/pyroscope/pull/685 – it was closed as the change didn't show a substantial performance improvement. However, recently we encountered an issue supposedly caused by very long locks of the cache. The strategy implemented in this PR was aimed at eliminating such harmful blocking.

TODO:
 - [x] Solve conflicts


kolesnikovae:
Let's revisit this later, if it will be still relevant

PR #1760, opened 2022-11-24
chore(webapp): make frontend aware of other app fields

eh-am:
Frontend is now aware of `spyName` and `units` field
Screen Shot 2022-11-24 at 11 57 42


korniltsev (approved):
I like that it looks like it would be slightly easier to add a new unit type next time

PR #1762, opened 2022-11-25
chore(webapp): Tag Explorer, select 1st tag automatically

pavelpashkovsky:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1745

## Changes
it automatically selects first available tag in 2 cases (DEMO: https://pr-1762.pyroscope.io/)
- user lands on this page, there's no selected tags
- user changes app in app selector


webapp/javascript/redux/reducers/continuous.ts
```diff
@@ -337,7 +337,7 @@ export const continuousSlice = createSlice({
     /*******************************/
 
     builder.addCase(fetchTagExplorerView.pending, (state) => {
-      switch (state.diffView.type) {
```
pavelpashkovsky, line 340: Is there any purpose to check loading state of diffView store section for tagExplorerView one? @eh-am

webapp/javascript/redux/reducers/continuous.ts
```diff
@@ -379,7 +379,7 @@ export const continuousSlice = createSlice({
       state.tagExplorerView = {
         ...state.tagExplorerView,
         activeTagProfile: action.payload.profile,
-        type: 'loaded',
+        type: state.tagExplorerView.type,
```
pavelpashkovsky, line 382: Changed this line because there is kind of race condition for tag explorer loading state. Because `fetchTagExplorerViewProfile` and `fetchTagExplorerView` have access to `state.tagExplorerView.type` at the same time. They fetch different data but share loading state between each other. It causes false `type: loaded` sometimes. This update isn't the best but working solution. Any thoughts? Ideally we need to separate loading state for them.

webapp/javascript/redux/reducers/continuous.ts
eh-am: most likely a typo

webapp/javascript/redux/reducers/continuous.ts
eh-am: I would separate it in different states, yes.

eh-am: See the labels loading for more or less how I would implement this. Each part has its own loading state.

webapp/javascript/pages/TagExplorerView.tsx
```diff
@@ -311,6 +316,13 @@ function TagExplorerView() {
       activeTagProfile.metadata.units
     );
 
+  const dataLoading =
+    groupsLoadingType === 'loading' ||
+    groupsLoadingType === 'reloading' ||
+    activeTagProfileLoadingType === 'loading';
+
+  console.log('dataLoading', dataLoading);
```
eh-am, line 324: whoops

webapp/javascript/redux/reducers/continuous/state.ts
```diff
@@ -25,33 +25,14 @@ type SingleView =
...
-      groupByTagValue: string;
-    };
+type TagExplorerView = {
+  groupsLoadingType: 'pristine' | 'loading' | 'loaded' | 'reloading';
+  groups: Groups;
+  groupByTag: string;
+  groupByTagValue: string;
+  activeTagProfile?: Profile;
+  activeTagProfileLoadingType: 'pristine' | 'loading' | 'loaded' | 'reloading';
+};
```
eh-am, line 35: I like the original way since it forces you to think about the different states. For example, if type is `pristine`, it means you have no data, so you have to design your UI around it.

Although with the way it was implemented, the types didn't really indicate that, so based on that your refactor makes sense.

eh-am (approved):

PR #1772, opened 2022-11-29
fix: refactor tags dropdown width

dogfrogfog:
## Changes 
- add min-width to left and right(base/comp) column
- add max-width to container component

https://user-images.githubusercontent.com/47758224/204571681-3bf2cf2b-50dc-43e5-b241-1be0fb47a8e5.mov


Rperry2174 (approved):
lgtm!

PR #1776, opened 2022-11-30
feat(flamegraph): allow sorting by diff percentage

dogfrogfog:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1770

## Changes
- add ability to sort by diff column


dogfrogfog:
is `infinity` value expected here ? 

https://user-images.githubusercontent.com/47758224/204783752-9887d6f1-efee-4016-ba71-d9f24c3c2197.mov

eh-am:
>is infinity value expected here ?
I don't think so, it looks like it should be considered "new" no?

dogfrogfog:
> > is infinity value expected here ?
> > I don't think so, it looks like it should be considered "new" no?

ye..i will fix it

eh-am (approved):

PR #1785, opened 2022-12-02
fix(webapp | flamegraph): toolbar overlaps annotation

pavelpashkovsky:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1780

## Changes

DEMO: https://pr-1785.pyroscope.io/

image

### DIFF
--- a/packages/pyroscope-flamegraph/src/Toolbar.module.scss
+++ b/packages/pyroscope-flamegraph/src/Toolbar.module.scss
@@ -11,7 +11,6 @@ $buttonHeight: 37px;
   border: 1px solid var(--ps-ui-border);
   padding: 4px;
   align-items: center;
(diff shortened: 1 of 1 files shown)


pavelpashkovsky:
I fixed this issue by removing z-index from Toolbar styles. My testing showed that it didn't cause any problems. @eh-am, do you have any reason to have z-index for flamegraph's Toolbar ? _bc I could forget about some edge case why we need setting z-index.._

eh-am:
It was introduced here https://github.com/pyroscope-io/pyroscope/pull/1754 by @dogfrogfog 
So @dogfrogfog, can we remove that z-index?

dogfrogfog:
i think this can be removed

dogfrogfog (approved):

PR #1790, opened 2022-12-05
chore(webapp): color mode for timeline [Single View]

pavelpashkovsky:
## Brief
- https://github.com/pyroscope-io/pyroscope/issues/1788

## Changes
- added color for Single View timeline chart on light mode (I set primary color for light mode as bars color, but it can be changed after review)
image

- added logic which tracks changing color mode for `<body data-theme="light | dark">`. There are 2 source where we set current color mode - redux store (`ui.colorMode`) and `<body data-theme="light | dark">`. I extended `useColorMode` logic to have synced both these source. When some logic changes color mode in `data-theme` there is react-side handler which updates UI according to new theme.

- DEMO: https://pr-1790.pyroscope.io/


Rperry2174 (approved):
lgtm thanks!

PR #1799, opened 2022-12-20
feat: pprof streaming parsing

korniltsev:


pkg/convert/pprof/decoder.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/pyroscope-io/pyroscope/pkg/storage/tree"
 )
 
-var bufPool = bytebufferpool.Pool{}
+
```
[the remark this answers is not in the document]
korniltsev: I wanted to reuse a buffer with a current parser. I now made it to share one non global pool between them.

pkg/convert/pprof/streaming/parser_streaming.go
```diff
@@ -0,0 +1,521 @@
...
+}
+
+func (p *MoleculeParser) SetSampleTypesFilter(f func(string) bool) {
+	p.sampleTypesFilter = f
+}
+func (p *MoleculeParser) ParsePprof(ctx context.Context, startTime, endTime time.Time, bs []byte) error {
+	p.startTime = startTime
+	p.endTime = endTime
+	p.ctx = ctx
+	defer func() { p.ctx = nil }()
```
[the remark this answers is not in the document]
korniltsev: it depends on how big is context retained size. there is no need to keep it. I rewrote without defer

pkg/convert/pprof/streaming/parser_streaming.go
```diff
@@ -0,0 +1,521 @@
...
+			return fmt.Errorf("failed to create pprof profile zip reader: %w", err)
+		}
+		defer gzipr.Close()
+		buf := bytebufferpool.Get()
+		defer bytebufferpool.Put(buf)
+		if _, err = io.Copy(buf, gzipr); err != nil {
+			return err
+		}
+		p.profile = buf.Bytes()
+		defer func() { p.profile = nil }()
```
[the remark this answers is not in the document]
korniltsev: the field is to avoid passing the buffer around multiple functions , there is no need to keep profile after parsing is done since it may contain garbage. I remove a reference to it just to prevent possible access to the garbage in the future. I also rewrite without defer. OK?

pkg/convert/pprof/streaming/parser_streaming.go
```diff
@@ -0,0 +1,521 @@
...
+	p.functions = make([]function, 0, nFunctions) //todo reuse these for consecutive parse calls? if cap is enough ?
+	p.locations = make([]location, 0, nLocations)
+
+	return err
+}
+
+// step 2
+// - parse locations
+// - parse functions
+func (p *MoleculeParser) parseStructs2() error {
```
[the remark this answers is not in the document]
korniltsev: It is now parseFunctionsAndLocations.

pkg/convert/pprof/streaming/labels_cache.go
```diff
@@ -0,0 +1,125 @@
...
+		if i != j {
+			c = append(c, copyLabel(label))
+		}
+	}
+	return c
+}
+
+func copyLabel(label label) label {
+	return label
+}
```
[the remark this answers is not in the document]
korniltsev: made it 
```
func CutLabel(labels Labels, i int) Labels {
	c := make(Labels, len(labels)-1)
	copy(c[:i], labels[:i])
	copy(c[i:], labels[i+1:])
	return c
}
```

ok?

pkg/convert/pprof/streaming/parser_streaming.go
```diff
@@ -0,0 +1,521 @@
...
+		defer func() { p.profile = nil }()
+	} else {
+		p.profile = bs
+	}
+	return p.parsePprofDecompressed()
+}
+
+func (p *MoleculeParser) parsePprofDecompressed() error {
+	var err error
+	p.strings = make([][]byte, 0, 256) // todo sane default? count? reuse?
```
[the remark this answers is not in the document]
korniltsev: re strings size: the whole parsing happens in 3 pass, now I count number of strings in first pass and allocate correct slice size in second

pkg/convert/pprof/profile.go
```diff
@@ -187,35 +190,46 @@ func (p *RawProfile) Parse(ctx context.Context, putter storage.Putter, _ storage
...
 
 		if p.PreviousProfile != nil {
 			// Ignore non-cumulative samples from the PreviousProfile
 			// to avoid duplicates: although, presence of PreviousProfile
 			// tells that there are cumulative sample types, it may also
 			// include regular ones.
-			filter := p.parser.sampleTypesFilter
-			p.parser.sampleTypesFilter = func(s string) bool {
+			filter := p.parser.GetSampleTypesFilter()
+			p.parser.SetSampleTypesFilter(func(s string) bool {
```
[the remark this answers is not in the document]
korniltsev: I wen with extra argument to ParsePprof

pkg/convert/pprof/parser.go
```diff
@@ -165,15 +170,23 @@ func (p *Parser) iterate(x *tree.Profile, fn func(vt *tree.ValueType, l tree.Lab
...
+		if p.sampleTypesFilter != nil && p.sampleTypesFilter(st) {
+			if cumulative {
+				if p.sampleTypes[st].Cumulative {
+					indexes = append(indexes, i)
+					types = append(types, s.Type)
+				}
+			} else {
+				indexes = append(indexes, i)
+				types = append(types, s.Type)
+			}
```
kolesnikovae, line 189:
```suggestion
			if !cumulative || (cumulative && p.sampleTypes[st].Cumulative) {
				indexes = append(indexes, i)
				types = append(types, s.Type)
			}
```

pkg/convert/pprof/parser.go
```diff
@@ -207,6 +220,7 @@ func (p *Parser) readTrees(x *tree.Profile, c tree.LabelsCache, f tree.Finder) {
 		}
 		// Insert tree nodes.
 		for i, vi := range indexes {
+			_ = i
```
kolesnikovae, line 223: Must be leftover after debugging

pkg/convert/pprof/profile.go
```diff
@@ -187,35 +190,34 @@ func (p *RawProfile) Parse(ctx context.Context, putter storage.Putter, _ storage
...
+				SkipExemplars:       p.SkipExemplars,
+				StackFrameFormatter: StackFrameFormatterForSpyName(md.SpyName),
+			})
+		}
 
 		if p.PreviousProfile != nil {
-			// Ignore non-cumulative samples from the PreviousProfile
-			// to avoid duplicates: although, presence of PreviousProfile
-			// tells that there are cumulative sample types, it may also
-			// include regular ones.
```
kolesnikovae, line 203: Please leave the comment :) I fixed this place two times in past and therefore decided to put it here: it is the only case when we want to grab only samples of cumulative type (also, it would be nice to rename the argument to something like `cumulativeOnly` just for better readability)

pkg/convert/pprof/streaming/parser_streaming.go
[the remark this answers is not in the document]
korniltsev: re pool: done

pkg/convert/pprof/streaming/profile_finder.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+		idx := id - 1
+		//if idx >= uint64(len(f.functions)) {
+		//	return nil, false
+		//}
+		return &f.functions[idx], true
+	}
+	ff, ok := f.functionsMap[id]
+	return ff, ok
+}
```
kolesnikovae, line 53: Why not to inline them?

korniltsev:
@kolesnikovae ptal, I rewrote the parse with code generated from VT, it is a bit faster and the code is a bit more reusable (except for the Profile message) and also more verbose.
I also addressed codereview issues
The pr is still work in progress, there are bunch of things I want to do here
- cleanup and commit a bench + comparing test
- implement stack formatter for ruby's pprof


I'd like to do the testing before merging the pr, do you think I need to do smth else for testing or I can try now

pkg/convert/pprof/streaming/structs.go
```diff
@@ -0,0 +1,114 @@
...
+		s.tmpValues = s.tmpValues[:0]
+	}
+}
+
+func (s *sample) resetSample() {
+	s.tmpValues = s.tmpValues[:0]
+	s.tmpLabels = s.tmpLabels[:0]
+	s.tmpStack = s.tmpStack[:0]
+	s.tmpStackLoc = s.tmpStackLoc[:0]
+}
```
kolesnikovae, line 102: I'm not sure I understand why we need two separate functions. Also, shouldn't `reset()` reset labels as well?

pkg/convert/pprof/streaming/vt_function.go
```diff
@@ -0,0 +1,155 @@
+package streaming
+
+import (
+	"fmt"
+	"io"
+)
+
+// revive:disable-next-line:cognitive-complexity,cyclomatic necessary complexity
+func (m *function) UnmarshalVT(dAtA []byte) error {
```
kolesnikovae, line 9: These unmarshal methods are very repetitive. Can we do anything with it?

kolesnikovae:
I think if unit tests are okay we can test it with real data in our own tenant and polish everything after

kolesnikovae:
One thing I'd like to clarify – do you plan to reuse parsers? My understanding is that we allocate bunch of various slices that we can reuse

pkg/convert/pprof/streaming/profile_finder.go
kolesnikovae: Also, please clean up the code from debug helpers

pkg/convert/pprof/profile.go
```diff
@@ -187,35 +191,41 @@ func (p *RawProfile) Parse(ctx context.Context, putter storage.Putter, _ storage
...
+				Putter:        putter,
+				SampleTypes:   sampleTypes,
+				SkipExemplars: p.SkipExemplars,
+			})
+			p.parser = parser
+			if p.PoolStreamingParser {
+				defer func() {
+					parser.ReturnToPool()
+					p.parser = nil
+				}()
```
kolesnikovae, line 208: Use of `streaming.VTStreamingParserFromPool` should depend on `p.PoolStreamingParser` as far as I understand

pkg/convert/pprof/streaming/vt_sample.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+	reverseStack(p.tmpSample.tmpStack)
+
+	p.createTrees(p.newCache)
+
+	return nil
+}
+
+// revive:disable-next-line:cognitive-complexity,cyclomatic necessary complexity
+func (s *sample) UnmarshalSampleVT(dAtA []byte, tmpLabel *label) error {
```
kolesnikovae, line 30: It does not seem we modify it here (we should not at least as far as I can see) so please consider passing it by value - this is way cheaper to pass label by value

kolesnikovae: I mean, we do not need a tmpLabel at all: it is better to allocate small types like below on the stack in the function body: 

```go

 type line struct {
 	functionID uint64
 }

 type label struct {
 	k, v int64
 }

```

pkg/convert/pprof/streaming/labels_cache.go
```diff
@@ -0,0 +1,116 @@
+package streaming
+
+import (
+	"encoding/binary"
+	"github.com/cespare/xxhash/v2"
+	"github.com/pyroscope-io/pyroscope/pkg/storage/tree"
+	"sort"
+)
+
+type Labels []label
```
kolesnikovae, line 10: Consider representing a label as a single uint64: we know for sure that both K and V << 2^32 so we can pack them `(k<<32)|v`. This reduces cost of:
1. sort
2. hashing <- I'm not even sure if we would need this: comparing underlying byte slices (of Labels) might be way cheaper

pkg/convert/pprof/streaming/labels_cache.go
```diff
@@ -0,0 +1,116 @@
...
+		binary.LittleEndian.PutUint64(t[0:8], uint64(x.k))
+		binary.LittleEndian.PutUint64(t[8:16], uint64(x.v))
+		_, _ = h.Write(t[:])
+	}
+	return h.Sum64()
+}
+
+// sample type -> labels hash -> entry
+
+type LabelsCache map[int64]map[uint64]*LabelsCacheEntry
```
kolesnikovae, line 33: In future we may want to optimise the structure – this is the hottest path.

We know that there aren't this many sample types and the set is dense, therefore we don't need the top level map, it should be replaces with a slice
```go
[]map[uint64]*LabelsCacheEntry
```

Next, `*LabelsCacheEntry` is actually a tree and associated Labels. We can represent them in Structure of Arrays way:

```go
type cache struct {
   // sample type -> labels hash -> index to labels and trees
   indices []map[uint64]int
   labels  []Labels
   trees   []*Tree
}
```

Another possible improvement is getting rid of Labels in favour of an arena:

```go
type cache struct {
   // sample type -> labels hash -> index to labelRefs and trees
   indices []map[uint64]int
   // A label reference points to the subset in labels:
   // Hight 32 bits is the start offset, lower 32 bits is the subset size.
   labelRefs []uint64
   labels    []Label
   trees     []*Tree
}
```

This allows reusing the cache easily. Also, `map[int]int can be optimised further (e.g https://github.com/brentp/intintmap).

pkg/convert/pprof/streaming/labels_cache.go
korniltsev: Nice one.
I thought about another possible optimization: representing string as a uint64(pair of uint32, [from, to)) indices into profile buffer

pkg/convert/pprof/profile.go
korniltsev: it is, it is in a nested if branch.
or i don't get the comment?

pkg/convert/pprof/streaming/structs.go
korniltsev: it is a leftover, thanks

pkg/convert/pprof/streaming/profile_finder.go
korniltsev: noinline is also leftover, will remove

pkg/convert/pprof/profile.go
korniltsev: it is just the comment is kinda related to a boolean arg and it is just "true", it felt it is not clear what the comment is reffering to
I guess if lave the comment and do cumulativeOnly:=true/false it would be better

korniltsev:
Actually let me try to implement these optimizations before the testing

pkg/convert/pprof/streaming/labels_cache.go
korniltsev: after our call I had an impression that with a new cloudstorage's ingesters we will not need the LabelsCache at all so I did not even try to optimize it

kolesnikovae: Honestly, I'm not 100% sure when (if) it will be released as part of the OSS version. Also, the cache is only needed to do diffs of cumulative profiles, so I think we can't remove it completely :/ Moving the logic to the storage is kinda problematic, though possible I guess

pkg/convert/pprof/streaming/labels_cache.go
kolesnikovae: This should be also helpful. However, we still need to allocate a string for a reference (copy these bytes), otherwise memory ownership will be violated since the bytes are taken from pool. Therefore, depending on how many times a particular reference is resolved, it might even deteriorate the overall performance – though, I don't have a full picture and may be missing something

pkg/convert/pprof/profile.go
kolesnikovae: I mean that a parser should be only taken from pool if `p.PoolStreamingParser` is true, else it should be created with `NewStreamingParser`. Otherwise we will pollute the pool because we never return to it, which may cause higher memory consumption and some extra pressure on GC. However, `sync.Pool` implementation is quite smart and the impact might be unnoticeable.

Also, it is a common practice (according to my observations) to explicitly call `Reset` before returning an item to the pool in order to release (and make them available to GC) objects that can't be reused.

Frankly, I don't think the current implementation is wrong or not optimal, but there are some things that you are used to and expect to see

pkg/convert/pprof/streaming/vt_function.go
korniltsev: I don't think it is a problem for small structs like function and locations
For Profile I had 3 huge almost identical copies for 3 passes, I merged them to a single function with an "operation" parameter

petethepig (approved):

PR #1800, opened 2022-12-22
chore: add file service discovery example

kolesnikovae:
The PR adds an example configuration of the [file service discovery mechanism](https://github.com/pyroscope-io/pyroscope/pull/662) which was implemented almost a year ago but has not been documented

### DIFF
--- a/examples/golang-pull/file/Dockerfile
+++ b/examples/golang-pull/file/Dockerfile
@@ -1,0 +1,14 @@
+FROM golang:1.18.3
+
+WORKDIR /go/src/app
+
+COPY main.go ./main.go
+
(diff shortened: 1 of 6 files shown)


Rperry2174 (approved):

.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
+# . "$(dirname "$0")/_/husky.sh"
```
eh-am, line 2: just don't forget to uncommit this

you can run with `git commit --no-verify` if you need to bypass the formatting

kolesnikovae: Yep, thanks for pointing this out - had to hack it when was doing a big rebase via Goland UI

PR #1804, opened 2023-01-10
feat(pprof): parsing arenas

korniltsev:
```
name                                                               old time/op  new time/op  delta
SmallStreaming/BenchmarkSmallStreaming_pool_false_arenas_false-32  1.23ms ±11%  1.18ms ± 8%     ~     (p=0.123 n=10+10)
SmallStreaming/BenchmarkSmallStreaming_pool_true_arenas_false-32   1.12ms ± 6%  1.13ms ± 3%     ~     (p=0.968 n=10+9)
SmallStreaming/BenchmarkSmallStreaming_pool_false_arenas_true-32   1.23ms ±15%  0.74ms ± 2%  -40.03%  (p=0.000 n=10+10)
BigStreaming/BenchmarkBigStreaming_pool_false_arenas_false-32      11.0ms ± 5%   9.4ms ± 5%  -14.95%  (p=0.000 n=8+9)
BigStreaming/BenchmarkBigStreaming_pool_true_arenas_false-32       10.3ms ± 3%   9.3ms ± 4%   -9.50%  (p=0.000 n=9+10)
BigStreaming/BenchmarkBigStreaming_pool_false_arenas_true-32       10.5ms ± 2%   6.5ms ± 1%  -38.28%  (p=0.000 n=10+10)
```


korniltsev:
consider assigning an arena to a tree

pkg/convert/pprof/streaming/parser_streaming.go
```diff
@@ -95,7 +101,7 @@ func (p *VTStreamingParser) ParsePprof(ctx context.Context, startTime, endTime t
 		if err != nil {
 			err = fmt.Errorf("failed to create pprof profile zip reader: %w", err)
 		} else {
-			buf := PPROFBufPool.Get()
+			buf := PPROFBufPool.Get() //todo consider decompression on arena
```
kolesnikovae, line 104: Well, if you need my opinion, I think pools and arenas pursue different goals:
 1. Pools allow reusing already allocated objects, reducing number of allocations.
 2. Arenas allow to allocate objects from a contiguous region of memory and free them all at once, reducing garbage collection overhead caused by large number of objects on heap. Objects allocated on arena can't/shouldn't be reused.

Therefore, I doubt moving the buffer to arena will improve preformance

pkg/convert/pprof/streaming/parser_streaming.go
```diff
@@ -213,20 +219,25 @@ func (p *VTStreamingParser) addStackFrame(l *line) error {
...
 		pFuncName := p.strings[f.name]
-		frame := p.profile[(pFuncName >> 32):(pFuncName & 0xffffffff)]
-		p.tmpSample.tmpStack = append(p.tmpSample.tmpStack, frame)
+		frame = p.profile[(pFuncName >> 32):(pFuncName & 0xffffffff)]
+	}
+	pSample := &p.tmpSample
+	if len(pSample.tmpStack) < cap(pSample.tmpStack) {
+		pSample.tmpStack = append(pSample.tmpStack, frame)
+	} else {
+		pSample.tmpStack = arenahelper.AppendA(pSample.tmpStack, frame, p.arena)
```
kolesnikovae, line 240: I saw this many times in the PR: I must me missing something but should not `arenahelper` be responsible of that?

pkg/convert/pprof/streaming/vt_location.go
```diff
@@ -10,12 +11,15 @@ type locationFunctions struct {
 	lines []line
 }
 
-func (l *locationFunctions) reset() {
+func (l *locationFunctions) reset(a arenahelper.ArenaWrapper, nLocations int) {
+	if l.lines == nil {
+		l.lines = arenahelper.MakeSlice[line](a, 0, nLocations*2)
+	}
 	l.lines = l.lines[:0]
```
kolesnikovae, line 18: I'm not sure we need this extra check: `l.lines = l.lines[:0]` should be suffice – even if `l.lines == nil`. It is better to allocate the slice lazily, on append, not on `reset` (at least one allocation is avoided)

pkg/convert/pprof/profile.go
```diff
@@ -210,7 +211,16 @@ func (p *RawProfile) Parse(ctx context.Context, putter storage.Putter, _ storage
...
-				p.parser = streaming.NewStreamingParser(config)
+				if p.ArenasEnabled {
+					config.ArenasEnabled = true
+				}
+				parser := streaming.NewStreamingParser(config)
+				p.parser = parser
+				if p.ArenasEnabled {
+					defer func() {
+						parser.FreeArena()
+					}()
```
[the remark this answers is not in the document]
kolesnikovae: Just a note:

My biggest concern is the lifetime of the trees we Put to OSS cache/storage. I'm almost sure that we reference the source object in some cases (e.g I wouldn't clone a tree when put it to a cache for the first time). This is not something that makes use of arenas for trees impossible, but if we are up to enabling them in OSS, we need to carefully analyse object ownership beforehand.

I think in future we will also need to add `tree.Free()` because currently you need to own the arena to do so, which is not always easy to achieve. For example, if we had this method and freed trees before evicting them from the cache, tree lifecycle would be pretty explicit. However, this also requires allocating an arena per tree.

kolesnikovae (approved):
Looks good. Added couple of minor notes

pkg/convert/pprof/streaming/parser_streaming.go
korniltsev: This is an obscure optimization attempt.
arenahelper.AppendA is too complex for inlining. 
and with this manual check it becomes just a bunch of instructions like `slice.data[slice.len++]=newVal` with no function call.

kolesnikovae: Out of curiosity: does PGO help in this particular case? What is the impact of the optimisation?

korniltsev: I've just tried to do pgo thing on a trivial program but could not make compiler inline the appendA function and no performance improvements found after pgo applied

I tried it on the following program:

https://gist.github.com/korniltsev/189f66aa2f2ca3706e2765ef884548f4

```
go build  . 
./main
go build -pgo=cpu.pprof .
./main
```


maybe Im missing smth or doing something wrong, did not spend a lot of time on it

korniltsev: I wrote a quick bench to demonstrate impact
````
package main

import (
	"arenaspg/arenahelper"
	"fmt"
	"testing"
)

var testcases = []int{16, 64, 256}

func BenchmarkOp(b *testing.B) {
	for _, n := range testcases {
		b.Run(fmt.Sprintf("n=%d", n), func(b *testing.B) {
			a := arenahelper.NewArenaWrapper()
			s := arenahelper.MakeSlice[int](a, 0, n)
			for i := 0; i < b.N; i++ {
				s = s[:0]
				for i := 0; i < n; i++ {
					if len(s) < cap(s) {
						s = append(s, i)
					} else {
						s = arenahelper.AppendA(s, i, a)
					}
				}
				bh(s)
			}
		})
	}
}

func BenchmarkNoOP(b *testing.B) {
	for _, n := range testcases {
		b.Run(fmt.Sprintf("n=%d", n), func(b *testing.B) {
			a := arenahelper.NewArenaWrapper()
			s := arenahelper.MakeSlice[int](a, 0, n)
			for i := 0; i < b.N; i++ {
				s = s[:0]
				for i := 0; i < n; i++ {
					s = arenahelper.AppendA(s, i, a)
				}
				bh(s)
			}
		})
	}
}
````


```
goos: darwin
goarch: arm64
pkg: arenaspg
BenchmarkOp
BenchmarkOp/n=16
BenchmarkOp/n=16-8         	183499448	         6.537 ns/op
BenchmarkOp/n=64
BenchmarkOp/n=64-8         	53859661	        22.25 ns/op
BenchmarkOp/n=256
BenchmarkOp/n=256-8        	12262503	        97.54 ns/op
BenchmarkNoOP
BenchmarkNoOP/n=16
BenchmarkNoOP/n=16-8       	47982326	        24.97 ns/op
BenchmarkNoOP/n=64
BenchmarkNoOP/n=64-8       	 9093646	       131.4 ns/op
BenchmarkNoOP/n=256
BenchmarkNoOP/n=256-8      	 2377095	       501.8 ns/op

```

File: pkg/convert/pprof/profile.go (hunk 4 above)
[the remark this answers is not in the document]
Comment (korniltsev) [reply pkg/convert/pprof/profile.go:222]: I believe we should not enable arenas for OSS. Arena is good for a short lived object with a well defined lifetime, but as you said a tree is kept in memory sometimes for unknown time span. It is hard reason about lifecycle of a tree. Also multiple trees will share a single arena so they will need to coordinate somehow in order to free when all of them are free which is error-prone and extends arena life time(increases memory).
Maybe we even should not introduce arena to a Tree at all - after "moving parsers" project we will not need a tree as an intermediate representation in most of cases.

File: pkg/convert/pprof/streaming/parser_streaming.go (hunk 1 above)
Comment (korniltsev) [reply pkg/convert/pprof/streaming/parser_streaming.go:104]: So, I dont know if will help or not, I just wanted to try it :)

My thought process is that I believe pool is cleaned on each gc cycle so allocations still happen sometime. At the same time arena chunk is 8m and if we don't use all of it during parsing, which I hope we dont(except maybe we do for cumulative profiles), we could use some of it for decompression. Also there is a huge chance I will not be able to measure the deference at all.

File: pkg/convert/pprof/streaming/vt_location.go (hunk 3 above)
Comment (korniltsev) [reply pkg/convert/pprof/streaming/vt_location.go:18]: I dont think I understand the comment. 
It looks like here I know the number of locations and the number of lines is at least the same, here I pre-allocate twice as much of locations in hope I never allocate again at all. I agree 2x is maybe too agressive for non arena allocations but for arena allocations it should not matter. At the same time if I allocate lazilly at append I will allocate multiple times: 2, 4, 8, 16, at least to next power of higher than nLocations.
So I dont understand in your comment which one allocation is avoided

Comment (kolesnikovae) [reply pkg/convert/pprof/streaming/vt_location.go:18]: Never mind – I did get it wrong. Please consider using `grow` – in that case the intent would be more clear

**Review by kolesnikovae — APPROVED** [review]

File: pkg/convert/pprof/streaming/parser_streaming.go (hunk 1 above)
Comment (kolesnikovae) [reply pkg/convert/pprof/streaming/parser_streaming.go:104]: I find this interesting – it seems that a object bigger than 2MB (given the chunk size) will be allocated on heap (I'm using 1.20rc3): 
```go
	// userArenaChunkMaxAllocBytes is the maximum size of an object that can
	// be allocated from an arena. This number is chosen to cap worst-case
	// fragmentation of user arenas to 25%. Larger allocations are redirected
	// to the heap.
	userArenaChunkMaxAllocBytes = userArenaChunkBytes / 4
```
```go
	if size > userArenaChunkMaxAllocBytes {
		// Redirect allocations that don't fit into a chunk well directly
		// from the heap.
		if cap >= 0 {
			return newarray(typ, cap)
		}
		return newobject(typ)
	}
```

pkg/convert/pprof/streaming/parser_streaming.go
korniltsev: haha I think pgo does not apply to generics either intentionally or because of a bug
pprof reports it as  `arenahelper.AppendA[...]` and there is only a `arenahelper.AppendA[go.shape.int]` in call graph

if I manually create a non generic copy of AppendA - AppendInt - it is inlined with pgo

ascheglovclick: Hi,
A little belated advice
It seems splitting method into fast and slow paths should solve inlining problem

```
func AppendA[T any](data []T, v T, a ArenaWrapper) []T {
	if len(data) < cap(data) {
		return append(data, v)
	}

	return appendASlow(data, v, a)
}

func appendASlow[T any](data []T, v T, a ArenaWrapper) []T {
	if a == nil { // move me to the AppendA method when the compiler gets smarter
		return append(data, v)
	}

	c := 2 * len(data)
	if c == 0 {
		c = 1
	}
	newData := arena.MakeSlice[T](a, len(data)+1, c)
	copy(newData, data)
	data = newData
	data[len(data)-1] = v
	return data
}
```

korniltsev: That's likely true.
Forthermore updating golang is likely to help. The PR was written at the time of golang 1.20 and there were fixes (using fully quallified generics name in profiles instead of `[...]`)
Note: this code is completely dead and not used right now (was used before grafana merge) .

PR #1813, opened 2023-01-18
feat(webapp): sync crosshair in different timelines

eh-am:


korniltsev (approved):
I completely dont understand how this all works but if this works it looks good to me

eh-am:
lmao all good as long as functionality is what you had in mind :)

PR #1816, opened 2023-01-25
feat: pprof write batch parsing

korniltsev:


pkg/convert/pprof/streaming/parser_streaming_writebatch.go
```diff
@@ -0,0 +1,288 @@
...
+		if wb == nil {
+			continue
+		}
+
+		sb := wb.wb.StackBuilder()
+		sb.Reset()
+		for _, frame := range p.tmpSample.tmpStack {
+			sb.Push(frame)
+		}
+		stackID := sb.Build()
```
kolesnikovae, line 112: Depending on the implementation, `Push` and `Build` may be quite expensive and we may want to minimize the number of calls. I think we can optimise it in feature(!): instead of iterating over frames, we can build an intermediate tree, with a node defined as follows:
```go
type node struct {
  loc    uint32 
  labels []uint64 // each entry points to the labels set (high 32) and offset in the value array (low 32); number of values is known
  values []int64  // groups of values (value idx within a group matches sample type index)
  // ...
}
```

Then, we can traverse the tree and build each frame and stack only once. The reason is that one of our popular "worst" cases assumes lots of unique label sets (e.g exemplars) pointing the same stacks. It _may_ also facilitate diff of cumulative profiles.

pkg/convert/pprof/profile.go
```diff
@@ -207,6 +188,59 @@ func (p *RawProfile) Parse(ctx context.Context, putter storage.Putter, _ storage
...
 
+func (p *RawProfile) ParseWithWriteBatch(c context.Context, wb stackbuilder.WriteBatchFactory, fallback storage.Putter, md ingestion.Metadata) error {
+	cont, err := p.handleRawData()
+	if err != nil || !cont {
+		return err
+	}
+	parser := streaming.NewStreamingParser(streaming.ParserConfig{
+		SpyName:       md.SpyName,
+		Labels:        md.Key.Labels(),
+		Putter:        fallback,
```
kolesnikovae, line 199: I'm not 100% sure I understand why we need this? Is it something outside OSS?

kolesnikovae (approved):
LGTM!

pkg/convert/pprof/profile.go
korniltsev: removed, it was a leftover from a first POC which could not do cumulative, so it used trees as fallback

PR #1840, opened 2023-02-08
feat(flamegraph): allow regex in search

eh-am:
image

It's useful when investigating a flamegraph, like highlighting only certain nodes, or highlighting all that don't match an expression (as in the screenshot).


petethepig (approved):
I'm slightly concerned about performance implications of creating a regex object for each node in the table. But other than that looks good.

eh-am:
@petethepig good catch, updated to cache that object.

PR #1854, opened 2023-02-14
fix: nil pointer while tag key contains invalid characters

nonbutAworker:


pkg/storage/labels/labels.go
```diff
@@ -20,6 +21,9 @@ func New(db *badger.DB) *Labels {
 func (ll *Labels) PutLabels(labels map[string]string) error {
 	return ll.db.Update(func(txn *badger.Txn) error {
 		for k, v := range labels {
+			if err := flameql.ValidateTagKeyWithoutReserved(k); err != nil {
+				return err
+			}
```
kolesnikovae, line 26: I like that now labels are validated here, but I'm afraid when we call `storage.Put` we need to validate them earlier, before the key is accessed for the first time in this section:
https://github.com/pyroscope-io/pyroscope/blob/65e1d6923461cb29ca46fe777bd6355bbaba1d06/pkg/storage/storage_put.go#L37-L66

Otherwise, we may store invalid app name or invalid labels / segment key (if it is an exemplar)

pkg/flameql/flameql.go
```diff
@@ -83,6 +83,19 @@ func ValidateTagKey(k string) error {
...
+		return ErrTagKeyIsRequired
+	}
+	for _, r := range k {
+		if !IsTagKeyRuneAllowed(r) {
+			return newInvalidTagKeyRuneError(k, r)
+		}
+	}
+	return nil
+}
+
```
kolesnikovae, line 98: I think we also need to validate the app name. What do you thing about moving this validation to [`segment key`](https://github.com/pyroscope-io/pyroscope/blob/main/pkg/storage/segment/key.go), e.g `segment.ValidateKey(k *Key)`?  We already validate syntax there when parse a string key representation, and a function for validating a structured representation is what the package lacks

kolesnikovae (commented):
@nonbutAworker thank you so much for the contribution! 🙌🏻  I've added a couple of suggestions, please give them a look

PR #1862, opened 2023-02-17
fix: add structed validation for tag key and app name in case nil pointer panic

nonbutAworker:
Hi, 
So sorry for my mistake, I want to revert my code so I click `sync fork`, never thought of that means GitHub will delete my old repo and close relevant [PR](https://github.com/pyroscope-io/pyroscope/pull/1854) automatically then create a new one

Just like the comments from my wrongly closed [PR](https://github.com/pyroscope-io/pyroscope/pull/1854) say, I add `segment.ValidateKey(k *Key)` function for both tag key and app name, add caller at where the key is assigned, so we can validate it from the start,  before it's used

after changing code, I ran some tests. now both client and server will report error about invalid characters:
img_v2_32ede418-bf3a-47d6-bb67-b9f95e81cf4g
img_v2_603d8ba7-3635-453e-87b0-8ba0a3a887dg

### DIFF
--- a/pkg/convert/pprof/streaming/parser_streaming.go
+++ b/pkg/convert/pprof/streaming/parser_streaming.go
@@ -321,7 +321,10 @@ func (p *VTStreamingParser) put(stIndex int, st *valueType, l Labels, t *tree.Tr
 			return false, err
 		}
 	}
-	pi.Key = p.buildName(sampleType, p.ResolveLabels(l))
+	pi.Key, err = p.buildName(sampleType, p.ResolveLabels(l))
+	if err != nil {
+		return false, err
+	}
 	err = p.putter.Put(p.ctx, &pi)
 	return sampleTypeConfig.Cumulative, err
 }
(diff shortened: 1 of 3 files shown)


pkg/storage/segment/key.go
```diff
@@ -55,6 +55,21 @@ func ParseKey(name string) (*Key, error) {
...
+			}
+		} else {
+			if err := flameql.ValidateTagKey(key); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
```
kolesnikovae, line 72: As an extra measure it would be nice to validate is the key is nil

pkg/convert/pprof/streaming/parser_streaming.go
```diff
@@ -361,12 +364,17 @@ func (p *VTStreamingParser) ResolveLabels(l Labels) map[string]string {
...
 		labels[k] = v
 	}
 	labels["__name__"] += "." + sampleTypeName
-	return segment.NewKey(labels)
+	key := segment.NewKey(labels)
+	err := segment.ValidateKey(key)
+	if err != nil {
+		return nil, err
+	}
+	return key, nil
```
kolesnikovae, line 377: One suggestion from my side – please consider moving the validation to [`storage.Put`](https://github.com/pyroscope-io/pyroscope/blob/c79a22fbe459f4c111b338ce8d15a247087243f1/pkg/storage/storage_put.go#L29), or better to [`parser.Ingest`](https://github.com/pyroscope-io/pyroscope/blob/c79a22fbe459f4c111b338ce8d15a247087243f1/pkg/parser/parser.go#L27): this will make it so that the key is validated regardless of the input format. Otherwise we need to add this validation to all the parsers we have

kolesnikovae (approved):
@nonbutAworker awesome work, thank you for the contribution!

pkg/convert/pprof/streaming/parser_streaming.go
kolesnikovae: Oh, I realised that we actually need to validate keys after [`parser.Ingest`](https://github.com/pyroscope-io/pyroscope/blob/c79a22fbe459f4c111b338ce8d15a247087243f1/pkg/parser/parser.go#L27). So, I would move the validation to [`storage.Put`](https://github.com/pyroscope-io/pyroscope/blob/c79a22fbe459f4c111b338ce8d15a247087243f1/pkg/storage/storage_put.go#L29)

PR #1865, opened 2023-02-21
chore: add consul discovery example

kolesnikovae:
Note that currently the example uses `pyroscope-dev:main` – it's better to release the current version and update the image


pkg/scrape/target.go
```diff
@@ -103,6 +103,13 @@ func (t *Target) offset(interval time.Duration) time.Duration {
...
+	for _, l := range t.labels {
+		if l.Name == model.SpyNameLabel && l.Value != "" {
+			return l.Value
+		}
+	}
+	// N.B: There is no need to check discovered labels:
+	// in most cases, the label is created at relabeling.
 	for _, l := range t.discoveredLabels {
 		if l.Name == model.SpyNameLabel && l.Value != "" {
 			return l.Value
```
kolesnikovae, line 115: The fix makes it possible to actually override `__spy_name__` in relabelling

kolesnikovae:
@eastfisher I've prepared an example of consul discovery setup so that everyone could run it locally – I would really appreciate if you could take a look

examples/golang-pull/consul/app/main.go
```diff
@@ -0,0 +1,108 @@
...
+			wg.Add(6)
+			go fastFunction(c, &wg)
+			go fastFunction(c, &wg)
+			go fastFunction(c, &wg)
+			go slowFunction(c, &wg)
+			go slowFunction(c, &wg)
+			go slowFunction(c, &wg)
+			wg.Wait()
+		}
+	})
```
eastfisher, line 107: What about block the main() function at the end? When the process exit, it will deregister from Consul, then Pyroscope cannot scrape the profile data from it.

```go
    ...
			wg.Wait()
		}
	})
    // add this line to block main()
	(chan struct{})(nil) <- struct{}{}
}
```

eastfisher: Executing fastFunction() and slowFunction() take enough time for Pyroscope to scrape the profile data, so it is not a problem.

eastfisher (approved):
LGTM!

examples/golang-pull/consul/app/main.go
kolesnikovae: Yeah, this is a standard Go program we use as an example (I borrowed it from `static`). The loop is synchronous and blocks main, therefore it should be fine.

Regarding the de-registration – definitely makes sense, thanks for pointing this out!

kolesnikovae:
Thank you for the review @eastfisher!

petethepig (approved):

PR #1867, opened 2023-02-22
fix(webapp): always render annotations in the viewport

eh-am:
Closes https://github.com/pyroscope-io/pyroscope/issues/1806

#Before
image

# After
image


Rperry2174 (approved):
Nice thx!

PR #1872, opened 2023-02-23
chore: make low level storage settings configurable

kolesnikovae:
Some of the constants we currently use are not suitable for all cases. Those are low level setting that should be changed with caution: I intentionally didn't make them available in the CLI help output.

Default values of newly added configuration params:
```yaml
badger-value-log-size: 1GB
badger-gc-diff-size: 1GB
badger-gc-discard-ratio: 0.7
badger-gc-task-interval: 5m
metrics-update-task-interval: 10s
writeback-task-interval: 1m
eviction-task-interval: 20s
retention-task-interval: 10m
```

In particular, too big default Badger value log size (1GB) in some cases prevents removing data from the disk for long time because of how garbage collection is implemented in Badger. One way to mitigate the problem is to reduce its size and set low discard ratio:
```yaml
badger-value-log-size: 100MB
badger-gc-diff-size: 100MB
badger-gc-discard-ratio: 0.1
```

These parameters make it so that GC starts rewrite value log files if they have more than 10% garbage (10MB). Please note that this comes at the cost of higher CPU consumption.


kolesnikovae:
Closing for now

PR #1873, opened 2023-02-23
fix: segment node truncation

kolesnikovae:
Currently, when segment nodes are removed because of the retention policy, the change is not persistent: unless the segment is modified at write, it won't be written to the disk. Thus, the segment is only deleted when all of its nodes have expired. In most cases, this is not a big problem: after the first truncation (run of the retention task) nodes are removed again.

The change makes it so that the segment is marked as modified after removing expired nodes, so that the write-back thread flushes it to the disk next time.


kolesnikovae:
Closing for now

PR #1875, opened 2023-02-23
chore: prune tree based on `MaxNodes` when converting json files

eh-am:
Closes https://github.com/pyroscope-io/pyroscope/issues/1874

Not quite sure about memory usage of this


kolesnikovae (approved):
LGTM!

PR #1877, opened 2023-02-24
feat(datasource-plugin): add support for label queries

kolesnikovae:
The implementation and query syntax is similar to the Prometheus datasource for convenience: https://grafana.com/docs/grafana/latest/datasources/prometheus/template-variables/#use-query-variables


kolesnikovae:
@eh-am if you have a chance, please give the PR a look – followed by our discussion, I took an attempt to implement label queries

packages/pyroscope-datasource-plugin/src/datasource.ts
```diff
@@ -47,32 +48,85 @@ export class DataSource extends DataSourceApi<
...
+    if (labelNamesQuery) {
+      return this.queryLabelNames(labelNamesQuery[1]);
+    }
+
+    const labelValuesQuery = expandedQuery.match(labelValuesRegex);
+    if (labelValuesQuery) {
+      return this.queryLabelValues(labelValuesQuery[1], labelValuesQuery[2]);
+    }
+
+    return [{ text: '' }];
```
eh-am, line 74: We gotta return empty arrays, otherwise we get a single empty item:
Screen Shot 2023-02-24 at 20 01 39

eh-am (commented):
This is pretty sweet!


Screen Shot 2023-02-24 at 20 13 26
Screen Shot 2023-02-24 at 20 21 24

https://user-images.githubusercontent.com/6951209/221282883-01957dd4-1853-425e-b146-cc8e9f22fdb0.mp4


One thing I find necessary though is to be able to choose no tag. In the example above I may want to look at `simple.app.cpu{}`, it would be nice to have some kind of catch all using `.*`, but that would require updating the endpoints etc I assume. 

One last thing is that we can't forget about updating the README with this new functionality.

eh-am:
>One thing I find necessary though is to be able to choose no tag. In the example above I may want to look at simple.app.cpu{}, it would be nice to have some kind of catch all using .*, but that would require updating the endpoints etc I assume.
One idea is to in the datasource, if both sides of the `=` are empty, to just discard it altogether? That way, if you select the 2 variables to be an empty value, it will be translated `myapp{=}` -> `myapp{}`

kolesnikovae:
> > One thing I find necessary though is to be able to choose no tag. In the example above I may want to look at simple.app.cpu{}, it would be nice to have some kind of catch all using .*, but that would require updating the endpoints etc I assume.
> 
> One idea is to in the datasource, if both sides of the = are empty, to just discard it altogether? That way, if you select the 2 variables to be an empty value, it will be translated myapp{=} -> myapp{}

Yeah, also was thinking about this. But honestly I doubt variable substitution mechanism in Grafana is good/designed for this – I can picture one may want to use multiple labels, use different operators, have a sophisticated logic of label dependencies, etc. In general, I can't even find a good example when to use `label_names`. I think we should implement a proper query builder instead.

I was playing with various scenarios and discovered that repeating panel or row does not seem to work as expected (both for apps and label values) – when I specify multiple values, our datasource gets an array of values:

image

The variable:
image 

The panel:
image

What do you think? Is it what we should handle in our panel plugin or I did miss something?

kolesnikovae:
Ok, I figured out how to [make it work](https://github.com/pyroscope-io/pyroscope/pull/1877/commits/4abbb6c8b20274d2a12d643fba0f69c5a832bbac) with multi-variable choice: 
image

packages/pyroscope-datasource-plugin/grafana-provisioning/dashboards/with-dynamic-variable.json
```diff
@@ -77,7 +77,7 @@
         "multi": false,
         "name": "appName",
         "options": [],
-        "query": ".*",
+        "query": "apps()",
```
kolesnikovae, line 80: This changes makes it so that variable queries like `.*` will stop working. I'm not sure if we need to maintain the backward compatibility

@eh-am what do you think?

eh-am: Nah go for it, this change hasn't even been released yet

eh-am (approved):

PR #1886, opened 2023-03-06
chore: remove duplicated expectFormats check

sfwn:
`expectFormats` check is already done inside `renderParametersFromRequest` at [pkg/server/render.go#L264](https://github.com/pyroscope-io/pyroscope/blob/main/pkg/server/render.go#L264).


kolesnikovae (approved):
LGTM. @sfwn Thanks for the contribution!

Issue #1885, opened 2023-03-03
Feature Request: CRI-O Support


isugimpy:
Attempted to try out Pyroscope today, was disappointed to find that the ebpf profile exits immediately with the message `Error: unknown cri cri-o://1.24.4`.  CRI-O is a standard and supported CRI, and conforms to the spec.  It'd be nice if support for it could be added.

korniltsev:
@isugimpy Could you share `/proc/${PID}/cgroup` of any pod in your cri-o cluster?

isugimpy:
Sure!

`0::/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-podb57320a0_e7eb_4ac8_a791_4c4472796867.slice/crio-0ecc7949cbaf17e883264ea1055f60b184a7cb264fd759c4a692e1155086fe2d.scope`

korniltsev:
@isugimpy I assumed containerID has a `cri-o://` prefix. Lets double check it before I merge the fix.
We can check it with `kubectl get pods -A -o json | jq '.items[].status.containerStatuses[].containerID'` .

isugimpy:
This won't have been the same pod from this morning, but `cri-o://9ba4b1139f3e37d7ad36dfcb49dd0ac79ac887c4c15eea2f59993d5ef86b9290` is a valid container ID from one of my clusters.

vishnuchalla:
@korniltsev I have come across the same issue, when trying to install eBPF agent on my openshift cluster.
```
time="2023-06-13T01:52:44.615741" level=info msg="to disable logging from pyroscope, specify -no-logging flag" file=" exec/cli.go:50"
Error: unknown cri cri-o://1.25.3-2.rhaos4.12.git592efcd.el8
```
Is there a workaround/alternative for this issue. So that I can use eBPF agent to profile my entire openshift cluster?

korniltsev:
https://github.com/grafana/pyroscope/pull/1890/files

PR #1890, opened 2023-03-13
fix: ebpf crio runtime container matching  #1885

korniltsev:


lbogdan:
@korniltsev Any reason this was closed?

korniltsev:
there is no ebpf profiler in the pyroscope binary now.

now it lives in grafana agent and it was fixed there

PR #1898, opened 2023-03-15
chore: adds clarifications RE Apache 2 packages to LICENSING.md

petethepig:
Adds clarifications that our frontend packages are Apache 2:
* `@pyroscope/flamegraph`
* `@pyroscope/models`


lizthegrey:
Thank you very much. This is required so that partners can have confidence that we can maintain our integrations with you and not need to reinvent the wheel for how data from Pyroscope/Pyroscope Cloud is shown.

PR #1912, opened 2023-03-28
fix(webapp): fix heatmap y axis

eh-am:
When querying exemplars and there's no data, the backend will return `minValue: 0, maxValue: 0`, which breaks the y axis calculation. Worse, it will generate duplicate values (a bunch of `0.0`s), which are used as the [`key`](https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key) in the React component. This confuses React, which upon querying exemplars that exist, will not be aware that the items must be removed (since the key is duplicated).

It's better exemplified in the following screenshots:
image
image

To fix this I just make it very clear that the Y axis can not be constructed, and default to a single value: 0.


kolesnikovae (approved):
Thank you for fixing this!

PR #1915, opened 2023-04-06
FlameGraph: Change buttons to grafana/ui

aocenas:
Added @grafana/ui and used buttons from there for the toolbar plus some additional small refactoring (see comments)

Screenshot from 2023-04-06 16-09-55

### DIFF
--- a/packages/pyroscope-flamegraph/src/Toolbar.module.scss
+++ b/packages/pyroscope-flamegraph/src/Toolbar.module.scss
@@ -23,61 +23,9 @@ $buttonHeight: 37px;
   color: var(--ps-toolbar-icon-color);
 }
 
-.toggleViewButton {
-  height: $buttonHeight;
-  min-width: auto;
-  width: 40px;
-  margin: 0 2px;
-  padding: 0;
-  border-color: transparent;
-  background-color: transparent;
-  display: flex;
-  justify-content: center;
-  align-items: center;
-
-  svg {
-    width: 22px;
-    height: 22px;
-  }
-
-  &.selected {
-    background-color: var(--ps-fl-toolbar-btn-bg);
-
-    &:hover {
-      background-color: var(--ps-fl-toolbar-btn-bg);
-    }
-
-    svg {
-      color: white;
-    }
-  }
-}
-
-.fitModeButton {
-  @extend .toggleViewButton;
-}
-
-.collapseNodeButton {
-  min-width: initial;
-  width: 40px;
-  height: $buttonHeight;
-  margin: 0 2px;
-  padding: 0;
-  background-color: transparent;
-  border-color: transparent;
-}
-
-.resetViewButton {
-  @extend .collapseNodeButton;
-  width: 40px;
-  min-width: initial;
-  margin: 0 2px;
-  padding: 0;
-}
-
 .divider {
   width: 1px;
-  margin: 0 2px;
+  margin: 0 4px;
   height: $buttonHeight;
   background-color: var(--ps-ui-border);
 }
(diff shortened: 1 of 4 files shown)


packages/pyroscope-flamegraph/src/Toolbar.module.scss
```diff
@@ -23,85 +23,17 @@ $buttonHeight: 37px;
   color: var(--ps-toolbar-icon-color);
 }
 
-.toggleViewButton {
```
aocenas, line 26: Styling for the buttons were removed.

packages/pyroscope-flamegraph/src/Toolbar.tsx
```diff
@@ -66,45 +53,60 @@ const calculateCollapsedItems = (
...
 
   return {
-    isCollapsed,
-    handleMoreClick,
-    collapsedItemsNumber,
+    isMenuOpen,
+    handleMoreClick: () => {
+      setMenuOpen((v) => !v);
+    },
+    hiddenItems: toolbarSections
```
aocenas, line 106: This now directly returns the visible and hidden items.

packages/pyroscope-flamegraph/src/Toolbar.tsx
```diff
@@ -293,31 +236,40 @@ const Toolbar = memo(
...
   return (
     <Tooltip placement="top" title="Collapse nodes above">
       <div>
-        <Button
+        <GButton
           disabled={!selectedNode.isJust}
-          onClick={onClick}
-          className={styles.collapseNodeButton}
+          onClick={() => {
+            if (selectedNode.isJust) {
```
aocenas, line 266: This may be a bit of a preference, but I think it's better to use as minimal API surface as possible so we don't have to lear too many libraries. This also probably has the same effect as just assigning empty function in case `!selectNode.isJust` because at the same time the button is disabled in that case.

packages/pyroscope-flamegraph/src/Toolbar.tsx
```diff
@@ -359,71 +310,60 @@ function FitMode({
...
+      value: 'both',
+      Icon: TablePlusFlamegraphIcon,
+    },
+    {
+      label: 'Flamegraph',
+      value: 'flamegraph',
+      Icon: FlamegraphIcon,
+    },
+  ];
+  if (flamegraphType === 'single') {
```
aocenas, line 355: Refactored this so options aren't defined 2 times.

packages/pyroscope-flamegraph/src/Toolbar.tsx
eh-am: Yeah that's fine, historically it was used when typescript hasn't been fully added, so it was a way to force people to deal with undefined cases.

eh-am:
Just one thing, the merge commit needs to follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) and needs to indicate a breaking change, so that learn publishes it as a new major version. 
https://github.com/grafana/pyroscope/blob/e7cf318b242c81342c84ae431558a338a14adc87/.github/workflows/publish-lib-to-npm.yml#L127

aocenas:
Changed the base branch to feature branch so we can do release later on when everything is in place

packages/pyroscope-flamegraph/src/Toolbar.tsx
```diff
@@ -293,31 +236,40 @@ const Toolbar = memo(
   }
 );
 
+function ToolbarButtons(props: { items: React.ReactNode[] }) {
+  return (
+    <>
+      {props.items.map((el, i) => (
+        // eslint-disable-next-line react/no-array-index-key
```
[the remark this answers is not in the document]
aocenas: Then the eslint isn't happy because of the `key={i}`.

PR #1916, opened 2023-04-10
docs(dotnet): add dynamic start/stop api example

korniltsev:
This PR adds an example of a dynamic start/stop API for Pyroscope in  .NET integration. 

The example includes several API endpoints for controlling Pyroscope's profiling settings and testing various scenarios that may cause performance issues. 

See docs [here](https://github.com/pyroscope-io/docs/pull/168)


korniltsev:
cc @Rperry2174 @petethepig

examples/dotnet/rideshare/example/Program.cs
```diff
@@ -26,12 +30,83 @@ public static void Main(string[] args)
...
+            Pyroscope.Profiler.Instance.SetAllocationTrackingEnabled(enable);
+            return "OK";
+        });
+        app.MapGet("/pyroscope/contention", (HttpRequest request) =>
+        {
+            var enable = request.Query["enable"] == "true";
+            Pyroscope.Profiler.Instance.SetContentionTrackingEnabled(enable);
+            return "OK";
+        });
+        app.MapGet("/pyroscope/contention", (HttpRequest request) =>
```
kolesnikovae, line 59:
```suggestion
        app.MapGet("/pyroscope/exception", (HttpRequest request) =>
```

examples/dotnet/rideshare/example/Program.cs
```diff
@@ -26,12 +30,83 @@ public static void Main(string[] args)
...
             return "Car ordered";
         });
+        
+        
+        app.MapGet("/pyroscope/cpu", (HttpRequest request) =>
+        {
+            var enable = request.Query["enable"] == "true";
+            Pyroscope.Profiler.Instance.SetCPUTrackingEnabled(enable);
+            return "OK";
+        });
```
kolesnikovae, line 46: A nit: for the sake of demo we could also allow disabling profilers

kolesnikovae (commented):
Looks great! One minor thing: HTTP route for the exception profiler collides with the contention profiler

examples/dotnet/rideshare/example/Program.cs
korniltsev: Sorry I don't understand the comment. if I pass something other then true to the query, it will disable profiling

kolesnikovae: Oh right, my bad!

PR #1935, opened 2023-04-24
docs: example remote writing to phlare

eh-am:
It doesn't do any authentication though. Which is added in https://github.com/grafana/pyroscope/pull/1934

Running phlare:
tab 1: make build && ./phlare
tab 2: yarn && yarn dev

tab 3:  On pr branch
To run this (assuming phlare is already running on `http://localhost:4100`):
```
cd examples/rideshare-phlare
docker-compose pull
docker-compose up
```


Rperry2174 (approved):
LGTM

Rperry2174 (approved):
LGTM

petethepig (approved):
what's the context? where are we going to use this? 

if it's without any context it's unclear that you should be running phlare separately on port 4100.

If we are going to be showing this to people I would maybe put an actual `phlare` service using phlare docker image into docker-compose.yml

Rperry2174:
This is internal use for testing Pyroscope ui, in phlare, with real data (which is why phlare is not included) ... I think the readme is a good idea to explain that

PR #1936, opened 2023-04-25
fix the typo in the gorm tag of the annotation table

chengjoey:
fix the typo in the gorm tag of the annotation table


kolesnikovae:
Hello @chengjoey. Thank you for the contribution!

@eh-am, I think it is a safe change that does not affect backward compatibility – do you see any risks?

kolesnikovae (approved):

PR #1947, opened 2023-05-10
benchmark: use the more efficient strings.Builder

cuishuang:
strings.Builder uses *(*string)(unsafe.Pointer(&b.buf)) to save the application memory operation when converting to a string


kolesnikovae:
Hello @cuishuang,

Indeed, use of `strings.Builder` is preferable here – nice catch! Thank you for the contribution!

kolesnikovae (approved):

PR #1963, opened 2023-05-30
fix(docs): update grafana-agent example

korniltsev:
remove phlare references in agent flow configuration
depends on https://github.com/grafana/agent/pull/4013


korniltsev:
I guess we should not merge until the agent is released

Rperry2174:
I think its fine to merge since it's a little hidden for now (we don't link to it in docs yet) grafana agent should be updated soon

PR #1969, opened 2023-06-14
fix(server): set `Content-Type` to JSON in label handlers

tdakkota:
For now,  `/labels` and `/label-values` handlers do not set `Content-Type` explictly, so `net/http` wrongly sets it to `text/plain`.

This PR fixes the problem by using `httpUtils.WriteResponseJSON` function.

### DIFF
--- a/pkg/server/labels.go
+++ b/pkg/server/labels.go
@@ -1,7 +1,6 @@
 package server
 
 import (
-	"encoding/json"
 	"net/http"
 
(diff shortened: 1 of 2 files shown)


Rperry2174:
hey @tdakkota thanks for the PR! Can you give some more information to the description on why you created this PR / what you're attempting to fix?

cyriltovena (approved):
LGTM

Rperry2174:
Sorry for the delay! Thanks @tdakkota !

PR #1998, opened 2023-07-14
fix: update and/or remove dependency lock files

darrenjaneczek:
Goal is to update dependencies as an initial flush out of security concerns

- example projects are treated as stand-alone and retain yarn.lock files
- removed package lock files from examples
- removed yarn.lock files from sub packages
- added suitable .gitignore entries for yarn v3.6


webapp/javascript/pages/exemplars/ExemplarsSingleView.tsx
```diff
@@ -145,21 +145,22 @@ function ExemplarsSingleView() {
...
+                  message="Select an area in the heatmap to get started"
+                />
+                <img
+                  className={styles.gif}
+                  src={heatmapSelectionPreviewGif}
+                  alt="heatmap-selection-gif"
+                />
+              </div>
+            </Box>
+          )}
```
darrenjaneczek, line 163: This is an example of a formatting fix; no functionality changed.

darrenjaneczek:
This branch has issues that break the build. It did too much at once.
I will likely start over one dependency at a time.

PR #2002, opened 2023-07-17
chore: fix d3-color version to 3.1.0

darrenjaneczek:


Rperry2174 (approved):
lgtm!

yarn.lock
```diff
@@ -8879,7 +8879,7 @@ d3-chord@3:
     d3-path "1 - 3"
 
 "d3-color@1 - 3", d3-color@3:
-  version "3.0.1"
+  version "3.1.0"
   resolved "https://registry.yarnpkg.com/d3-color/-/d3-color-3.0.1.tgz#03316e595955d1fcd39d9f3610ad41bb90194d0a"
```
abannachGrafana, line 8883: - Is d3-color a transitive dependency?
- does the version in this `resolved` url need updating?

abannachGrafana: Could you use the `resolutions` block in the `package.json` to codify this change? Just concerned that a different package.json change will re-resolve this back to the old version.

PR #2006, opened 2023-07-17
fix: repace word-wrap with secure variant

darrenjaneczek:
- Force version of:
  ```
  "optionator": "^0.9.3",
  ```
  to ensure import of an alternative implementation of `word-wrap`, which doesn't trigger security concerns.


Rperry2174 (approved):
lgtm!

PR #2008, opened 2023-07-17
fix: limit time picker presets

### LINKED ISSUE #1999: Update presets on time picker
The initial time picker had a bunch of defaults that we copied from Grafana's old time picker. In reality, it is definitely not recommended for people to query, for example, 5 years of profiling data. 
image

We should update the date picker to look / function similar to the new grafana date picker except limit the presets to no longer than   6 months (maybe even shorter) 
image

darrenjaneczek:
- Progress toward #1999

The new maximum time range preset if 90 days.

image


Rperry2174:
cc @cyriltovena whats the longest preset we should support here?

simonswine (approved):
I am ok with leaving the 90 days for now

LGTM

PR #2011, opened 2023-07-18
WIP: Seek all pages at once (or maybe not)

simonswine:
This uses the reader returned from the bucket directly.

Ideally this will also allow to fetch bigger chunks for the bucket to reduce object store query latency.


kolesnikovae (commented):
Awesome stuff. Also, we may want to use a read-ahead https://github.com/klauspost/readahead reader

simonswine:
I will be closing this, I think the way forward it make the parquet reader more aware of the structure of the parquet file, by keeping its own index. #2392

PR #2014, opened 2023-07-18
clarify licensing

petethepig:


Rperry2174:
is the idea to remove everything from the og folder eventually? Why not just move those packages out of the og folder since they don't overlap with phlare stuff?

petethepig:
@Rperry2174 This is just a PR that clarifies licensing, the location of it is a separate conversation imo.

RE location, you're right about the eventual goal of removing `/og` directory. But if I just move `/og/pkg/*` to `pkg`, things will overlap some, for example there's an `agent` package in both, and it means different things in each project. Also, I would like to be able to easily tell if a package is from og or not og pyroscope so I can delete og ones more easily, and if I put everything under `/pkg` that's going to be a mess.

So, to make it easier for everyone, I'm going to do these step by step.

Rperry2174 (approved):
lgtm

Issue #2010, opened 2023-07-18
Merge grafana/phlare repo into grafana/pyroscope


simonswine:
## Problem
End-Users, contributors and developers face a fair bit of confusion with the two projects and especially the wording of the Phlare README is misleading

## Goals
* Keep commit history intact
* Ensure commit messages that reference to issues and PRs stay correct (Avoid that #123 in a message from grafana/phlare will link to a different PR in the grafana/pyroscope repo)
* Do not loose the stars
* Be able to release OG and Pyroscope weeklies

### Phase 1

* [x] Script to do the merge of the repos: https://github.com/grafana/phlare/pull/808 

### Phase 2: Move repo atomically 

* [x] Run the script on 2023-07-18 13:00 UTC, ensure PRs no longer can be merged into grafana/phlare, make Grafana Pyroscope available in the `next` branch, switch default branch to next.

```[tasklist]
### Phase 3
- [x] Improve README.md in both projects.
- [ ] https://github.com/grafana/pyroscope/issues/2000
- [x] Rename phlare binaries to pyroscope https://github.com/grafana/pyroscope/pull/2023
- [ ] https://github.com/grafana/pyroscope/issues/2149
- [ ] @petethepig  Clean up the `og/` folder, decide what to keep what to remove and what to move out.
- [ ] Determine the future of the `pyroscope` subcommands. (currently phlare only supports `pyroscope server`)
- [ ] https://github.com/grafana/pyroscope/issues/2151
- [x] Ensure the `make generate` works again @cyriltovena https://github.com/grafana/pyroscope/pull/2140
- [x] Fix weekly builds @simonswine
- [ ] tag api/ and ebpf/ folder
- [ ] Merge and/or rename helm chart for pyroscope and phlare
- [x] [Update MAINTAINERS.md](https://github.com/grafana/pyroscope/pull/2155)
```

```[tasklist]
### Phase 4
* [ ] [Update examples](https://github.com/grafana/pyroscope/issues/2188)
* [ ]  The `og/` folder has disappeared
* [ ] No code imports of github.com/pyroscope-io/pyroscope anmymore
* [ ] First release of Grafana Pyroscope
```

petethepig:
Little update from me, I deleted some things from `/og` directory, and started figuring out what to do with go packages (I will move some that we use and delete the rest).

PR #2017, opened 2023-07-18
OG Go code cleanup

petethepig:
Part of #2010 

This PR:
* moves OG pyroscope go code that we use from `og/pkg` to `pkg/og`
* removes `pyroscope-io/pyroscope` dependency
* deletes all other OG pyroscope go code


go.work.sum
```diff
@@ -343,7 +344,6 @@ github.com/chzyer/readline v1.5.1/go.mod h1:Eh+b79XXUwfKfcPLepksvw2tcLE/Ct21YObk
 github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1 h1:q763qf9huN11kDQavWsoZXJNW3xEE4JJyHa5Q25/sd8=
 github.com/chzyer/test v1.0.0 h1:p3BQDXSxOhOG0P9z6/hGnII4LGiEPOYBhs8asl/fC04=
 github.com/chzyer/test v1.0.0/go.mod h1:2JlltgoNkt4TW/z9V/IzDdFaMTM2JPIi26O1pF38GC8=
-github.com/cilium/ebpf v0.11.0/go.mod h1:WE7CZAnqOL2RouJ4f1uyNhqr2P4CCvXFIqdRDUgWsVs=
```
korniltsev, line 346: I wonder why is it deleted? o_O

I'd expect it to be still used https://github.com/grafana/pyroscope/blob/next/ebpf/go.mod#L7

simonswine: Personally not too familiar with workspaces, but according to [stackoverflow](https://stackoverflow.com/questions/71611196/what-does-the-go-work-sum-file-track-in-go-1-18) `go.work.sum`. 

It would only contain things not contained/different from the referenced modules in `go.work`.

As `epbf/go.sum` contains this particular sum it can be removed. I also think our PR ci doesn't do a `go mod tidy` check yet on `ebpf`. That would also explain why it has been added in the first place.

simonswine (approved):
LGTM. 

Thanks for your work here, I am all for removing as much as we can. Once something is still required it can always make a comeback.

There are still a few linting issues to sort out. (I disabled linting of `og/` before)

kolesnikovae (approved):

go.work.sum
simonswine: This should prevent it going forward: https://github.com/grafana/pyroscope/pull/2020

PR #2018, opened 2023-07-19
Adds Export to flamegraph.com

cyriltovena:
Now that we have CORS setup on the flamegrah.com side, we can simply call the API from  here.

I also merged improvement to use `saveAs` library


simonswine (approved):
LGTM


I have given that a try and it works for me

cyriltovena:
I'm adding more commit as I'm testing it in Grafana too :/

PR #2023, opened 2023-07-19
Update binary name to pyroscope

simonswine:
Also updates makefile targets and some folder names


.github/workflows/test.yml
```diff
@@ -84,7 +84,7 @@ jobs:
           node-version: 18.15.0
           cache: yarn
       - name: Build image Phlare
-        run: make docker-image/phlare/build "BUILDX_ARGS=--cache-from=type=gha --cache-to=type=gha"
+        run: make docker-image/pyroscope/build "BUILDX_ARGS=--cache-from=type=gha --cache-to=type=gha"
```
kolesnikovae, line 87: `.goreleaser` needs some fixes as well but I'm not sure if it is supposed to be part of this step. Same for helm chart

cyriltovena (approved):
LGTM

.github/workflows/test.yml
simonswine: Will try to get weekly builds going next, there is some more work there yes

simonswine: Most recent commit fixed `make deploy`, we should not touch the helm chart for now

PR #2132, opened 2023-07-19
Ensure loser tree closes underlying sequences properly on error

simonswine:
When an error happened on a single one of the iterators during initialisation, then `Close()` would call `close` with a `null/nil` value,  causing some panic. It also would not close the remaining iterators.

This adds tests and the fix


cyriltovena (approved):
LGTM

PR #2136, opened 2023-07-20
Compact symdb

kolesnikovae:
The PR contains implementation of the symbols rewriter which is used in compaction. The design of the implementation is heavily influenced by the following factors:
 - On compaction, we may want to get only a subrange of profiles from a source block. The range may be limited to a specific series set, time range, or both.
 - Symbolic information of a source block must be loaded into memory entirely, because there is no efficient way to load it in parts.

TODO:
 - [ ] Validity verification.
 - [ ] Performance tests.

Currently, the complete compaction flow is incomplete which complicates further testing. Specifically, I didn't manage to query any data from a storage with a compacted block: it either returns empty result, or fails with the error `panic: label hash conflict`.

Nevertheless, I think that we should pull this change into the main compaction branch and continue work there. Moreover, I'd like to get the compaction branch merged into main (`next`).

### DIFF
--- a/pkg/phlaredb/block_querier.go
+++ b/pkg/phlaredb/block_querier.go
@@ -435,6 +435,11 @@ func (b *singleBlockQuerier) Index() IndexReader {
 	return b.index
 }
 
+func (b *singleBlockQuerier) Symbols() SymbolsReader {
+	// TODO(kolesnikovae)
+	return nil
+}
+
 func (b *singleBlockQuerier) Meta() block.Meta {
 	if b.meta == nil {
 		return block.Meta{}
--- a/pkg/phlaredb/block_querier.go
+++ b/pkg/phlaredb/block_querier.go
@@ -305,7 +305,7 @@ type singleBlockQuerier struct {
 type StacktraceDB interface {
 	Open(ctx context.Context) error
 	Close() error
-	Resolve(ctx context.Context, mapping uint64, locs locationsIdsByStacktraceID, stacktraceIDs []uint32) error
+	Resolve(ctx context.Context, partition uint64, locs symdb.StacktraceInserter, stacktraceIDs []uint32) error
 }
 
 type stacktraceResolverV1 struct {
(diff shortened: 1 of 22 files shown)


pkg/phlaredb/block_querier.go
```diff
@@ -321,18 +327,33 @@ func (r *stacktraceResolverV1) Close() error {
...
+	s.StacktracesTotal = int(r.stacktraces.file.NumRows())
+	s.MaxStacktraceID = s.StacktracesTotal
+}
+
+func (r *stacktraceResolverV1) Load(context.Context) error {
+	// FIXME(kolesnikovae): Loading all stacktraces from parquet file
+	//  into memory is likely a bad choice. Instead we could convert
+	//  it to symdb first.
+	return nil
+}
```
kolesnikovae, line 355: Compacting v1 stacktraces might be a bit challenging. It is easy if a source block is compacted entirely, when we read all of its profiles sequentially. However, in practice I guess we will need to filter based on time and series which changes the access pattern to random access.

@cyriltovena what do you think?

pkg/phlaredb/compact.go
```diff
@@ -534,3 +473,463 @@ func prepareIndexWriter(ctx context.Context, path string, readers []BlockReader)
...
+		r := s.rewriters[profile.blockReader]
+		if err = r.rewriteStacktraces(profile.row.StacktracePartitionID(), s.stacktraces); err != nil {
+			return
+		}
+		s.numSamples += uint64(len(values))
+		for i, v := range values {
+			// FIXME: the original order is not preserved, which will affect encoding.
+			values[i] = parquet.Int64Value(int64(s.stacktraces[i])).Level(v.RepetitionLevel(), v.DefinitionLevel(), v.Column())
+		}
+	})
```
kolesnikovae, line 527: When we translate stack trace IDs, the original order will be invalidated:

Before compaction:
```
IDs: 1, 2, 3, 4, 42, 43
```
After compaction:
```
IDs: 1, 2, 6, 4, 453, 12
```

It will deteriorate delta encoding compression ratio. We clearly should fix this

pkg/phlaredb/compact.go
```diff
@@ -534,3 +473,463 @@ func prepareIndexWriter(ctx context.Context, path string, readers []BlockReader)
...
+	return s.profiles.Err()
+}
+
+func (s *symbolsRewriter) Next() bool {
+	if !s.profiles.Next() {
+		return false
+	}
+	var err error
+	profile := s.profiles.At()
+	profile.row.ForStacktraceIDsValues(func(values []parquet.Value) {
```
kolesnikovae, line 516: We may also want to rewrite `Comments` field, which is basically a reference to strings

pkg/phlaredb/block_querier.go
cyriltovena: I suggest we don't compact v1.

Issue #2137, opened 2023-07-20
Investigate IteratorResult pool usage and performance


kolesnikovae:
I noticed rather suspicious code path that might affect the read path performance – I suspect that we use the `query.iteratorResultPoolGet` pool incorrectly:
![image](https://github.com/grafana/pyroscope/assets/12090599/85b26f72-f3a0-4f02-ad80-560f64d9de61)

kolesnikovae:
I guess it is caused by `defer` in the loop:
https://github.com/grafana/pyroscope/blob/e17b0d0164e0e0806c33a2c12de9b63f6b1e872a/pkg/phlaredb/query/iters.go#L457-L467

simonswine:
Good catch, yes that wouldn't work like I imagined it 🙁

PR #2138, opened 2023-07-20
Fix IteratorResult pool usage

kolesnikovae:
I'm not fully sure this solves the problem mentioned https://github.com/grafana/pyroscope/issues/2137 but should alleviate it at least


simonswine (approved):
LGTM 

Yes that should make IteratorResult reuse a lot more likely

PR #2139, opened 2023-07-20
Avoid BufferedRowReaderIterator buffer re-allocation

kolesnikovae:
The change should make use of BufferedRowReaderIterator buffer more efficient

### DIFF
--- a/pkg/parquet/row_reader.go
+++ b/pkg/parquet/row_reader.go
@@ -12,7 +12,7 @@ import (
 )
 
 const (
-	defaultRowBufferSize = 1024
+	defaultRowBufferSize = 64
 )
 
 var (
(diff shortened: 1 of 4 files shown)


cyriltovena (approved):
LGTM

pkg/parquet/row_reader.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 const (
-	defaultRowBufferSize = 1024
+	defaultRowBufferSize = 64
 )
```
kolesnikovae, line 16: This revealed one bug: when we read via `SliceRowReader`, the deconstruction function _appends_ values to the row, given that we don't reset it, profile rows with more than buff length values (1024) are getting malformed because they _include_ samples of another row. I think that we have not been experiencing it, because we have not  been using the buffer in practice, although I am not 100% sure

PR #2140, opened 2023-07-20
Fixes make generate

cyriltovena:


simonswine (approved):
Do we think there is any interoperability problem?

I am a bit unsure what is in those description bytes


Otherwise LGTM

cyriltovena:
No worried for those description we don't use reflection

PR #2142, opened 2023-07-20
Forward port updates from main to next

simonswine:
I don't plan to do this regularly, but it basically just takes the version updates @darrenjaneczek has been doing to main into next.


darrenjaneczek (approved):
Thank you.

PR #2143, opened 2023-07-20
fix: remove dependency on pyroscope-oss git branch

darrenjaneczek:
Now that they live in the same branch on this repository, remove the layer of cross repository dependency on the "og main" branch of "pyroscope-oss" and instead use the subpackages in the new `/og` subdirectory.


tsconfig.json
```diff
@@ -5,17 +5,7 @@
...
-      "@pyroscope/webapp/*": ["./node_modules/pyroscope-oss/webapp/*"],
-      "@webapp/redux/*": [
-        "./node_modules/pyroscope-oss/webapp/javascript/redux/*"
-      ],
-      "@pyroscope/flamegraph/*": [
-        "./node_modules/pyroscope-oss/packages/pyroscope-flamegraph/*"
-      ],
-      "@pyroscope/models/*": [
-        "./node_modules/pyroscope-oss/packages/pyroscope-models/*"
-      ],
```
eh-am, line 17: Note these refer to the `packages` dir, I think they still need to be aliased? Or does the `workspaces` bit in package.json handles that somehow?

eh-am (commented):
👍 

Don't forget to update [scripts/webpack/webpack.common.ts too!](https://github.com/grafana/pyroscope/blob/next/scripts/webpack/webpack.common.js) too!

tsconfig.json
darrenjaneczek: > Or does the workspaces bit in package.json handles that somehow?

Yes, that appears to be the case. But I am getting the sense that some of the aliases are clashing with the workspaces in certain circumstances. I will continue trying things.

>Don't forget to update [scripts/webpack/webpack.common.ts too!](https://github.com/grafana/pyroscope/blob/next/scripts/webpack/webpack.common.js) too!

Thanks for the note. However, my attempts to fix `webpack.common.ts` are meeting with failure locally.
I am thinking some old "node_modules" imports are still cached, so the current "passed tests" might be a misleading positive.

darrenjaneczek: I think I will get rid of the "workspace" entries, and just use aliases at this stage.

darrenjaneczek:
I will verify this a little more closely tomorrow.
This feels like a good initial step... assuming everything still works.

darrenjaneczek:
@eh-am, the frontend code that is now in "og" has a lot of elaborate testing which I don't suspect has been replicated for the new root of the project. What approach would you recommend for adapting it over to work with the new project root, and all of its overrides?

Rperry2174 (approved):
lgtm!

PR #2144, opened 2023-07-20
Move examples folder out of og

Rperry2174:
Moves examples out of og folder. There is potentially a conversation about moving them to a total repo altogether, but if we do that I think we should do it separately.


og/examples/java-jfr/rideshare/bin/main/org/example/rideshare/OrderService.class
petethepig, line 1: feels like these should stay

korniltsev: feels like class files should be deleted

Rperry2174: lol should they stay or should they go? I ended up closing this PR because I couldn't figure out why some files were getting deleted and most of the problem files were the .class files for some reason (I think maybe has to do with .class not being in .gitignore)

PR #2157, opened 2023-07-21
Adds push limits on profiles data

cyriltovena:
This adds 5 new limits.

- Total size of a profile uncompressed
- Maximum stacktrace depth/frames. Truncates them instead of erroring.
- Maximum amount of labels per sample (although we don't use them)
- Maximum size of a symbol, the symbol will be truncated.
- Maximum amount of samples per profile.

Closes https://github.com/grafana/pyroscope-squad/issues/16


pkg/validation/limits.go
```diff
@@ -81,6 +87,12 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.Var(&l.QuerySplitDuration, "querier.split-queries-by-interval", "Split queries by a time interval and execute in parallel. The value 0 disables splitting by time")
 
 	f.IntVar(&l.MaxQueryParallelism, "querier.max-query-parallelism", 0, "Maximum number of queries that will be scheduled in parallel by the frontend.")
+
+	f.IntVar(&l.MaxProfileSizeBytes, "validation.max-profile-size-bytes", 2*1024*1024, "Maximum size of a profile in bytes. This is based off the uncompressed size. 0 to disable.")
+	f.IntVar(&l.MaxProfileStacktraceSamples, "validation.max-profile-stacktrace-samples", 2000, "Maximum number of samples in a profile. 0 to disable.")
```
kolesnikovae, line 92: I wonder if this value is overly conservative. With the default 15s duration of a profiling session, and 100Hz of sample rate, we may expect that a single CPU core may emit up to 1500 of samples (edge case). Of course, most of them point to the same stack trace, therefore they are aggregated, but a 32 core machine may easily generate a profile containing a few thousand of samples.

Also, senders of cumulative profiles may always hit this limit. This is especially true for Go heap profiles.

Another important factor here is pprof labels: they may dramatically increase cardinality of samples. We should either drop them and dedup samples, or take them into account. I'd go for the former.

Should we set it to something in between 4K and 16K?

pkg/validation/validate.go
```diff
@@ -122,6 +127,47 @@ func ValidateLabels(limits LabelValidationLimits, userID string, ls []*typesv1.L
...
+		depthLimit        = limits.MaxProfileStacktraceDepth(userID)
+		labelsLimit       = limits.MaxProfileStacktraceSampleLabels(userID)
+		symbolLengthLimit = limits.MaxProfileSymbolValueLength(userID)
+	)
+	for _, s := range prof.Sample {
+
+		if depthLimit != 0 && len(s.LocationId) > depthLimit {
+			// truncate the deepest frames
+			s.LocationId = s.LocationId[:depthLimit]
+		}
```
kolesnikovae, line 158: It would be nice to have a stub like "other" – I think it could be a separate PR

kolesnikovae (approved):
LGTM! I'd maybe adjust some of the default values

pkg/validation/limits.go
cyriltovena: Yeah sounds good let me update that.

pkg/validation/validate.go
```diff
@@ -122,6 +127,47 @@ func ValidateLabels(limits LabelValidationLimits, userID string, ls []*typesv1.L
...
+	var (
+		depthLimit        = limits.MaxProfileStacktraceDepth(userID)
+		labelsLimit       = limits.MaxProfileStacktraceSampleLabels(userID)
+		symbolLengthLimit = limits.MaxProfileSymbolValueLength(userID)
+	)
+	for _, s := range prof.Sample {
+
+		if depthLimit != 0 && len(s.LocationId) > depthLimit {
+			// truncate the deepest frames
+			s.LocationId = s.LocationId[:depthLimit]
```
kolesnikovae, line 157: I just have realised that this truncates parent locations... This may make the flamegraph unreadable. I guess we could achieve better results with cutting the stack tip:
```go
s.LocationId = s.LocationId[len(s.LocationId)-depthLimit:]
```

cyriltovena: Yeah that wasn't intentional, let's fix it.

PR #2159, opened 2023-07-21
Profile proto serialization with slice reused

cyriltovena:
I realized we have quite a lot of alloc in marshalling proto specially a profile.

So I took a look and realize some slice are reused some are not when reseting (which is done when placing back to pool)

https://github.com/planetscale/vtprotobuf/issues/64 @joe-elliott is also questioning this behaviour.

We having also a discussion here https://github.com/planetscale/vtprotobuf/pull/35#issuecomment-1133384043

Result speak for themselves

```
❯ benchstat before.txt after.txt                                                                                           
name             old time/op    new time/op    delta
FromRawBytes-16    8.18ms ± 3%    6.14ms ± 6%  -24.98%  (p=0.000 n=10+10)

name             old alloc/op   new alloc/op   delta
FromRawBytes-16    6.50MB ± 2%    1.75MB ± 6%  -73.12%  (p=0.000 n=10+10)

name             old allocs/op  new allocs/op  delta
FromRawBytes-16     81.5k ± 0%     32.9k ± 2%  -59.69%  (p=0.000 n=10+10)
```


kolesnikovae:
@korniltsev implemented a very efficient pprof parser with arenas and pools once – I'd consider resurrecting it

cyriltovena:
Yeah that sounds good, the VT proto is a quick win though.

cyriltovena:
Switched to a fork with the fix.

```
❯ benchstat before.txt after.txt
name             old time/op    new time/op    delta
FromRawBytes-16    8.07ms ± 2%    4.76ms ± 6%  -41.03%  (p=0.000 n=10+10)

name             old alloc/op   new alloc/op   delta
FromRawBytes-16    6.38MB ± 1%    0.21MB ±13%  -96.73%  (p=0.000 n=10+10)

name             old allocs/op  new allocs/op  delta
FromRawBytes-16     81.5k ± 0%      3.0k ± 0%  -96.29%  (p=0.000 n=10+8)
````

kolesnikovae (approved):
LGTM!

PR #2161, opened 2023-07-23
Update datasources.yml

hamedprog:


simonswine (approved):
Good catch! Thanks for fixing this right away 🙂 

LGTM

Rperry2174:
Let me double check before merging. I think the reason we use a custom data source here has something to do with being able to view exemplars (traces + profiles linked together)

Rperry2174:
Yeah it turns out this was so that we could do profiling exmplars and show them in grafana:
image
 
relevant links:
- [image example](https://github.com/grafana/pyroscope/pull/1402#issuecomment-1215125294) 

But looks like the datasource doesn't work anymore anyway so I think this is fine

PR #2162, opened 2023-07-24
Ensure we only query ingesters in active state

simonswine:
With ring.Read we query ingesters in any state. This change is taken
from Mimir and it will ensure to ignore not active ingesters.

https://github.com/grafana/mimir/blob/d99cb3c081022e0752ee3db8c2cc39af46d09ba8/pkg/distributor/query.go#L33-L35

This should improve the behaviour of Pyroscope during rollouts and
partial outages.


kolesnikovae (approved):
LGTM

PR #2163, opened 2023-07-24
Push down label_name filter on Series API

simonswine:
This will filter label names on the ingesters directly.


pkg/querier/querier.go
```diff
@@ -229,11 +229,13 @@ func (q *Querier) Series(ctx context.Context, req *connect.Request[querierv1.Ser
...
 	responses, err := forAllIngesters(ctx, q.ingesterQuerier, func(childCtx context.Context, ic IngesterQueryClient) ([]*typesv1.Labels, error) {
 		res, err := ic.Series(childCtx, connect.NewRequest(&ingestv1.SeriesRequest{
-			Matchers: req.Msg.Matchers,
+			Matchers:   req.Msg.Matchers,
+			LabelNames: req.Msg.LabelNames,
 		}))
 		if err != nil {
 			return nil, err
 		}
+		// TODO: Remove this once we have the ingesters returning the filtered response, has been rolled out (f25).
```
simonswine, line 238: Once rolled out, we should remove this filtering

pkg/querier/querier.go
```diff
@@ -229,11 +229,13 @@ func (q *Querier) Series(ctx context.Context, req *connect.Request[querierv1.Ser
 
 	responses, err := forAllIngesters(ctx, q.ingesterQuerier, func(childCtx context.Context, ic IngesterQueryClient) ([]*typesv1.Labels, error) {
 		res, err := ic.Series(childCtx, connect.NewRequest(&ingestv1.SeriesRequest{
-			Matchers: req.Msg.Matchers,
+			Matchers:   req.Msg.Matchers,
+			LabelNames: req.Msg.LabelNames,
```
simonswine, line 233: This doesn't seem to make problems with older ingesters, that do not know about the LabelNames field

cyriltovena (approved):
LGTM

PR #2164, opened 2023-07-24
Update connect-go

simonswine:
This will make the connect-go Handlers/Client in JSON less strict and allow for easier adding of new fields.


cyriltovena (approved):
LGTM

PR #2166, opened 2023-07-25
Reset all items in ResetVT and do not discard in RemoveInPlace

cyriltovena:
We found a bug where `p.Normalized()` would mutate the profile and duplicate pointer in the same array, but also some values in case of a resliced were not reset.


simonswine (approved):
Nice work getting to the bottom of this @kolesnikovae @cyriltovena 


LGTM

PR #2168, opened 2023-07-25
Fix goroutine leak in BufferedIterator

cyriltovena:
This PR fix a goroutine on the query path that can trigger if we don't call Close correctly or leave the buffer full (no calling next).

It adds changes:
- Close all iterator in case of failure when building them
- Add cancellation when pushing to channel.


simonswine (approved):
LGTM

(Apart from the linting problem)

PR #2169, opened 2023-07-25
chore: integrate og/webapp into public/app (wip)

darrenjaneczek:


Rperry2174 (approved):
I've meticulously reviewed all 11,440 lines of code... (and by that I mean I ran it locally and tested everything I could find and didn't see any issues) 

LGTM🥇 🙏 


Documenting this just so we don't forget I'll create an issue as well
image

PR #2174, opened 2023-07-26
Move to the new parquet-go location

simonswine:
This is in preparation before forking for #2011 to be merged


kolesnikovae (approved):
LGTM

PR #2176, opened 2023-07-26
Fixes a compaction bug in series rewriting

cyriltovena:
This also adds a bunch of tests that were missing.

I still needs to add Compact Test with querying and deduping.


kolesnikovae (approved):
LGTM. Nice job!

PR #2178, opened 2023-07-27
Fixes compaction incorrect profile seriesRef column

cyriltovena:
This adds even more tests and fixes 2 bugs surfaced with them:
- We were writing twice the last series.
- We were not correctly setting the column seriesIndex in the profile parquet file.


kolesnikovae (approved):
LGTM

PR #2184, opened 2023-07-28
Correctly setup compaction sources in block meta files

cyriltovena:
The original block should have all source IDS which is itself, then compacted should merge source of parents.

Also adds test for both case.

### DIFF
--- a/pkg/phlaredb/head.go
+++ b/pkg/phlaredb/head.go
@@ -16,6 +16,7 @@ import (
 	"github.com/gogo/status"
 	"github.com/google/pprof/profile"
 	"github.com/google/uuid"
+	"github.com/oklog/ulid"
 	"github.com/opentracing/opentracing-go"
 	otlog "github.com/opentracing/opentracing-go/log"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


pkg/phlaredb/head.go
```diff
@@ -1003,6 +1004,7 @@ func (h *Head) flush(ctx context.Context) error {
 	h.meta.Files = files
 	h.meta.Stats.NumProfiles = uint64(h.profiles.index.totalProfiles.Load())
 	h.meta.Stats.NumSamples = h.totalSamples.Load()
+	h.meta.Compaction.Sources = []ulid.ULID{h.meta.ULID}
```
cyriltovena, line 1007: Original block should have their own sources this is how the Prometheus format expect it.

cyriltovena: cc @pracucci if we don't have local compaction that is correct right ?

cyriltovena: Confirmed this with Marco and reading more Prometheus.

kolesnikovae (approved):
LGTM

pkg/phlaredb/head.go
simonswine, line 1007: When prometheus does this it also sets compaction level to `1`

simonswine: https://github.com/prometheus/prometheus/blob/76dd9b5470e6f49330a6420c5d6133bea40a71ea/tsdb/compact.go#L499C1-L500

PR #2185, opened 2023-07-28
Adds a clone function for meta file

cyriltovena:
This is going to be used for creating meta of splitted sharded blocks.


simonswine (approved):
LGTM

PR #2187, opened 2023-07-28
fix(example): rideshare java build

korniltsev:


Rperry2174 (approved):
lgtm! thanks for the quick fix

PR #2190, opened 2023-08-01
chore: port change (4100 -> 4040)

petethepig:
This was originally many things put into one PR. I split other work into separate PRs:
* https://github.com/grafana/pyroscope/pull/2234
* https://github.com/grafana/pyroscope/pull/2235

Right now it's just port change PR. It does include helm changes.


simonswine:
I just gave that a try using `make build run 'PARAMS=--config.file ./cmd/pyroscope/pyroscope.yaml --querier.max-query-length=30d'`


When I go via the browser to `http://localhost:4040`, a few things are not working

- Memberlist and Overrides-Exporter on `/admin`
- Sometimes `/admin/` shows the normal UI instead of the admin interface
- Can't query anything through the normal UI (curl seems to be fine) 
   image


Not looked in detail through the other changes but in general looking good to me

simonswine:
This port should probably also changed:

https://github.com/grafana/pyroscope/blob/a3d53793154f89afa0e0acb7073363d494858ba2/operations/phlare/helm/phlare/values.yaml#L67

And in turn it needs to be set back by deployment tools

petethepig:
@simonswine RE helm changes — I intentionally left those ports untouched for now, because CI tries to use old helm chart with new code and config. I think we should change helm stuff in a separate PR.

petethepig:
I moved things that are not port related to separate PRs

simonswine (approved):
Sorry just catching up,  LGTM worked locally.

I did not test upgrade behaviour though

PR #2191, opened 2023-08-01
Use app auth for release/weekly-release/deploy

simonswine:
This will avoid running into the rate limiting, which we hit on the bot account


petethepig (approved):
LGTM

Issue #2147, opened 2023-07-21
Incorrect timeline boundaries


kolesnikovae:
I captured this video at 14:00 local time – note how the timeline end changes from 14:05 to 14:15 for different profile types, and the data (ingestion is ongoing in the real time) is shown at wrong position.

Works as expected in Explore view. I suspect that the backend (`/render` and maybe other endpoints) responds with malformed timeline.

https://github.com/grafana/pyroscope/assets/12090599/5b70818f-31d4-4257-9822-fd27a544e2e0

simonswine:
I have also noticed this at some point.

korniltsev:
I think there's off by one at least here, and it should be `i = i + int(backfillNum)  + 1`, otherwise it overwrites correct values.
https://github.com/grafana/pyroscope/blob/2f3a1e14b6c7e57eff81a5426a945c945b5b7d59/pkg/querier/timeline/timeline.go#L48


Here's an ugly testcase to reproduce https://gist.github.com/korniltsev/2cda5e4c8eda8ddad573a9bc6eb4f2f1

And maybe there are more issues, I've stopped looking into it, since @Rperry2174 said @bryanhuhta is already looking into it

PR #2200, opened 2023-08-02
Fix timeline backfills

bryanhuhta:
closes https://github.com/grafana/pyroscope/issues/2147

This PR fixes the timeline bug where some timelines would have lots of 0 values towards the end. It also fixes some slice copy bugs, where slices would be copied incorrectly back on to other slices, adding erroneous values.

This also has fixes for small alignment issues in the timeline values, though these are unlikely to ever be noticed as it previously would only miss at most 2 padded values (one at the beginning and one at the end).

### DIFF
--- a/pkg/querier/timeline/timeline.go
+++ b/pkg/querier/timeline/timeline.go
@@ -18,56 +18,57 @@ func New(series *v1.Series, startMs int64, endMs int64, durationDeltaSec int64) 
 	points := series.GetPoints()
 	res := make([]uint64, len(points))
 
+	timeline := &flamebearer.FlamebearerTimelineV1{
+		StartTime:     startSec,
+		DurationDelta: durationDeltaSec,
+		Samples:       []uint64{},
+	}
+
 	if len(points) < 1 {
-		return &flamebearer.FlamebearerTimelineV1{
-			StartTime:     startSec,
-			DurationDelta: durationDeltaSec,
-			Samples:       backfill(startMs, endMs, durationDeltaSec),
+		if n := sizeToBackfill(startMs, endMs, durationDeltaSec); n > 0 {
+			timeline.Samples = make([]uint64, n)
 		}
+		return timeline
 	}
 
 	i := 0
 	prev := points[0]
 	for _, curr := range points {
+		res[i] = uint64(curr.Value)
+
 		backfillNum := sizeToBackfill(prev.Timestamp, curr.Timestamp, durationDeltaSec)
-
 		if backfillNum > 0 {
-			// backfill + newValue
-			bf := append(backfill(prev.Timestamp, curr.Timestamp, durationDeltaSec), uint64(curr.Value))
+			// Subtract 1 to account for the current value already being added
+			// to the result slice.
+			backfillNum--
 
-			// break the slice
-			first := res[:i]
-			second := res[i:]
+			// Insert backfill.
+			bf := make([]uint64, backfillNum)
+			res = append(res[:i], append(bf, res[i:]...)...)
 
-			// add new backfilled items
-			first = append(first, bf...)
-
-			// concatenate the three slices to form the new slice
-			res = append(first, second...)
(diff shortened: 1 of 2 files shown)


pkg/querier/timeline/timeline.go
```diff
@@ -18,56 +18,57 @@ func New(series *v1.Series, startMs int64, endMs int64, durationDeltaSec int64)
 	points := series.GetPoints()
 	res := make([]uint64, len(points))
```
korniltsev, line 19: nit: I wonder if we can pre allocate the whole buffer here with zero length and `sizeToBackfill(startMs, endMs, durationDeltaSec)` capacity

And then down the line instead of alloc + append backfill buffer, just extend the length of the res. 

This is probably premature optimization. feel free to ignore the comment.

pkg/querier/timeline/timeline_test.go
```diff
@@ -82,12 +118,57 @@ func Test_Backfill_All(t *testing.T) {
...
+		69, // 2000 ms
+		83, // 3000 ms
+		0,  // 4000 ms
+		0,  // 5000 ms
+		85, // 6000 ms
+		0,  // 7000 ms
+		91, // 8000 ms
+		0,  // 9000 ms
+	}, tl.Samples)
+}
```
korniltsev, line 174: It looks like we don't add a trailing zero for a sample at endMs

```go
func Test_Backfill_LastSample(t *testing.T) {
	startMs := int64(3000)
	step := int64(10)
	series := &typesv1.Series{
		Points: []*typesv1.Point{
			{Timestamp: 23000, Value: 69},
			{Timestamp: 53000, Value: 91},
		},
	}

	tl := timeline.New(series, startMs, int64(53000), step)

	assert.Equal(t, startMs/1000, tl.StartTime)
	assert.Equal(t, []uint64{
		0, 0, 69, 0, 0, 91,
	}, tl.Samples)

	tl = timeline.New(series, startMs, int64(63000), step)
	assert.Equal(t, startMs/1000, tl.StartTime)
	assert.Equal(t, []uint64{
		0, 0, 69, 0, 0, 91, 0,
	}, tl.Samples)
}

```

korniltsev (approved):
Nice!

simonswine (approved):
LGTM, nice work, good to see a bit more test coverage there 🙂

pkg/querier/timeline/timeline_test.go
bryanhuhta: Is a valid input for `timeline.New` allowed to be equal to `endMs`? The comment for the function seems to indicate this shouldn't be allowed:

> New generates a FlamebearerTimeline,
> backfilling any missing data with zeros
> It assumes:
> * Ordered
> * startMs is earlier than the first series value
> * **endMs is after the last series value**

Though, admittedly the assumptions outlined there seem to suggest `startMs` needs to be strictly before the first series value. I was operating under the assumption series should have timestamps in the range `[startMs, endMs)`.

If this is true, then the 53000ms test case has an invalid series range since it has a timestamp that's not in the range (91). That value has a timestamp equal to `endMs`. While the 63000ms test case has a valid series range, but we shouldn't expect the final zero, since that would correspond to a timestamp outside of `[startMs, endMs)`.

At the end of the day, the logic for how we pad either side of the timeline will depend on what is considered "valid input". If the valid series timestamp range is `[startMs, endMs)`, we'll have the current behavior. If the valid series timestamp range is `[startMs, endMs]`, then we'll have the behavior of this test case above.

This brings up another point, if we are given a series with out of bounds timestamps, should we truncate it to fit within `[startMs, endMs)`?

korniltsev: I dont know answers for these questions. I just felt it may be a good idea to keep both cases consistent, either add a zero or drop value. We can merge as is and improve later.

petethepig (approved):
LGTM, Great job!

PR #2201, opened 2023-08-03
feat(ebpf): add pyperf

korniltsev:
This PR adds a python profiling to ebpf module
It is based on https://github.com/iovisor/bcc/blob/master/examples/cpp/pyperf/PyPerfBPFProgram.cc , rewriten without bcc, added support for other python versions and musl libc 

Future improvements:
- ~fix fork/execve race - if process forks we may observe it before exec, therefore selecting wrong profiling type, for example python before execve /bin/sh or the other way around~
- fix class name retrieval for a strange case 3.9 see NullCls in pyperf.bpf.c
- support for python 3.12
- support for python  2.7
-  CI step to verify if the PR has correctly regenerated the files
- e2e tests (maybe using cmd/playground)


korniltsev:
it is 3.5 to 3.11

ebpf/bpf/pthread.bpf.h
```diff
@@ -0,0 +1,80 @@
+//
+// Created by korniltsev on 31/7/2566.
```
cyriltovena, line 2: Back from the future !

ebpf/bpf/pyperf.bpf.c
```diff
@@ -0,0 +1,548 @@
...
+        py_symbol *symbol,
+        // ctx is only used to call helper to clear symbol, see documentation below
+        void *ctx) {
+    void *code_ptr;
+    void *cur_frame = *frame_ptr;
+    if (!cur_frame) {
+        return 0;
+    }
+
+    //        py311
```
cyriltovena, line 434: Should that be commented ? may be add why.

ebpf/cmd/playground/main.go
```diff
@@ -3,57 +3,153 @@
...
 	"github.com/grafana/pyroscope/ebpf/sd"
 	"github.com/grafana/pyroscope/ebpf/symtab"
 	"github.com/grafana/pyroscope/ebpf/symtab/elf"
 	"github.com/prometheus/client_golang/prometheus"
+	commonconfig "github.com/prometheus/common/config"
 )
 
+const sampleRate = 99 // times per second
+
+// todo remove this cmd or move to a separate go module or out of repository, because it adds a bunch of dependencies
```
cyriltovena, line 31: I'd love a CLI to activate the profiler locally.

ebpf/python/procinfo_test.go
```diff
@@ -0,0 +1,256 @@
...
+	info, err = GetProcInfo(bufio.NewScanner(bytes.NewReader([]byte(maps))))
+	require.NoError(t, err)
+	require.Nil(t, info.Musl)
+	require.Equal(t, info.Version, Version{3, 7, 0})
+	require.NotNil(t, info.PythonMaps)
+	require.Nil(t, info.LibPythonMaps)
+}
+
+//func TestMusl(t *testing.T) {
+//	fs := []string{
```
cyriltovena, line 132: Seems like good tests to have.

ebpf/python/pyperf.go
```diff
@@ -0,0 +1,279 @@
...
+	go func() {
+		defer res.wg.Done()
+		res.loop()
+	}()
+	return res, nil
+}
+
+// AddPythonPID return python init error, second return value is true if error occured and process is stil alive,
+// third return value - python process data
+func (s *Perf) AddPythonPID(pid uint32, serviceName string) (error, bool, *PerfPyPidData) {
```
cyriltovena, line 63: err should be the last returned element not the first. It's also not clear where you add the PID.

ebpf/python/pyperf.go
```diff
@@ -0,0 +1,279 @@
...
+			}
+			_ = level.Error(s.logger).Log("msg", "[pyperf] reading from perf event reader", "err", err)
+			continue
+		}
+
+		if record.LostSamples != 0 {
+			s.metrics.LostSamples.Add(float64(record.LostSamples))
+			_ = level.Debug(s.logger).Log("msg", "[pyperf] perf event ring buffer full, dropped samples", "n", record.LostSamples)
+		}
+		//_ = level.Debug(s.logger).Log("msg", "[pyperf] perf event sample", "n", len(record.RawSample))
```
cyriltovena, line 103:
```suggestion
```

ebpf/python/pyperf.go
```diff
@@ -0,0 +1,279 @@
...
+	copy(eventsCopy, s.events)
+	for i := range s.events {
+		s.events[i] = nil
+	}
+	s.events = s.events[:0]
+
+	return eventsCopy
+}
+
+func (s *Perf) GetSymbolsLazy() LazySymbols {
```
cyriltovena, line 140:
```suggestion
func (s *Perf) GetLazySymbols() LazySymbols {
```

ebpf/cmd/playground/main.go
korniltsev: this main file is exactly that.
The comment is about that this cli is using more dependencies that is required (for example grpc)
Im going to just remove this comment until we have someone complain

ebpf/bpf/pyperf.bpf.c
korniltsev: removed

ebpf/bpf/pthread.bpf.h
korniltsev: :D

ebpf/python/procinfo_test.go
korniltsev: added

ebpf/cmd/playground/main.go
```diff
@@ -3,75 +3,327 @@
...
 	ebpfspy "github.com/grafana/pyroscope/ebpf"
+	"github.com/grafana/pyroscope/ebpf/pprof"
 	"github.com/grafana/pyroscope/ebpf/sd"
 	"github.com/grafana/pyroscope/ebpf/symtab"
 	"github.com/grafana/pyroscope/ebpf/symtab/elf"
 	"github.com/prometheus/client_golang/prometheus"
+	commonconfig "github.com/prometheus/common/config"
+)
+
+var configFile = flag.String("config", "", "config file path")
```
cyriltovena, line 36: super cool do you have an config file example ?

cyriltovena: Nevermind I see the default !

ebpf/cmd/python_dwarfdump/main.go
```diff
@@ -0,0 +1,399 @@
+// based on https://sourcegraph.com/github.com/grafana/ebpf-autoinstrument/-/blob/pkg/goexec/structmembers.go
```
cyriltovena, line 1: getting a 404 on this one.

Can you add a readme for what this program is for ?

ebpf/cmd/playground/main.go
korniltsev: I've used locally like this. This is only for "testing/prototyping" locally.  
```
{
  "RelabelConfig": [
    {
      "SourceLabels": [
        "__meta_process_exe"
      ],
      "Regex": ".*python.*",
      "Action": "keep"
    },
    {
      "SourceLabels": [
        "__meta_process_cwd"
      ],
      "TargetLabel": "service_name",
      "Regex": ".*pythonProjectAsync.*",
      "Replacement": "python/ProjectAsync",
      "Action": "replace"
    },
    {
      "SourceLabels": [
        "__meta_process_cwd"
      ],
      "TargetLabel": "service_name",
      "Regex": ".*examples/python/rideshare/flask.*",
      "Replacement": "python/flask/rideshare",
      "Action": "replace"
    },
    {
      "SourceLabels": [
        "service_name"
      ],
      "Regex": "python/.*",
      "Action": "keep"
    }
  ]
}
```

We could create something better if we ever decide to put it into profilecli

ebpf/cmd/python_dwarfdump/main.go
korniltsev: https://github.com/grafana/beyla/blob/6b46732da73f2f2cb84e41efdc74789509a7fa2b/pkg/internal/goexec/structmembers.go

fixed the link

ebpf/Makefile
```diff
@@ -0,0 +1,12 @@
+
+.phony: python/dwarfdump
+python/dwarfdump:
```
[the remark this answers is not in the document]
korniltsev: agree, lets do it in a separate pr

ebpf/python/pyperf.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+	res.wg.Add(1)
+	go func() {
+		defer res.wg.Done()
+		res.loop()
+	}()
+	return res, nil
+}
+
+// StartPythonProfiling return python init error, second return value is true if process is still alive,
```
cyriltovena, line 61: The comment is outdated.

ebpf/python/pyperf.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+}
+
+func (s *Perf) Close() {
+	_ = s.rd.Close()
+	s.wg.Wait()
+}
+
+func (s *Perf) ResetEvents() []*PerfPyEvent {
```
cyriltovena, line 114:
```suggestion
func (s *Perf) CollectEvents() []*PerfPyEvent {
```

ebpf/python/pyperf.go
```diff
@@ -0,0 +1,264 @@
...
+	if errors.Is(err, ebpf.ErrKeyNotExist) {
+		return nil, nil
+	}
+	// batch not supported
+
+	// try iterating if batch failed
+	it := m.Iterate()
+
+	v := uint32(0)
+	for {
```
cyriltovena, line 172: why not `for it.Next(k, &v) {` ?

ebpf/python/pyperf.go
```diff
@@ -0,0 +1,264 @@
...
+
+func (s *Perf) RemoveDeadPID(pid uint32) {
+	s.pidCache.Remove(pid)
+	err := s.pidDataHashMap.Delete(pid)
+	if err != nil && !errors.Is(err, ebpf.ErrKeyNotExist) {
+		_ = level.Error(s.logger).Log("msg", "[pyperf] deleting pid data hash map", "err", err)
+	}
+}
+
+func ReadPyEvent(raw []byte) (*PerfPyEvent, error) {
```
cyriltovena, line 201: I would use the pattern as json here

``
 ReadPyEvent(data []byte, v *PerfPyEvent) error
```

This will be easier for the caller to reuse the object.

File: ebpf/session.go [hunk 15]
```diff
@@ -357,45 +499,250 @@ func (s *session) walkStack(sb *stackBuilder, stack []byte, pid uint32, stats *s
 	}
 }
 
-func (s *session) updateSampleRate(sampleRate int) error {
-	if s.options.SampleRate == sampleRate {
-		return nil
+func (s *session) readEvents(events *perf.Reader,
```
cyriltovena, line 502: There's no test for session.go and I think it would be great to think about it since you have many locks and goroutine, this means most likely add a interface for the perf package used.

Just to make sure it calls correctly perf and doesn't leak and has a race.

cyriltovena (approved):
LGTM

I think we should have more test for the session.go file.

ebpf/session.go
korniltsev: totally agree, I want this a lot, however mocking out the whole cilium/ebpf library is not trivial, there's a lot of structures used, like maps, programs, perf reader.  Lets create an issue and improve later

korniltsev: https://github.com/grafana/pyroscope/issues/2576

ebpf/python/pyperf.go
korniltsev: applied the function prototype change, not reusing the event just yet

ebpf/python/pyperf.go
korniltsev: removed it

ebpf/python/pyperf.go
```diff
@@ -0,0 +1,264 @@
...
+	s.wg.Wait()
+}
+
+func (s *Perf) ResetEvents() []*PerfPyEvent {
+	s.eventsLock.Lock()
+	defer s.eventsLock.Unlock()
+	if len(s.events) == 0 {
+		return nil
+	}
+	eventsCopy := make([]*PerfPyEvent, len(s.events))
```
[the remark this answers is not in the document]
korniltsev: the reason I chose to use a pointer is because PerfPyEvent is quite big - 320 bytes
and there could be quite a lot of them 100 times per second * 15 seconds * 16 cores * 320 bytes =~7M
And i did not want to move this memory twice. It is likely a premature optimization already. At the same time I dont see this function hot, so I suggest we keep it as is and optimize if we have a problem or any evidence it can be optimized

PR #2202, opened 2023-08-03
fix: render flamebearer sample rate

korniltsev:


pkg/model/flamegraph.go
```diff
@@ -128,7 +128,7 @@ func ExportToFlamebearer(fg *querierv1.FlameGraph, profileType *typesv1.ProfileT
 		unit = metadata.ObjectsUnits
 	case "cpu":
 		unit = metadata.SamplesUnits
-		sampleRate = uint32(100000000)
+		sampleRate = uint32(1000000000)
```
simonswine, line 131: This might make it easier to read:
```suggestion
		sampleRate = uint32(1_000_000_000)
```

simonswine (approved):
LGTM

PR #2206, opened 2023-08-03
refactor: reduce allocs when building timeline

bryanhuhta:
In https://github.com/grafana/pyroscope/pull/2200 ([comment](https://github.com/grafana/pyroscope/pull/2200#discussion_r1282708713)), it was noted that there could be a performance improvement when allocating slices. This PR changes how slices are allocated such that only one allocation is necessary at the beginning of the function.

Here's a quick benchmark that generally describes the performance change compared to the old method:

```
$ go test -bench=. -run=xxx -benchtime=5s -benchmem                                      
goos: darwin
goarch: arm64
pkg: github.com/grafana/pyroscope/pkg/querier/timeline
Benchmark_Backfill/New-12               5614            961086 ns/op         3942035 B/op          6 allocs/op
Benchmark_Backfill/Old-12                  1       72707588750 ns/op    953187418008 B/op     500043 allocs/op
PASS
ok      github.com/grafana/pyroscope/pkg/querier/timeline       79.087s
```

The benchmark mapped 3 different sizes of series data:
- short (~100 points)
- medium (~10,000 points)
- long (~100,000 points)

This benchmark probably doesn't precisely mimic a production workload, but in the synthetic workload there is massive improvements with the new approach.

### DIFF
--- a/pkg/querier/timeline/timeline.go
+++ b/pkg/querier/timeline/timeline.go
@@ -16,7 +16,7 @@ func New(series *v1.Series, startMs int64, endMs int64, durationDeltaSec int64) 
 	// ms to seconds
 	startSec := startMs / 1000
 	points := series.GetPoints()
-	res := make([]uint64, 0, sizeToBackfill(startMs, endMs, durationDeltaSec))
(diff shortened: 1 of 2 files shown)


kolesnikovae (approved):
LGTM. Nice improvement!

PR #2207, opened 2023-08-04
Upgrade dskit to remove direct dependency on weaveworks/common

charleskorn:
This PR upgrades dskit and replaces use of packages from weaveworks/common with their migrated equivalents in dskit. See https://github.com/grafana/dskit/pull/342 for more details.

Note that there is still an indirect dependency on weaveworks/common through the use of `github.com/grafana/mimir/pkg/storegateway` in `pkg/querier`. This will be resolved by upgrading the dependency on Mimir once https://github.com/grafana/mimir/pull/5672 is merged.

If this PR needs to be rebuilt, I used `rewrite.sh` ([source](https://gist.github.com/charleskorn/48efe62a09d6d70f3de30327003df5c5#file-rewrite-sh)) to generate most of these changes.


simonswine (approved):
Thanks for updating. LGTM

PR #2217, opened 2023-08-07
Skip image validation in doc-validator

jdbaldry:
The check is specific to storing assets in the repository as part of Hugo leaf bundles. Current advice is to instead store assets in the website asset uploader service following the instructions in https://grafana.com/docs/writers-toolkit/writing-guide/image-guidelines/#where-to-store-media-assets. This makes the check raise false errors.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -17,4 +17,4 @@ test:
 
 .PHONY: validate
 validate:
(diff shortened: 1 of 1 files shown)


Rperry2174 (approved):
lgtm. Thanks!

PR #2240, opened 2023-08-11
Disable pools for locations, functions, and lines

kolesnikovae:
Reusing tiny pprof protobuf objects is more expensive than allocating new: `(*Profile).ReturnToVTPool` consumes 23% of total CPU time. It makes sense to revisit the decision to use pools for pprof proto.

image


korniltsev (approved):
LGTM 
is markdown change relevant to the pb gen?

kolesnikovae:
> is markdown change relevant to the pb gen?

I think it comes from the dependencies

PR #2243, opened 2023-08-11
Bump golangci-lint to 1.54.0

hi-rustin:
Bumped the golangci-lint to 1.54.0. The main change is we need to upgrade the depguard from v1 to v2. You can see more at: https://github.com/golangci/golangci-lint/pull/3795/files


simonswine (approved):
LGTM

PR #2244, opened 2023-08-14
fix(docs): remove 'native' from python docs

korniltsev:


docs/sources/configure-client/language-sdks/python.md
```diff
@@ -42,7 +42,6 @@ pyroscope.configure(
   sample_rate         = 100, # default is 100
```
korniltsev, line 42: @Rperry2174 should we remove auth_token from this example to not confuse users?
auth_token is only needed for og pyroscope, seems like we should not advertise it.

Rperry2174 (approved):
Lgtm yeah let's remove auth token too if that's the one for og pyroscope can we replace it with new authentication?

PR #2246, opened 2023-08-14
fix(ebpf): disable gotable fallback by default, make gotablefallback smarter

korniltsev:
for golang binaries we create symbol table from `.gopclntab` section. it is useful for striped binaries.
It may be usefull to still look into `.sym`  sections for non go functions ie cgo . Thus there's a GoTableWithFallback. 
However for unstripped binaries it now keeps in memory duplicate symbols for go functions  (both from   `.gopclntab` and  `.sym` )

This PR:
- disable gotable fallback by default
- when  gotable fallback is enabled, do not duplicate symbols between gotable and `.sym` table

### DIFF
--- a/ebpf/symtab/elf.go
+++ b/ebpf/symtab/elf.go
@@ -165,7 +165,7 @@ func (et *ElfTable) createSymbolTable(me *elf2.MMapedElfFile) (SymbolNameResolve
 	}
 	symbolOptions := elf2.SymbolsOptions{}
 	if goErr == nil && goTable.Index.Entry.Length() > 0 {
-		symbolOptions.FilterTo = goTable.Index.Entry.Get(0)
+		symbolOptions.FilterFrom = goTable.Index.Entry.Get(0)
 		symbolOptions.FilterTo = goTable.Index.End
 	}
 	symTable, symErr := me.NewSymbolTable(&symbolOptions)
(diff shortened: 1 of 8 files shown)


ebpf/symtab/elf.go
```diff
@@ -148,8 +159,16 @@ func (et *ElfTable) load() {
...
 func (et *ElfTable) createSymbolTable(me *elf2.MMapedElfFile) (SymbolNameResolver, error) {
-	symTable, symErr := me.NewSymbolTable()
 	goTable, goErr := me.NewGoTable()
+	if !et.options.SymbolOptions.GoTableFallback && goErr == nil {
+		return goTable, nil
+	}
+	symbolOptions := elf2.SymbolsOptions{}
+	if goErr == nil && goTable.Index.Entry.Length() > 0 {
+		symbolOptions.FilterTo = goTable.Index.Entry.Get(0)
+		symbolOptions.FilterTo = goTable.Index.End
```
kolesnikovae, line 169: There must be a typo:
```go
symbolOptions.FilterFrom = goTable.Index.Entry.Get(0)
```

korniltsev: yes!
I guess my test is broken as well :(

korniltsev: actually I dont know how to easily catch it with a test. the test works because it filters `[0;last_go_symbol]`
And in order to catch this typo I need to have a symbol in `.sym` in range `[0;first_go_symbols]` which I dont know how to do without messing with a linker.

I will just fix the typo and not improve test

Thanks!

kolesnikovae (approved):

Issue #2045, opened 2023-05-31
Figure how to run the store-gateway when using single-binary


cyriltovena:
Right now the store gateway is not enable when using single binary, but we should definitively have it if you run using an cloud object storage.

simonswine:
Just to add a bit more context: For different components we use the [dskit/modules](https://pkg.go.dev/github.com/grafana/dskit@v0.0.0-20230808012443-3f5683fdb201/modules) package, you can find the original[ design for Cortex ](https://cortexmetrics.io/docs/proposals/generalize-modules/)(proper ancient).

The role of the store-gateway is very similar how it is explained in Grafana Mimir https://grafana.com/docs/mimir/latest/references/architecture/components/store-gateway/

hi-rustin:
I am working on this.

hi-rustin:
Oh, I guess the failure was caused by my env.

After I changed the config to:
```yaml
# Do not use this configuration in production.
# It is for demonstration purposes only.
scrape_configs:
  - job_name: "default"
    scrape_interval: "15s"
    static_configs:
      - targets: ["127.0.0.1:4100"]
        labels:
          service_name: "service-demo"

# storage:
#   backend: s3
#   s3:
#     endpoint: 127.0.0.1:9000
#     bucket_name: test
#     insecure: true
#     access_key_id: minioadmin
#     secret_access_key: minioadmin


storage:
  backend: filesystem
  filesystem:
    dir: /tmp/pyroscope

```

Then it worked well.

simonswine:
Looking at the potential combinations and what I would expect:

* storage.backend not configured and `target=all` (single binary): 
  * As right now we won't initialize the store-gatway and shipper.
* storage.backend configured and `target=all` (single binary):
  * It is valid to use all type of storage.backend 
  * Start the storage gateway and the shipper.
* storage.backend configured and `target!=all` (micro services):
  * It is valid to use all type of storage.backend (e.g. filesystem could be a NFS mount, shared by all components) 
  * Start the storage gateway and the shipper.
* storage.backend configured and `target!=all` (micro services):
  * Not valid we should error out, as the microservices need to have a shared storage between them

kolesnikovae:
> * storage.backend not configured and target=all (single binary):
>   * As right now we won't initialize the store-gatway and shipper.
> * storage.backend configured and target!=all (micro services):
>   * Not valid we should error out, as the microservices need to have a shared storage between them

[...]

Issue #2029, opened 2023-07-12
Blocks iteration by prefix on filesystem


cyriltovena:
see https://github.com/grafana/phlare/pull/849

Currently the `IterBlockMetas` doesn't work on filesystem because we use prefix listing and that is not supported on filesystem client.

PR #2247, opened 2023-08-14
Run the store gateway if using single binary

hi-rustin:
close https://github.com/grafana/pyroscope/issues/2045

Enable store gateway when using single binary.

### DIFF
--- a/pkg/phlare/modules.go
+++ b/pkg/phlare/modules.go
@@ -329,21 +329,15 @@ func (f *Phlare) initRing() (_ services.Service, err error) {
 
 func (f *Phlare) initStorage() (_ services.Service, err error) {
 	objectStoreTypeStats.Set(f.Cfg.Storage.Bucket.Backend)
-	if cfg := f.Cfg.Storage.Bucket; cfg.Backend != "filesystem" {
-		b, err := objstoreclient.NewBucket(
-			f.context(),
-			cfg,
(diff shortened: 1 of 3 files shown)


simonswine:
@hi-rustin thanks for your work on this, I think it is a bit more complicated depending if storage is configured or single-binary vs. micro-service mode.

https://github.com/grafana/pyroscope/issues/2045#issuecomment-1682242265

Let me know if that makes sense to you or if you think otherwise.

pkg/phlare/modules.go
```diff
@@ -329,21 +329,15 @@ func (f *Phlare) initRing() (_ services.Service, err error) {
 
 func (f *Phlare) initStorage() (_ services.Service, err error) {
 	objectStoreTypeStats.Set(f.Cfg.Storage.Bucket.Backend)
-	if cfg := f.Cfg.Storage.Bucket; cfg.Backend != "filesystem" {
```
cyriltovena, line 332: I think we should keep this for now as this won't work because of https://github.com/grafana/pyroscope/issues/2029

cyriltovena: I'll check if I can resolve it.

cyriltovena:
https://github.com/grafana/pyroscope/pull/2318

Thanks create another PR for now that will support everything but filesystem storage.

PR #2248, opened 2023-08-14
fix(ebpf): do not clear ElfCache

korniltsev:
fix bug: ElfCache.Cleanup should clear the whole cache, only Close all files, and remove entries witch did not update for couple of rounds .

In this PR:
- fix the bug
- improve debug info a bit - add pointers to the expensive objects, add last_updated_round bits.

### DIFF
--- a/ebpf/symtab/elf_cache.go
+++ b/ebpf/symtab/elf_cache.go
@@ -75,12 +75,6 @@ func (e *ElfCache) NextRound() {
 func (e *ElfCache) Cleanup() {
 	e.BuildIDCache.Cleanup()
 	e.SameFileCache.Cleanup()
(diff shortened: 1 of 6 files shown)


ebpf/symtab/elf_cache.go
```diff
@@ -75,12 +75,6 @@ func (e *ElfCache) NextRound() {
 func (e *ElfCache) Cleanup() {
 	e.BuildIDCache.Cleanup()
 	e.SameFileCache.Cleanup()
-	e.BuildIDCache.Each(func(k elf.BuildID, v SymbolNameResolver) {
```
korniltsev, line 78: The bug was here.

kolesnikovae (approved):
LGTM

PR #2250, opened 2023-08-14
Ensure local micro-services still works after rename

simonswine:
This was affected by the helm charts renaming


Rperry2174 (approved):
lgtm cc @petethepig

PR #2252, opened 2023-08-15
Update dotnet docs

Rperry2174:
Updates dotnet docs


korniltsev (changes requested):
we say here to replace <User> <Password> and <TenantID> placeholders, however the placeholders are only used later in the separate section in a separate example

It feels like in the table it should not suggest to replace anyhting , especially in further examples and just state. "this is http basic auth. you need it if you send profiles to atuhenticated pyroscope server, for example grafana cloud"

korniltsev (approved):

PR #2255, opened 2023-08-15
fix(ebpf): improve pcindex binary search speed

korniltsev:
PCIndex.FindIndex is not super hot, it is only 3%. However the change is not hard so I do it anyway. "It ain't much, but it's honest work"

In this PR I've inlined sort.Search in PCIndex.FindIndex.

```
goos: linux
goarch: amd64
pkg: github.com/grafana/pyroscope/ebpf/symtab/gosym
cpu: AMD Ryzen 9 5950X 16-Core Processor            
             │   old.txt    │               new.txt               │
             │    sec/op    │   sec/op     vs base                │
BinSearch-32   3.354m ± 11%   1.716m ± 1%  -48.83% (p=0.000 n=10)
```

### DIFF
--- a/ebpf/symtab/gosym/pcindex.go
+++ b/ebpf/symtab/gosym/pcindex.go
@@ -2,6 +2,8 @@ package gosym
 
 import (
 	"math"
+
+	"golang.org/x/exp/slices"
 )
 
 type PCIndex struct {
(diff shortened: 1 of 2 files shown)


kolesnikovae (approved):
Yeah, `sort.Search` sometimes makes me sad. I wonder if [`slices.BinarySarch`](https://cs.opensource.google/go/go/+/master:src/slices/sort.go;l=124) would be of use here – it was released as part of 1.21.0, for versions before there is [`x/exp/slices`](https://cs.opensource.google/go/x/exp/+/master:slices/sort.go;l=125) package

PR #2260, opened 2023-08-16
chore: renames phlaredb to pyroscopedb

petethepig:
As far as I can tell we don't use this config parameter anywhere. I left some names untouched, because that would break hard, so I think we should change those later e.g:
* https://github.com/pyroscope-io/pyroscope/blob/d97de8a608786a23270b167de18e495f61cfeceb/pkg/storegateway/bucket.go#L45
* https://github.com/pyroscope-io/pyroscope/blob/5dbd4fb1fe294f57651fa70405acf5a666b13d2e/pkg/phlaredb/profile_store.go#L68

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,7 +170,7 @@ define deploy
 		--set pyroscope.podAnnotations."profiles\.grafana\.com\/cpu\.port_name"=http-metrics \
 		--set pyroscope.podAnnotations."profiles\.grafana\.com\/goroutine\.port_name"=http-metrics \
 		--set pyroscope.extraEnvVars.JAEGER_AGENT_HOST=jaeger.monitoring.svc.cluster.local. \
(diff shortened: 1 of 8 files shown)


simonswine:
This is should be fine to change https://github.com/pyroscope-io/pyroscope/blob/5dbd4fb1fe294f57651fa70405acf5a666b13d2e/pkg/phlaredb/profile_store.go#L68

simonswine:
Probably while you in the file: https://github.com/grafana/pyroscope/blob/5dbd4fb1fe294f57651fa70405acf5a666b13d2e/pkg/phlaredb/profile_store.go#L69

Rperry2174 (approved):
lgtm

simonswine (approved):
LGTM

PR #2264, opened 2023-08-17
use go 1.21 for pgo testing

korniltsev:


simonswine (commented):
We usually try to target a go version one behind the latest release, which would be 1.20.0.

I think there is certainly an argument to get 1.21.0 PGO features, I do believe once we like to use PGO we would need different build pipelines anyhow for (e.g. to optimize for different targets).

In any way, can you use the script, because that make sure we don't forget updates at random places `./tools/upgrade-go-version.sh 1.21.0` 🙂

PR #2266, opened 2023-08-17
Handle OG Pyroscope endpoints in query frontend

### LINKED ISSUE #2133: Parellize render queries
Currently the query frontend doesn't split by time and execute in parallel.

I see two solutions:
- Either we deprecate those render APIs.
- Or we use we push up in the frontend the render logic, which basically queries for SelectSeries and SelectMergeStacktraces in the frontend. Technically we should be able to use the same code.

cc @kolesnikovae

kolesnikovae:
Resolves #2133. The change allows to deprecate the [`querier.Diff`](https://github.com/grafana/pyroscope/blob/1596f8a119d74d9272a523444dbb2ff0f84cecf4/api/querier/v1/querier.proto#L16) method.

### DIFF
--- a/pkg/frontend/frontend_diff.go
+++ b/pkg/frontend/frontend_diff.go
@@ -2,13 +2,52 @@ package frontend
 
 import (
 	"context"
+	"net/http"
 
 	"github.com/bufbuild/connect-go"
+	"golang.org/x/sync/errgroup"
 
 	querierv1 "github.com/grafana/pyroscope/api/gen/proto/go/querier/v1"
+	"github.com/grafana/pyroscope/api/gen/proto/go/querier/v1/querierv1connect"
+	phlaremodel "github.com/grafana/pyroscope/pkg/model"
 	"github.com/grafana/pyroscope/pkg/util/connectgrpc"
+	"github.com/grafana/pyroscope/pkg/util/math"
 )
 
-func (f *Frontend) Diff(ctx context.Context, c *connect.Request[querierv1.DiffRequest]) (*connect.Response[querierv1.DiffResponse], error) {
-	return connectgrpc.RoundTripUnary[querierv1.DiffRequest, querierv1.DiffResponse](ctx, f, c)
+func (f *Frontend) Diff(ctx context.Context,
+	c *connect.Request[querierv1.DiffRequest]) (
+	*connect.Response[querierv1.DiffResponse], error,
+) {
+	ctx = connectgrpc.WithProcedure(ctx, querierv1connect.QuerierServiceDiffProcedure)
+	g, ctx := errgroup.WithContext(ctx)
+
+	var left, right *phlaremodel.Tree
+	g.Go(func() error {
+		resp, err := f.SelectMergeStacktraces(ctx, connect.NewRequest(c.Msg.Left))
+		m := phlaremodel.NewFlameGraphMerger()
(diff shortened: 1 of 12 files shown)


pkg/util/connectgrpc/connectgrpc.go
```diff
@@ -115,17 +115,31 @@ func decodeRequest[Req any](req *httpgrpc.HTTPRequest) (*connect.Request[Req], e
...
 	// e.g. initial Content-Type could be 'application/json'.
 	h := removeContentHeaders(req.Header().Clone())
 	h.Set("Content-Type", "application/proto")
 	out := &httpgrpc.HTTPRequest{
 		Method:  http.MethodPost,
-		Url:     req.Spec().Procedure,
+		Url:     url,
 		Headers: connectHeaderToHTTPGRPCHeader(h),
 	}
 	var err error
```
kolesnikovae, line 145: I don't really like it but I doubt there is a better way to ensure that the `Procedure` member of the `*connect.Request` is always present. If the request comes from the inside (i.e. not from the connect-generated handler), the caller does not have an option to specify it.

kolesnikovae:
I also want to check if we affected by [this bug](https://github.com/pyroscope-io/flamegraph.com/issues/333)

pkg/util/connectgrpc/connectgrpc.go
simonswine: Yes I think I hit that problem before, I think it is a reasonable approach.

pkg/util/connectgrpc/connectgrpc.go
```diff
@@ -115,17 +115,31 @@ func decodeRequest[Req any](req *httpgrpc.HTTPRequest) (*connect.Request[Req], e
...
+
+func ProcedureFromContext(ctx context.Context) string {
+	s, _ := ctx.Value(connectURLCtxKey{}).(string)
+	return s
+}
+
+func encodeRequest[Req any](ctx context.Context, req *connect.Request[Req]) (*httpgrpc.HTTPRequest, error) {
+	url := req.Spec().Procedure
+	if url == "" {
+		if url = ProcedureFromContext(ctx); url == "" {
```
simonswine, line 132: Unsure which one of the two should take precedence, might be useful to override in a test using the context, but it is really arbitrary.

simonswine (approved):
LGTM

pkg/frontend/frontend_diff.go
```diff
@@ -2,13 +2,52 @@ package frontend
...
+		m := phlaremodel.NewFlameGraphMerger()
+		m.MergeFlameGraph(resp.Msg.Flamegraph)
+		right = m.Tree()
+		return err
+	})
+	if err := g.Wait(); err != nil {
+		return nil, err
+	}
+
+	maxNodes := int(math.Max(c.Msg.Left.GetMaxNodes(), c.Msg.Right.GetMaxNodes()))
```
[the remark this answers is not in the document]
kolesnikovae: That's a good point. Despite the fact that at the moment both left and right values source from the same HTTP query parameter, it may change in the future. I guess the right way to tackle this would be adding an explicit filed to the `DiffRequest`, because currently the semantics is pretty ambiguous. OTOH, I'm in doubts if we need to care about the limit at all – given that the trees are already truncated.

PR #2270, opened 2023-08-17
Upgrade to dskit without log.Interface and logrus

duricanikolic:
This PR upgrades `dskit` and uses `dskit`'s `log` package for the creation of the global logger.
For the main in the `dskit` see https://github.com/grafana/dskit/pull/359.


simonswine (approved):
LGTM. Thanks for fixing. 

Works as far as I can tell

PR #2271, opened 2023-08-18
chore: fixes release pipeline's GITHUB_TOKEN

petethepig:
cc @simonswine 

I figured this might be the issue we saw with weekly releases — but they work now, so I'm not sure.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -59,7 +59,7 @@ jobs:
           version: latest
           args: release --rm-dist --timeout 60m
         env:
-          GITHUB_TOKEN: ${{ steps.app-releaser.outputs.token }}
+          GITHUB_TOKEN: ${{ steps.app-goreleaser.outputs.token }}
           # Your GoReleaser Pro key, if you are using the 'goreleaser-pro'
           # distribution:
           # GORELEASER_KEY: ${{ secrets.GORELEASER_KEY }}


kolesnikovae (approved):

.github/workflows/release.yml
```diff
@@ -59,7 +59,7 @@ jobs:
           version: latest
           args: release --rm-dist --timeout 60m
         env:
-          GITHUB_TOKEN: ${{ steps.app-releaser.outputs.token }}
+          GITHUB_TOKEN: ${{ steps.app-goreleaser.outputs.token }}
```
petethepig, line 62: Id on line 50 is `app-goreleaser`

simonswine (approved):
LGTM

Unsure what weekly release issue you mean, but this workflow is not run during weekly release, it is only run for tagged OSS releases.  I also would not expect this to be able to run with the id mismatch this PR fixes 🙂 

The last weekly release seems to have ran fine: https://github.com/grafana/pyroscope/actions/runs/5880951980/job/15948379744

petethepig:
@simonswine Yeah, sorry I should have clarified what I meant.

What I mean is that there's the same ID mismatch in the weekly-release file as well: https://github.com/grafana/pyroscope/blob/next/.github/workflows/weekly-release.yml#L59

But you're right, that one does work, so I didn't update it. It's a bit of a mystery to me why the weekly one works, because the regular one did not, and that's why i'm fixing it here.

I guess I won't touch the other one since it works.

PR #2289, opened 2023-08-22
Add pod disruption budget

simonswine:
This makes sure no more than a single pod of a stateful set can be interrupted by e.g. node updates.


cyriltovena (approved):
LGTM

PR #2292, opened 2023-08-23
Fix reference counting

### LINKED ISSUE #2291: Panic when resolving stacktraces
Found this panic recently

image

this is on latest I suspect we're reading but the tree has been released.

cc @kolesnikovae

kolesnikovae:
Fixes #2291

Additionally we could check if the tree/table is nil, but in case if ref counting is not respected it will lead to a memory leak

### DIFF
--- a/pkg/phlaredb/symdb/resolver_test.go
+++ b/pkg/phlaredb/symdb/resolver_test.go
@@ -81,8 +81,7 @@ func Test_Resolver_Unreleased_Failed_Partition(t *testing.T) {
 	s := newBlockSuite(t, [][]string{{"testdata/profile.pb.gz"}})
 	defer s.teardown()
 	ctx, cancel := context.WithCancel(context.Background())
-	// Pass canceled context to make partition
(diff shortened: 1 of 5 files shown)


pkg/phlaredb/symdb/resolver_test.go
```diff
@@ -76,3 +76,25 @@ func Benchmark_block_Resolver_ResolveTree(t *testing.B) {
 		_, _ = r.Tree()
 	}
 }
+
+func Test_Resolver_Unreleased_Failed_Partition(t *testing.T) {
+	s := newBlockSuite(t, [][]string{{"testdata/profile.pb.gz"}})
+	defer s.teardown()
+	ctx, cancel := context.WithCancel(context.Background())
+	// Pass canceled context to make partition
+	// initialization to fail and.
```
kolesnikovae, line 85:
```suggestion
	// Pass canceled context to make partition initialization to fail.
```

cyriltovena (approved):
LGTM

PR #2296, opened 2023-08-23
Fixes a panic when flamegraph is nil 

cyriltovena:
cc @kolesnikovae


simonswine (approved):
LGTM

kolesnikovae:
Thanks for fixing this! I think the nil comes from 
https://github.com/grafana/pyroscope/blob/d1db88d60cf2869fbb2a85161e80f20fa33804be/pkg/frontend/frontend_select_merge_stacktraces.go#L31-L37

And for the default read path via gRPC endpoints it did not cause issues. But once we started handling [OG API in the frontend](https://github.com/grafana/pyroscope/commit/e69682a379206153b1a349ea3ca44d6476da1d99) that became a problem – I'll fix this

Issue #2189, opened 2023-07-31
Stack trace depth limit handling


kolesnikovae:
Follow-up to https://github.com/grafana/pyroscope/pull/2157:
 - [ ] Truncated stack frames should be replaced with `other` stub.
 - [ ] Stacks should be truncated from the top to preserve the original flamegraph structure.

PR #2299, opened 2023-08-24
Clarify profile validation errors

### LINKED ISSUE #2215: Make invalid_profile validation error more specific
Currently we may reject a profile if it violates configured limits. Rejected ingesting responses are accompanied by a corresponding "reason".

In some cases we use "invalid_profile" reason "by default" as we don't have a specific one: https://github.com/grafana/pyroscope/blob/next/pkg/validation/validate.go#L143. We should distinguish these cases and have individual "reasons" for each of them as currently it might be not obvious which limit exactly is being breached.

kolesnikovae:
Resolves #2215. Also makes stack trace truncation to cut deepest frames as per #2189 (but without a stub frame, like "other")

### DIFF
--- a/pkg/validation/validate_test.go
+++ b/pkg/validation/validate_test.go
@@ -252,7 +252,7 @@ func TestValidateProfile(t *testing.T) {
 			MockLimits{
 				MaxProfileStacktraceSampleLabelsValue: 2,
 			},
-			NewErrorf(SampleLabelsLimit, ProfileTooManyLabelsErrorMsg, `{foo="bar"}`, 3, 2),
+			NewErrorf(SampleLabelsLimit, ProfileTooManySampleLabelsErrorMsg, `{foo="bar"}`, 3, 2),
 			nil,
 		},
 		{
--- a/pkg/validation/validate_test.go
+++ b/pkg/validation/validate_test.go
@@ -224,7 +224,7 @@ func TestValidateProfile(t *testing.T) {
 			MockLimits{
 				MaxProfileSizeBytesValue: 1,
 			},
-			NewErrorf(InvalidProfile, ProfileTooBigErrorMsg, `{foo="bar"}`, 3, 1),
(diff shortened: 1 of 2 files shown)


pkg/validation/validate_test.go
```diff
@@ -252,7 +252,7 @@ func TestValidateProfile(t *testing.T) {
 			MockLimits{
 				MaxProfileStacktraceSampleLabelsValue: 2,
 			},
-			NewErrorf(InvalidProfile, ProfileTooManyLabelsErrorMsg, `{foo="bar"}`, 3, 2),
+			NewErrorf(SampleLabelsLimit, ProfileTooManyLabelsErrorMsg, `{foo="bar"}`, 3, 2),
```
simonswine, line 255: We probably should call the variable `ProfileTooManySampleLabelsErrorMsg`, to make it clear which labels

simonswine (approved):
Thanks for adding this, I would like to make the error messages even better, by mentioning which is the limit name to adjust. In some future I want there to be a URL to learn more about the limit and what can be done to avoid it:

```
	QueryTooLongErrorMsg           = "the query time range exceeds the limit (query length: %s, limit: %s, name: max_query_length)"
	ProfileTooBigErrorMsg          = "the profile with labels '%s' size exceeds the limit (profile size: %d, limit: %d, name: max_profile_size_bytes)"
	ProfileTooManySamplesErrorMsg  = "the profile with labels '%s' exceeds the samples count limit (actual: %d, limit: %d, name: max_profile_stacktrace_samples)"
	ProfileTooManyLabelsErrorMsg   = "the profile with labels '%s' label count exceeds the limit (actual: %d, limit: %d, name: max_profile_stacktrace_sample_labels)"
```

kolesnikovae:
> In some future I want there to be a URL to learn more about the limit and what can be done to avoid it:

Yes, that would be very helpful, for most people these messages tell nothing.

PR #2303, opened 2023-08-24
Imporve limit exceed messages

simonswine:


kolesnikovae (approved):
Oh, right, thanks for fixing this!

PR #2304, opened 2023-08-24
helm: Replace internal agent with Grafana Agent

cyriltovena:
This replace the old agent module with Grafana Agent (as a subchart) in helm.

I decided to keep the agent on by default as it was before. I suspect this whole logic might be better in the agent helm chart.

I still needs to verify if the documentation needs edit.


Makefile
```diff
@@ -356,15 +356,15 @@ trunk/fmt: $(BIN)/trunk
 
 .PHONY: helm/check
 helm/check: $(BIN)/kubeconform $(BIN)/helm
-	$(BIN)/helm repo add --force-update minio https://charts.min.io/
```
simonswine, line 359: You need to keep this and actually add the grafana repo for the agent.

operations/pyroscope/relabeling_test.go
```diff
@@ -1,230 +0,0 @@
...
-	"github.com/prometheus/common/model"
-	"github.com/prometheus/prometheus/discovery/targetgroup"
-	"github.com/prometheus/prometheus/model/labels"
-	"github.com/stretchr/testify/require"
-	"gopkg.in/yaml.v3"
-
-	"github.com/grafana/pyroscope/pkg/agent"
-)
-
-func Test_Relabeling(t *testing.T) {
```
simonswine, line 20: Wonder if we can test the river config in a similar way

cyriltovena: Not sure we could lint the river file may be.

cyriltovena: I added a test that verify the river file is valid and formatted.

operations/pyroscope/helm/pyroscope/templates/configmap-agent.yaml
```diff
@@ -0,0 +1,186 @@
...
+    		action        = "replace"
+    		source_labels = ["__meta_kubernetes_pod_container_name"]
+    		target_label  = "container"
+    	}
+    }
+  {{- $profileTypes := list "memory" "cpu" "goroutine" "block" "mutex"  }}
+  {{- range $profileTypes }}
+
+    discovery.relabel "kubernetes_pods_{{.}}_default_name" {
+    	targets = concat(discovery.relabel.kubernetes_pods.output)
```
simonswine, line 71: Worth leaving a comment why those need to be tabs, vs. the ones in lines 36-65 are spaces

operations/pyroscope/helm/pyroscope/templates/NOTES.txt
```diff
@@ -1,19 +1,50 @@
...
+For that, you'll need to add the Pyroscope data source to your Grafana instance and configure the query URL accordingly.
+See https://grafana.com/docs/grafana/latest/datasources/grafana-pyroscope/ for more details.
 
-To forward the query API to your localhost you can use:
+The in-cluster query URL for the data source in Grafana is:
 
+```
 {{- if hasKey .Values.pyroscope.components "query-frontend" }}
-kubectl --namespace {{ .Release.Namespace }} port-forward svc/{{ include "pyroscope.fullname" . }}-query-frontend {{ .Values.pyroscope.service.port }}:{{ .Values.pyroscope.service.port }}
+http://{{ include "pyroscope.fullname" . }}-query-frontend.{{ .Release.Namespace }}.svc.cluster.local.:{{ .Values.pyroscope.service.port }}
```
simonswine, line 23: This port is wrong when you override the port only for that one component.

You would need to do the default thing here, too.

simonswine (approved):
LGTM

operations/pyroscope/helm/pyroscope/templates/configmap-agent.yaml
cyriltovena: Yeah this is because of the river linter.

operations/pyroscope/helm/pyroscope/templates/NOTES.txt
cyriltovena: Yeah not easy with the dash. :thinking:

cyriltovena:
```
{{ $port := ((index .Values.pyroscope.components "query-frontend").service).port | default .Values.pyroscope.service.port }}
```

did the trick :mindblowing:

Issue #2302, opened 2023-08-24
unrecognised module name: store-gateway


mvkrishna86:
#### Describe the bug
Getting the below error in pyroscope-store-gateway pod. 
failed running phlare: unrecognised module name: store-gateway

#### To Reproduce

Start phlare in microservice mode with operations/pyroscope/helm/pyroscope/values-micro-services.yaml

#### Expected behavior

It should be up without any issues

#### Environment

- Infrastructure:Kubernetes
- Deployment tool: helm

![image](https://github.com/grafana/pyroscope/assets/15136443/c8dc408a-cdea-421f-8d42-0dbea70d1ec6)

mvkrishna86:
@petethepig  and @simonswine can you please help?

simonswine:
Do you mind sharing your values.yaml file and the version of the helm chart (e.g. in your `helm install` command?)

My suspicion is the docker image used is too old.

mvkrishna86:
Here is my HelmRelease file I am using with kustomization and flux:


`apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: pyroscope
  namespace: profiling
spec:
  interval: 1h0m0s
  releaseName: pyroscope
  serviceAccountName: devops
  timeout: 10m
  maxHistory: 2
  upgrade:
    cleanupOnFail: true
  chart:
    spec:
      chart: pyroscope
      version: 0.5.4
      sourceRef:
        kind: HelmRepository
        name: grafana
        namespace: flux-system
  values:

    # Default values for pyroscope.
    # This is a YAML-formatted file.
    # Declare variables to be passed into your templates.

    pyroscope:
      nodeSelector:
        workload: monitoring
      tolerations:
        - operator: Exists
      components:
        querier:
          kind: Deployment
          replicaCount: 3
          resources:
            limits:
              memory: 1Gi
            requests:
              memory: 256Mi
              cpu: 1
        query-frontend:
          kind: Deployment
          replicaCount: 2
          resources:
            limits:
              memory: 1Gi
            requests:
              memory: 256Mi
              cpu: 100m
        query-scheduler:
          kind: Deployment
          replicaCount: 2
          resources:
            limits:
              memory: 1Gi
            requests:
              memory: 256Mi
              cpu: 100m
        distributor:
          kind: Deployment
          replicaCount: 2
          resources:
            limits:
              memory: 1Gi
            requests:
              memory: 256Mi
              cpu: 500m
[...]

PR #2306, opened 2023-08-25
Update helm chart release to be 1.0.0

simonswine:
Update helm chart to be 1.0.0, as otherwise it is not used as the default and the broken chart version 0.5.4 is used.

See #2302


operations/pyroscope/helm/pyroscope/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: pyroscope
 description: 🔥 horizontally-scalable, highly-available, multi-tenant continuous profiling aggregation system
 type: application
-version: 1.0.0-rc.0
+version: 1.0.0
```
cyriltovena, line 5: technically we could go with 0.9 for now.

cyriltovena (approved):
LGTM

PR #2311, opened 2023-08-25
Fix head flush orchestration

### LINKED ISSUE #2284: Investigate test timeouts
From time to time Go test CI job fails because of the timeout (10m): https://github.com/grafana/pyroscope/actions/runs/5922518424/job/16056603953

The very first time it happened on the main/next branch is: https://github.com/grafana/pyroscope/actions/runs/5888100969/job/15968636669

I think that https://github.com/grafana/pyroscope/commit/29fb2b36d1224826d8abb58e6129932865764987 could make things worse therefore the problem pops up more frequently

On the other hand, `make test` takes not longer than 100s if run locally

kolesnikovae:
The change makes `PhlareDB` responsible for tracking the `Head` block flush conditions.

Resolves #2284

### DIFF
--- a/pkg/phlaredb/phlaredb.go
+++ b/pkg/phlaredb/phlaredb.go
@@ -181,7 +181,7 @@ func (f *PhlareDB) maxBlockDuration() time.Duration {
 
 func (f *PhlareDB) maxBlockBytes() uint64 {
 	maxBlockBytes := defaultParquetConfig.MaxBlockBytes
-	if f.cfg.Parquet != nil && f.cfg.Parquet.MaxBlockBytes > maxBlockBytes {
+	if f.cfg.Parquet != nil && f.cfg.Parquet.MaxBlockBytes > 0 {
 		maxBlockBytes = f.cfg.Parquet.MaxBlockBytes
 	}
 	return maxBlockBytes
--- a/pkg/phlaredb/phlaredb.go
+++ b/pkg/phlaredb/phlaredb.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/bufbuild/connect-go"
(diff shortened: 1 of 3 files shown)


pkg/phlaredb/phlaredb.go
```diff
@@ -146,47 +157,34 @@ func (f *PhlareDB) loop() {
...
 }
 
-func (f *PhlareDB) headFlushCh() chan struct{} {
-	f.headLock.RLock()
-	defer f.headLock.RUnlock()
-	if h := f.head; h != nil {
-		return h.flushCh
+func (f *PhlareDB) maxBlockBytes() uint64 {
+	maxBlockBytes := defaultParquetConfig.MaxBlockBytes
+	if f.cfg.Parquet != nil && f.cfg.Parquet.MaxBlockBytes > maxBlockBytes {
```
simonswine, line 184: Not too sure how much sense it makes to limit the block size to the default. Maybe someone wanting to run really small clusters might benefit from setting it lower. (Currently we don't expose the limit in CLI/config at all, so this is a theoretical question)

simonswine (approved):
Thanks for getting to the bottom of that. 

LGTM

pkg/phlaredb/phlaredb.go
kolesnikovae: Agreed, it makes no sense, overlooked that. Thanks for pointing out!

kolesnikovae:
I have just realised that the block size limit is 12.8 GB, this is something we should re-asses. One one hand, it is good, because we don't want to flush symbols too often. On the other, well, someone may want to set some memory limits (although, I wouldn't rely on this given we don't have actual direct control over memory).

Issue #2317, opened 2023-08-28
Storage: SelectMergeProfile (pprof) is not queried from store-gateway


cyriltovena:
#### Describe the bug

I guess we forgot to implement this but when querying for pprof format, we currently only query ingesters so long term data that has been removed from ingesters won't appear.

We should query for both ingesters and store-gateway if available.

PR #2318, opened 2023-08-28
Activate store-gateway when using single binary mode.

cyriltovena:
This activate the store-gateway when using single binary mode with object storage configured properly.

This doesn't work for filesystem because of https://github.com/grafana/pyroscope/issues/2029.

I'm still testing this and found https://github.com/grafana/pyroscope/issues/2317 that affect PGO feature when using single binary and querying for historical timerange.

In this PR I also changed the default for replication factor of the store-gateway to 1.

Closes #2247

### DIFF
--- a/cmd/pyroscope/pyroscope.yaml
+++ b/cmd/pyroscope/pyroscope.yaml
@@ -1,2 +1,10 @@
 server:
   http_listen_port: 4040
+
+storage:
+  backend: gcs
+  gcs:
+    bucket_name: cyriltovena
+
+phlaredb:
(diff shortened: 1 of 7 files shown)


cmd/pyroscope/pyroscope.yaml
```diff
@@ -1,2 +1,10 @@
 server:
   http_listen_port: 4040
+
```
cyriltovena, line 3: Still testing and will remove once I'm done

petethepig (approved):
LGTM

PR #2329, opened 2023-08-29
Switch to v1 Golang SDK

cyriltovena:
If we want to switch to the v1 SDK once is merged https://github.com/grafana/pyroscope-golang/pull/39

We'll need to tag and reupdate the PR with 1.0.


kolesnikovae (approved):
LGTM

PR #2334, opened 2023-08-30
docs: Removes the old scrape_configs block documentation

cyriltovena:
This is a left over of the internal agent.


kolesnikovae (approved):
LGTM

Issue #2315, opened 2023-08-25
"Single" button link is broken on "Tag Explorer" page


bryanhuhta:
#### Describe the bug

When selecting the "Single" button on the "Tag Explorer" page, it redirects to the Grafana home page.

#### To Reproduce

1. Navigate to the "Tag Explorer" page
2. Click on the "Single" button from the "Tag Breakdown" section
  <img width="1418" alt="Screenshot 2023-08-25 at 5 36 06 PM" src="https://github.com/grafana/pyroscope/assets/32787160/ab20c721-96c1-4b60-a9af-f2318f2464d4">
3. The button link redirects to Grafana home.

#### Expected behavior

The button link should redirect to the "Single View" page.

Rperry2174:
Similar to this for the app plugin: https://github.com/grafana/pyroscope-app-plugin/issues/23

I believe this happened as a result of the merge and what the root should link to

bryanhuhta:
Fixed in https://github.com/grafana/pyroscope/pull/2337

PR #2337, opened 2023-08-30
fix: make URLs modifiable in code

darrenjaneczek:
- Replaced `PAGES` `enum` with a simple keyed object to represent the URLs of each page
  - Renamed file from `constants` to `urls` since consumers of this code can now modify the URL values (not constant anymore `;)` ).
- Supports #2315


Rperry2174 (approved):
Lgtm

Issue #2241, opened 2023-08-11
go1.21.0 pprof function names


kolesnikovae:
Starting from go 1.21.0 function/method names in pprof profiles become somewhat cumbersome. This is true at least for heavily _generalised_ types. 

<img width="1344" alt="image" src="https://github.com/grafana/pyroscope/assets/12090599/d4e1d291-926b-447e-a762-7e508755692e">

korniltsev:
I bet this is to fix pgo for generics

It did not work because pprof had [...] in names and compiler new the exact type

petethepig:
we should maybe shrink those names at render time.

PR #2343, opened 2023-08-31
Remove type parameters from go names

kolesnikovae:
Fixes https://github.com/grafana/pyroscope/issues/2241.

Replaces Go type parameters with ellipses as it was before go 1.21.0. 

image


pkg/validation/limits.go
```diff
@@ -92,7 +92,7 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&l.MaxProfileStacktraceSamples, "validation.max-profile-stacktrace-samples", 16000, "Maximum number of samples in a profile. 0 to disable.")
 	f.IntVar(&l.MaxProfileStacktraceSampleLabels, "validation.max-profile-stacktrace-sample-labels", 100, "Maximum number of labels in a profile sample. 0 to disable.")
 	f.IntVar(&l.MaxProfileStacktraceDepth, "validation.max-profile-stacktrace-depth", 1000, "Maximum depth of a profile stacktrace. Profiles are not rejected instead stacktraces are truncated. 0 to disable.")
-	f.IntVar(&l.MaxProfileSymbolValueLength, "validation.max-profile-symbol-value-length", 1024, "Maximum length of a profile symbol value (labels, function names and filenames, etc...). Profiles are not rejected instead symbol values are truncated. 0 to disable.")
+	f.IntVar(&l.MaxProfileSymbolValueLength, "validation.max-profile-symbol-value-length", 65535, "Maximum length of a profile symbol value (labels, function names and filenames, etc...). Profiles are not rejected instead symbol values are truncated. 0 to disable.")
```
kolesnikovae, line 95: The limit is increased as length of 1024 is too short for Go:  
image

In practice we already have profile size limit therefore this one might not be required. Because of the string truncation it's easy to get unreadable profile (e.g such profiles are useless for PGO)

pkg/model/fmt.go
```diff
@@ -0,0 +1,24 @@
...
+	var modified strings.Builder
+	prevEnd := 0
+	for _, indices := range matchesIndices {
+		start, end := indices[0], indices[1]
+		modified.WriteString(input[prevEnd:start] + "[...]")
+		prevEnd = end
+	}
+	modified.WriteString(input[prevEnd:])
+	return modified.String()
+}
```
kolesnikovae, line 24: Note that we don't know if it is a Go name therefore it is applied to all nodes. `ReplaceAllString` would look better here but it allocates a new string.

Issue #2341, opened 2023-08-31
Not able to send profiling data to pyroscope


mvkrishna86:
#### Describe the bug
Not able to send profiling data to pyroscope

#### To Reproduce

Run Java application with below config:
export PYROSCOPE_FORMAT=jfr
export PYROSCOPE_PROFILER_ALLOC=0
export PYROSCOPE_TENANT_ID=dev
export PYROSCOPE_LOG_LEVEL=debug
export PYROSCOPE_SERVER_ADDRESS=https://pyroscope.xxxxxx.click/
export PYROSCOPE_APPLICATION_NAME=PetClinic

java -javaagent:/home/vamsi/pyroscope.jar -jar spring-petclinic-3.1.0-SNAPSHOT.jar

#### Expected behavior

Profiling data should be exported to Pyroscope server

#### Environment
Client:

OS: Ubuntu 22.04.1
Java: openjdk version "17.0.8"

Server: 
- Infrastructure: Kubernetes
- Deployment tool: helm
- chart version: 1.0.1

korniltsev:
can you share logs and pyroscope sdk & server version?

mvkrishna86:
Server version: 1.0.0 ( image: grafana/pyroscope:1.0.0)

Errors in Client Logs:

```
2023-08-31 10:49:30.025 [DEBUG] Uploading started
2023-08-31 10:49:30.027 [DEBUG] Config Config{applicationName='PetClinic', profilingInterval=PT0.01S, profilingEvent=ITIMER, profilingAlloc='0', profilingLock='', uploadInterval=PT10S, logLevel=DEBUG, serverAddress='https://pyroscope.xxxxxx.click', authToken='null', timeseriesName='PetClinic', timeseries=PetClinic, format=JFR, pushQueueCapacity=8, labels={}, ingestMaxTries=8, compressionLevelJFR=1, compressionLevelLabels=1, allocLive=false, httpHeaders={}, samplingDuration=null, tenantID=dev}
2023-08-31 10:49:30.054 [INFO] Profiling started


              |\      _,,,--,,_
             /,`.-'`'   ._  \-;;,_
  _______ __|,4-  ) )_   .;.(__`'-'__     ___ __    _ ___ _______
 |       | '---''(_/._)-'(_\_)   |   |   |   |  |  | |   |       |
 |    _  |    ___|_     _|       |   |   |   |   |_| |   |       | __ _ _
[...]

PR #2352, opened 2023-09-01
fix(docs): warn java users setting alloc/lock to 0 is not a good idea

korniltsev:
Add the following sentence

```
Setting to 0 registers every event, causing significant overhead and results in large profiles, making it not ideal for production. We recommend a starting value of 512k, adjusting as needed.
```

Maybe it will stop users setting alloc=0

https://github.com/grafana/pyroscope/issues/2341


docs/sources/configure-client/language-sdks/java.md
```diff
@@ -116,26 +116,25 @@ for configuration in multiple sources: system properties, environment variables,
...
+|---------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `PYROSCOPE_FORMAT`                          | sets the profiler output format. The default is `collapsed`, but in order to support multiple formats it must be set to `jfr`.                                                                                                                                                                                                                                                                              |
```
petethepig, line 121: side note: we should probably switch to jfr by default

docs/sources/configure-client/language-sdks/java.md
```diff
@@ -116,26 +116,25 @@ for configuration in multiple sources: system properties, environment variables,
...
+| `PYROSCOPE_PROFILER_EVENT`                  | sets the profiler event. With JFR format enabled, this event refers to one of the possible CPU profiling events: `itimer`, `cpu`, `wall`. The default is `itimer`.                                                                                                                                                                                                                                          |
+| `PYROSCOPE_PROFILER_ALLOC`                  | sets the allocation threshold to register the events, in bytes (equivalent to `--alloc=` in `async-profiler`). The default value is `""` - empty string, which means that allocation profiling is disabled. Setting to 0 registers every event, causing significant overhead and results in large profiles, making it not ideal for production. We recommend a starting value of 512k, adjusting as needed. |
```
petethepig, line 123:
```suggestion
| `PYROSCOPE_PROFILER_ALLOC`                  | sets the allocation threshold to register the events, in bytes (equivalent to `--alloc=` in `async-profiler`). The default value is `""` - empty string, which means that allocation profiling is disabled. Setting it to 0 will register every event, causing significant CPU and network overhead, making it not suitable production environments. We recommend setting a starting value of 512k and adjusting it as needed. |
```

petethepig (approved):
made it sound a bit better

docs/sources/configure-client/language-sdks/java.md
korniltsev: hehe, I already did it this morning https://github.com/grafana/pyroscope-java/pull/112/files#diff-99f8fe330b16fd778c4e54482fe3f5e9ba425b5dd458fbcf9858e3ef74e0d786R70

PR #2362, opened 2023-09-05
Remove live alloc profiling from java docs

Rperry2174:
Seems like we don't support this option so should remove it from docs:
image


korniltsev (approved):
I'd maybe go ahead and remove it even from the code, I dont know how to make it work with continuous profiling without patchhing async-profiler

PR #2364, opened 2023-09-05
add ability to annotate services separately

rabunkosar-dd:
Adding ability to annotate services separately. Currently all services share the same labels and annotations from the main block. Not sure how this will impact the monolith annotations?


operations/pyroscope/helm/pyroscope/templates/services.yaml
```diff
@@ -9,6 +9,13 @@ metadata:
   labels:
     {{- include "pyroscope.labels" . | nindent 4 }}
     app.kubernetes.io/component: {{ $component | quote }}
+    {{- if $cfg.labels }}
+      {{- toYaml $cfg.labels | nindent 4 }}
+    {{- end }}
+  {{- if $cfg.annotations }}
+  annotations:
+    {{- toYaml $cfg.annotations | nindent 4 }}
+  {{- end }}
```
kolesnikovae, line 18: I think we should be more specific and specify `serviceAnnotations`

operations/pyroscope/helm/pyroscope/templates/services.yaml
```diff
@@ -9,6 +9,13 @@ metadata:
   labels:
     {{- include "pyroscope.labels" . | nindent 4 }}
     app.kubernetes.io/component: {{ $component | quote }}
+    {{- if $cfg.labels }}
+      {{- toYaml $cfg.labels | nindent 4 }}
+    {{- end }}
```
kolesnikovae, line 14: Could you please tell more about the use case? I'm asking because I think that `app.kubernetes.io/component: {{ $component | quote }}` identifies the service object unambiguously and additional labels should not be required

operations/pyroscope/helm/pyroscope/templates/services.yaml
kolesnikovae, line 18: Consider naming the attribute `serviceAnnotations`

kolesnikovae (commented):
Thank you for the contribution! 🎉 

In monolith mode there going to be just [one virtual component `all`](https://github.com/grafana/pyroscope/blob/b4fb0e65a02f072ba601066b42dff7dcb4983fe4/operations/pyroscope/helm/pyroscope/templates/_helpers.tpl#L90-L92) without annotations, but it should be possible to override it

operations/pyroscope/helm/pyroscope/templates/services.yaml
rabunkosar-dd: we had a case where we used a specific label to identify the pod but like you mentioned, with a change we switched to the component label, I'll remove that.

rabunkosar-dd:
@kolesnikovae addressed your feedback, PTAL when you have time, thank you very much

kolesnikovae:
Thank you so much @rabunkosar-dd! Could you please run `helm/check` to make the linter happy?

kolesnikovae (approved):
LGTM

JakeCooper:
Just so we're clear, this PR doesn't seem to allow specifying service annotations for the `all` component when in non-microservice mode right?

Trying to add a LoadBalancer to ingest profiles from our fleet of VM instances

JakeCooper:
Posted new thread here

https://github.com/grafana/pyroscope/issues/2689

PR #2365, opened 2023-09-06
Add option to disable RBAC

### LINKED ISSUE #2360: Helm: make RBAC configurable
Pyroscope helm chart does not allow configuring RBAC, which may be a problem in some environments where e.g creation of ClusterRole requires priviliges.

kolesnikovae:
Resolves #2360 

We should revisit the RBAC setup, I don't think we need a cluster role.


cyriltovena (approved):
LGTM

PR #2366, opened 2023-09-06
storage: Refactor block compaction to allow shard-splitting

cyriltovena:
This refactors the current compaction function `Compact` to create a new `CompactWithSplitting(...,shardCount uint64)` function.

The new function compacts and deduplicates data away from input blocks while splitting it into `shardCount` shards new blocks. Some output blocks might be emtpy and not returned.

Right now the shards key is the labels finguerprint (hash of all labels pair). I suspect we might want to add different sharding strategy in the future after some experiment.

To split into multiple blocks we don't hook into the reader anymore, we read row by row and push to the correct output block, turns out to be much cleaner. Specially since we don't have to write row groups ourselves anymore I discovered that if you configure correctly the parquet writer it will automatically handle flushing row groups.

I'm still working on adding more tests now.

PS: I suggest to review this in split view.


pkg/phlaredb/compact.go
```diff
@@ -469,73 +571,36 @@ func (it *dedupeProfileRowIterator) Next() bool {
...
-	err         error
 
 	numSamples uint64
 }
 
-func newSymbolsRewriter(it iter.Iterator[profileRow], blocks []BlockReader, w *symdb.SymDB) *symbolsRewriter {
-	sr := symbolsRewriter{
-		profiles:  it,
-		rewriters: make(map[BlockReader]*symdb.Rewriter, len(blocks)),
+func newSymbolsRewriter(path string) *symbolsRewriter {
```
cyriltovena, line 582: @kolesnikovae 

We now creates  rewriters lazily as soon as I see the first row for a given block. Rewriters are not shared across multiple output blocks.

WDYT ?

kolesnikovae: I like it :)

cyriltovena:
Just need to finish writing some more robust tests then we should be good. I spotted a bug where start/end time is from original block and not actual profiles so I'll have to fix this.

pkg/phlaredb/compact_test.go
```diff
@@ -85,6 +88,150 @@ func TestCompact(t *testing.T) {
...
+
+	// Finally test some stacktraces resolution.
+	it, err = queriers[1].SelectMatchingProfiles(ctx, matchAll)
+	require.NoError(t, err)
+	res, err := queriers[1].MergeByStacktraces(ctx, it)
+	require.NoError(t, err)
+
+	expected := new(phlaremodel.Tree)
+	expected.InsertStack(10, "baz", "bar", "foo")
+	require.Equal(t, expected.String(), res.String())
```
cyriltovena, line 199: @kolesnikovae Seems like I must be doing something wrong with symbols, only the first  block (shard 0) can correctly resolve the stacktraces, there rest are showing `nanoseconds` as part of the stacktrace.

Let me know if you see anything wrong w/r/t symbols.

pkg/phlaredb/compact.go
```diff
@@ -469,73 +573,36 @@ func (it *dedupeProfileRowIterator) Next() bool {
...
-	}
+func (s *symbolsRewriter) ReWriteRow(profile profileRow) error {
 	var err error
-	profile := s.profiles.At()
 	profile.row.ForStacktraceIDsValues(func(values []parquet.Value) {
 		s.loadStacktracesID(values)
-		r := s.rewriters[profile.blockReader]
+		r, ok := s.rewriters[profile.blockReader]
+		if !ok {
+			r = symdb.NewRewriter(s.w, profile.blockReader.Symbols())
```
cyriltovena, line 603: I suspect this is where I'm doing something wrong.

kolesnikovae: It looks okay, this is how I'd have done it. I need to take a closer look

pkg/phlaredb/compact_test.go
kolesnikovae: Right, I guess the problem is that each rewriter thinks that it owns the symbols of the block reader and therefore modifies it in-place. I'll fix this. But it will make it even more expensive resource-wise :/

kolesnikovae: https://github.com/grafana/pyroscope/pull/2371 should do the trick. Locally tests pass

PR #2371, opened 2023-09-07
Fix symbols split-compaction

kolesnikovae:
The fix allows using the same block symbols reader by multiple stack trace rewrites. Note that this will cause excessive memory usage. We likely need a better way of handling this.

The failing test is irrelevant to the change.


kolesnikovae:
@cyriltovena this should fix [the problem](https://github.com/grafana/pyroscope/pull/2366#discussion_r1317822271).

cyriltovena:
Awesome thanks I'll check out the race now.

Issue #2359, opened 2023-09-04
Avoid converting input pprof profiles to tree in the pyroscope adapter


kolesnikovae:
Currently, `/ingest` handler converts input pprof profiles to intermediate tree and then to pprof again for PushService (https://github.com/grafana/pyroscope/blob/main/pkg/ingester/pyroscope). There are two problems:
 - file paths, mappings, and some other information is dropped at conversion.
 - the conversion is expensive resource-wise.

However, this adapter allows to split a pprof profile into multiple profiles based on the sample labels. This might be useful in certain cases (e.g. to annotate HTTP routes, span names, etc), but may also cause issues because of high cardinality of the labels.

korniltsev:
I am already doing it for [JFR](https://github.com/grafana/pyroscope/pull/2355), not touching pprof though yet

PR #2376, opened 2023-09-07
fix: pprof ParseToPprof

korniltsev:
address https://github.com/grafana/pyroscope/issues/2359


pkg/og/convert/pprof/profile.go
```diff
@@ -275,3 +129,69 @@ func (p *RawProfile) loadPprofFromForm() error {
...
+	}
+	if strings.Contains(st, "alloc_") || strings.Contains(st, "inuse_") {
+		return "memory"
+	}
+	if strings.Contains(st, "mutex_") {
+		return "mutex"
+	}
+	if strings.Contains(st, "mutex_") {
+		return "mutex"
+	}
```
kolesnikovae, line 154:
```suggestion
	if strings.Contains(st, "mutex_") {
		return "mutex"
	}
```

pkg/og/convert/pprof/profile.go
```diff
@@ -259,10 +109,14 @@ func (p *RawProfile) loadPprofFromForm() error {
...
 	}
-	p.PreviousProfile, err = form.ReadField(f, formFieldPreviousProfile)
+	PreviousProfile, err := form.ReadField(f, formFieldPreviousProfile)
 	if err != nil {
 		return err
 	}
+	if PreviousProfile != nil {
+		return fmt.Errorf("cumulative profile is nolonger supported. " +
+			"Update clients to the latest versions of pyroscope-golang/godeltaprof")
+	}
```
kolesnikovae, line 119: I guess `godeltaprof` is our internal dependency users might not be aware of

```suggestion
	if PreviousProfile != nil {
		return fmt.Errorf("unsupported client version. " +
			"Please update github.com/grafana/pyroscope-go to the latest version")
	}
```

kolesnikovae (approved):
🎉

korniltsev:
Btw this breaks all the rideshare examples
cc @grafana/pyroscope-team

korniltsev:
@kolesnikovae ptal again. I've adde two fixes since your last review. 
1 - there was a wrong time in pyspy pprofs
2 - I had to rewrite function names for pyspy and rbspy to include function name. There used to be a line number in the past, I decided not to add it for now.

korniltsev:
pyspy issue https://github.com/grafana/pyroscope-rs/pull/134

korniltsev:
o_O I did not expect automerge with no second approve

kolesnikovae:
The new changes look reasonable 👍🏻

kolesnikovae:
> Btw this breaks all the rideshare examples

I was under impression that only Go sends delta profiles (not sure about nodejs), which would indeed stop working. But the example in thte repo (go-push) already uses pyroscope-go with `deltagoprof`, and AFAIK apart from that we don't have anything. Old rideshare examples (demo.pyroscope.io) should not be affected because in pyroscope cloud we use the old parser. Am I missing something?

korniltsev:
the dynamic tags( for example vehicle in rideshare or [this](https://github.com/grafana/pyroscope/blob/main/examples/python/simple/main.py) ) will be broken, not the delta profiles

PR #2379, opened 2023-09-08
Fix pprof merge

### LINKED ISSUE #2377: Panic when merging 2 pprof
image

Seems to happen here in the google pprof library https://github.com/grafana/pyroscope/blob/main/pkg/phlaredb/symdb/resolver.go#L180

kolesnikovae:
Fixes #2377


korniltsev (approved):
Is this the nil deref was

PR #2363, opened 2023-09-05
feat: add params for initContainers, and hostNetwork

bodji:

PR #2380, opened 2023-09-08
Enable PDB for deployments

### LINKED ISSUE #2367: Add disruption budget for deployments
Recently we added PodDisruptionBudget for stateful sets (#2289). It makes sense to add for deployments as well.

kolesnikovae:
Resolves #2367 

PDBs are created without configuration options. Although for deployments it may make more sense to specify `minAvailable` rather than `maxUnavailable=1`, in practice the rollout duration is determined by the ingester stateful set, therefore I think it should not be a problem.

There is also no option to disable PDBs, which might be a problem for old Kubernetes clusters (< 1.21) that don't support `policy/v1` API. I've been thinking about a workaround like https://github.com/grafana/pyroscope/pull/2363#discussion_r1316771362 but haven't included it into the PR.

/cc @simonswine 

UPD: made PDBs configurable. They are enabled by default for all deployments and stateful sets with `maxUnavailable=1`


cyriltovena (approved):
LGTM

simonswine:
> Resolves #2367
> 
> PDBs are created without configuration options. Although for deployments it may make more sense to specify `minAvailable` rather than `maxUnavailable=1`, in practice the rollout duration is determined by the ingester stateful set, therefore I think it should not be a problem.
>

I am also not a total expert on this, but I think for some components loosing them has higher user impact than others. And we need to balance the pdb guaranteed stability with the ability to deschedule workloads at scale and bin-pack them tighter. 

So I think we should not have a pdb for distributors/queriers. They are a large part of our footprint and should be stateless enough to still serve traffic correctly after a sigterm.

The impact of loosing query-frontends and query-schedulers is high, as they would require all queriers to reconnect and affect fairness.

> There is also no option to disable PDBs, which might be a problem for old Kubernetes clusters (< 1.21) that don't support `policy/v1` API. I've been thinking about a workaround like [#2363 (comment)](https://github.com/grafana/pyroscope/pull/2363#discussion_r1316771362) but haven't included it into the PR.

I am not too sure if and how well those `.Capabilities` work when we are using helm charts via tanka. Maybe that's also something that can be easily address/workedaround. But we had issues setting minKubernetes version because of that:

simonswine:
I think it is worth checking what components Mimir/Loki sets PDBs on and if they use minAvailable/maxUnavailable.

I think for querier/distributors a `minAvailable` with a percentage might work quite well.

cyriltovena:
For the race see https://github.com/grafana/pyroscope/pull/2382

PR #2381, opened 2023-09-08
Introduce a meta fetcher that synchronize filtered metadata locally

cyriltovena:
Also introduces markers for deletions

Most of the code is adapted from Mimir.

### DIFF
--- a/pkg/phlaredb/block/fetcher.go
+++ b/pkg/phlaredb/block/fetcher.go
@@ -1,0 +1,606 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+// Provenance-includes-location: https://github.com/grafana/mimir/blob/main/pkg/storage/tsdb/block/fetcher.go
+// Provenance-includes-license: Apache-2.0
+// Provenance-includes-copyright: The Thanos Authors.
+
+package block
+
+import (
+	"context"
+	"encoding/json"
+	"io"
+	"os"
+	"path"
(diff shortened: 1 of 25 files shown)


pkg/phlaredb/block/fetcher.go
```diff
@@ -0,0 +1,606 @@
...
+		}
+	}
+
+	// todo(cyriltovena): we should use ReaderWithExpectedErrs(f.bkt.IsObjNotFoundErr) here, to avoid counting IsObjNotFoundErr as an error
+	// since this is expected
+	r, err := f.bkt.Get(ctx, metaFile)
+	if f.bkt.IsObjNotFoundErr(err) {
+		// Meta.json was deleted between bkt.Exists and here.
+		return nil, errors.Wrapf(ErrorSyncMetaNotFound, "%v", err)
+	}
```
kolesnikovae, line 232: I guess it should be fine. The only caller is `fetchMetadata` which does handle `ErrorSyncMetaNotFound` error properly so we probably don't have to do anything special here

kolesnikovae (approved):
LGTM

pkg/phlaredb/block/fetcher.go
cyriltovena: Yeah the original code was doing this, mostly for metrics, which is not urgent, but I think I can roll it back I was able to make ReaderWithExpectedErrs work I think

PR #2384, opened 2023-09-08
Adds a tenant scanner

cyriltovena:
Adds a way to list all tenants, thoses that requires deletion.

Also adds objstore mock for facilitating tests.

This is required for the compactor that will need to list/delete users.


pkg/phlaredb/bucket/tenant_scanner_test.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	s := NewTenantsScanner(bucketClient, isOwned, log.NewNopLogger())
+	actual, deleted, err := s.ScanTenants(context.Background())
+	require.NoError(t, err)
+	assert.Equal(t, expected, actual)
+	assert.Empty(t, deleted)
+}
+
+func TestUsersScanner_ScanUsers_ShouldNotReturnPrefixedUsedByMimirInternals(t *testing.T) {
```
kolesnikovae, line 57:
```suggestion
func TestUsersScanner_ScanUsers_ShouldNotReturnPrefixedUsedByPyroscopeInternals(t *testing.T) {
```

pkg/phlaredb/bucket/tenant_scanner.go
```diff
@@ -0,0 +1,74 @@
...
+		logger:       logger,
+		isOwned:      isOwned,
+	}
+}
+
+// ScanTenants returns a fresh list of users found in the storage, that are not marked for deletion,
+// and list of users marked for deletion.
+//
+// If sharding is enabled, returned lists contains only the users owned by this instance.
+func (s *TenantsScanner) ScanTenants(ctx context.Context) (users, markedForDeletion []string, err error) {
```
kolesnikovae, line 41: I'm unsure if this is always true. As far as I can see we may return users that are not owned by this instance if `isOwned` fails. I'm pretty sure this is intended behaviour, judging by the tests. It should be okay only if the tenant scanner consumers can handle the same users concurrently (e.g if the only use case is removing tenants). @cyriltovena are there other use cases?

kolesnikovae (approved):
LGTM

pkg/phlaredb/bucket/tenant_scanner.go
cyriltovena: Yes I didn't realize this I'll check the usage of the scanner, I think it was just for cleaning blocks and tenants.

cyriltovena: Seems to only be used for cleaning yes, so we should be fine, nice catch though

PR #2386, opened 2023-09-10
chore(examples): add exampels updater, update examples

korniltsev:


tools/update_examples.go
```diff
@@ -0,0 +1,306 @@
...
+	"strings"
+
+	"github.com/grafana/regexp"
+	"golang.org/x/exp/slices"
+)
+
+var ghToken string
+
+// todo run it by cron & create a PR if needed
+func main() {
```
cyriltovena, line 21: I suggest a make target first.

cyriltovena (approved):
LGTM

tools/update_examples.go
korniltsev: This seems to be recurring theme for my code review :D
:heavy_check_mark:  Added

PR #2387, opened 2023-09-11
Derive series from pprof sample labels

kolesnikovae:
Currently, distributor does not preserve sample labels at normalisation. For example, samples
```
stack trace A, {foo=bar,baz=quz}, 1
stack trace A, {foo=fred,baz=zoo}, 1
```

cut to:
```
stack trace A, {foo=bar,baz=quz}, 2
```
---

With this change, labeled pprof samples are exported as indpendent series. Thus, if a profile with external label set 
```
{service_name="my-service", instance="x"}
```
contains samples labeled with `span_name="A"` and `span_name="B"`, distributor will extract them and send to ingesters in accordance to the newly created series labels:

```
{service_name="my-service", instance="x"} // Contains no samples with labels
{service_name="my-service", instance="x", span_name="A"} // Only contains samples with the span_name="A" label
{service_name="my-service", instance="x", span_name="B"} // Only contains samples with the span_name="B" label
```

In practice this allows for more granular queries. For example, profile of an HTTP/gRPC endpoint:

https://github.com/grafana/pyroscope/assets/12090599/a78dcd3a-f1b7-4551-8789-98c9410a3044


cyriltovena (approved):
LGTM

korniltsev:
nit: maybe add  go.cpu.labels.pprof to pkg/test/integration/ingest_pprof_test.go

PR #2388, opened 2023-09-11
Adds bucketindex package

cyriltovena:
This package allow to create and update a file that index all blocks for a given tenant.

This will be maintained by the compactor and used for cleaning, caching and fetching blocks. In the future we might also want the store-gateway to use this so that they don't list the whole bucket on every startup, etc....

The code is mostly adapted from Mimir.


pkg/phlaredb/bucketindex/index.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+	}
+
+	for i := 0; i < len(idx.BlockDeletionMarks); i++ {
+		if idx.BlockDeletionMarks[i].ID == id {
+			idx.BlockDeletionMarks = append(idx.BlockDeletionMarks[:i], idx.BlockDeletionMarks[i+1:]...)
+			break
+		}
+	}
+}
```
kolesnikovae, line 68: A nit: an object is still retained when removed from `idx.Blocks` and `idx.BlockDeletionMarks`, we should nil it

pkg/phlaredb/bucketindex/index.go
```diff
@@ -0,0 +1,189 @@
...
+const (
+	IndexFilename           = "bucket-index.json"
+	IndexCompressedFilename = IndexFilename + ".gz"
+	IndexVersion1           = 1
+	IndexVersion2           = 2 // Added CompactorShardID field.
+	SegmentsFormatUnknown   = ""
+
+	// SegmentsFormat1Based6Digits defined segments numbered with 6 digits numbers in a sequence starting from number 1
+	// eg. (000001, 000002, 000003).
+	SegmentsFormat1Based6Digits = "1b6d"
```
kolesnikovae, line 30: Nit: I haven't found any mentions of segments in the package. Should we maybe remove the constants?

kolesnikovae (approved):
LGTM

pkg/phlaredb/bucketindex/index.go
cyriltovena: You're right we should remove all of this.

PR #2390, opened 2023-09-11
Use the optimized querier for symdb

simonswine:
This will remove one get_range request to the bucket per table and block in the SymDB.

Unsure if this was left out on purpose. But that saves a 4 byte round trip to object store.


kolesnikovae (approved):
LGTM

PR #2391, opened 2023-09-11
Refactoring: Create a common ring configuration

cyriltovena:
I was about to copy the code again for compaction and decided to end it here.

This basically refactors all component to use the same `RingConfig` struct. Added bonus the generated doc is now more consistent.

### DIFF
--- a/cmd/pyroscope/help.txt.tmpl
+++ b/cmd/pyroscope/help.txt.tmpl
@@ -35,18 +35,8 @@ Usage of ./pyroscope:
     	Etcd password.
   -distributor.ring.etcd.username string
     	Etcd username.
-  -distributor.ring.heartbeat-period duration
-    	Period at which to heartbeat to the ring. 0 = disabled. (default 5s)
-  -distributor.ring.heartbeat-timeout duration
-    	The heartbeat timeout after which distributors are considered unhealthy within the ring. 0 = never (timeout disabled). (default 1m0s)
-  -distributor.ring.instance-addr string
-    	IP address to advertise in the ring.
-  -distributor.ring.instance-id string
-    	Instance ID to register in the ring. (default "<hostname>")
   -distributor.ring.instance-interface-names string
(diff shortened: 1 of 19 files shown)


pkg/validation/limits.go
```diff
@@ -12,6 +12,10 @@ import (
 
 const (
 	bytesInMB = 1048576
+
+	// MinCompactorPartialBlockDeletionDelay is the minimum partial blocks deletion delay that can be configured in Mimir.
+	// Partial blocks are blocks that are not having meta file uploaded yet.
+	MinCompactorPartialBlockDeletionDelay = 4 * time.Hour
```
cyriltovena, line 18: I think this is a left over from another branch.

cmd/pyroscope/help.txt.tmpl
```diff
@@ -35,18 +35,8 @@ Usage of ./pyroscope:
...
-  -distributor.ring.heartbeat-timeout duration
-    	The heartbeat timeout after which distributors are considered unhealthy within the ring. 0 = never (timeout disabled). (default 1m0s)
-  -distributor.ring.instance-addr string
-    	IP address to advertise in the ring.
-  -distributor.ring.instance-id string
-    	Instance ID to register in the ring. (default "<hostname>")
   -distributor.ring.instance-interface-names string
-    	Name of network interface to read address from. (default [<private network interfaces>])
-  -distributor.ring.instance-port int
-    	Port to advertise in the ring (defaults to server.http-listen-port). (default 4040)
```
kolesnikovae, line 49: Just to double check: we're hiding these config options from the sort output, not removing them, right?

kolesnikovae (approved):
LGTM

cmd/pyroscope/help.txt.tmpl
cyriltovena: Correct those a either remove from the normal output (all has it) or it's re-added with common description.

PR #2392, opened 2023-09-11
Reduce roundtrips to object store, when opening parquet files

simonswine:
Opening a parquet files currently involves 6 RTT to the object store bucket, this PR will reduce this to one. (🤞)

- [x] Refactor parquet file opening from to use the same code for all parquet acceess
- [x] Retrieve n ending bytes (a percentage of the full parquet size) to cover footer-magic, footersize, offset- and column- indexes. Limit maximum download size to 256KiB. Cache this until the block is fully opened.
- ~Consider storing the correct size in `meta.json` and use those~ #2419

### DIFF
--- a/pkg/objstore/parquet/file.go
+++ b/pkg/objstore/parquet/file.go
@@ -13,29 +13,31 @@ import (
 type File struct {
 	*parquet.File
 	reader objstore.ReaderAtCloser
-	path   string
-	size   int64
+	meta   block.File
 }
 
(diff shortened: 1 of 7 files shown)


cyriltovena (approved):
LGTM Great work !

I suspect we'll have to make sure the request cancellation is properly working through the reader when reading but this is unrelated to the PR.

kolesnikovae (approved):
LGTM 🚀 Awesome improvement!

simonswine:
>  request cancellation is properly working through the reader

I think the propagation stops quite early, we are not even get that propagated in the parquet code currently. (Might need a `ReaderAtWithContext` or something)

PR #2394, opened 2023-09-12
fix: log tenant id

korniltsev:
https://github.com/grafana/pyroscope/issues/2393

```
ts=2023-09-12T10:30:57.130450314Z caller=http.go:190 level=info traceID=1cfcc3ccda0a068b tenant=238 msg="POST /ingest?aggregationType=&from=1694514641&name=simple.golang.app-new%7B%7D&sampleRate=0&spyName=gospy&units=&until=1694514656 (200) 556.059µs"
```

### DIFF
--- a/pkg/util/http.go
+++ b/pkg/util/http.go
@@ -14,7 +14,7 @@ import (
 	"time"
 
 	"github.com/felixge/httpsnoop"
-	log "github.com/go-kit/log"
+	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/gorilla/mux"
 	"github.com/grafana/dskit/instrument"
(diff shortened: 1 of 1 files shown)


pkg/util/http.go
```diff
@@ -109,8 +109,11 @@ func (l Log) logWithRequest(r *http.Request) log.Logger {
 			localLog = log.With(localLog, "sourceIPs", ips)
 		}
 	}
-
-	return user.LogWith(r.Context(), localLog)
+	orgID := r.Header.Get(user.OrgIDHeaderName)
```
cyriltovena, line 112: Can you fallback on getting in the context then on `anonymous` ? There's some cases where the header won't be there specially on the query path.

korniltsev: I've added fallback. Don't know how to test it though.

cyriltovena (approved):
LGTM

PR #2396, opened 2023-09-12
block: Adds Download and MarkForNoCompact

cyriltovena:
Download basically sync a dir locally and MarkForNoCompact allow to make a block that should not be compacted.

I've also added a bunch of missing tests.


pkg/phlaredb/block/block.go
```diff
@@ -52,6 +52,26 @@ func DownloadMeta(ctx context.Context, logger log.Logger, bkt objstore.Bucket, i
 	return m, nil
 }
 
+// Download downloads directory that is meant to be block directory. If any of the files
+// have a hash calculated in the meta file and it matches with what is in the destination path then
+// we do not download it. We always re-download the meta file.
+func Download(ctx context.Context, logger log.Logger, bucket objstore.Bucket, id ulid.ULID, dst string, options ...objstore.DownloadOption) error {
```
kolesnikovae, line 58: A nit: just to be clear as we are not actually calculating hashes

```suggestion
// Download downloads directory that is meant to be block directory.
func Download(ctx context.Context, logger log.Logger, bucket objstore.Bucket, id ulid.ULID, dst string, options ...objstore.DownloadOption) error {
```

pkg/util/labels.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"github.com/cespare/xxhash/v2"
+	"github.com/prometheus/prometheus/model/labels"
+)
+
+var seps = []byte{'\xff'}
+
+// StableHash is a labels hashing implementation which is guaranteed to not change over time.
+// This function should be used whenever labels hashing backward compatibility must be guaranteed.
+func StableHash(ls labels.Labels) uint64 {
```
kolesnikovae, line 12: Can't find use of the function. If we need it, should we maybe place it in the `model` package?

kolesnikovae (approved):
LGTM! Left a few minor notes

pkg/util/labels.go
cyriltovena: We'll use it for the compactor yes ! I can move it to model.

PR #2397, opened 2023-09-12
Adds a way to validate local blocks

cyriltovena:
This will allow us to ensure blocks downloaded locally are valid before uploading or compacting.

For now I'm using file system bucket storage for simplicity but it might be more efficient in the future to just use the filesystem.


simonswine (commented):
LGTM

simonswine (approved):
LE(ven)B(etter)TM

PR #2398, opened 2023-09-12
fix: Rings are using the incorrect port

bryanhuhta:
In https://github.com/grafana/pyroscope/pull/2391, a change was made to factor out a common ring config to reduce duplication of ring settings across components. In doing so, a change was added to set default port values if no value was provided:

https://github.com/grafana/pyroscope/commit/2c73692f18f99b00dfb336088b7211b953a9ae12#diff-e94eca41c9ef18ab2a324780565931df687280a46d570e6d08ebc54f73bf541dR167-R170

In doing so, this fouled up a separate process where (if no default was provided for instance-port) it would fallback to the http listen port.

This PR removes the change where a default port value would be used, allowing the fallback mechanism to work correctly.


cyriltovena (approved):
LGTM

Issue #2295, opened 2023-08-23
Cache meta iteration in store-gateway


cyriltovena:
Right now each store-gateway download the last 31d metadata blocks on initial startup, on ring change (new member), and periodically (default 15m).

This can causes high CPU usage by store-gateways since there's can be many blocks (thousand).

<img width="2475" alt="image" src="https://github.com/grafana/pyroscope/assets/1053421/7b330c58-9d76-46d4-bdd5-293083c26e1e">

<img width="2475" alt="image" src="https://github.com/grafana/pyroscope/assets/1053421/14f58546-3b71-4e44-b4c4-865c92338ab6">

I think we can improve this, by only updating the latest block prefixes and caching meta files on disk/memory.

cyriltovena:
We also seems allocate a lot of memory while iterating 
<img width="2475" alt="image" src="https://github.com/grafana/pyroscope/assets/1053421/01aeb7aa-6de0-4676-bf1f-ea397d248c25">

cyriltovena:
Seems related to restarting container because of 
<img width="2475" alt="image" src="https://github.com/grafana/pyroscope/assets/1053421/c7256e2d-76a2-4852-98ae-b3c5a30ec538">
 and https://github.com/grafana/pyroscope/issues/2291

kolesnikovae:
Apparently it also causes store-gateway restarts at deployment time, when ingesters start uploading data to the object store, and store-gateways sometimes fail to load meta because of the race. In the log we can see record like:

> caller=phlare.go:405 level=error msg="module failed" module=store-gateway error="invalid service state: Failed, expected: Running, failure: initial blocks synchronization: failed to synchronize Pyroscope blocks for user X: sync block: iter block metas: storage: object doesn't exist"

<img width="1304" alt="image" src="https://github.com/grafana/pyroscope/assets/12090599/f58c0dfa-e52b-4887-b017-d267c71fbb1c">

PR #2401, opened 2023-09-14
Fixes IterBlockMeta to skip dir without meta

cyriltovena:
This will fixes issue seen in https://github.com/grafana/pyroscope/issues/2295#issuecomment-1696846301 with the message:

```
caller=phlare.go:405 level=error msg="module failed" module=store-gateway error="invalid service state: Failed, expected: Running, failure: initial blocks synchronization: failed to synchronize Pyroscope blocks for user X: sync block: iter block metas: storage: object doesn't exist"
```

This can basically happens when a block is being uploaded and not finished. Currently this will make the whole sync operation on store-gateway to fails.

I wanted to write tests on filesystem but the prefix iteration is not working on filesystem, so I went and fixed it.

Fixes https://github.com/grafana/pyroscope/issues/2029
Fixes https://github.com/grafana/pyroscope/issues/2045


kolesnikovae (approved):
LGTM

pkg/phlaredb/block/list.go
```diff
@@ -63,8 +63,12 @@ func IterBlockMetas(ctx context.Context, bkt phlareobj.Bucket, from, to time.Tim
 		for _, id := range ids {
 			id := id
 			g.Go(func() error {
-				r, err := bkt.Get(ctx, id+MetaFilename)
+				r, err := bkt.Get(ctx, path.Join(id, MetaFilename))
 				if err != nil {
+					if bkt.IsObjNotFoundErr(err) {
+						level.Info(util.Logger).Log("msg", "block meta.json not found while iterating", "id", id)
```
simonswine, line 69: Maybe that makes it clearer, that this is not an error:

```suggestion
						level.Info(util.Logger).Log("msg", "skipping block as meta.json not found", "id", id)
```

simonswine (approved):
LGTM

PR #2402, opened 2023-09-14
fix: fix 2 races

korniltsev:
- add integration tests, which do {ingest,push} then {render,SelectMergeProfile}
- fix two races


cyriltovena (approved):
LGTM

korniltsev:
I hope the tests will not be flaky 🤞🏼

PR #2403, opened 2023-09-14
fix:  dotnet ingested pprofs

korniltsev:


pkg/og/convert/pprof/profile.go
```diff
@@ -245,3 +250,27 @@ func FixFunctionNamesForScriptingLanguages(p *pprof.Profile, md ingestion.Metada
...
+	for _, function := range profile.Function {
+		if function.Id != 0 {
+			return
+		}
+	}
+	if len(profile.Function) != len(profile.Location) {
+		return
+	}
+	for i := range profile.Location {
+		profile.Function[i].Id = profile.Location[i].Id
```
cyriltovena, line 274: Always on location per function ?

korniltsev: in current dotnet versions - yes

PR #2407, opened 2023-09-14
fix: retain profile name and app when clicking certain buttons or refreshing page

### LINKED ISSUE #2383: URL state gets overwritten by default profile
#### Describe the bug

Using a URL with preselected profile and label selector gets overwritten (at least partily)

#### To Reproduce

Steps to reproduce the behavior on latest main checkout:

- I last tried it in diff view, but I also encountered in the others.
- Select a non default profile and service name. Now add custom selectors.
- Copy the URL and use a new browser tap to load the page again. (Example URL: http://localhost:4040/comparison-diff?query=memory%3Aalloc_space%3Abytes%3Aspace%3Abytes%7Bservice_name%3D%22default%2Fingester%22%7D&from=1694164281744&rightQuery=memory%3Aalloc_space%3Abytes%3Aspace%3Abytes%7Bservice_name%3D%22default%2Fingester%22%2C+pod%3D%22pyroscope-micro-services-ingester-1%22%7D&leftQuery=memory%3Aalloc_space%3Abytes%3Aspace%3Abytes%7Bservice_name%3D%22default%2Fingester%22%2C+pod%3D%22pyroscope-micro-services-ingester-0%22%7D&until=1694167891744&leftFrom=1694164285&leftUntil=1694167885&rightFrom=1694164285&rightUntil=1694167885)
- For a split second you can see the UI using the correct selectors e.g. `memory:inuse_space:bytes:space:bytes{service_name="default/ingester", pod="pyroscope-micro-services-ingester-0"}`, but then it switches to the default one:

Screenshot 2023-09-08 at 11 34 29

#### Expected behavior

Keep both queries correctly and don't overwrite them with the default query.

#### Additional Context

Also observed in Grafana Cloud plugin

darrenjaneczek:
- Fixes #2383
  - Prevent the query from automatically resetting to default if a query is already present
- Ensure that the Single, Comparison, and Diff buttons result in a query/URL that retains the "service_name"
- Deleted a file not being used


Rperry2174 (approved):
lgtm

bryanhuhta (approved):
LGTM :shipit:

Issue #2368, opened 2023-09-06
Wrong status code for query time range validation 


cyriltovena:
Found this today while investigating logs

<img width="2105" alt="image" src="https://github.com/grafana/pyroscope/assets/1053421/31030b7e-c25e-4db9-a0e6-3a14ab9d47d6">

We shouldn't 500 for a bad/limited user request.

PR #2408, opened 2023-09-14
api: Improve http errors

cyriltovena:
At first, I only wanted to fix some wrong connect error translation but then I realized we have many different ways of doing HTTP errors.

This PR makes sure we use the same function `util.WriteError` or `util.WriteErrorWithStatus` which generates the same HTTP JSON response as the connect API.

Fixes https://github.com/grafana/pyroscope/issues/2368


pkg/frontend/frontend_select_merge_profile.go
```diff
@@ -20,11 +19,11 @@ func (f *Frontend) SelectMergeProfile(ctx context.Context, c *connect.Request[qu
 	ctx = connectgrpc.WithProcedure(ctx, querierv1connect.QuerierServiceSelectMergeProfileProcedure)
 	tenantIDs, err := tenant.TenantIDs(ctx)
 	if err != nil {
-		return nil, connect.NewError(http.StatusBadRequest, err)
+		return nil, connect.NewError(connect.CodeInvalidArgument, err)
```
cyriltovena, line 22: This is what pages me today.

cyriltovena:
Fixes https://github.com/grafana/pyroscope/issues/2368

pkg/util/error.go
```diff
@@ -22,8 +27,35 @@ const (
...
+			http.Error(w, writeErr.Error(), http.StatusInternalServerError)
+		}
+		return
+	}
 	status, cerr := ClientHTTPStatusAndError(err)
-	http.Error(w, cerr.Error(), status)
+	WriteErrorWithStatus(cerr, w, status)
+}
+
+func WriteErrorWithStatus(err error, w http.ResponseWriter, status int) {
```
bryanhuhta, line 44: nit: a more natural signature here might be

```suggestion
func WriteErrorWithStatus(w http.ResponseWriter, status int, err error) {
```

since it's much more common in Go to have the writer first, but keeping this signature consistent with `WriteError(err error, w http.ResponseWriter)` makes sense.

bryanhuhta (approved):
LGTM :shipit:

pkg/util/error.go
cyriltovena: Yeah I also wonder if we should also move it to `util/http` and do `Errorr..(` for those function names.

so `httputil.Error(w,error)`

pkg/util/error.go
```diff
@@ -22,8 +27,35 @@ const (
...
+	w.Header().Set("Content-Type", "text/plain; charset=utf-8")
+	w.Header().Set("X-Content-Type-Options", "nosniff")
+	w.Header().Set("Content-Type", "application/json")
```
simonswine, line 47: I don't think you want to Set `Content-Type` twice. There are also cases where headers have been set before, so maybe we only want to set something if nothing is set:

https://github.com/grafana/pyroscope/blob/dfc13514e978c69f63ca347d1fc5d266e7adb852/pkg/querier/http.go#L66-L69

I also remember for some browsers you need `text/plain` to get them to show the text as monospaced font or something.

simonswine (commented):
Like to get the content type situation sorted before merge 🙂

pkg/util/error.go
cyriltovena: did the whole refactoring now so we don't have it more painful in the future

PR #2410, opened 2023-09-15
fix: various bugs that occur when selection is too small

darrenjaneczek:
It was detected that for very small selections on the timeline, the date range would be left in an invalid state. This is often due to there being no data points in the new selection, resulting in a corruption of subsequent range selection.

This PR avoids the issue by doing a few checks before accepting a range selection on the time line. If there are too few points in the new selection, it is aborted, thus preventing the resulting bugs.


simonswine (approved):
LGTM, nothing obviously wrong, but the code feels long enough to warrant a test case (maybe)

darrenjaneczek:
@simonswine 
> ...the code feels long enough to warrant a test case (maybe)

Shortened the code slightly, extrapolated the evaluation to a util function, added testing for that util function.

Rperry2174 (approved):
Lgtm!

PR #2420, opened 2023-09-18
feat: add metric to track number of times a request fails to forward

bryanhuhta:
In https://github.com/grafana/pyroscope/pull/2391, a bug was introduced that stopped the scheduler from sending requests to the querier. This was reverted in https://github.com/grafana/pyroscope/pull/2398.

This PR adds metrics to track the number of times a frontend worker fails to forward a request to a querier. We can then create alerting based off those metrics to catch failures earlier.

### DIFF
--- a/pkg/frontend/frontend_scheduler_worker.go
+++ b/pkg/frontend/frontend_scheduler_worker.go
@@ -253,7 +253,7 @@ func newFrontendSchedulerWorker(conn *grpc.ClientConn, schedulerAddr string, fro
 		requestCh:                    requestCh,
 		cancelCh:                     make(chan uint64, schedulerWorkerCancelChanCapacity),
 		enqueuedRequests:             enqueuedRequests,
-		requestsForwardedFailedStats: usagestats.NewInt("pyroscope_query_frontend_requests_forwarded_failed"),
+		requestsForwardedFailedStats: usagestats.NewInt("query_frontend_forwarded_requests_failed"),
 		maxLoopDuration:              maxLoopDuration,
 	}
 	w.ctx, w.cancel = context.WithCancel(context.Background())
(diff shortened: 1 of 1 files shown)


pkg/frontend/frontend_scheduler_worker.go
```diff
@@ -309,8 +315,10 @@ func (w *frontendSchedulerWorker) runOne(ctx context.Context, client schedulerpb
 	backoff := backoff.New(ctx, backoffConfig)
 	for backoff.Ongoing() {
 		if !attemptLoop() {
+			w.forwardedRequestsFailedStats.Add(1)
 			backoff.Wait()
 		} else {
+			w.forwardedRequestsFailedStats.Set(0)
 			backoff.Reset()
 		}
 	}
```
kolesnikovae, line 324: It would be great if we could distinguish between failed retry attempts and failures, I guess (assuming we want to setup alerting). `backoff.Err()` indicates whether we have exhausted all the attempts.  

I'm not very sure about resetting the metric with `w.forwardedRequestsFailedStats.Set(0)`, as this is a shared object; how does it affect concurrently occurring failures?

pkg/frontend/frontend_scheduler_worker.go
```diff
@@ -235,20 +237,24 @@ type frontendSchedulerWorker struct {
 	// Number of queries sent to this scheduler.
 	enqueuedRequests prometheus.Counter
 
+	// Number of queries that failed to forward to the querier.
+	forwardedRequestsFailedStats *expvar.Int
```
kolesnikovae, line 241: I'm wondering if we should use usagestats for this. Would a `prometheus.Counter` be more suitable?

simonswine: Yes I agree. We use usage stats for the "phone home" functionality and wouldn't collect operational metrics to that detail there.

bryanhuhta:
I don't have the bandwidth to pursue this at the moment. I'm going to close this PR in favor of this issue https://github.com/grafana/pyroscope/issues/2523. When there is more time, we can come back to that issue for a more thorough improvement.

PR #2413, opened 2023-09-18
Use centralized `make docs` procedure

jdbaldry:
Source:  https://github.com/grafana/writers-toolkit/tree/main/docs.

The centralized procedure ensures a consistent local development experience across a growing number of repositories.

Documentation is available in https://grafana.com/docs/writers-toolkit/review/run-a-local-webserver/.

The updated workflow renames some of the targets in the Makefile but I've done my best to support your existing workflows like `make docs/docs` (or `cd docs; make docs`) and `make docs/test`. `make docs/validate` has been renamed to `make docs/doc-validator` for consistency with other repositories.

The `doc-validator` GitHub workflow has been updated to comment suggestions on PRs. The errors it outputs are documented in https://grafana.com/docs/writers-toolkit/review/doc-validator/errata/. If you experience confusing errors or buggy reports, let me know and I'll happily help out on the PR and fix any bugs in the linter.

PR #2430, opened 2023-09-20
docs: add note about object storage

Rperry2174:
Adding note to warn people about where/when querying from object storage works

### DIFF
--- a/docs/sources/configure-server/configure-object-storage-backend.md
+++ b/docs/sources/configure-server/configure-object-storage-backend.md
@@ -12,6 +12,8 @@ aliases:
 Pyroscope can use different object storage services to persist blocks containing the profiles data.
 Blocks are flushed by ingesters [on disk]({{<relref "./configure-disk-storage.md">}}) first then are uploaded to the object store.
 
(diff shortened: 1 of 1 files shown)


docs/sources/configure-server/configure-object-storage-backend.md
```diff
@@ -12,6 +12,8 @@ aliases:
 Pyroscope can use different object storage services to persist blocks containing the profiles data.
 Blocks are flushed by ingesters [on disk]({{<relref "./configure-disk-storage.md">}}) first then are uploaded to the object store.
 
+At the moment, querying from object storage only works in microservices mode and for labels being actively ingested. In the future, we plan to support querying from object storage in standalone mode as well.
+
```
simonswine, line 16: As per slack the stand-alone mode has support for querying from object store.

The limitation for labels still exists for both stand-alone & micro-services. But once #2413 merges it also is gone.

Rperry2174: yeah will just close this then

Rperry2174:
closing because this will all be fixed very soon

Issue #2230, opened 2023-08-10
Allow series/profileTypes/labelValues to be querier from block/store


simonswine:
We currently only support label retrieval from the Head of each ingester, we should add `start`/`end` parameters to those methods and support querying blocks on ingesters and store-gateways. 

We need to implement this for these APIs:

- [x] Series labels https://github.com/grafana/pyroscope/pull/2431
- [x] Label names https://github.com/grafana/pyroscope/pull/2498
- [x] Label values https://github.com/grafana/pyroscope/pull/2594
- [x] Profile types https://github.com/grafana/pyroscope/pull/2617

bryanhuhta:
With https://github.com/grafana/grafana/pull/77523 merged, this is complete.

PR #2431, opened 2023-09-20
feat: get series labels from store gateway

bryanhuhta:

Previously series were only queried from the head block in the ingester. This adds a time window to the `Series` RPC and queries the store gateway if necessary.

~This PR also adds a rudimentary UI update to capture the selected window start/end and passes it along the RPC.~ Edit: I reverted this in https://github.com/grafana/pyroscope/pull/2431/commits/c72f5c6ef554cc7153edd3d2310ab2de92596c0b to reduce the complexity of this PR. This will be added in a follow-up PR where more time and energy can be used to focus on a "non-rudimentary" implementation.


pkg/phlaredb/block_querier.go
```diff
@@ -905,6 +938,21 @@ func (b *singleBlockQuerier) SelectMatchingProfiles(ctx context.Context, params
...
+	sp, ctx := opentracing.StartSpanFromContext(ctx, "Series Block")
+	defer sp.Finish()
+
+	// TODO(bryan) do we need to close this?
+	err := b.Open(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO(bryan) get labels from block and filter by matchers/label_names
```
bryanhuhta, line 951: Not exactly sure how to get label pairs from a block. I was following how it was queried from the ingester, but there is some differences in the head block format and a block that's been flushed to storage.

I was browsing `profiles.parquet` because it had a `labels` column, but none of the rows had any labels.

simonswine: As per our slack discussion, there two different kind of labels, in this case we are interested in the external labels, which can be found in `index.tsdb` and are mappend into profiles.tsdb over the series index

public/app/redux/reducers/continuous/index.ts
```diff
@@ -75,8 +76,14 @@ const initialState: ContinuousState = {
 export const reloadAppNames = createAsyncThunk(
   'names/reloadAppNames',
   async (_, thunkAPI) => {
+    const start = getUTCdate(formatAsOBject(thunkAPI.getState()['continuous'].from), 0);
+    const end = getUTCdate(formatAsOBject(thunkAPI.getState()['continuous'].until), 0);
+
+    const startMs = toUnixTimestamp(start) * 1000;
+    const endMs = toUnixTimestamp(end) * 1000;
```
bryanhuhta, line 83: Getting the start/end in milliseconds for a page seems like a pretty common action, but I couldn't find a better way to get this information. Ideally we should extract this out into a function.

public/app/redux/reducers/settings.ts
```diff
@@ -88,7 +88,8 @@ export const reloadUsers = createAsyncThunk(
 export const reloadApps = createAsyncThunk(
   'newRoot/reloadApps',
   async (_, thunkAPI) => {
-    const res = await fetchApps();
+    // TODO(bryan) What is this?
+    const res = await fetchApps(0, 0);
```
bryanhuhta, line 92: I'm unsure how to get the start and end here. In fact, I don't really know where in the UI a trigger for this reducer lives.

pkg/phlaredb/head_queriers.go
```diff
@@ -142,6 +143,14 @@ func (q *headOnDiskQuerier) MergeByLabels(ctx context.Context, rows iter.Iterato
 	return seriesByLabels.normalize(), nil
 }
 
+func (q *headOnDiskQuerier) Series(ctx context.Context, params *ingestv1.SeriesRequest) ([]*typesv1.Labels, error) {
```
kolesnikovae, line 146: As of now we keep TSDB in memory until the block is flushed on disk, therefore the head does not store any labels/series on disk. So I think we could return an empty response here

pkg/phlaredb/block_querier.go
```diff
@@ -386,6 +386,7 @@ type Querier interface {
 	MergeByStacktraces(ctx context.Context, rows iter.Iterator[Profile]) (*phlaremodel.Tree, error)
 	MergeByLabels(ctx context.Context, rows iter.Iterator[Profile], by ...string) ([]*typesv1.Series, error)
 	MergePprof(ctx context.Context, rows iter.Iterator[Profile]) (*profile.Profile, error)
+	Series(ctx context.Context, params *ingestv1.SeriesRequest) ([]*typesv1.Labels, error)
```
kolesnikovae, line 389: The method is supposed to be used here: https://github.com/grafana/pyroscope/blob/sg-labels/pkg/phlaredb/phlaredb.go#L333 (using the `Series` function defined below). Otherwise, local blocks on disk will not be queried

pkg/phlaredb/block_querier.go
```diff
@@ -905,6 +938,75 @@ func (b *singleBlockQuerier) SelectMatchingProfiles(ctx context.Context, params
...
 
+func (b *singleBlockQuerier) Series(ctx context.Context, params *ingestv1.SeriesRequest) ([]*typesv1.Labels, error) {
+	sp, ctx := opentracing.StartSpanFromContext(ctx, "Series Block")
+	defer sp.Finish()
+
+	// TODO(bryan) do we need to close this?
+	err := b.Open(ctx)
+	if err != nil {
+		return nil, err
+	}
```
kolesnikovae, line 949: `Open` is very expensive operation, therefore and we do want to keep blocks open. Unfortunately, at the moment we don't have a straight-forward caching / eviction policy, which may easily become a problem. We will change this in the future

kolesnikovae (commented):
Looks great!

Next we need to update [`PhlareDB.Series`](https://github.com/grafana/pyroscope/blob/sg-labels/pkg/phlaredb/phlaredb.go#L333) to use the new implementation

pkg/phlaredb/block_querier.go
```diff
@@ -905,6 +938,75 @@ func (b *singleBlockQuerier) SelectMatchingProfiles(ctx context.Context, params
 	return iter.NewMergeIterator(maxBlockProfile, false, iters...), nil
 }
 
+func (b *singleBlockQuerier) Series(ctx context.Context, params *ingestv1.SeriesRequest) ([]*typesv1.Labels, error) {
```
simonswine, line 941: The big question will be how accurate do we want to be. Without looking at the `profiles.parquet`, we will have a label resolution for the block length only. 

Let's say we only have profile for pod-a from t1 until t1+5min. Label queries will still show pod-a if we query t1+1h until t1+2, just because a block containing the is going from t1-1h until t1+2h.  

I do think this only a minor problem and would rather keep things simple and index only.

We should add a comment about this though.

pkg/phlaredb/block_querier.go
simonswine: Potentially we also what it to have a state where only the TSDB index is there and parquet files are not even looked at.

pkg/phlaredb/block_querier.go
bryanhuhta: Yeah, I did think about that too. The block-length query resolution is unfortunate, but I think something we can iterate on.

Added a comment about this in: https://github.com/grafana/pyroscope/pull/2431/commits/cc05423aed7d23e234515c0a96de77ebea91756d

api/querier/v1/querier.proto
```diff
@@ -33,6 +33,8 @@ message ProfileTypesResponse {
 message SeriesRequest {
   repeated string matchers = 1;
   repeated string label_names = 2;
+  int64 start = 3; // milliseconds since epoch
+  int64 end = 4; // milliseconds since epoch
```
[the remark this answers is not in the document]
bryanhuhta: Fixed in https://github.com/grafana/pyroscope/pull/2431/commits/f4eec65962a8e28c6c80674d6260d2ffbc218113 and https://github.com/grafana/pyroscope/pull/2431/commits/e0c8640c12c25f88e3f6ce299a2f63517f792bbd.

pkg/phlaredb/block_querier.go
bryanhuhta: It makes sense we try keep things open for as long as possible if opens are expensive. I like @simonswine's idea of doing a partial open vs loading the entire block.

pkg/phlaredb/block_querier.go
bryanhuhta: Good catch 🤦

Fixed in https://github.com/grafana/pyroscope/pull/2431/commits/a9182acbc7d3e33580eed95c36fce2d8940c78b5.

pkg/phlaredb/head_queriers.go
bryanhuhta: Fixed in https://github.com/grafana/pyroscope/pull/2431/commits/d671a6a2ed6b0e7efb027ec052aa646d972d7fef

pkg/querier/querier.go
```diff
@@ -193,55 +193,76 @@ func (q *Querier) LabelNames(ctx context.Context, req *connect.Request[typesv1.L
...
 		)
 		sp.Finish()
 	}()
 
-	// build up map of label names
-	labelNameMap := make(map[string]struct{}, len(req.Msg.LabelNames))
-	for _, labelName := range req.Msg.LabelNames {
-		labelNameMap[labelName] = struct{}{}
+	if req.Msg.Start > req.Msg.End {
+		return nil, connect.NewError(connect.CodeInvalidArgument, errors.New("start must be before end"))
```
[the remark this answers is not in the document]
bryanhuhta: In https://github.com/grafana/pyroscope/pull/2431/commits/d8df8654d6767a0ba78266c2a37fe48aacb0a102, I moved the validation from the querier to the query-frontend.

Should we do validation in both places?

pkg/phlaredb/block_querier.go
```diff
@@ -767,6 +768,38 @@ func MergeProfilesPprof(ctx context.Context, stream *connect.BidiStream[ingestv1
...
+			return nil
+		}))
+	}
+	err = group.Wait()
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO(bryan) dedupe here?
+	return &ingestv1.SeriesResponse{LabelsSet: labelsSet}, nil
```
kolesnikovae, line 800: Well, strictly speaking, we do not have to since the caller will deduplicate series either way

kolesnikovae (approved):
LGTM!

pkg/querier/querier.go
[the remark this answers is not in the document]
kolesnikovae: I think doing it in the query-frontend should be enough. We only need to make sure `req.Msg.Start > req.Msg.End` condition is actually validated there

pkg/phlaredb/block_querier.go
```diff
@@ -905,6 +938,78 @@ func (b *singleBlockQuerier) SelectMatchingProfiles(ctx context.Context, params
...
+		if !ok {
+			return false
+		}
+
+		if selectors.matchesAll() {
+			return true
+		}
+
+		for _, selector := range selectors {
+			for _, matcher := range selector {
```
cyriltovena, line 980: This doesn't seem right, you should be applying matcher for the right matching label name. I'd like to see a method that can be tested that only takes an IndexReader, matches, and returns all series.

I think you should be using `PostingsForMatchers` and `Series(id storage.SeriesRef, lbls *phlaremodel.Labels, chks *[]ChunkMeta)` see `(b *singleBlockQuerier) SelectMatchingProfiles` for some examples.

Again `matcher.Matches` is supposed to be run on values, not names.

kolesnikovae: I think Cyril is right, we should get postings via `PostingsForMatchers` and iterate through the series. I haven't tested it but I'd expect something like:

```go
	// get all relevant series.
	for postings.Next() {
		if _, err = b.index.Series(postings.At(), &lbls, &chks); err != nil {
			return nil, err
		}
		if len(params.LabelNames) > 0 {
			if x := lbls.WithLabels(params.LabelNames...); len(x) > 0 {
				// TODO:
				//  - Not sure if we want to keep empty label sets
				//  - Filter out duplicates
				resp.LabelsSet = append(resp.LabelsSet, &typesv1.Labels{
					// TODO: Double check lbls' memory in not reused after Clone.
					Labels: x.Clone(),
				})
			}
			continue
		}
		resp.LabelsSet = append(resp.LabelsSet, &typesv1.Labels{
			Labels: lbls.Clone(),
		})
	}
```

pkg/querier/querier.go
```diff
@@ -193,55 +193,76 @@ func (q *Querier) LabelNames(ctx context.Context, req *connect.Request[typesv1.L
...
-	// build up map of label names
-	labelNameMap := make(map[string]struct{}, len(req.Msg.LabelNames))
-	for _, labelName := range req.Msg.LabelNames {
-		labelNameMap[labelName] = struct{}{}
+	// Some clients may not be sending us timestamps. If start or end are 0, then
+	// mark this a legacy request. Legacy requests only query the ingesters.
+	legacyRequest := req.Msg.Start == 0 || req.Msg.End == 0
+
+	if q.storeGatewayQuerier == nil || legacyRequest {
+		responses, err := q.seriesFromIngesters(ctx, &ingestv1.SeriesRequest{
```
simonswine, line 213: The legacy behaviour is was query from ingesters - Head index only.

With this PR it changes to Head Index and all of the blocks on the ingesters. I fear that this will be fairly slow. So I think something similar should also be done in the ingester Serires. 

```suggestion
		responses, err := q.seriesFromIngesters(ctx, &ingestv1.SeriesRequest{
```

simonswine (commented):
Nice work so far. 

I do think before those changes go in, as they're in the critical path to explore and the app, we need a bit more (read any 😆 ) test coverage.

 I was trying to get some Series calls going locally and I struggled to get any working both on StoreGateway / Querier. 😂 

```
$ curl -v -L -d '{}' -H "content-type: application/json" http://localhost:4040/querier.v1.QuerierService/Series
{"code":"unknown","message":"{\"code\":\"internal\",\"message\":\"unknown: matchers parameter is not implemented: [\\u003cnil\\u003e]\"}"}
```
Can't really explain where this is coming from.

pkg/phlaredb/block_querier.go
```diff
@@ -905,6 +938,90 @@ func (b *singleBlockQuerier) SelectMatchingProfiles(ctx context.Context, params
...
+		}
+	} else {
+		for _, matchers := range selectors {
+			iter, err := PostingsForMatchers(b.index, nil, matchers...)
+			if err != nil {
+				return nil, err
+			}
+			postingsIters = append(postingsIters, iter)
+		}
+	}
```
bryanhuhta, line 995: Depending on the selectors provided, collect posting iterators one of two ways:

1. If no selectors, use `b.index.Postings` for every label name in the request
2. If selectors, use `PostingsForMatchers` for each selector

This is done because `PostingsForMatchers` requires at least 1 selector. Sometimes none are provided. In that case, I collected an iterator for each label name.

pkg/phlaredb/block_querier_test.go
```diff
@@ -175,3 +177,148 @@ func TestSelectMatchingProfilesCleanUp(t *testing.T) {
 	})
 	require.Error(t, err)
 }
+
+func TestSeries(t *testing.T) {
```
bryanhuhta, line 181: The test case for getting series labels from a single block. I still need to try create an e2e test case.

pkg/phlaredb/block_querier.go
cyriltovena: You might be fine using `PostingForMatchers` all the time see https://github.com/grafana/pyroscope/blob/15e9b770dd4a920431d9a74fd4af85a1f9e6894a/pkg/phlaredb/querier.go#L126-L127 which gets all posting in case no matcher is given

pkg/frontend/frontend_series.go
```diff
@@ -4,13 +4,38 @@ import (
...
+	}
+	validated, err := validation.ValidateRangeRequest(f.limits, tenantIDs, interval, model.Now())
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, err)
+	}
+	if validated.IsEmpty {
+		return connect.NewResponse(&querierv1.SeriesResponse{}), nil
+	}
+	c.Msg.Start = validated.Start.Unix()
+	c.Msg.End = validated.End.Unix()
```
kolesnikovae, line 38: In case if the time range is not provided, we probably should set some sane default one. I'd set it to 3 hours from now (not trying to figure out the actual block duration)

pkg/phlaredb/block_querier.go
bryanhuhta: That will only run if `notIts` has data. If you pass no matchers, then both `its` and `notIts` will be empty and that if block will never run.

pkg/frontend/frontend_series.go
bryanhuhta: In https://github.com/grafana/pyroscope/pull/2431/commits/4c83f3674e7c2c245de319cf1df9b26b363add13 I set it to 4 hours. It was a relatively arbitrary decision. I could see an argument for 1h since that's the default time span when loading the plugin. We can always update this.

bryanhuhta:
I tested these changes in dev (using https://github.com/grafana/pyroscope/pull/2431/commits/02ec93982829bee6aa91d55c6c83979dd26ac30a) and the behavior of querying from the store gateway works as intended. We also maintain legacy query behavior if no interval is sent.

I also ran a quick performance check and the results where, well, substandard at best. I made 20 requests with a variety of intervals to the query frontend and collected the _server-side_ time of the response. I tested the existing query pattern, the new one with only synchronous requests (basically a best-case scenario), and the new query pattern with asynchronous requests (the "typical" scenario).

| Interval | Old avg. (ms) | New avg. sync (ms) | New avg. async (ms) |
|--------:|--------:|----------:|----------:|
| 1h     | 215.35 | 177.85   | 390.84   |
| 6h     | 218.00 | 253.25   | 427.29   |
| 24h    | 200.45 | 2837.50  | 6582.29  |
| 2d     | 183.35 | 5199.00  | 12021.38 |
| 7d     | 195.60 | 15423.00 | 13842.05 | 
| 14d    | 196.15 | 27824.16 | 23781.07 |

chart (2)

As the interval increases, the response time increases exponentially in the synchronous case (with an R<sup>2</sup> of 0.998 according to Google Sheets). The asynchronous case increase almost like the synchronous ones, but taper off towards the end. This, however, is due to the pods OOM and crashing, cancelling requests, and causing artificially lower request times.

Any request interval above 24h are almost guaranteed to make pods crash and requests to fail with the new pattern. Any interval above 24h works, but the application is more or less unusable due to the high latency. Less than 24h interval, the requests latency is largely unnoticeable.

I suspect the OOMs are a result of opening an unbounded number of blocks and scanning them for series labels (though that's just a hunch). We could mitigate this reducing parallelism on a single pod, but that would in turn increase the total request duration. Another obvious solution would be to try cache the series labels so we can avoid opening a block altogether, but caching is always easier to propose than it is to implement.

kolesnikovae:
Cross-posting from slack:

I'm surprised the performance is so low. The TSDB index is downloaded into memory entirely on the first access to the block and kept there forever (sort of an in-memory cache). I'm wondering if the performance of subsequent queries is any better. Judging by the flamegraph, it is not since we spend most of the time decoding data (in-memory):

image

We should:
 1. As we have all the data in memory, the query should take no longer than tens of  milliseconds. Most likely, there is a better way to iterate the series (rather than decoding series by series). If it does not help, I think we may want to keep decoded data in-memory. Our indices are tiny, a few thousands of series each, this should not be much worse than we have now.
 2. Implement block compaction and block deduplication. We process lots of duplicate data (3-6x). @cyriltovena and @simonswine are working on it.
 3. Make the cache more robust. Maybe pre-load some blocks into memory on start (so it won't happen at the first query) and evict old TSDB indices.

I guess we could try to fix this before merging.
1. The very first thing I'd do is the late materialisation: when we iterate series via [`index.(*Decoder).Series`](https://github.com/grafana/pyroscope/blob/main/pkg/phlaredb/tsdb/index/index.go#L1961), all label keys and values are resolved with `reader.lookupSymbol`, which contributes most to the overall CPU consumption. I believe we should only resolve symbol references after we collected and deduplicated, so that lookup is performed just once for each unique symbol.
2. Only fetch labels listed in the request, rather than doing it afterwards with `model.Labels.WithLabels` (which is responsible for 5%).
3. Optimize the allocations. `index.(*Decoder).Series` makes lots of allocations, but I fail to explain them (line numbers can be found in the pprof profile you can get with profilecli). I'm 99% percent sure we can eliminate 99.99% of them.
4. Skip the chunk-related part. We don't use it.

pkg/phlaredb/block_querier.go
simonswine: I am kind of surprised by this too. The reason why we never faced this issue, is because we do always use a `__profile_type__=` matcher for our SelectSeries queries.

I think we could potentially refactor the method to be like the prometheus where this was originally copied from  https://github.com/prometheus/prometheus/blob/0de7f39e6ad8ab43d1fac377e25f3b76cfaf616a/tsdb/querier.go#L240 then it would support it and also come with other optimizations.

pkg/querier/querier.go
```diff
@@ -193,53 +194,92 @@ func (q *Querier) LabelNames(ctx context.Context, req *connect.Request[typesv1.L
...
 		if err != nil {
 			return nil, err
 		}
-		// TODO: Remove this once we have the ingesters returning the filtered response, has been rolled out (f25).
-		filterSeriesByLabelNames(res.Msg.LabelsSet, labelNameMap)
 
-		return res.Msg.LabelsSet, nil
-	})
+		return connect.NewResponse(&querierv1.SeriesResponse{
+			LabelsSet: lo.UniqBy(
```
simonswine, line 224: UniqBy does not sort the slice (as far I can tell). The Head querier returned sorted labels, I think that would be better for consistency.

pkg/frontend/frontend_series.go
```diff
@@ -2,15 +2,46 @@ package frontend
...
+	tenantIDs, err := tenant.TenantIDs(ctx)
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, err)
+	}
+
+	if c.Msg.Start == 0 || c.Msg.End == 0 {
+		const defaultInterval = 4 * time.Hour
+		now := model.Now()
+		c.Msg.Start = now.Add(-defaultInterval).Unix()
+		c.Msg.End = now.Unix()
```
simonswine, line 29: I would prefer to not do that, ideally we should maintain the legacy behaviour of only  returning `head.Series` result when no Start/End supplied.

pkg/phlaredb/block_querier.go
```diff
@@ -905,6 +938,90 @@ func (b *singleBlockQuerier) SelectMatchingProfiles(ctx context.Context, params
...
+
+	var labelsSets []*typesv1.Labels
+	fingerprints := make(map[uint64]struct{})
+	for _, postings := range postingsIters {
+		for postings.Next() {
+			labels := phlaremodel.Labels{}
+			chunks := []index.ChunkMeta{} // Ignored.
```
simonswine, line 1002: allocs: this should only be allocated once

pkg/phlaredb/block_querier.go
```diff
@@ -905,6 +938,90 @@ func (b *singleBlockQuerier) SelectMatchingProfiles(ctx context.Context, params
...
+	labelNamesFilter := make(map[string]struct{}, len(params.LabelNames))
+	for _, n := range params.LabelNames {
+		labelNamesFilter[n] = struct{}{}
+	}
+	names = lo.Filter(names, func(name string, _ int) bool {
+		_, ok := labelNamesFilter[name]
+		return ok
+	})
+
+	var postingsIters []iter.SeekIterator[storage.SeriesRef, storage.SeriesRef]
```
simonswine, line 973: I would iterate through the postings right away, rather than keep the iterators open and do it later.

pkg/phlaredb/block_querier.go
```diff
@@ -767,6 +769,37 @@ func MergeProfilesPprof(ctx context.Context, stream *connect.BidiStream[ingestv1
...
 
+func Series(ctx context.Context, req *ingestv1.SeriesRequest, blockGetter BlockGetter) (*ingestv1.SeriesResponse, error) {
+	queriers, err := blockGetter(ctx, model.TimeFromUnix(req.Start), model.TimeFromUnix(req.End))
+	if err != nil {
+		return nil, err
+	}
+
+	var labelsSet []*typesv1.Labels
+	var lock sync.Mutex
+	group, ctx := errgroup.WithContext(ctx)
```
simonswine, line 780: We probably should bound the amount of parallelism  with some constant.

pkg/phlaredb/block_querier.go
```diff
@@ -767,6 +769,37 @@ func MergeProfilesPprof(ctx context.Context, stream *connect.BidiStream[ingestv1
...
+			lock.Unlock()
+			return nil
+		}))
+	}
+	err = group.Wait()
+	if err != nil {
+		return nil, err
+	}
+
+	return &ingestv1.SeriesResponse{LabelsSet: labelsSet}, nil
```
simonswine, line 800: This will contain many duplicates, I would sugest at least here we should order and deduplicate. Maybe even at every step.

pkg/querier/querier.go
bryanhuhta: Fixed in https://github.com/grafana/pyroscope/pull/2431/commits/59f4ed20bfd55a39ac4df2204e3708fe751c9355

pkg/phlaredb/block_querier_test.go
```diff
@@ -175,3 +177,148 @@ func TestSelectMatchingProfilesCleanUp(t *testing.T) {
...
+			"pyroscope_spy",
+			"service_name",
+			"target",
+			"version",
+		}
+		got, err := q.index.LabelNames()
+		assert.NoError(t, err)
+		assert.Equal(t, want, got)
+	})
+
```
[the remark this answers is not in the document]
bryanhuhta: Added in https://github.com/grafana/pyroscope/pull/2431/commits/bd1c70d9d99a8eded9f845c415711887843dc5d8

pkg/phlaredb/block_querier.go
```diff
@@ -905,6 +938,90 @@ func (b *singleBlockQuerier) SelectMatchingProfiles(ctx context.Context, params
...
+	}
+
+	labelNamesFilter := make(map[string]struct{}, len(params.LabelNames))
+	for _, n := range params.LabelNames {
+		labelNamesFilter[n] = struct{}{}
+	}
+	names = lo.Filter(names, func(name string, _ int) bool {
+		_, ok := labelNamesFilter[name]
+		return ok
+	})
```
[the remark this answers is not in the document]
bryanhuhta: Fixed in https://github.com/grafana/pyroscope/pull/2431/commits/bd1c70d9d99a8eded9f845c415711887843dc5d8

pkg/frontend/frontend_series.go
bryanhuhta: I reverted this in https://github.com/grafana/pyroscope/pull/2431/commits/f508ad37d627a4ccecd1193786544dcad2ee5907

pkg/phlaredb/block_querier.go
bryanhuhta: I set a parallelism constant to 50 in https://github.com/grafana/pyroscope/pull/2431/commits/28d98bb50fc695d4c89e2285143ae5ed953778c6. 50 is only a SWAG.

pkg/phlaredb/block_querier.go
bryanhuhta: Fixed in https://github.com/grafana/pyroscope/pull/2431/commits/a9dc85efb2565b51c8affb9b825b4c55dcfe0e66

bryanhuhta:
After implementing some of the suggestions raised by @kolesnikovae and @simonswine, here are the new performance metrics:

| Interval | Old avg. (ms) | New (v1) avg. (ms) | New (current) avg (ms) |
|--------:|--------:|----------:|----------:|
| 1h     | 215.35 | 177.85   | 118.00   |
| 6h     | 218.00 | 253.25   | 141.97   |
| 24h    | 200.45 | 2837.50  | 506.45  |
| 2d     | 183.35 | 5199.00  | 938.38 |
| 7d     | 195.60 | 15423.00 | 2400.54 |
| 14d    | 196.15 | 27824.16 | 4908.47 |

chart (6)

As you can see, there are significant decreases in response time from the first iteration. Even at quite long intervals, the response time hovers around 5s (while not optimal, still manageable). More typical intervals have an adequate request duration, broadly staying under 1s (until it's longer than 2d).

@kolesnikovae had a number of other [suggestions](https://github.com/grafana/pyroscope/pull/2431#issuecomment-1736653067) we could pursue to improve performance, but at this point, I think these numbers are healthy enough to ship and iterate on.

kolesnikovae:
Great work @bryanhuhta! I think you're right, we should merge it now, and only optimise it if we see the actual need

pkg/phlaredb/block_querier.go
```diff
@@ -767,6 +769,48 @@ func MergeProfilesPprof(ctx context.Context, stream *connect.BidiStream[ingestv1
...
+	var lock sync.Mutex
+	group, ctx := errgroup.WithContext(ctx)
+
+	// TODO(bryan) Verify this limit is ok
+	const concurrentQueryLimit = 50
+	group.SetLimit(concurrentQueryLimit)
+
+	for _, q := range queriers {
+		group.Go(util.RecoverPanic(func() error {
+			labels, err := q.Series(ctx, req)
```
cyriltovena, line 788: if we change that API to send back finguerprint ID we should gain even more perf boost

cyriltovena (approved):
LGTM 

something to investigate if you could use later https://github.com/grafana/pyroscope/blob/ac9da904ac2998f5231390aecc449a7e353a901b/pkg/phlaredb/tsdb/index/index.go#L1738

pkg/frontend/frontend_series.go
```diff
@@ -4,13 +4,38 @@ import (
...
+
+	tenantIDs, err := tenant.TenantIDs(ctx)
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, err)
+	}
+
+	if c.Msg.Start != 0 && c.Msg.End != 0 {
+		interval := model.Interval{
+			Start: model.TimeFromUnix(c.Msg.Start),
+			End:   model.TimeFromUnix(c.Msg.End),
```
simonswine, line 27: In order to be consistent with the rest of the api this needs to be milliseconds, something like that:

```suggestion
			Start: model.Time(int64(c.Msg.Start)),
			End: model.Time(int64(c.Msg.End)),
```

pkg/frontend/frontend_series.go
```diff
@@ -4,13 +4,38 @@ import (
...
+		}
+		validated, err := validation.ValidateRangeRequest(f.limits, tenantIDs, interval, model.Now())
+		if err != nil {
+			return nil, connect.NewError(connect.CodeInvalidArgument, err)
+		}
+		if validated.IsEmpty {
+			return connect.NewResponse(&querierv1.SeriesResponse{}), nil
+		}
+		c.Msg.Start = validated.Start.Unix()
+		c.Msg.End = validated.End.Unix()
```
simonswine, line 37: Also needs to stay in milliseconds

simonswine (approved):
LGTM. Thanks @bryanhuhta for you patience going though all of our feedback. 

Before you merge (sorry), I would like you to change the start/stop time unit from seconds to milliseconds, as the other query endpoints use.

pkg/frontend/frontend_series.go
bryanhuhta: I fixed all the timestamps in https://github.com/grafana/pyroscope/pull/2431/commits/47fca7040ae2f850364eaf1da82c06a7ad849630.

pkg/frontend/frontend_series.go
bryanhuhta: I fixed all the timestamps in https://github.com/grafana/pyroscope/pull/2431/commits/47fca7040ae2f850364eaf1da82c06a7ad849630.

PR #2432, opened 2023-09-21
fix(helm): set correct namespace for all resources

jhedev:
This PR adds namespace values for each resource created by the helm chart. Previously, everything would be installed in the default namespace, e.g. `helm -n pyroscope-test install pyroscope grafana/pyroscope` would not install into the `pyroscope-test` namespace. 

Not sure if there was a specfic reason for this.

### DIFF
--- a/operations/pyroscope/helm/pyroscope/rendered/micro-services.yaml
+++ b/operations/pyroscope/helm/pyroscope/rendered/micro-services.yaml
@@ -4,6 +4,7 @@ apiVersion: policy/v1
 kind: PodDisruptionBudget
 metadata:
   name: pyroscope-dev-distributor
+  namespace: default
   labels:
     helm.sh/chart: pyroscope-1.0.3
     app.kubernetes.io/name: pyroscope
@@ -24,6 +25,7 @@ apiVersion: policy/v1
 kind: PodDisruptionBudget
 metadata:
   name: pyroscope-dev-ingester
+  namespace: default
   labels:
     helm.sh/chart: pyroscope-1.0.3
     app.kubernetes.io/name: pyroscope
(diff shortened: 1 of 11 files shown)


simonswine:
Are you actually facing an issue with this. Just tried it myself with helm  `version.BuildInfo{Version:"v3.11.3", GitCommit:"v3.11.3", GitTreeState:"", GoVersion:"go1.20.7"}`:

```
$ kubectl create ns test-ns
$ helm install --namespace test-ns pyroscope grafana/pyroscope
$ kubectl get pods -n test-ns
NAME                READY   STATUS    RESTARTS   AGE
pyroscope-0         1/1     Running   0          87s
pyroscope-agent-0   2/2     Running   0          87s
```

jhedev:
It looks like it is working for `helm install` but not for `helm template`. I mostly use them with `helm template` and apply the files with `kubectl`. 

When I run `helm template` the resources do not have the right namespace set:

```shell
$ helm -n pyroscope-test template pyroscope grafana/pyroscope
...
# Source: pyroscope/templates/deployments-statefulsets.yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: pyroscope
  labels:
    helm.sh/chart: pyroscope-1.0.3
    app.kubernetes.io/name: pyroscope
    app.kubernetes.io/instance: pyroscope
    app.kubernetes.io/version: "1.0.0"
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/component: "all"
spec:
...
$ helm version
version.BuildInfo{Version:"v3.12.3", GitCommit:"3a31588ad33fe3b89af5a2a54ee1d25bfe6eaa5e", GitTreeState:"clean", GoVersion:"go1.20.7"}
```

cyriltovena (approved):
LGTM

PR #2433, opened 2023-09-21
f32 fixes

korniltsev:
- enable logging in the /ingest - it was NoOpLogger
- fix division by zero panic leading to 422 for a jfr ingested with sampleRate=0 parameter
- detect old dotnetspy by looking into  sample type config. (these pprofs have function.id = 0, which lead to "function id is 0" error from the distributor validation
- do not try to push to distributor if profile has zero series (there is some tenant with misconfigured jfr producing zero profiles)

### DIFF
--- a/pkg/og/convert/jfr/pprof.go
+++ b/pkg/og/convert/jfr/pprof.go
@@ -35,11 +35,17 @@ const (
 func newJfrPprofBuilders(p *parser.Parser, jfrLabels *LabelsSnapshot, piOriginal *storage.PutInput) *jfrPprofBuilders {
 	st := piOriginal.StartTime.UnixNano()
 	et := piOriginal.EndTime.UnixNano()
+	var period int64
+	if piOriginal.SampleRate == 0 {
+		period = 0
+	} else {
+		period = 1e9 / int64(piOriginal.SampleRate)
+	}
 	res := &jfrPprofBuilders{
 		timeNanos:     st,
 		durationNanos: et - st,
 		labels:        make([]*v1.LabelPair, 0, len(piOriginal.Key.Labels())+5),
-		period:        1e9 / int64(piOriginal.SampleRate),
+		period:        period,
 		appName:       piOriginal.Key.AppName(),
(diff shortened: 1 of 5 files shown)


pkg/og/convert/jfr/pprof.go
```diff
@@ -35,11 +35,17 @@ const (
 func newJfrPprofBuilders(p *parser.Parser, jfrLabels *LabelsSnapshot, piOriginal *storage.PutInput) *jfrPprofBuilders {
 	st := piOriginal.StartTime.UnixNano()
 	et := piOriginal.EndTime.UnixNano()
+	var period int64
+	if piOriginal.SampleRate == 0 {
+		period = 0
```
korniltsev, line 40: div by zero if jfr is ingested with sampleRate=0 param

pkg/og/convert/pprof/profile.go
```diff
@@ -82,6 +82,14 @@ func (p *RawProfile) ParseToPprof(_ context.Context, md ingestion.Metadata) (res
 	return
 }
 
+func (p *RawProfile) isDotnetspy(md ingestion.Metadata) bool {
+	if md.SpyName == "dotnetspy" {
+		return true
+	}
+	stc := p.getSampleTypes()
+	return stc != nil && stc["inuse-space"] != nil
```
korniltsev, line 90: slightly better dotnetspy detection, afaik "inuse-space" was only used in dotnet

pkg/phlare/phlare.go
```diff
@@ -492,7 +492,7 @@ func initLogger(logFormat string, logLevel dslog.Level) log.Logger {
 }
 
 func (f *Phlare) initAPI() (services.Service, error) {
-	a, err := api.New(f.Cfg.API, f.Server, f.grpcGatewayMux, util.Logger)
```
korniltsev, line 495: util.Logger is NoOplogger, we should rename it or remove alltogether

korniltsev:
new branch https://github.com/grafana/pyroscope/pull/2434

PR #2434, opened 2023-09-21
Fix ingestionAdapter bugs surfaced in f32

korniltsev:
-    enable logging in the /ingest - it was NoOpLogger
 -   fix division by zero panic leading to 422 for a jfr ingested with sampleRate=0 parameter
 -   detect old dotnetspy by looking into sample type config. (these pprofs have function.id = 0, which lead to "function id is 0" error from the distributor validation
-    do not try to push to distributor if profile has zero series (there is some tenant with misconfigured jfr producing zero profiles)


pkg/ingester/pyroscope/ingest_adapter.go
```diff
@@ -164,6 +167,13 @@ func (p *pyroscopeIngesterAdapter) parseToPprof(ctx context.Context, in *ingesti
 	if err != nil {
 		return fmt.Errorf("parsing IngestInput-pprof failed %w", err)
 	}
+	if len(plainReq.Series) == 0 {
+		tenantID, _ := tenant.ExtractTenantIDFromContext(ctx)
```
korniltsev, line 171: do not try to push pprofs to distributor if there no pprofs (for exampel misconfigured jfr with zero samples)

pkg/og/convert/pprof/profile.go
```diff
@@ -82,6 +82,14 @@ func (p *RawProfile) ParseToPprof(_ context.Context, md ingestion.Metadata) (res
 	return
 }
 
+func (p *RawProfile) isDotnetspy(md ingestion.Metadata) bool {
+	if md.SpyName == "dotnetspy" {
+		return true
+	}
+	stc := p.getSampleTypes()
+	return stc != nil && stc["inuse-space"] != nil
```
korniltsev, line 90: try to detect old dotnetspy. afaik "inuse-space" was used only in dotnet.

korniltsev: maybe I should only use inuse-space if SPyName==""

pkg/phlare/phlare.go
```diff
@@ -492,7 +492,7 @@ func initLogger(logFormat string, logLevel dslog.Level) log.Logger {
 }
 
 func (f *Phlare) initAPI() (services.Service, error) {
-	a, err := api.New(f.Cfg.API, f.Server, f.grpcGatewayMux, util.Logger)
```
korniltsev, line 495: this util.Logger is NoOpLogger

simonswine: That's annoying, thanks for spotting

pkg/og/convert/pprof/profile.go
simonswine: I think the SpyName == ""  check might be useful.

pkg/phlare/phlare.go
korniltsev: we should rename util.Logger to NoOpLoggerPleasNoUseIt or  remove it. there's still 27 usages of it
Screenshot 2023-09-21 at 7 27 52 PM

pkg/ingester/pyroscope/ingest_handler.go
```diff
@@ -46,17 +46,17 @@ func NewIngestHandler(l log.Logger, p ingestion.Ingester) http.Handler {
...
+		_ = h.log.Log("msg", "bad request", "err", err, "orgID", tenantID)
 		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
 
-	tenantID, _ := tenant.ExtractTenantIDFromContext(r.Context())
 	err = h.ingester.Ingest(r.Context(), input)
 	if err != nil {
-		_ = h.log.Log("msg", "pyroscope ingest", "err", err, "tenant_id", tenantID)
+		_ = h.log.Log("msg", "pyroscope ingest", "err", err, "orgID", tenantID)
```
simonswine, line 59: Would give it an error level

```suggestion
		_ = level.Error(h.log).Log("msg", "pyroscope ingest", "err", err, "orgID", tenantID)
```

pkg/phlare/phlare.go
simonswine: Yeah Mimir uses that global, can you create an issues to try to get rid of it.

simonswine:
Long term, we need some tests acutally trying to ingest profiles of that type (e.g. with sampleRate 0)

korniltsev:
> Long term, we need some tests acutally trying to ingest profiles of that type (e.g. with sampleRate 0)
Totally agree. I will add e2e testcases tomorrow

pkg/ingester/pyroscope/ingest_handler.go
korniltsev: the thing is that the h.log is already wrapped with level.Error in the constructor function :shrug:

pkg/ingester/pyroscope/ingest_handler.go
```diff
@@ -46,17 +46,17 @@ func NewIngestHandler(l log.Logger, p ingestion.Ingester) http.Handler {
 }
 
 func (h ingestHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	tenantID, _ := tenant.ExtractTenantIDFromContext(r.Context())
 	input, err := h.ingestInputFromRequest(r)
 	if err != nil {
-		_ = h.log.Log("msg", "bad request", "err", err)
+		_ = h.log.Log("msg", "bad request", "err", err, "orgID", tenantID)
```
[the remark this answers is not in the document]
korniltsev: https://github.com/grafana/pyroscope/pull/2434#discussion_r1332912733

pkg/phlare/phlare.go
korniltsev: Yeah already created 
https://github.com/grafana/pyroscope/issues/2435

> eah Mimir uses that globa
o_O mimir? it is pyroscope/pkg/util/logger.go

simonswine: 😆 we obviously copy pasted it all from mimir

pkg/ingester/pyroscope/ingest_handler.go
simonswine: Then forget about it 🙂

simonswine:
> OK, will do tomorrow, together with e2e reproducers

Let's do main first, otherwise it is the recipe of regression for f33

pkg/og/convert/pprof/profile.go
korniltsev: added

Krithika3:
@korniltsev @simonswine I am seeing 422 even when the samplerate is 100
/ingest?name=my-project-java&units=samples&aggregationType=sum&sampleRate=100&from=1695444827&until=1695444842&spyName=javaspy&format=jfr (422) 15.138099ms"

korniltsev:
Create a separate issue. Provide your configuration, debug logs. Your versions of sdk and and server. Share an example of jfr causing 422.

korniltsev:
Update your server to the latest version, you will have more error logs. 

Update your java sdk to the latest version - it will handle retries better (it will not retry 422)

One of the reasons you can get 422 is out if order ingestion - if you have multiple clients sending same series at the same time. This could happen if one client sends profiles for 10:00:10 period of time and then the other sends for 10:00:00, this one will be rejected with 422. If this is actually the reason you get 422 then you could fix it by adding an unique per client label, for example a pod name or hostname or something else unique. You can set the label in many ways from env variables, system props or from java code - see documentation. 

This is just a guess. As I said above - create a separate issue and provide information.

Krithika3:
Sure, I can create a separate issue

Krithika3:
Created https://github.com/grafana/pyroscope/issues/2447

PR #2441, opened 2023-09-22
run make tools/update_examples

korniltsev:


cyriltovena (approved):
LGTM you need to `make go/mod`

PR #2449, opened 2023-09-23
feature: prefer cpu app default if none selected

darrenjaneczek:
- If no app is selected, choose the first cpu option as default
- If it can't find a cpu app, just choose the first as it did before


simonswine (approved):
LGTM

PR #2452, opened 2023-09-25
Distributor/Ingestor: Replace out of order validation with ingestion window

simonswine:
This change removes of order validation and instead make sure the ingested profiles are within the ingestion window, which will by default reach from now-1h to now+10m.

This will ensure a smoother experience with workloads that don't have unique labels sets

### DIFF
--- a/cmd/pyroscope/help-all.txt.tmpl
+++ b/cmd/pyroscope/help-all.txt.tmpl
@@ -899,5 +899,9 @@ Usage of ./pyroscope:
     	Maximum number of samples in a profile. 0 to disable. (default 16000)
   -validation.max-profile-symbol-value-length int
     	Maximum length of a profile symbol value (labels, function names and filenames, etc...). Profiles are not rejected instead symbol values are truncated. 0 to disable. (default 65535)
+  -validation.reject-newer-than duration
+    	This limits how far into the future profiling data can be ingested. This limit is enforced in the distributor. 0 to disable, defaults to 10m. (default 10m)
+  -validation.reject-older-than duration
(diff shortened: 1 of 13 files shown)


kolesnikovae:
I'm wondering if we should do anything about the in-memory profiles order, because I think this is required for the deduplication. I see that profiles [are sorted](https://github.com/grafana/pyroscope/blob/3d9b990b0afca84aafbb0b7f3d37f2c5740bf264/pkg/phlaredb/profile_store.go#L357) when we flush a row group on disk, but I couldn't find if we sort in-memory profiles explicitly. I expected to find it [here](https://github.com/grafana/pyroscope/blob/3d9b990b0afca84aafbb0b7f3d37f2c5740bf264/pkg/phlaredb/head_queriers.go#L201), but maybe I don't fully understand how it works.

korniltsev (approved):
looks good

~should we remove delta as well?~ just read the design doc

simonswine:
> I'm wondering if we should do anything about the order of the profiles in memory because I think it's necessary for deduplication. I see that profiles [are sorted](https://github.com/grafana/pyroscope/blob/3d9b990b0afca84aafbb0b7f3d37f2c5740bf264/pkg/phlaredb/profile_store.go#L357) when we flush a row group on disk, but I couldn't find if we sort in-memory profiles explicitly at query time. I expected to find it [here](https://github.com/grafana/pyroscope/blob/3d9b990b0afca84aafbb0b7f3d37f2c5740bf264/pkg/phlaredb/head_queriers.go#L201), but maybe I don't fully understand how it works.

@kolesnikovae you have raised a good point, I have added a binary search for the cases when profiles get added out of order.

cyriltovena (approved):
LGTM

Yep we need to keep the in memory sorted.

PR #2454, opened 2023-09-26
Hardcode grafana agent version

simonswine:
This will make sure grafana agent only updates, when we like it to.


Rperry2174 (approved):
lgtm

PR #2458, opened 2023-09-27
Tracing integration API

kolesnikovae:
The PR contains a new API for querying profile samples with a filter on span identifiers. Only in-memory profile representation is affected, the persistence layer is not changed.

---

It's a bit clunky, but the new API can be tested manually:

https://github.com/grafana/pyroscope/assets/12090599/46b5560b-003b-4009-8767-0ee53833d06b


pkg/phlaredb/schemas/v1/profiles.go
```diff
@@ -246,6 +247,9 @@ type InMemoryProfile struct {
 type Samples struct {
 	StacktraceIDs []uint32
 	Values        []uint64
+	// Span associated with samples.
+	// Optional: Spans == nil, if not present.
+	Spans []uint64
```
cyriltovena, line 252: Nice !

cyriltovena (approved):
LGTM

PR #2459, opened 2023-09-27
doc: use correct memory profile endpoint in Go pull example

mlcdf:
Hello there,

I believe the memory profile is available at `/debug/pprof/heap` and not `/debug/pprof/memory` on Golang applications.

This PR fixes a doc that would suggest otherwise (got stuck for a while on this haha).

https://cs.opensource.google/go/go/+/refs/tags/go1.21.1:src/runtime/pprof/pprof.go;l=189


simonswine (approved):
LGTM.

Sorry for that unnecessary obstacle and thanks for contributing!

mlcdf:
No worries. Have a great day!

PR #2461, opened 2023-09-28
Add script to test packaging artifacts

simonswine:
This will make it easier to validate RPM/DEB pacakges.

I noticed a problem with the systemd unit after install: Basically the data directory was not created and also not specified correctly.

With this it will use `/var/lib/pyroscope` and create it as part of the `./postinstall`

cc @wilfriedroset


cyriltovena (approved):
Following @wilfriedroset LGTM I have no idea what's going on :)

PR #2465, opened 2023-09-29
Disable protobuf pools for all but profiles

### LINKED ISSUE #2457: Disable VT pool for pprof samlpes
We should consider disabling pool for pprof Samples as it might be more expensive that allocating new ones. 

https://flamegraph.com/share/b9ba6057-5d12-11ee-b135-0a855ce7b751
image

kolesnikovae:
I would prefer the PR merged after the next weekly release.

Resolves #2457. I'll attach a flamegraph-diff once I get some representative data but the gist is that GC overhead will likely be less that the pool overhead (~15% of the total CPU time). 

https://github.com/grafana/pyroscope/assets/12090599/49ca2f13-cca5-425f-a623-bed2d1427a1f

### DIFF
--- a/api/buf.gen.yaml
+++ b/api/buf.gen.yaml
@@ -17,9 +17,6 @@ plugins:
       - paths=source_relative
       - features=marshal+unmarshal+size+pool+grpc+pool+clone
       - pool=github.com/grafana/pyroscope/api/gen/proto/go/google/v1.Profile
-      - pool=github.com/grafana/pyroscope/api/gen/proto/go/google/v1.Sample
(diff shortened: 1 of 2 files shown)


cyriltovena (approved):
Sounds good let's give it go when you're ready.

kolesnikovae:
Okay, we got -10% of total CPU time in `ResetVT`:
image
 
And +3% of total CPU time cause by pressure on GC:
image

https://flamegraph.com/share/0e1f6a48-61ad-11ee-b135-0a855ce7b751

5-7% effortless improvement

PR #2466, opened 2023-09-29
feat: Scalable Compactor

cyriltovena:
This PR introduces a scalable compactor component that can split and merge Pyroscope blocks to improve performance and deduplicate replication factor.

This is heavily inspired by Mimir compactor but adapted for Pyroscope. One of the main difference for instance is that Pyroscope maxT are inclusive.

The design and behaviour is documented here https://grafana.com/docs/mimir/latest/references/architecture/components/compactor/

Splitting is optional and only required for large tenants where output blocks would be too big.

## Remaining tasks:
- [x] Add the compactor component (right now only the package is there)
- [x] Update the store-gateway to only download compacted blocks and use the block index file.
- [x] Tweak the configuration to fit Pyroscope needs (specially the first range is 2h where our blocks is 3h by default)
- [x] Test in a real environment and iterate on it.
  - [x] Test vertical compaction
  - [x] Test split and merge compaction
- [x] Document operations https://github.com/grafana/pyroscope/issues/2540

## Previous related PR:
- https://github.com/grafana/pyroscope/pull/2401
- https://github.com/grafana/pyroscope/pull/2397
- https://github.com/grafana/pyroscope/pull/2396
- https://github.com/grafana/pyroscope/pull/2391
- https://github.com/grafana/pyroscope/pull/2388
- https://github.com/grafana/pyroscope/pull/2384
- https://github.com/grafana/pyroscope/pull/2381
- https://github.com/grafana/pyroscope/pull/2366
- https://github.com/grafana/pyroscope/pull/2184
- https://github.com/grafana/pyroscope/pull/2178
- https://github.com/grafana/pyroscope/pull/2176
- https://github.com/grafana/pyroscope/pull/2134
- https://github.com/grafana/pyroscope/pull/2467
- https://github.com/grafana/pyroscope/pull/2521

## Testing Result
See how block were before level 1 not aligned, now you can see the 8h range being compacted.
image


pkg/compactor/compactor.go
```diff
@@ -118,7 +118,7 @@ type Config struct {
 func (cfg *Config) RegisterFlags(f *flag.FlagSet, logger log.Logger) {
 	cfg.ShardingRing.RegisterFlags(f, logger)
 
-	cfg.BlockRanges = DurationList{2 * time.Hour, 12 * time.Hour, 24 * time.Hour}
+	cfg.BlockRanges = DurationList{4 * time.Hour}
```
cyriltovena, line 121: I will start with this but tweak it as we gain operational experience.

cmd/pyroscope/help-all.txt.tmpl
```diff
@@ -4,13 +4,137 @@ Usage of ./pyroscope:
...
+  -compactor.compaction-concurrency int
+    	Max number of concurrent compactions running. (default 1)
+  -compactor.compaction-interval duration
+    	The frequency at which the compaction runs (default 1h0m0s)
+  -compactor.compaction-jobs-order string
+    	The sorting to use when deciding which compaction jobs should run first for a given tenant. Supported values are: smallest-range-oldest-blocks-first, newest-blocks-first. (default "smallest-range-oldest-blocks-first")
+  -compactor.compaction-retries int
+    	How many times to retry a failed compaction within a single compaction run. (default 3)
+  -compactor.compaction-split-by string
+    	The strategy to use when splitting blocks during compaction. Supported values are: fingerprint, stacktracePartition. (default "fingerprint")
```
simonswine, line 37: Do we really want to expose this?

cmd/pyroscope/help-all.txt.tmpl
```diff
@@ -4,13 +4,137 @@ Usage of ./pyroscope:
...
+  -compactor.compactor-tenant-shard-size int
+    	Max number of compactors that can compact blocks for single tenant. 0 to disable the limit and use all compactors.
+  -compactor.data-dir string
+    	Directory to temporarily store blocks during compaction. This directory is not required to be persisted between restarts. (default "/data")
```
simonswine, line 41:
```suggestion
    	Directory to temporarily store blocks during compaction. This directory is not required to be persisted between restarts. (default "./data-compactor")
```

I don't really want it to clash with the ingester local folder. Also that one is configure relative to the working dir.

cmd/pyroscope/help-all.txt.tmpl
```diff
@@ -4,13 +4,137 @@ Usage of ./pyroscope:
...
+  -compactor.ring.wait-active-instance-timeout duration
+    	Timeout for waiting on compactor to become ACTIVE in the ring. (default 10m0s)
+  -compactor.ring.wait-stability-max-duration duration
+    	Maximum time to wait for ring stability at startup. If the compactor ring keeps changing after this period of time, the compactor will start anyway. (default 5m0s)
+  -compactor.ring.wait-stability-min-duration duration
+    	Minimum time to wait for ring stability at startup. 0 to disable.
+  -compactor.split-and-merge-shards int
+    	The number of shards to use when splitting blocks. 0 to disable splitting.
+  -compactor.split-groups int
+    	Number of groups that blocks for splitting should be grouped into. Each group of blocks is then split separately. Number of output split shards is controlled by -compactor.split-and-merge-shards. (default 1)
```
simonswine, line 135: Can we mention that split-groups should be >= split-merge-shards?

cmd/pyroscope/help-all.txt.tmpl
```diff
@@ -4,13 +4,137 @@ Usage of ./pyroscope:
...
+  -compactor.tenant-cleanup-delay duration
```
simonswine, line 136: I don't think we support tenant deleteion or do we?

cmd/pyroscope/help-all.txt.tmpl
```diff
@@ -406,7 +530,7 @@ Usage of ./pyroscope:
   -querier.max-query-parallelism int
     	Maximum number of queries that will be scheduled in parallel by the frontend.
   -querier.query-store-after duration
-    	The time after which a metric should be queried from storage and not just ingesters. 0 means all queries are sent to store. If this option is enabled, the time range of the query sent to the store-gateway will be manipulated to ensure the query end is not more recent than 'now - query-store-after'. (default 4h0m0s)
+    	The time after which a metric should be queried from storage and not just ingesters. 0 means all queries are sent to store. If this option is enabled, the time range of the query sent to the store-gateway will be manipulated to ensure the query end is not more recent than 'now - query-store-after'. (default 12h0m0s)
```
simonswine, line 533: Can you explain this change?

pkg/compactor/compactor.go
```diff
@@ -0,0 +1,882 @@
...
+	f.BoolVar(&cfg.NoBlocksFileCleanupEnabled, "compactor.no-blocks-file-cleanup-enabled", false, "If enabled, will delete the bucket-index, markers and debug files in the tenant bucket when there are no blocks left in the index.")
+	// compactor concurrency options
+	f.IntVar(&cfg.MaxOpeningBlocksConcurrency, "compactor.max-opening-blocks-concurrency", 1, "Number of goroutines opening blocks before compaction.")
+
+	f.Var(&cfg.EnabledTenants, "compactor.enabled-tenants", "Comma separated list of tenants that can be compacted. If specified, only these tenants will be compacted by compactor, otherwise all tenants can be compacted. Subject to sharding.")
+	f.Var(&cfg.DisabledTenants, "compactor.disabled-tenants", "Comma separated list of tenants that cannot be compacted by this compactor. If specified, and compactor would normally pick given tenant for compaction (via -compactor.enabled-tenants or sharding), it will be ignored instead.")
+}
+
+func (cfg *Config) Validate() error {
+	// Each block range period should be divisible by the previous one.
```
simonswine, line 153: I would also check that the first block range is a divisible for MaxBlockDuration.

cmd/pyroscope/help-all.txt.tmpl
```diff
@@ -4,13 +4,137 @@ Usage of ./pyroscope:
...
     	Directory to store synchronized pyroscope block headers. This directory is not required to be persisted between restarts, but it's highly recommended in order to improve the store-gateway startup time. (default "./data/pyroscope-sync/")
   -blocks-storage.bucket-store.sync-interval duration
     	How frequently to scan the bucket, or to refresh the bucket index (if enabled), in order to look for changes (new blocks shipped by ingesters and blocks deleted by retention or compaction). (default 15m0s)
   -blocks-storage.bucket-store.tenant-sync-concurrency int
     	Maximum number of concurrent tenants synching blocks. (default 10)
+  -compactor.block-ranges value
+    	List of compaction time ranges. (default 3h0m0s,6h0m0s,12h0m0s)
+  -compactor.block-sync-concurrency int
+    	Number of Go routines to use when downloading blocks for compaction and uploading resulting blocks. (default 8)
+  -compactor.blocks-retention-period duration
```
simonswine, line 22: Is that working? if not we should remove it

pkg/phlaredb/compact.go
```diff
@@ -417,9 +417,6 @@ func compactMetas(src ...block.Meta) block.Meta {
 			labels[k] = v
 		}
 	}
-	if hostname, err := os.Hostname(); err == nil {
-		labels[block.HostnameLabel] = hostname
-	}
 	meta.Source = block.CompactorSource
 	meta.Compaction = tsdb.BlockMetaCompaction{
 		Deletable: false,
```
simonswine, line 422: Again wrong line, look more 404/405. The MaxTime and MinTime is not consistent with the one of the block itself. 

While we use integer milliseconds for parents, we use a floating point number for MaxTime, MinTime for the block itself. We should fix that:

```
{
        "ulid": "01HCW5635RGVR5DV2HEQJNPXCQ",
        "minTime": 1697316520.195,
        "maxTime": 1697317199.969,
[...]
        "compaction": {
                "parents": [
                        {
                                "ulid": "01HCQZ5HA3DHKSJPK5EGJ0YKDK",
                                "minTime": 1697316520195,
                                "maxTime": 1697316599969
                        },
[...]
```

simonswine (commented):
I have not made it through every line of this PR, but I was able to get it running on a local cluster.

cmd/pyroscope/help-all.txt.tmpl
cyriltovena: We can remove it later, the idea is to allow experimentation, @kolesnikovae also want to try by service name.

cmd/pyroscope/help-all.txt.tmpl
cyriltovena: Nope we don't yet

cmd/pyroscope/help-all.txt.tmpl
cyriltovena: Not sure it's intentional, I would actually prefer 4h for the future.

cmd/pyroscope/help-all.txt.tmpl
cyriltovena: Didn't tried it yet but it should, it's pretty simple, on compaction run it mark block outside of retention deleted, the cleaner will eventually finish it.

cyriltovena: It's disabled by default btw.

cmd/pyroscope/help-all.txt.tmpl
cyriltovena: Commenting it.

cmd/pyroscope/help-all.txt.tmpl
cyriltovena: I'll add an experimental note.

cmd/pyroscope/help-all.txt.tmpl
cyriltovena: Yeah this was a hack for not having to handle helm, let me see what I can do.

pkg/compactor/compactor.go
cyriltovena: I did it but it's a bit inconsistent with ingesters now, their ULID is not based on the minTime but when the head was creared.

pkg/phlaredb/compact.go
cyriltovena: Made the change, I had to fork the compaction meta struct from tsdb.

pkg/phlaredb/compact.go
```diff
@@ -417,9 +417,6 @@ func compactMetas(src ...block.Meta) block.Meta {
 			labels[k] = v
 		}
 	}
-	if hostname, err := os.Hostname(); err == nil {
```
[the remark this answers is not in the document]
cyriltovena: Done, not this is a bit inconsistent with ingester behaviour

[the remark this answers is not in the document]
simonswine: Yeah that's true, maybe we should the minTime of the maxBlockDuration?  So something like `(time.Now() / maxBlockDuration) * maxBlockDuration`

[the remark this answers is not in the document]
cyriltovena: It's ok I did the change now, for ingester block it means the ULID should be close to minT in most case. We can figure something later.

cmd/pyroscope/help-all.txt.tmpl
cyriltovena: It does not need to be, actually should be smaller or equal, splitgroup is only used during split stage.

It means first create n (split-groups) group from src blocks, each group is a compaction with the split factor.

So you end up with n*m blocks. This is mostly when the first range has too many block and loading them all at once will be pressuring memory.

cyriltovena: The next merge stage will group then by shard ID and so each group will have `n` (split group block)

pkg/phlaredb/block/block_test.go
```diff
@@ -84,7 +84,7 @@ func TestDelete(t *testing.T) {
 
 	bkt := objstore.NewInMemBucket()
 	{
-		meta, dir, err := block_testutil.CreateBlock(t, func() []*testhelper.ProfileBuilder {
```
simonswine, line 87: I think block delete is not deleting the symdb prefix within the block. Can you double check:

```
gsutil ls gs://dev-us-central-0-profiles-dev-001-data/1218/phlaredb/01H9YTWD4JS697B9F2RRYGB9VA/symbols
gs://dev-us-central-0-profiles-dev-001-data/1218/phlaredb/01H9YTWD4JS697B9F2RRYGB9VA/symbols/index.symdb
gs://dev-us-central-0-profiles-dev-001-data/1218/phlaredb/01H9YTWD4JS697B9F2RRYGB9VA/symbols/stacktraces.symdb
```

cyriltovena: yeah I can see the Delete being not recursive. Let me work on that. Nice catch !

cyriltovena: Can't find any issue I wonder if this is because that cluster is in a crashloop

PR #2467, opened 2023-09-29
Fixes a race detected when generating ULID

cyriltovena:
Stumble on this while working on compactor.

the entropy must be thread safe and the `crypto/rand` is.


simonswine (approved):
LGTM

PR #2469, opened 2023-09-29
Create leaky goroutine pool

bryanhuhta:
This modifies the rideshare example to include leaky goroutines. This should in turn simulate increased memory usage as the goroutines hang and are never cleaned up by the Go runtime.

I haven't tested this yet, so the leaking goroutines may take too long to become noticeable. We could also add a parameter which causes the leaked goroutine to allocate some memory to speed memory allocations.


bryanhuhta:
Using https://github.com/grafana/pyroscope/pull/2610 in favor of this one, as the implementation is more straightforward.

PR #2474, opened 2023-10-02
profilecli: Add query series command

simonswine:
This command allows to query the profile series available from pyroscope. It support querier/ingester/store-gateway APIs.

The results are written to stdout, so they e.g. can be processed further with `jq`.


Rperry2174 (approved):
lgtm

PR #2478, opened 2023-10-03
update rideshare

korniltsev:
- update examples
- make loadgen hit each target concurrently
- update java rideshare example docker to build the app on BUILDPLATFORM with multistage for faster builds
- update java rideshare docker file to use grafana url instead of pyroscope-io
- update  tools/update_examples.go to update rideshare dockerfile


Rperry2174 (approved):
lgtm

Issue #2443, opened 2023-09-22
Low cardinality series distribution


kolesnikovae:
In Pyroscope, profiles distributed evenly among ingesters/shards based on the profile label set (series). However, in some cases where there are not enough unique series in the profile source, this strategy results in skewed data. This also causes out-of-order drop (see https://github.com/grafana/pyroscope/issues/2198).

This makes it practically impossible to use Pyroscope is some scenarios, such as serverless profiling (aka Function as a Service) with [Pyroscope AWS Lambda Extension](https://github.com/grafana/pyroscope-lambda-extension).

We should figure out which distribution strategy might work better for this specific use case. I see the following options:
 - [Preferable] Require the sender to ensure sufficient cardinality by adding labels to profiles:
    - In SDK, we can generate random value at start and add it to the series labels (e.g `generation`/`session`).
    - Cardinality of the token should be limited (ideally it should match the number of shards).
 - Add "shard" labels in the distributors (randomly on based on some attributes), before sending the profile to ingesters:
    - This could be a parameter of the ingestion API (applicaton-wide).
    - Or a configuration parameter (tenant-wise).
    - Or we could determine the need to randomise profile series dynamically with some heuristics.

kolesnikovae:
Implemented in https://github.com/grafana/pyroscope/pull/2480

PR #2480, opened 2023-10-03
Allow `__session_id__` label

kolesnikovae:
Please see https://github.com/grafana/pyroscope/issues/2443: we add a new mechanism that allows increasing series cardinality, which might be useful in certain cases (such as server-less profiling) The mechanism is disabled by default


cyriltovena (approved):
LGTM

PR #2482, opened 2023-10-03
Pass start/end when fetching series

bryanhuhta:

This is the UI component of https://github.com/grafana/pyroscope/pull/2431.


public/app/redux/reducers/continuous/index.ts
```diff
@@ -72,27 +74,34 @@ const initialState: ContinuousState = {
...
-        type: 'danger',
-        title: 'Failed to load app names',
-        message: res.error.message,
-      })
-    );
-
-    return Promise.reject(res.error);
+export const reloadAppNames = createAsyncThunk<
+  App[],
+  null,
```
darrenjaneczek, line 79: If you replace this `null` with `void` you can revert passing `null` in the dispatch calls to `reloadAppNames` above.

public/app/services/apps.ts
```diff
@@ -82,9 +82,10 @@ function removeDuplicateApps(app: App[]) {
   });
 }
 
-export async function fetchApps(): Promise<
-  Result<App[], RequestError | ZodError>
-> {
+export async function fetchApps(
+  start: number,
+  end: number
```
darrenjaneczek, line 87: Not all calls to `fetchApps` are passing these parameters in `apps.spec.ts`
- Do we want to consider making them optional args?
- Will settings the 
- Everywhere else in pyroscope oss, we have `from` and `until` instead of `start` and `end`. Do we want to keep the different terminology exclusive to the POST method body, or do we want it to propagate through the function signature and the variables in calling code.
  > ```  
  >   const start = formatAsOBject(state.continuous.from);
  >   const end = formatAsOBject(state.continuous.until);
  >   const startMs = start.getTime();
  >   const endMs = end.getTime();
  >    ```
- What affect will `start` and `end` being `0` have on the back end? What about `null` or `undefined`? Will they be treated equivalently?

darrenjaneczek (changes requested):
A few recommendations and questions.

public/app/redux/reducers/continuous/index.ts
bryanhuhta: Awesome! This is exactly what I was trying to do yesterday but couldn't figure it out. This also means [this PR](https://github.com/grafana/pyroscope-app-plugin/pull/86) is not relevant anymore. Fixed in https://github.com/grafana/pyroscope/pull/2482/commits/4e932cc9c6edab2db9dae299819ec527dd1acfbe and https://github.com/grafana/pyroscope/pull/2482/commits/d2012bd2c6cd546bab146bd47bb43eb811e6c672.

public/app/services/apps.ts
bryanhuhta: > Do we want to consider making them optional args?

That's a good idea! Fixed in https://github.com/grafana/pyroscope/pull/2482/commits/f597fe19be9d2b362b117011848f13d0945a16cf.

> Will settings the

🤔 

> Everywhere else in pyroscope oss, we have from and until instead of start and end. Do we want to keep the different terminology exclusive to the POST method body, or do we want it to propagate through the function signature and the variables in calling code.

Keeping `from` and `to` as conventions in the UI is probably best. I'll change references to `start` and `end`. Fixed in https://github.com/grafana/pyroscope/pull/2482/commits/4e932cc9c6edab2db9dae299819ec527dd1acfbe and https://github.com/grafana/pyroscope/pull/2482/commits/f597fe19be9d2b362b117011848f13d0945a16cf.

> What affect will start and end being 0 have on the back end? What about null or undefined? Will they be treated equivalently?

If either `start` or `end` are `0`, `null`, or `undefined`, then the backend will fallback to the legacy behavior of only querying the head block and not querying the block store.

public/app/services/apps.ts
```diff
@@ -96,6 +104,8 @@ export async function fetchApps(): Promise<
         '__type__',
         '__name__',
       ],
+      start: fromMs,
+      end: toMs,
```
darrenjaneczek, line 108: Please consider enjoying this syntactic sugar. 
```suggestion
      start: fromMs || 0,
      end: toMs || 0,
```

If anything before `||` resolves to false, it will default to what follows, which will allow you to remove the if checks above.

darrenjaneczek (approved):
LGTM.
Some suggestions for syntactic sugar.

public/app/services/apps.ts
bryanhuhta: Ah nice! Added in https://github.com/grafana/pyroscope/pull/2482/commits/cd03f458510c08cd278fc6c5bd89518e1a02a681

PR #2476, opened 2023-10-02
Add otel logs to rideshare

simonswine:
In order to show better how the data signals connect I am adding a otel log pusher to our rideshare demo.

Note: The upgrade of otel required an upgrade of dskit, which needed some refactoring in commit f241af5.

Unfortunately the rideshare-app does not log at present, so I had to add a bit of logging as well

https://github.com/grafana/pyroscope/assets/223048/0acb4842-f036-4369-bd8d-3723646c4546

PR #2484, opened 2023-10-04
Refactor after dskit upgrade

simonswine:
Upgraded it for a different work item that required upgrading grpc and hence required a new dskit. (#2476)

This uses the interface PoolFactory instead of the `func`


kolesnikovae (approved):
LGTM

PR #2485, opened 2023-10-04
[packaging] Create pyroscope unix group

wilfriedroset:
This PR create a unix group for pyroscope user. This will facilitate permission management down the road.
Bonus point: it is similar to [mimir packaging](https://github.com/grafana/mimir/blob/main/packaging/nfpm/mimir/postinstall.sh#L19). This could be an improvement for [loki](https://github.com/grafana/loki/blob/main/tools/packaging/loki-postinstall.sh)


tools/packaging/postinstall.sh
```diff
@@ -11,12 +11,15 @@ fi
 cleanInstall() {
     printf "\033[32m Post Install of a clean install\033[0m\n"
 
+    if ! getent group pyroscope >/dev/null 2>&1; then
+        groupadd -r pyroscope
+    fi
     # Create the user
     if ! id pyroscope > /dev/null 2>&1 ; then
-        adduser --system --home /var/lib/pyroscope --shell /bin/false "pyroscope"
+        adduser --create-home --system --home /var/lib/pyroscope -g pyroscope --shell /bin/false "pyroscope"
```
simonswine, line 19: I think debian doesn't like `create-home` and only shares `--gid` with the redhat variant:

```suggestion
        adduser --system --home /var/lib/pyroscope --gid $(getent group | grep pyroscope: | awk -F ":" '{ print $3 }') --shell /bin/false "pyroscope"
```

tools/packaging/postinstall.sh
```diff
@@ -11,12 +11,15 @@ fi
 cleanInstall() {
     printf "\033[32m Post Install of a clean install\033[0m\n"
 
+    if ! getent group pyroscope >/dev/null 2>&1; then
+        groupadd -r pyroscope
```
simonswine, line 15: Long arguments are always nicer to understand:

```suggestion
        groupadd --system pyroscope
```

simonswine (commented):
Thanks for adding group, I tried to add it, but it was too hard 🙂 .

See my comments, why the change you were proposing is not working.

I tested on docker images `debian:11` and `fedora:38`

tools/packaging/postinstall.sh
wilfriedroset: I've made a little improvement to your suggestion.
WDYT?

simonswine (approved):
LGTM

Issue #2483, opened 2023-10-04
Deduplication must respect profiles with identical timestamps within the same series


kolesnikovae:
[At deduplication](https://github.com/grafana/pyroscope/blob/main/pkg/querier/select_merge.go#L222) we discard profiles if their timestamps and labels match:

https://github.com/grafana/pyroscope/blob/8efd7c176a0b6de02ebcc9aa37c5aab1bd7c7dde/pkg/querier/select_merge.go#L254-L258

Sometimes distinct profiles of the same series may have the same timestamps (millisecond precision), which results in partial reads. Potentially, this could be fixed by #2480 (#2443)

/cc @cyriltovena

cyriltovena:
Presicion should actually be nanoseconds but you're correct, we also have a unique ID that we could use to ultimately decide if this is a real duplicate.

cyriltovena:
We could also use the IngesterAddr is it's the same as the previous one it's not a dupe, duplicate will only exists across ingester not within.

kolesnikovae:
I think we can have duplicate timestamps on the same ingesters, unfortunately. It comes down to the case when multiple instances form a single series, e.g:
- host A sends `{service_name="my-app"}` at time X
- host B sends `{service_name="my-app"}` at time X 
- host C sends `{service_name="my-app"}` at time X 

And so on

kolesnikovae:
https://github.com/grafana/pyroscope/pull/2489 makes this issue way less severe

PR #2489, opened 2023-10-05
Preserve profile timestamp

kolesnikovae:
Overriding the profile time with the request `StartTime` causes precision loss since `pyroscope-go` specifies time in seconds, which triggers #2483. I'm not sure if we need to update our SDKs, but created PR for Go: https://github.com/grafana/pyroscope-go/pull/54.


simonswine (approved):
LGTM

PR #2498, opened 2023-10-06
feat: get label names from store gateway

bryanhuhta:

Previously label names were only queried from the head block in the ingester. This adds a time window to the LabelNames RPC and queries the store gateway if necessary.


api/gen/proto/go/types/v1/extension.go
```diff
@@ -0,0 +1,6 @@
+package typesv1
+
+// IsLegacy is true if this is a legacy label names request.
+func (r *LabelNamesRequest) IsLegacy() bool {
```
bryanhuhta, line 7: I'm not sure if this is the best place to add this.

pkg/phlaredb/block_querier.go
```diff
@@ -949,6 +987,67 @@ func (b *singleBlockQuerier) SelectMatchingProfiles(ctx context.Context, params
...
+				case index == len(names): // Add name to the end of the slice.
+					names = append(names, name)
+				case names[index] != name: // Insert name into the middle of the slice.
+					names = append(names, name)
+					copy(names[index+1:], names[index:])
+					names[index] = name
+				}
+			}
+		}
+	}
```
bryanhuhta, line 1046: My thinking was:

1. Get all the label names for this index
2. If the selector was "matches all", then we can simply return all the label names
3. If there are matchers, then clear the label names slice, but keep the allocated memory the labels
4. For every posting that satisfies the matchers, find the label names for that posting
5. Select only label names we haven't seen yet

pkg/phlaredb/block_querier.go
```diff
@@ -949,6 +987,67 @@ func (b *singleBlockQuerier) SelectMatchingProfiles(ctx context.Context, params
...
+			if err != nil {
+				if err == storage.ErrNotFound {
+					continue
+				}
+				return nil, err
+			}
+
+			// Select all the label names which we don't already have.
+			for _, name := range labelNames {
+				index := sort.SearchStrings(names, name)
```
cyriltovena, line 1035: I think you could just use a map here to keep unique names.

cyriltovena: Actually I think you need to keep only names that intersect with both matchers. Something is weird here.

api/gen/proto/go/types/v1/extension.go
bryanhuhta: Done in https://github.com/grafana/pyroscope/pull/2498/commits/a10f131663ee055e670248988da4830cb707c8aa

pkg/phlaredb/block_querier.go
bryanhuhta: New logic is in https://github.com/grafana/pyroscope/pull/2498/commits/4e3295c94f3028d4ac0155ccfaee9915f6efb2bb.

pkg/util/time.go
```diff
@@ -98,3 +98,13 @@ func NewDisableableTicker(interval time.Duration) (func(), <-chan time.Time) {
...
+
+type TimeRangeRequest interface {
+	GetStart() int64
+	GetEnd() int64
+}
+
+// HasTimeRange is true if the request has a start and end set.
+func HasTimeRange(req TimeRangeRequest) bool {
+	return req.GetStart() == 0 || req.GetEnd() == 0
+}
```
bryanhuhta, line 110: I had to move this helper here, instead of adding a method directly on `typesv1.LabelNamesRequest`. If I tried to add it directly on the type, the definition got deleted every time we regenerated the protobuf code.

cyriltovena: Yep that make sense, may be `model` makes also sense up to you.

pkg/frontend/frontend_label_names.go
```diff
@@ -4,13 +4,39 @@ import (
...
 
 func (f *Frontend) LabelNames(ctx context.Context, c *connect.Request[typesv1.LabelNamesRequest]) (*connect.Response[typesv1.LabelNamesResponse], error) {
 	ctx = connectgrpc.WithProcedure(ctx, querierv1connect.QuerierServiceLabelNamesProcedure)
+
+	tenantIDs, err := tenant.TenantIDs(ctx)
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, err)
+	}
+
+	if util.HasTimeRange(c.Msg) {
```
cyriltovena, line 25: I guess it should return ok, interval and be named model.RequestTimeRange() ?

pkg/phlaredb/phlaredb.go
```diff
@@ -329,20 +344,17 @@ func (f *PhlareDB) ProfileTypes(ctx context.Context, req *connect.Request[ingest
...
 // Series returns labels series for the given set of matchers.
 func (f *PhlareDB) Series(ctx context.Context, req *connect.Request[ingestv1.SeriesRequest]) (*connect.Response[ingestv1.SeriesResponse], error) {
-	sp, ctx := opentracing.StartSpanFromContext(ctx, "PhareDB Series")
+	sp, ctx := opentracing.StartSpanFromContext(ctx, "PhlareDB Series")
 	defer sp.Finish()
 
+	if !util.HasTimeRange(req.Msg) {
+		return withHeadForQuery(f, func(head *Head) (*connect.Response[ingestv1.SeriesResponse], error) {
+			return head.Series(ctx, req)
+		})
```
cyriltovena, line 355: nit: might be better to use f.headQuerier but it should be the same.

cyriltovena (approved):
LGTM well done !

pkg/frontend/frontend_label_names.go
bryanhuhta: Done in https://github.com/grafana/pyroscope/pull/2498/commits/0d5f92cc2cd98ac9b9e81080ac882e59026b1ae7.

pkg/phlaredb/phlaredb.go
bryanhuhta: No longer a problem after your querier changes 😄

bryanhuhta:
I put https://github.com/grafana/pyroscope/pull/2498/commits/38f39cbae808b53f97062e5ed1d2ab3cbf71fbb6 in dev and ran some quick tests and I couldn't find any appreciable difference for querying label names for the store gateway regardless of the time window. For a 2d to 14d windows it takes about 100ms from my browser and about 1 - 10ms server side.

I'll merge this in and let it cook.

PR #2502, opened 2023-10-08
feat(chart): `ingressClassName` and `tls` for ingress

jjangga0214:


operations/pyroscope/helm/pyroscope/templates/ingress.yaml
```diff
@@ -11,6 +11,13 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
+  {{- if and .Values.ingress.className (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion) }}
+  ingressClassName: {{ .Values.ingress.className }}
+  {{- end }}
+  {{- with .Values.ingress.tls }}
+  tls:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
```
kolesnikovae, line 20: I'm not very sure we want the capability version check as we unconditionally set the API version to `networking.k8s.io/v1` (1.19)

kolesnikovae (commented):
Thank you for the contribution @jjangga0214! 🎉 

Could you please add empty attributes to the value file and run `helm/docs`?

kolesnikovae:
Hello @jjangga0214, could you please resolve the review comments? I think that the change might be also interesting for others, therefore I'd like to get it merged

jjangga0214:
@kolesnikovae Done :) 
Thank you for the review and maintenance!

kolesnikovae:
Thank you @jjangga0214! Could you also please run `helm/check` and push the updated templates?

simonswine (approved):
It was useful to have that sooner, so I ran `helm check` and commited the result.

lgtm

Thanks for the contriubtion @jjangga0214  🙂

Issue #461, opened 2021-10-12
Server config reload


eh-am:
As per https://github.com/pyroscope-io/pyroscope/pull/404#issuecomment-924899621

Now that we have configs that one may want to change while server is running (TLS Cert, or [metrics export](https://github.com/pyroscope-io/pyroscope/pull/444)), it's a good idea to allow the server to refresh the config:

- Given a SIGHUP signal, or call to `/reload`
- Check the config will work (ie it won't crash the server)
- If so
  - restart the server with the new config (with graceful shutdown and whatnot)
  - update metrics, [similar to what prometheus does](https://github.com/prometheus/prometheus/blob/48a101be1b59dce6f9285b40d686be24f7793473/cmd/prometheus/main.go#L78)
- If not, print to the log and [update metrics to 0](https://github.com/prometheus/prometheus/blob/48a101be1b59dce6f9285b40d686be24f7793473/cmd/prometheus/main.go#L986)

eh-am:
Now that we have kubernetes integration this is even more required


Another feature that's similar is being able to validate the config has at least valid syntax. Similar to what `promtool` does https://www.robustperception.io/how-to-check-your-prometheus-yml-is-valid

For now that feature could be baked into the pyroscope server.

cyriltovena:
This is still relevant with Pyroscope 1.0 and could be useful.

see https://github.com/grafana/pyroscope/pull/2506

PR #2506, opened 2023-10-09
[packaging] Configure systemd to allow pyroscope to be reloaded via kill -HUP

wilfriedroset:
This PR aim to fix the following error:

```
# systemctl reload pyroscope
Failed to reload pyroscope.service: Job type reload is not applicable for unit pyroscope.service.
```


wilfriedroset:
If pyroscope does not support reloading my PR does not make sense.
For the context: when running on baremetal it would be nice to be able to reload pyroscope instead of stop/start. This could be a little improvement

cyriltovena:
Closing until https://github.com/grafana/pyroscope/issues/461 is implemented. Thank you @wilfriedroset

PR #2519, opened 2023-10-11
Handle symdb partition initialization error

kolesnikovae:


pkg/phlaredb/symdb/block_reader.go
```diff
@@ -350,12 +351,7 @@ func (c *stacktraceChunkReader) fetch(ctx context.Context) (err error) {
 	// it after it is fully loaded. Use of atomics here
 	// for reference counting is not sufficient.
 	c.m.Lock()
-	defer func() {
-		if err != nil {
-			c.r--
-		}
-		c.m.Unlock()
-	}()
```
simonswine, line 358: Nil: When an error happens in fetch, the next call to fetch will silently `return nil`.

This is not a problem, with the way the fetch is used in an `errgroup` today, but might become one in the future.

simonswine (approved):
LGTM

pkg/phlaredb/symdb/block_reader.go
kolesnikovae: Refactored the reference counter to a separate package so this does not look so ugly

PR #2520, opened 2023-10-11
Update helm chart to use latest version of pyroscope image

simonswine:
This updates helm to use the latest helm chart from yesterday's release


Rperry2174 (approved):
lgtm

PR #2521, opened 2023-10-11
Feat: ingest align block ranges.

cyriltovena:
This aligns block ranges when flushing heads.

To do this and still support out of order, we actually maintain multiple head and flush them after a certain period once they go stale.

Still needs to add some tests but it basically work nicely here is a screenshot showing 1m and 30m block perfectly aligned now

image

## Todos

- [x] Cleanup and better default config
- [x] Tests
- [ ] Reword config parameters


cyriltovena:
cc @bryanhuhta check out the small refactoring it should make your life easier.

pkg/phlaredb/phlaredb.go
```diff
@@ -161,16 +156,88 @@ func (f *PhlareDB) loop() {
 			f.runBlockQuerierSync(ctx)
 		case <-headSizeCheck.C:
 			if f.headSize() > maxBlockBytes {
-				f.flushHead(ctx, flushReasonMaxBlockBytes)
+				f.Flush(ctx, true, flushReasonMaxBlockBytes)
 			}
-		case <-f.forceFlush.C:
-			f.flushHead(ctx, flushReasonMaxDuration)
+		case <-staleHeadTicker.C:
+			f.Flush(ctx, false, flushReasonMaxDuration)
```
kolesnikovae, line 162: Would be nice to have some sort of jitter here to avoid thundering herd problem

pkg/phlaredb/phlaredb.go
```diff
@@ -161,16 +156,88 @@ func (f *PhlareDB) loop() {
...
 
+func (f *PhlareDB) Flush(ctx context.Context, force bool, reason string) (err error) {
+	// Ensure this is the only Flush running.
+	f.flushLock.Lock()
+	defer f.flushLock.Unlock()
+
+	currentSize := f.headSize()
+	// Create a new head and evict the old one. Reads and writes
+	// are blocked – after the lock is released, no new ingestion
+	// requests will be arriving to the old head.
```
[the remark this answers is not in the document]
cyriltovena: I'll update thanks !

pkg/phlaredb/head.go
```diff
@@ -161,6 +163,16 @@ func (h *Head) loop() {
...
+func (h *Head) isStale(maxT int64, now time.Time) bool {
+	// todo less aggressive this is just for testing.
+	// Newly blocks are not stale
+	if now.Sub(h.createdAt) <= 1*time.Minute {
+		return false
+	}
+	// Blocks that have pass their maxT range by 5min are stale
+	return now.Sub(time.Unix(0, maxT)) >= 1*time.Minute
+}
+
```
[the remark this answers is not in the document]
cyriltovena: Yes I was planning to work on that, thank you for the suggestions !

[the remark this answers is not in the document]
cyriltovena: I'm going down a simple path for now, making it more dynamic based on last ingested sample. We can improve later if this is not enough.

simonswine (approved):
Great work!

I have tried hard to break it but could not find a way 😆 

LGTM

pkg/ingester/ingester.go
```diff
@@ -268,7 +267,7 @@ func (i *Ingester) Flush(ctx context.Context, req *connect.Request[ingesterv1.Fl
 	i.instancesMtx.RLock()
 	defer i.instancesMtx.RUnlock()
 	for _, inst := range i.instances {
-		if err := inst.Flush(ctx); err != nil {
+		if err := inst.Flush(ctx, true, "api"); err != nil {
```
bryanhuhta, line 270: Should `"api"` be a constant somewhere? Just like `flushReasonMaxBlockBytes` and `flushReasonMaxDuration`.

pkg/phlaredb/phlaredb.go
```diff
@@ -346,236 +391,38 @@ func (f *PhlareDB) Series(ctx context.Context, req *connect.Request[ingestv1.Ser
 	f.headLock.RLock()
 	defer f.headLock.RUnlock()
 
-	res, err := Series(ctx, req.Msg, f.queriers().ForTimeRange)
-	if err != nil {
-		return nil, err
+	if req.Msg.Start == 0 || req.Msg.End == 0 {
+		return f.headQueriers().Series(ctx, req)
```
bryanhuhta, line 395: 👍

pkg/phlaredb/block_querier.go
```diff
@@ -405,6 +423,19 @@ func InRange(q Querier, start, end model.Time) bool {
 	return true
 }
 
+type ReadAPI interface {
```
bryanhuhta, line 426: This type is here strictly to enforce `Queriers` implements these methods, correct? I don't see it used anywhere else.

bryanhuhta (approved):
:shipit:

pkg/ingester/ingester.go
cyriltovena: may be ? It doesn't add much if it used once, but +1 on consistency.

pkg/phlaredb/block_querier.go
cyriltovena: Yeah we don't really need it, but I think this is the start of a bigger refactoring, we in fact should only implement this in Queriers.

PR #2524, opened 2023-10-11
feat: get label values from store gateway

bryanhuhta:

Previously label values were only queried from the head block in the ingester. This adds a time window to the LabelValues RPC and queries the store gateway if necessary.

https://github.com/grafana/pyroscope/pull/2498 should probably merge first, then the commit history of this PR will get cleaned up quite a bit


bryanhuhta:
Closing this in favor of https://github.com/grafana/pyroscope/pull/2594, which has a generally cleaner commit history and reduces the likelihood of a bug being introduced through a bad merge conflict resolution.

PR #2537, opened 2023-10-15
Update go_pull.md with new module link

yuseferi:
the old module has been deprecated and it's been moved inside, so I updated it with the correct link.


kolesnikovae:
Thank you @yuseferi!

kolesnikovae (approved):

PR #2543, opened 2023-10-16
refactor(Frontend): Extract the logic to determine the default app for continuous profiling pages

grafakus:
In this PR, we separate the business logic from the rest of the code to support code reusability and extensibility.


Rperry2174 (approved):
lgtm!

PR #2544, opened 2023-10-16
docs: adds server api docs

Rperry2174:
original server api docs https://github.com/pyroscope-io/docs/blob/main/docs/server-api-reference.mdx

TODO: 
- How does authentication work in new server API docs (with grafana cloud) 
- add more examples to examples folder

### DIFF
--- a/docs/sources/configure-server/about-server-api.md
+++ b/docs/sources/configure-server/about-server-api.md
@@ -1,0 +1,196 @@
+---
+description: Learn about server api
+menuTitle: About server API
+title: About Grafana Pyroscope tenant IDs
+weight: 20
+---
+
+# HTTP API Reference
+
+Pyroscope server exposes an HTTP API that can be used to query profiling data and to ingest profiling data from other sources.
+
+## Ingestion
+
+Currently there's just one endpoint: `POST /ingest`. It takes profile data in request body and metadata as query params.
+
+It takes multiple query parameters:
+
+| Name               | Description                             | Notes                           |
+|:-------------------|:----------------------------------------|:--------------------------------|
+| `name`             | application name                        | required |
+| `from`             | UNIX time of when the profiling started | required                        |
+| `until`            | UNIX time of when the profiling stopped | required                        |
+| `format`           | format of the profiling data            | optional (default is `folded`)  |
+| `sampleRate`       | sample rate used in Hz                  | optional (default is 100 Hz)    |
+| `spyName`          | name of the spy used                    | optional                        |
+| `units`            | name of the profiling data unit         | optional (default is `samples`  |
+| `aggregrationType` | type of aggregration to merge profiles  | optional (default is `sum`)     |
+
+* `name` specifies application name. It may be specified in the fully qualified application name notation as described in [the specification](/docs/flameql#notation). For example:
+```
+my.awesome.app.cpu{env=staging,region=us-west-1}
+```
+
+Request body contains the profiling data, and request header `Content-Type` may also be alongside `format` to determine profiling data format.
+
+Some of the query parameters depend on the format of profiling data. Pyroscope currently supports three major ingestion formats.
+
+### Text Formats
+
+Most simple ingestion formats send a single type of profiling data (for example, `cpu` samples) and don't usually support any metadata (e.g labels) within the format to indicate which kind of data they are sending.
+In these cases, all the metadata is taken from the query parameters, and the format itself is defined by `format` query parameter.
+The formats that work this way are:
+
+* `folded`. Some software also call this format `collapsed`. This is the default format. With this format you put one stacktrace per line with a number of samples you've captured for that particular stacktrace, for example:
+```
+foo;bar 100
+foo;baz 200
+```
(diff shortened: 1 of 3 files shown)


docs/sources/configure-server/about-server-api.md
```diff
@@ -0,0 +1,196 @@
...
+| `until`            | UNIX time of when the profiling stopped | required                        |
+| `format`           | format of the profiling data            | optional (default is `folded`)  |
+| `sampleRate`       | sample rate used in Hz                  | optional (default is 100 Hz)    |
+| `spyName`          | name of the spy used                    | optional                        |
+| `units`            | name of the profiling data unit         | optional (default is `samples`  |
+| `aggregrationType` | type of aggregration to merge profiles  | optional (default is `sum`)     |
+
+* `name` specifies application name. It may be specified in the fully qualified application name notation as described in [the specification](/docs/flameql#notation). For example:
+```
+my.awesome.app.cpu{env=staging,region=us-west-1}
```
Rperry2174, line 31: how does this work now with `{service_name="xyz"}`

kolesnikovae: Our `/ingestion` API won't be able to handle labels in the new format.

I think it's better to add a new parameter (to avoid clashes with the old logic), like `labels` and not to mention `name` and flameql here as to not confuse users. We will handle both params it in the backend, but only `labels` should be part of the public API

docs/sources/configure-server/about-server-api.md
```diff
@@ -0,0 +1,194 @@
...
+| Name               | Description                             | Notes                           |
+|:-------------------|:----------------------------------------|:--------------------------------|
+| `name`             | application name                        | required |
+| `from`             | UNIX time of when the profiling started | required                        |
+| `until`            | UNIX time of when the profiling stopped | required                        |
+| `format`           | format of the profiling data            | optional (default is `folded`)  |
+| `sampleRate`       | sample rate used in Hz                  | optional (default is 100 Hz)    |
+| `spyName`          | name of the spy used                    | optional                        |
+| `units`            | name of the profiling data unit         | optional (default is `samples`  |
+| `aggregrationType` | type of aggregration to merge profiles  | optional (default is `sum`)     |
```
kolesnikovae, line 30: We probably want to review the list of params, and remove some of them:
 - `name` – as Ryan mentioned, we need to make sure the new series labels format is supported;
 - `spyName` – could be part of series labels;
 - `units` - should be part of the profile type (a label);
 - `aggegationType` – unused.

docs/sources/configure-server/about-server-api.md
```diff
@@ -0,0 +1,194 @@
...
+title: Pyroscope Server HTTP API Reference
+weight: 20
+---
+
+# Pyroscope Server HTTP API Reference
+
+Pyroscope server exposes an HTTP API that can be used to query profiling data and to ingest profiling data from other sources.
+
+## Authentication
+TODO - how does authentication work with new Grafana cloud stuff?
```
Rperry2174, line 13: how does authentication work with new API

kolesnikovae, line 13:
```suggestion
Grafana Pyroscope does not come with any included authentication layer. Operators are expected to run an authenticating reverse proxy in front of your services.

Note that when using Pyroscope in multi-tenant mode, Pyroscope requires the HTTP header `X-Scope-OrgID` to be set to a string identifying the tenant; the responsibility of populating this value should be handled by the authenticating reverse proxy. Please refer to the [multi-tenancy documentation](https://grafana.com/docs/pyroscope/latest/configure-server/about-tenant-ids) for more information.
```

I'm not sure we should cover Grafana Cloud auth here. Should we maybe update the hosted instance page (where the credentials are obtained) with some examples?

kolesnikovae (approved):
LGTM

petethepig (approved):

docs/sources/configure-server/about-server-api.md
```diff
@@ -0,0 +1,196 @@
...
+---
+
+# Pyroscope Server HTTP API Reference
+
+Pyroscope server exposes an HTTP API that can be used to query profiling data and to ingest profiling data from other sources.
+
+## Authentication
+Grafana Pyroscope does not come with any included authentication layer. Operators are expected to run an authenticating reverse proxy in front of your services.
+
+Note that when using Pyroscope in multi-tenant mode, Pyroscope requires the HTTP header `X-Scope-OrgID` to be set to a string identifying the tenant; the responsibility of populating this value should be handled by the authenticating reverse proxy. Please refer to the [multi-tenancy documentation](https://grafana.com/docs/pyroscope/latest/configure-server/about-tenant-ids) for more information.
```
petethepig, line 15:
```suggestion
Note that when using Pyroscope in multi-tenant mode, Pyroscope requires the HTTP header `X-Scope-OrgID` to be set to a string identifying the tenant; the responsibility of populating this value should be handled by the authenticating reverse proxy. Please refer to the [multi-tenancy documentation]({{< relref "./about-tenant-ids" >}}) for more information.
```

PR #2547, opened 2023-10-17
fix(helm): add more ingress paths for query-frontend and distributor

wilfriedroset:
This PR adds the following paths for the ingress:

- /pyroscope/render/
- /pyroscope/render-diff/
- /ingest/
- /pyroscope/ingest/

Those paths are used by Pyroscope Language SDKs such as python.


operations/pyroscope/helm/pyroscope/templates/ingress.yaml
```diff
@@ -27,6 +27,28 @@ spec:
               number: {{ $.Values.pyroscope.service.port }}
         path: /querier.v1.QuerierService/
         pathType: Prefix
+      - backend:
```
wilfriedroset, line 30: An alternate way to this proposal is to enable [Regular Expression Support ](https://kubernetes.github.io/ingress-nginx/user-guide/ingress-path-matching/). With it we could have something like
`path: /(querier.v1.QuerierService|pyroscope/render|pyroscope/render-diff)/`

simonswine: I would like to keep the Path type if possible, as that might be compatible with more than nginx

operations/pyroscope/helm/pyroscope/templates/ingress.yaml
```diff
@@ -27,6 +27,28 @@ spec:
...
+          service:
+            {{- if gt (len $.Values.pyroscope.components) 1}}
+            name: {{ include "pyroscope.fullname" $ }}-query-frontend
+            {{- else }}
+            name: {{ include "pyroscope.fullname" $ }}
+            {{- end }}
+            port:
+              number: {{ $.Values.pyroscope.service.port }}
+        path: /pyroscope/render
```
simonswine, line 39:
```suggestion
        path: /render
```

operations/pyroscope/helm/pyroscope/templates/ingress.yaml
```diff
@@ -27,6 +27,28 @@ spec:
...
+      - backend:
+          service:
+            {{- if gt (len $.Values.pyroscope.components) 1}}
+            name: {{ include "pyroscope.fullname" $ }}-query-frontend
+            {{- else }}
+            name: {{ include "pyroscope.fullname" $ }}
+            {{- end }}
+            port:
+              number: {{ $.Values.pyroscope.service.port }}
+        path: /pyroscope/render-diff
```
simonswine, line 50:
```suggestion
        path: /render-diff
```

operations/pyroscope/helm/pyroscope/templates/ingress.yaml
```diff
@@ -38,5 +60,27 @@ spec:
...
+          service:
+            {{- if gt (len $.Values.pyroscope.components) 1}}
+            name: {{ include "pyroscope.fullname" $ }}-distributor
+            {{- else }}
+            name: {{ include "pyroscope.fullname" $ }}
+            {{- end }}
+            port:
+              number: {{ $.Values.pyroscope.service.port }}
+        path: /pyroscope/ingest
+        pathType: Prefix
```
simonswine, line 84: Let's not use the ones with `/pyroscope/xyz` anymore

```suggestion

```

simonswine (commented):
Thanks for your work on this, when testing those I noticed we also don't have an ingress class, will add a separate PR for that.

Looking mostly good, if could take a look at my changes

simonswine (approved):
There was already a separate PR #2502 so that merged. And I also think that is ready to go now 

LGTM

simonswine:
Ok I gave it another run, just for my own sanity this are steps how I tested ingress:

```
# Deploy microservices
$ make deploy-micro-services "HELM_ARGS=--set ingress.enabled=true --set ingress.hosts={pyroscope.localhost} --set ingress.className=nginx"

# Install nginx ingress
$ helm upgrade --install --version 4.7.3 ingress-nginx ingress-nginx/ingress-nginx

# port-forward, keep it running while typing the next commands
$ kubectl port-forward deployments/ingress-nginx-controller 8080:80

# test upload endpoint
curl -v --resolve -X POST  --data-binary @pprof/pprof.alloc_objects.alloc_space.inuse_objects.inuse_space.001.pb.gz  pyroscope.localhost:8080:127.0.0.1 http://pyroscope.localhost:8080/ingest\?format\=pprof\&from\=1697546758\&until\=1697546768\&name\=my-app-name-pprof
```

PR #2553, opened 2023-10-17
Update microservices values for ci tests

simonswine:
We acutally didn't update this in a while and so don't fully test those services.

### DIFF
--- a/operations/pyroscope/helm/pyroscope/ci/micro-services-values.yaml
+++ b/operations/pyroscope/helm/pyroscope/ci/micro-services-values.yaml
@@ -13,6 +13,24 @@ pyroscope:
         requests:
           memory: 32Mi
           cpu: 10m
+    query-frontend:
+      kind: Deployment
+      replicaCount: 2
+      resources:
+        limits:
+          memory: 1Gi
+        requests:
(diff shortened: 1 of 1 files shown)


operations/pyroscope/helm/pyroscope/ci/micro-services-values.yaml
```diff
@@ -27,10 +45,23 @@ pyroscope:
...
+    store-gateway:
+      kind: StatefulSet
+      replicaCount: 3
+      persistence:
+      # The store-gateway needs not need persistent storage, but we still run it as a StatefulSet
+      # This is to avoid having blocks of data being
+        enabled: false
+      resources:
+        limits:
+          memory: 16Gi
```
kolesnikovae, line 61: Should we make it lower as in case of ingester?

kolesnikovae (approved):
LGTM

operations/pyroscope/helm/pyroscope/ci/micro-services-values.yaml
simonswine: Yes that was an oversight thanks for catching 👍

PR #2556, opened 2023-10-18
docs: improves loadgen program

petethepig:
this makes it so that go version resembles the [python version](https://github.com/grafana/pyroscope/blob/main/examples/python/rideshare/django/load-generator.py) more — that one produces slightly more interesting looking graphs in demos

cc @Rperry2174 @korniltsev


kolesnikovae (approved):
LGTM

PR #2562, opened 2023-10-20
Improve flush spread

cyriltovena:
This should improve the spread of flushes per ingester. Not only I've increase the jitter, but it's re-evaluated on each run, making it even more randomized.

We still need a flush queue for large multitenant cluster.


kolesnikovae (approved):
LGTM

PR #2564, opened 2023-10-20
Tweak compaction configuration and performance.

cyriltovena:
This now set 1h the block duration default and compaction range to 1h, 2h, and 8h.

This should be much better default based on my overall experience with the compactor.

The commit also include a big performance improvement where compacting large amount of blocks. This is done by reducing the row buffer  from 1024 to 32.

### DIFF
--- a/pkg/compactor/compactor.go
+++ b/pkg/compactor/compactor.go
@@ -121,7 +121,7 @@ type Config struct {
 func (cfg *Config) RegisterFlags(f *flag.FlagSet, logger log.Logger) {
 	cfg.ShardingRing.RegisterFlags(f, logger)
 
-	cfg.BlockRanges = DurationList{3 * time.Hour, 6 * time.Hour, 12 * time.Hour}
+	cfg.BlockRanges = DurationList{1 * time.Hour, 4 * time.Hour, 8 * time.Hour}
 	cfg.retryMinBackoff = 10 * time.Second
 	cfg.retryMaxBackoff = time.Minute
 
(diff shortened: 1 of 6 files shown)


pkg/compactor/compactor.go
```diff
@@ -121,7 +121,7 @@ type Config struct {
 func (cfg *Config) RegisterFlags(f *flag.FlagSet, logger log.Logger) {
 	cfg.ShardingRing.RegisterFlags(f, logger)
 
-	cfg.BlockRanges = DurationList{3 * time.Hour, 6 * time.Hour, 12 * time.Hour}
+	cfg.BlockRanges = DurationList{1 * time.Hour, 4 * time.Hour, 8 * time.Hour}
```
kolesnikovae, line 124: In the description it's said the block ranges are to be: 1, 2, and 8 hours. I don't know exactly what is most optimal configuration, but assume that the earlier blocks get compacted, the better, therefore the 1, 2, 8 setup intuitively looks more appealing

pkg/phlaredb/phlaredb.go
```diff
@@ -50,7 +50,7 @@ type ParquetConfig struct {
 
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.DataPath, "pyroscopedb.data-path", "./data", "Directory used for local storage.")
-	f.DurationVar(&cfg.MaxBlockDuration, "pyroscopedb.max-block-duration", 3*time.Hour, "Upper limit to the duration of a Pyroscope block.")
+	f.DurationVar(&cfg.MaxBlockDuration, "pyroscopedb.max-block-duration", 1*time.Hour, "Upper limit to the duration of a Pyroscope block.")
```
kolesnikovae, line 53: 👍🏻 

Just thinking out-loud: now that we have compactor, smaller source blocks are preferable. I'm also wondering how it relates to the ingestion window. I suppose those should be "aligned", e.g the current default window duration is one hour, therefore 1h max block duration fits it nicely (so that at most two blocks of the same time range co-exist)

kolesnikovae (approved):
LGTM

pkg/phlaredb/phlaredb.go
simonswine: Yes a validation for this would be nice, or at least a warning when the windows is totally out of boudns (e.g. more than X head blocks required)

simonswine (approved):
LGTM

pkg/compactor/compactor.go
cyriltovena: I think it's fair to have the 2h mark in case the 4h is very slow. I will change it.

PR #2570, opened 2023-10-23
Fixes compaction range block grouping

cyriltovena:
This fixes an issue where if the minTime of a block would fall exactly on the end of a fixed range it would be considered overlapping (discarded).

In fact if that is the case it only means the block doesn't start within that range and we should move on to the next range.


simonswine (approved):
LGTM

PR #2575, opened 2023-10-24
Increase default compaction interval

cyriltovena:
I think 1h was too high based on the default block duration 1h too.

Also we can open more block at once without any problem.


simonswine (approved):
LGTM

What is the cost of compaction interval? Is it mainly listing the bucket and resolving the dependencies?

I probably would even go to 10/15min

PR #2577, opened 2023-10-25
chore(ebpf)  py312 support

korniltsev:
I've tried to document things I've done to enable support for new python version. I hope in the future only first two steps are required(in case only offset values changed). However I can not promise and dont know how to avoid the other changes if something unexpected changed in cpython (field removals, field renames, struct removals)

- added testdata libpython3.12.so.1.0
- run `python/dwarfdump` to regenerate python/version_gen.go
- profiling init failed because it detected  StringSize(PyAsciiObject) is changed and I've previously hardcoded it's size in the ebpf program, I think I will have to unhardcode it
- add String_size field to py_offset_config struct in pyperf.bpf.c, use the new field instead of hardcoded value
- regenerate userspace code and bpf program by `make bpf/gen`
- pass string size from userspace to bpf via new field
- profiling init failed again because it could not read autoTSSkey. it was moved from PyRuntimeState->gilstate->runtime_state.autoTSSkey to PyRuntimeState->autoTSSkey
- adjust tss reading code to read it from new place for 3.12
- unhardcoded more offsets (PY_OFFSET_PyVarObject_ob_size PY_OFFSET_PyObject_ob_type PY_OFFSET_PyTypeObject_tp_name), "just in case"


ebpf/bpf/pyperf.bpf.c
```diff
@@ -39,7 +39,6 @@ enum {
 
 };
 
-#define PY_OFFSET_String_size 48
 #define PY_OFFSET_PyVarObject_ob_size 16
```
korniltsev, line 42: I wonder if I should unhardcode these as well just in case

kolesnikovae (approved):
LGTM 👍🏻

PR #2580, opened 2023-10-25
ingest: drop empty pprof profiles

korniltsev:


kolesnikovae (approved):
Looks good! Should we add a test?

PR #2582, opened 2023-10-25
Rideshare: Make sure loki extracts the right labels with OTLP

simonswine:
This will use relevant OTLP tags as loki labels


cyriltovena (approved):
LGTM

PR #2586, opened 2023-10-25
Create a physical plan for block querying

simonswine:
This PR will reduce the work necessary to query blocks, by planning out
which blocks are fetched from which ingester/store-gateway.

That way queries will potentially not need profile-by-profile
deduplication for blocks that had been deduplicated by the compactor.

TODO:
- [x] respect limit the BlockMetacall to start/end in
  ingester/store-gateway
- [x] respect hints in ingester
- [ ] ~Reuse the compactor block mark deletion algorithm in part to disqualify blocks~ (It is a bit more nuanced, so keeping it separate for now)


pkg/querier/replication.go
```diff
@@ -46,26 +73,300 @@ func forGivenReplicationSet[Result any, Querier any](ctx context.Context, client
...
+		// set the selected replica
+		r.m[blockID] = []string{selectedReplica}
+	}
+
+	// adapt the plan to make sure all replicas will deduplicate
+	if deduplicate {
+		for _, hints := range plan {
+			hints.Deduplication = deduplicate
+		}
+	}
```
kolesnikovae, line 369: Does it mean that if we got a query like "last 2 days" we will force deduplication as the plan also includes blocks that have not been compacted yet? I guess, this should be mitigated by query time range split – I'm wondering if we should set to the block duration by default

api/ingester/v1/ingester.proto
```diff
@@ -63,6 +68,8 @@ message MergeProfilesStacktracesRequest {
   optional int64 max_nodes = 3;
   // On a batch of profiles, the client sends the profiles to keep for merging.
   repeated bool profiles = 2;
+  // On a batch of blocks, the client sends the profiles to keep for merging.
+  repeated bool blocks = 4;
```
kolesnikovae, line 72: Must be a leftover

pkg/querier/replication.go
simonswine: Yes this is true. The first step is for the query time range split to alight with roughly the processing time for the compacted block. So it should be slightly higher than block duration, just because we introduce a fair bit of latency before a block is compacted/deduplicated and available to store-gateways.

Ideally in a future where the query-frontend will also be aware of the plan, we can split accordingly to what the block layout is.

pkg/querier/ingester_querier.go
```diff
@@ -76,23 +89,36 @@ func (q *Querier) selectTreeFromIngesters(ctx context.Context, req *querierv1.Se
...
 	g, gCtx := errgroup.WithContext(ctx)
-	for _, r := range responses {
-		r := r
+	for idx := range responses {
+		r := responses[idx]
+		hints, ok := plan[r.addr]
+		if !ok && plan != nil {
+			return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("no hints found for replica %s", r.addr))
+		}
+
```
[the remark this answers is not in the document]
simonswine: Regarding the nil check, not too sure I can follow, why I need to do the plan-nil check before?

When a plan is not nil, we will only query ingesters which are part of the plan (`forAllPlannedIngesters`), so we should never be in the situation to hit a replica not part of the planreally, so I think it should be an error.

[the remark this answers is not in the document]
kolesnikovae: You're right, I overlooked that. Indeed, now I see that this may only happen if `forAllPlannedIngesters` removed the replica from plan, or got a response from unexpected one, which would be a bug

RE nil check: never mind, it should be okay

pkg/phlaredb/block_querier.go
```diff
@@ -771,54 +809,90 @@ func MergeProfilesStacktraces(ctx context.Context, stream *connect.BidiStream[in
...
+				if err = stream.Send(&ingestv1.MergeProfilesStacktracesResponse{}); err != nil {
+					return err
+				}
+
+				iters, err := querier.SelectMatchingProfiles(ctx, request)
+				if err != nil {
+					return err
+				}
+
+				// TODO: To sort or not?
```
cyriltovena, line 843: You need sorting if we don't change how `SelectMatchingProfiles` orders data, currently it reorders by time and then series.  

@kolesnikovae wanted to change this, and I couldn't remember if this was important or not.

simonswine: This is fixed in the meantime

kolesnikovae (approved):
LGTM 🚀 Looking forward to seeing this in action!

pkg/phlaredb/block_querier.go
```diff
@@ -769,56 +811,107 @@ func MergeProfilesStacktraces(ctx context.Context, stream *connect.BidiStream[in
...
+				if err != nil {
+					return err
+				}
+
+				if len(profiles) == 0 {
+					return nil
+				}
+
+				// TODO(simonswine): Split profiles per row group and run the MergeByStacktraces in parallel.
+				merge, err := querier.MergeByStacktraces(ctx, iter.NewSliceIterator(querier.Sort(profiles)))
```
cyriltovena, line 862: I think we can avoid sorting and pulling actual labels. This could be big savings.

simonswine: So you mean all in one step: a `SelectAndMergeProfiles` runs one iterator through data?

PR #2593, opened 2023-10-26
Removes useless logs in the hot write path

cyriltovena:
This is a big useless and causing contentions at high scale.


simonswine (approved):
LGTM

Rperry2174 (approved):
lgtm

PR #2594, opened 2023-10-26
feat: get label values from store gateway

bryanhuhta:
Supersedes: https://github.com/grafana/pyroscope/pull/2524

Previously label values were only queried from the head block in the ingester. This adds a time window to the LabelValues RPC and queries the store gateway if necessary.

I had to make some changes to the redux thunk which fetched label names to make querying actually work.


bryanhuhta:
I ran some tests for performance in dev. There's very little performance change between this implementation and the head-block-only implementation. I suspect @cyriltovena's compactor changes are a significant part to improving query performance 🎉

kolesnikovae (approved):
LGTM

pkg/phlaredb/block_querier.go
```diff
@@ -519,45 +581,19 @@ func (queriers Queriers) SelectMatchingProfiles(ctx context.Context, params *ing
...
-			mutex.Lock()
-			defer mutex.Unlock()
-			if res != nil {
-				for _, name := range res.Msg.Names {
-					uniqValues[name] = struct{}{}
-				}
-			}
-			return nil
-		})
+	// todo: verify empty timestamp request should return all series
```
cyriltovena, line 584: !?

PR #2595, opened 2023-10-27
chore(ebpf) python 3.13-rc support

korniltsev:
- `PyThreadState.cframe` was removed and `PyThreadState.current_frame` added
- `PyInterpreterFrame.f_code` was renamed to` PyInterpreterFrame.f_executable`


ebpf/session_python.go
```diff
@@ -71,6 +71,9 @@ func (s *session) collectPythonProfile(cb func(t *sd.Target, stack []string, val
 					}
 					classname := cStringFromI8Unsafe(sym.Classname[:])
 					name := cStringFromI8Unsafe(sym.Name[:])
+					if classname == "" && filename == "" && name == "" {
+						continue
+					}
```
kolesnikovae, line 76: I'm curious how happen it is happening and whether we should keep track of it, like metric or something

kolesnikovae (approved):
LGTM

ebpf/session_python.go
korniltsev: I saw this happening only for `frame->owner == FRAME_OWNED_BY_CSTACK`, for example the very first frame started by thread.

The python runtime silently discards them https://github.com/python/cpython/blob/e7331365b488382d906ce6733ab1349ded49c928/Python/traceback.c#L991
I could do same thing in ebpf, but that would require tracking extra structure field offset `frame->owner` which I am lazy to do.

korniltsev: I think we can ignore them for now

PR #2597, opened 2023-10-27
Increase max stack trace nodes per chunk limit

kolesnikovae:
I propose to increase the number of stack trace nodes per chunk from 1 million to 4.

The purpose of chunking was to allow for locations lookup with the symdb partition loaded into memory partially. The concern was high memory utilisation. However, at the moment, a query loads all the chunks concurrently in favour of lower query latency. In the meantime, chunking causes some duplication: as a single stack trace can't be split, we copy all its nodes into a new chunk, even if part of the stack trace is already known (but stored in another chunk and can't be referenced). 

Each node takes up 16 bytes memory in ingesters, and 8 bytes when read from the block. Existing data sets I have analysed show that a single partition may contain 7 chunks and more (6+ millions of stacktraces); after compacting them into a single chunk of a larger size, they fit 4M limit.


cyriltovena:
Sounds good to me !

cyriltovena (approved):
LGTM

PR #2598, opened 2023-10-27
Load symbols into memory at compaction

kolesnikovae:
When we compact blocks, it is preferable to load all symbols into memory

/cc @cyriltovena


cyriltovena (approved):
LGTM

PR #2606, opened 2023-10-30
Cleanup failed segment before flushing.

cyriltovena:
We've found instances where if an error happens during cut of rowgroups, there will be a file left over that will make next attempts fail.

This PR adds a test for this and fixes this by removing the segment if it exist, it will also log an error.


pkg/phlaredb/profile_store.go
```diff
@@ -240,7 +240,11 @@ func (s *profileStore) cutRowGroup(count int) (err error) {
 		s.path,
 		fmt.Sprintf("%s.%d%s", s.persister.Name(), s.rowsFlushed, block.ParquetSuffix),
 	)
-
+	// Removes the file if it exists. This can happen if the previous
+	// cut attempt failed.
+	if err := os.Remove(path); err == nil {
+		level.Error(s.logger).Log("msg", "old rowgroup segment found", "path", path)
```
simonswine, line 246: I would downgrade to a warning and make the message clearer

```suggestion
		level.Warning(s.logger).Log("msg", "deleting row group segment of a failed previous attempt found", "path", path)
```

simonswine (approved):
LGTM

PR #2609, opened 2023-10-30
Fallback to manual listing, when bucket index is not available

simonswine:
Since we merged the compactor, store-gateways rely on it to write the block index to quickly list blocks.

@cyriltovena not particularly pretty or performant, but I think might do its job


cyriltovena (approved):
Lgtm

PR #2610, opened 2023-10-30
Add worker pool that can leak a byte buffer

bryanhuhta:
Creates a worker pool that will keep a byte buffer around, causing memory in `us-east` to inflate. Tuning the buffer size and the number of workers that hang will adjust the impact of the memory leak.

Supersedes https://github.com/grafana/pyroscope/pull/2469 as this implementation is more straightforward.

Screenshot 2023-10-30 at 3 52 15 PM


cyriltovena (approved):
LGTM

PR #2613, opened 2023-10-31
fix: remove `request` which is handled by `fetchAndHandleErrors`

darrenjaneczek:
- Remove old `request`, which is virtually identical to `fetchAndHandleErrors`
- Rename `requestWithOrgID` to `request`
- Make non-imported functions in `base` not exportable


grafakus (approved):

grafakus:
Approved after looking at the code. I'm making some tests locally.

grafakus:
Did a swipe over the app and there were no errors.

grafakus (approved):

PR #2615, opened 2023-11-01
fix: discard profiles without samples

kolesnikovae:
Profiles are rejected in ingesters if they don't have sample values. However, it only happens after we create symdb partition for the profile and write symbolic data (such as mappings, that are always present), which in rare cases results in empty partitions. In turn, loading symdb (with `Load`) that has empty partitions causes panic.


simonswine (approved):
LGTM

petethepig (approved):

PR #2618, opened 2023-11-01
fix: don't show `Other` aggregate on Tag explorer timeline

darrenjaneczek:
When there are more than N timelines in the tag explorer, the timelines that don't meet the top N criteria, are aggregated into a pseudo timeline called "Other."

Primarily, this is for calculations used in the pie chart and the table that appear below the timeline. In creating a pseudo timeline, this was causing an "Other" entry to show up in the legend, and passed invalid `startTime` and `durationDelta` values, they have `N-totalNumberOfTimelines` times more samples than the other timelines, and thus cannot be charted properly anyway.

How it looked before:
image

How it looks now:
image

### DIFF
--- a/public/app/pages/TagExplorerView.tsx
+++ b/public/app/pages/TagExplorerView.tsx
@@ -292,10 +292,10 @@ function TagExplorerView() {
     sortedGroupsByTotal.length
   );
 
-  const groups =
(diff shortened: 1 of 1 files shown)


Rperry2174 (approved):
lgtm

PR #2619, opened 2023-11-02
remove old api key format from example docs

Rperry2174:
removes old api key format from examples

### DIFF
--- a/examples/rust/rideshare/server/src/main.rs
+++ b/examples/rust/rideshare/server/src/main.rs
@@ -30,12 +30,16 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let app_name = std::env::var("PYROSCOPE_APPLICATION_NAME")
         .unwrap_or_else(|_| "rust-ride-sharing-app".to_string());
 
-    let auth_token = std::env::var("PYROSCOPE_AUTH_TOKEN")
+    let auth_user = std::env::var("PYROSCOPE_BASIC_AUTH_USER")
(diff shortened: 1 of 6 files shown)


examples/rust/rideshare/server/src/main.rs
```diff
@@ -30,12 +30,16 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    let auth_user = std::env::var("PYROSCOPE_BASIC_AUTH_USER")
+        .unwrap_or_else(|_| "".to_string());
+
+    let auth_password = std::env::var("PYROSCOPE_BASIC_AUTH_PASSWORD")
         .unwrap_or_else(|_| "".to_string());
 
     // Configure Pyroscope client.
     let agent = PyroscopeAgent::builder(server_address, app_name.to_owned())
-        .auth_token(auth_token)
+        .auth_user(auth_user)
```
Rperry2174, line 41: @korniltsev couldn't find docs on this... is this right? we should also add docs

korniltsev: it should be `        .basic_auth("username", "password")`

korniltsev (approved):
LGTM

note: `PYROSCOPE_AUTH_TOKEN` is used by the og pyroscope examples cluster ingesting data to pyroscope cloud. once we merge the PR it will be harder to update og pyroscope examples cluster

knylander-grafana (approved):
Thank you for updating the java.md file.

PR #2621, opened 2023-11-02
fix: Series query race condition

kolesnikovae:
Discovered by accident while running Pyroscope built with `-race`. I checked other places (`errgroup.Go` arg capturing in a loop), it seems this is the only case


korniltsev (approved):

simonswine (approved):
Thanks for catching this.

LGTM

PR #2622, opened 2023-11-02
fix(ebpf): support python unicode strings

korniltsev:
This adds support for reading ucs-1 ucs-2 ucs-4 function names and file names and utf8 class names
Screenshot_20231102_133235

I hope this will fix a problem causing ingesting invalid utf8 strings


kolesnikovae (approved):
LGTM

Issue #2444, opened 2023-09-22
Profile series pre-aggregation


kolesnikovae:
As it was mentioned in https://github.com/grafana/pyroscope/issues/2443, there are certain write patterns not aligned with the distribution strategy. Specifically, large ingestion traffic may be represented by a single profile series. A typical example is profiling of a serverless workload with [Pyroscope AWS Lambda Extension](https://github.com/grafana/pyroscope-lambda-extension). Please also see https://github.com/grafana/pyroscope/issues/2273.

Besides the low cardinality issue, this write pattern can also cause very poor query performance. It's expected that profiles are sent at some rate, and 1/15s per series is a normal expected value if the profile source is Grafana Agent. Nevertheless, SDK and direct use of the ingestion API allow ingesting thousands of profiles per second. Although Pyroscope is capable of writing this high amount of profiles (millions per hour), querying the data might be impossible even in a large-scale deployments.

I propose to implement pre-aggregation of profiles over time. In the case described above, an individual profile is usually quite small and does not play a big role in itself, therefore aggregating with other profiles of the series within e.g one second on write won't drastically affect the accuracy (should not at all). But it will significantly reduce the read overhead.

kolesnikovae:
The optimal solution would be aggregation and downsampling during / after block compaction.

Issue #2273, opened 2023-08-18
MergeProfilesStacktraces fails with i/o timeout when querying data from 3 hosts @ 100k Hz for 3h, even with a large cluster


luisgerhorst:
#### Describe the bug

When running Pyroscope in the default cluster configuration from the Helm chart, I get the following error when querying the data for 3 hosts (each being profiled at 100k Hz by Grafana Flow Agent) for 3 hours or more:

```
level=warn caller=gokit.go:83 caller=http.go:197 traceID=48..cd msg="GET
       /pyroscope/render?format=json&from=now-3h&query=process_cpu%3Acpu%3Ananoseconds%3Acpu%3Ananoseconds%7Bservice_name%3D%22Linux%22%7D&until=now (500) 30.019743983s Response: \"unknown:
       {\\\"code\\\":\\\"internal\\\",\\\"message\\\":\\\"stream error: stream ID 7; INTERNAL_ERROR; received from peer\\\"}\\n\" ws: false; Accept: */*; Accept-Encoding: gzip, deflate, br; Accept-Language: en-US,en;q=0.5;
       Content-Length: 0; Forwarded: for=<...>;host=<...>.com;proto=https; Referer:
       http://localhost:4040/?query=process_cpu%3Acpu%3Ananoseconds%3Acpu%3Ananoseconds%7Bservice_name%3D%22Linux%22%7D&from=now-3h; Sec-Fetch-Dest: empty; Sec-Fetch-Mode: cors; Sec-Fetch-Site: same-origin; User-Agent:
       Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:109.0) Gecko/20100101 Firefox/115.0; Via: HTTP/1.1 <...>; X-Amzn-Trace-Id: Root=1-<...>; X-Forwarded-Port: 443; "
level=warn caller=gokit.go:83 caller=http.go:178 traceID=48...cd msg="POST /ingester.v1.IngesterService/MergeProfilesStacktraces 44.367168526s, error: 2 errors: i/o timeout; i/o timeout ws: false; Accept-Encoding:
       identity; Connect-Protocol-Version: 1; Content-Type: application/connect+proto; Uber-Trace-Id: <...>:0; User-Agent: connect-go/1.9.0 (go1.19.12); X-Scope-Orgid: anonymous; "
```

In the UI, I get `503 Service Unavailable` for the request to `/pyroscope/render`.

[...]

PR #2623, opened 2023-11-02
feat: profiles pre-aggregation

kolesnikovae:
The PR introduces a new feature that allows pre-aggregating profiles in distributors. Please see #2444:

> As it was mentioned in https://github.com/grafana/pyroscope/issues/2443, there are certain write patterns not aligned with the distribution strategy. Specifically, large ingestion traffic may be represented by a single profile series. A typical example is profiling of a serverless workload with [Pyroscope AWS Lambda Extension](https://github.com/grafana/pyroscope-lambda-extension). Please also see https://github.com/grafana/pyroscope/issues/2273.
> 
> Besides the low cardinality issue, this write pattern can also cause very poor query performance. It's expected that profiles are sent at some rate, and 1/15s per series is a normal expected value if the profile source is Grafana Agent. Nevertheless, SDK and direct use of the ingestion API allow ingesting thousands of profiles per second. Although Pyroscope is capable of writing this high amount of profiles (millions per hour), querying the data might be impossible even in a large-scale deployments.
> 
> I propose to implement pre-aggregation of profiles over time. In the case described above, an individual profile is usually quite small and does not play a big role in itself, therefore aggregating with other profiles of the series within e.g one second on write won't drastically affect the accuracy (should not at all). But it will significantly reduce the read overhead.

The feature is disabled by default, and is controlled with the following parameters (tenant-wide):
* `-distributor.aggregation-window`. Profiles of the same series with timestamps within the window are getting aggregated.
* `-distributor.aggregation-period`. Once an aggregated profile is created, profiles arriving within the duration of the period will be aggregated as well. Note that the period determines the lowest latency of the ingestion requests. 

I'd like to emphasise that the only scenario where this makes sense is server-less profiling, and we should discourage users from using it.

On the other hand, aggregation does mitigate the issue of spiky server-less traffic:

image

image

What is more important is that querying of the aggregated profiles takes substantially less time / resources. For the chart above, it was not possible to query a time range longer than 3 hours of raw data. After enabling aggregation, a 12-hour query takes slightly longer than 25 seconds.


pkg/distributor/aggregator/aggregator.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+// Stop the aggregator. It does not wait for ongoing aggregations
+// to complete as no aggregation requests expected during shutdown.
+func (a *Aggregator[T]) Stop() {
+	close(a.close)
+	<-a.done
+}
+
+type AggregateFn[T any] func(T) (T, error)
```
cyriltovena, line 73: This could be also used during compaction for doing sampling, really like the direction of that code.

pkg/distributor/aggregator/aggregator_test.go
```diff
@@ -0,0 +1,224 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_Aggregation(t *testing.T) {
+	w := time.Second * 15
+	d := time.Millisecond * 10
+
+	fn := func(i int) (int, error) {
```
cyriltovena, line 17: See using generics here, makes your testing life so much easier. Sounds like a good reason for using generics even if you don't specially need it.

pkg/distributor/distributor.go
```diff
@@ -309,6 +313,41 @@ func (d *Distributor) PushParsed(ctx context.Context, req *distributormodel.Push
...
+		}
+	}
+
+	if len(req.Series) == 1 && len(req.Series[0].Samples) == 1 {
+		series := req.Series[0]
+		// Actually all series profiles can be merged before aggregation.
+		// However, it's not expected that a series has more than one profile.
+		profile := series.Samples[0].Profile.Profile
+		var aggregated aggregator.AggregationResult[*pprof.ProfileMerge]
+		if aggregated, err = d.maybeAggregate(tenantID, series.Labels, profile); err != nil {
```
cyriltovena, line 330: Distributors are randomly assigned series of profiles, so I think this affect the effectiveness of aggregating profiles ?

pkg/distributor/aggregator/aggregator.go
```diff
@@ -0,0 +1,278 @@
...
+		t.shards[i] = &shard{v: make(map[uint64]int64, shardSize)}
+	}
+	return &t
+}
+
+func (t *tracker) shard(k uint64) *shard          { return t.shards[k%uint64(len(t.shards))] }
+func (t *tracker) update(k uint64, n int64) int64 { return t.shard(k).update(k, n) }
+
+// prune removes keys with values less than n and
+// reports the number of items remaining.
```
cyriltovena, line 232: doesn't have a return value actually.

pkg/distributor/distributor.go
cyriltovena: >  Note that the period determines the lowest latency of the ingestion requests

I'm also trying to understand something, are you holding all pusher until the aggregated value is pushed ?

cyriltovena (approved):
Overall LGTM.

Agree that this would be much better on the compactor side, but I also see the value for FaaS as you suggested.

pkg/distributor/distributor.go
kolesnikovae: > Distributors are randomly assigned series of profiles, so I think this affect the effectiveness of aggregating profiles ?

That's very true. Therefore aggregation only makes sense with low cardinality data. Moreover, `max_sessions_per_series` should be finely tuned, if enabled. Otherwise, aggregation may not trigger. 

> I'm also trying to understand something, are you holding all pusher until the aggregated value is pushed ?

All queries contributing to the aggregate are blocked until it is processed. This is the only way to guarantee at-least-once delivery: if we fail to send the unit to the receivers, we need to pass the error to the clients so they can handle it accordingly (e.g., try again). However, there is one danger: if a failure occurs on the client side after processing the aggregate and it repeats the sending, this will lead to duplication of data. For a serverless scenario this isn't as important as they should not (or can't) retry.

Now, as I write this, I wonder: If we tell clients not to try again, what's the point of blocking them? Perhaps we should release them immediately if we can guarantee they won't try again.

kolesnikovae:
Thank you for the review @cyriltovena! I'm still in doubts whether we need pre-aggregation or not: I'm not 100% convinced if it's worth it. We should only use it as a last resort. All the aggregations must be done in background.

You pointed out a very important moment – should we block clients or not. Basically, this comes down to a delivery guarantee matter: do we want to ensure at-least-once or at-most-once semantics? Given that we can't ensure idempotent handling, at-least-once does not make much sense, and I guess at-most-once (best effort) is what we actually need.

Frankly speaking, we need to sort out with retries in our SDKs first. For example, AFAIK, Go SDK does not retry.

PR #2624, opened 2023-11-02
fix: disable metadata fetcher on-disk cache

kolesnikovae:
For backward compatibility we decided to not use on-disk cache in the metadata fetcher. It does affect performance, as metas are cached in memory.


simonswine (approved):
LGTM

PR #2629, opened 2023-11-02
chore(DevX): Add lint-staged to launch Prettier when commiting

grafakus:
Just to prevent Prettier to fail the build.


simonswine (approved):
LGTM

PR #2631, opened 2023-11-02
Add profilecli docs

Rperry2174:
adds docs on how to use profilecli

### DIFF
--- a/docs/sources/configure-server/about-server-api.md
+++ b/docs/sources/configure-server/about-server-api.md
@@ -7,18 +7,19 @@ weight: 20
 
 # Pyroscope Server HTTP API Reference
 
-Pyroscope server exposes an HTTP API that can be used to query profiling data and to ingest profiling data from other sources.
+Pyroscope server exposes an HTTP API for querying profiling data and ingesting profiling data from other sources.
 
 ## Authentication
-Grafana Pyroscope does not come with any included authentication layer. Operators are expected to run an authenticating reverse proxy in front of your services.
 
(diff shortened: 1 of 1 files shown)


docs/sources/configure-server/about-server-api.md
```diff
@@ -61,55 +64,119 @@ foo;bar
...
+
+3. **Specify any Extra Labels (Optional).**
+
+   - You can add additional labels to your uploaded profile using the `--extra-labels` flag.
+
+4. **Construct and Execute the Upload Command.**
+
+   - Here's a basic command template:
+     ```
+     ./profilecli upload --url=<pyroscope_server_url> --username=<username> --password=<password> path/to/your/pprof-file.pprof
```
simonswine, line 159: Maybe we should be adding a service_name label so the app UI can handle it?

Potentially we should maybe also just do that as default in profilecli?

simonswine (approved):
LGTM

docs/sources/configure-server/about-server-api.md
Rperry2174: since we don't have docs not actually sure where that goes / how to add it can you make a commit here to add it?

cyriltovena: Created an issue for this https://github.com/grafana/pyroscope/issues/2642

Rperry2174: merged this so @aleks-p can update docs in his PR where this is added

PR #2635, opened 2023-11-03
Helm: Fix persistence for compactor's data path

simonswine:
This will use a subvolume of volume `data` for the compactor instead of a separate volume. This will ensure we are able to handle persistence options in the same way for the compactor.


kolesnikovae (approved):
LGTM

PR #2637, opened 2023-11-03
Use versioned action to update `make-docs` procedure

jdbaldry:
https://github.com/grafana/writers-toolkit/blob/main/update-make-docs/action.yml


simonswine (approved):
LGTM

PR #2641, opened 2023-11-06
Adds rideshare memory leak

Rperry2174:
Adding memory for example purposes 
image

### DIFF
--- a/examples/golang-push/rideshare/docker-compose.yml
+++ b/examples/golang-push/rideshare/docker-compose.yml
@@ -6,8 +6,8 @@ services:
     environment:
       - REGION=us-east
       - PYROSCOPE_SERVER_ADDRESS=http://pyroscope:4040
-      - PARAMETERS_POOL_SIZE=1000
-      - PARAMETERS_POOL_BUFFER_SIZE_KB=1000
+      - PARAMETERS_POOL_SIZE=10000
+      - PARAMETERS_POOL_BUFFER_SIZE_KB=10000
     build:
       context: .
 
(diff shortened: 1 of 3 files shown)


examples/golang-push/rideshare/docker-compose.yml
```diff
@@ -6,8 +6,8 @@ services:
     environment:
       - REGION=us-east
       - PYROSCOPE_SERVER_ADDRESS=http://pyroscope:4040
-      - PARAMETERS_POOL_SIZE=1000
-      - PARAMETERS_POOL_BUFFER_SIZE_KB=1000
+      - PARAMETERS_POOL_SIZE=10000
+      - PARAMETERS_POOL_BUFFER_SIZE_KB=10000
```
bryanhuhta, line 10: Adjusting the sizes here will work when running locally, but in order to get these to reflect in sedemo, you'll need to update them [here](https://github.com/grafana/deployment_tools/blob/f016110f8fcf6d82f51bab49a2c623b31d9932bd/ksonnet/environments/pyroscope-rideshare/main.jsonnet#L55-L56). There's additional knobs to tune there as well. The memory limit is 64Mi per pod, so using sizes this big would OOM those pods right away.

Of course, if this is just intended to be shown with a local version of the rideshare demo then adjusting these settings will but 👌

bryanhuhta (approved):
LGTM 👍  added one note on adjusting parameters

aleks-p:
This is out of date and it doesn't look like we intend on merging it. Feel free to re-open and update if you disagree.

PR #2643, opened 2023-11-06
chore(ebpf): add  python e2e tests

korniltsev:
- add e2e python tests running rideshare with different python versions in docker, profiling it. Requires root.
- allow to run these tests in qemu (not used in ci because it is slow). May be useful to run tests on darwin or to use specific linux kernel
- changes python profiler to skip "trampoline" frames in ebpf, not userspace
- changes python profiler to skip "__init__ __init__" frames in userspace


simonswine (approved):
Nice work! LGTM

Not too sure if this example even works, but it might be useful to only run it on changes to `ebpf/**`

https://github.com/orgs/community/discussions/25669#discussioncomment-3606300

PR #2652, opened 2023-11-07
fix(jfr): merge equal samples 

korniltsev:
Before this PR a long running profile could contain N equal samples with value=1 
After this PR we use jfr stacktrace id (StackTraceRef) to find equal samples and add values to an existing sample, giving us just one sample with value=N

This will produces smalle pprofs and make it faster, especially for long running profiles


kolesnikovae (approved):
LGTM

Issue #2650, opened 2023-11-06
[DOC] Extend profilecli docs


aleks-p:
_Originally posted by @Rperry2174 in https://github.com/grafana/pyroscope/issues/2645#issuecomment-1795717619_

I have a few suggestions that could help improve the clarity and usability of the docs for users across different platforms.

### Installation Instructions:
Could you clarify the installation steps for profile-cli? It would be beneficial to include explicit instructions on how users can install this utility on their system.
- For macOS users, is it `brew install pyroscope-io/brew/pyroscope` or brew `install pyroscope-io/brew/profile-cli` or both?
- For Linux users or non-macOS CI pipelines, could you provide the equivalent installation commands?

Adding a section that walks through the installation process before diving into the usage would be helpful.

### Working Example:
A working example is always helpful as well. Could you add a complete, executable example to the [examples/api folder](https://github.com/grafana/pyroscope/tree/main/examples/api) on GitHub? This would give users a quick way to see profile-cli in action.

### Separate Installation Documentation:
The installation instructions might be better suited for a separate page. While this could be a subsequent improvement, let's ensure that the current document is as complete and stand-alone as possible. We can later refactor the content into separate, more detailed page

The goal is to make the documentation intuitive enough so that when users ask "How do I install/use profile-cli?", they can be directed to both the docs and a comprehensive guide that works for their specific setup without additional explanations

knylander-grafana:
@aleks-p Do you think this issue is finished? You've done some great work towards it!

aleks-p:
Yeah, closing as completed. Thank you for following up!

PR #2654, opened 2023-11-07
Move profilecli docs to a dedicated page, add installation instructions 

aleks-p:
Improves profilecli docs as suggested in #2650 

A few remarks (cc. @Rperry2174):

1. I kept the installation instructions simple for now. We can technically use package managers on Linux as well but in my tests they also start a Pyroscope server as well which is not what we want in this context.
2. After looking at the [examples/api folder](https://github.com/grafana/pyroscope/tree/main/examples/api) more carefully, I am not sure examples for a CLI tool belong there. One option could be a bash script, perhaps in a new `examples/cli` directory. This will however be a one-liner that repeats what the docs already say. Any thoughts?

cc. @knylander-grafana


Rperry2174:
> After looking at the [examples/api folder](https://github.com/grafana/pyroscope/tree/main/examples/api) more carefully, I am not sure examples for a CLI tool belong there. One option could be a bash script, perhaps in a new examples/cli directory. This will however be a one-liner that repeats what the docs already say. Any thoughts?

Good point. Let's leave it like this for now.. I do think we should still at least mention and link to this page somewhere on the "server api" page since a lot of people may start on server api page and ultimately what they're really looking for / need is profilecli

kolesnikovae:
It would be great to also add instructions for `query` sub-command – that is super useful if one wants to integrate PGO into a CI pipeline

docs/sources/configure-server/profile-cli.md
```diff
@@ -0,0 +1,127 @@
...
+weight: 80
+---
+
+# Profile CLI
+
+`profilecli` is a command-line utility that enables various productivity flows such as:
+- interacting with a running Pyroscope Server to upload profiles manually, run queries, etc.
+- inspecting [Parquet](https://parquet.apache.org/docs/) files.
+
+Hint: Use the `help` command (`profilecli help`) to get a full list of capabilities as well as additional help information.
```
cyriltovena, line 17:
```suggestion
> Hint: Use the `help` command (`profilecli help`) to get a full list of capabilities as well as additional help information.
```

aleks-p:
@Rperry2174 @kolesnikovae I added a short section in the server API page with a link to the profile CLI docs
 as well as instructions for the `query` command. Feel free to update any wording as needed.

Also, in my last commit I switched from flags to environment variables for `url`, `username` and `password` because this is how we are using the commands from what I have seen and is usually the recommended way. Let me know if you have any concerns.

docs/sources/configure-server/profile-cli.md
```diff
@@ -0,0 +1,234 @@
...
+```
+
+### Other
+
+Download the `profilecli`` release asset from https://github.com/grafana/pyroscope/releases/latest for your operating system and architecture and make it executable.
+
+For example, for Linux with the AMD64 architecture:
+
+```bash
+curl -fL https://github.com/grafana/pyroscope/releases/download/v1.0.0/profilecli_1.0.0_linux_amd64.tar.gz | tar xvz
```
cyriltovena, line 34: Should we point to 1.1.5 ?

docs/sources/configure-server/profile-cli.md
```diff
@@ -0,0 +1,234 @@
...
+
+You can use the `profilecli query merge` command to retrieve a merged (aggregated) profile from a Pyroscope server. The command merges all samples found in the profile store for the specified query time range. By default it looks for samples within the last hour, though this can be controlled with the `--from` and `--to` flags. The source data can be narrowed down with the `--query` flag, in the same way as with the `series` command.
+
+#### Query Merge Steps
+
+1. **Specify optional flags.**
+
+   - You can provide a label selector using the `--query` flag (e.g., `--query='{service_name="my_application_name"}'`)
+   - You can provide a custom time range using the `--from` and `--to` flags (e.g., `--from="now-3h" --to="now"`)
+   - You can specify the profile type via the `--profile-type` flag. The available profile types are listed in the output of the `profilecli query series` command.
```
cyriltovena, line 208: Makes me think we should make this simpler, with flag like `-mem` or `-cpu` to simplify the usage of this.

We should also most likely show using go how to open and read those merged profiles. I suspect most people don't know how to use pprof.

cyriltovena (approved):
LGTM, left some suggestions for follow up.

docs/sources/configure-server/profile-cli.md
aleks-p: did it for now, though this will get out of date quickly unless we generate these parts of the docs

PR #2658, opened 2023-11-09
Make max flamegraph nodes limit configurable

kolesnikovae:
Please also see https://github.com/grafana/pyroscope-app-plugin/issues/170


cyriltovena (approved):
LGTM

PR #2660, opened 2023-11-09
Make deduplication decision more robust

simonswine:
A follow up to  #2586, which will make the deduplication more robust.

(All blocks should have Compaction, a compaction Level of 0 means a head blcok)


kolesnikovae (approved):
LGTM

PR #2661, opened 2023-11-09
fix: reduce redundant tag explorer fetches

darrenjaneczek:
- Corrected the dependency list and boolean check for a useEffect which resulted in an additional fetch under certain conditions
  - The condition was when `groupBy` and `groupByValue` URL parameters are undefined (but everything else is defined)

### DIFF
--- a/public/app/pages/TagExplorerView.tsx
+++ b/public/app/pages/TagExplorerView.tsx
@@ -193,10 +193,10 @@ function TagExplorerView() {
       return () => fetchData.abort('cancel');
     }
     return undefined;
-  }, [from, until, query, groupByTagValue, dispatch]);
+  }, [from, until, query, groupByTagValue, refreshToken, dispatch]);
 
   useEffect(() => {
-    if (from && until && query) {
+    if (from && until && query && groupByTag) {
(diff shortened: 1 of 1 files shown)


public/app/pages/TagExplorerView.tsx
```diff
@@ -193,10 +193,10 @@ function TagExplorerView() {
       return () => fetchData.abort('cancel');
     }
     return undefined;
-  }, [from, until, query, groupByTagValue, dispatch]);
+  }, [from, until, query, groupByTagValue, refreshToken, dispatch]);
```
darrenjaneczek, line 196: Allows the fetch above to be re-fetched in the event of an error or some hang

public/app/pages/TagExplorerView.tsx
```diff
@@ -193,10 +193,10 @@ function TagExplorerView() {
...
 
   useEffect(() => {
-    if (from && until && query) {
+    if (from && until && query && groupByTag) {
```
darrenjaneczek, line 199: Waits for `groupByTag` to be defined before trying to fetch, avoiding a redundant fetch when this is undefined.

Rperry2174 (approved):
lgtm

PR #2663, opened 2023-11-10
feat: compaction staging

kolesnikovae:
Number of compactor split shards is limited to the number of blocks we can write concurrently. There are two significant factors:
 - Symbols size in memory. Varies depending on the sharding key.
 - Memory utilized by parquet writer / row group buffer. Varies depending on the row group size and contents.

I propose to compact blocks in stages to circumvent these limitations. Potentially, we can mitigate them (and we will, I believe), however we can't eliminate them.

There are multiple ways to implement staging. The PR contains the simplest of them: at compaction, we only populate limited number of output blocks (stage size), and re-read source blocks as many times as many stages specified. For example, if we have 16 blocks at input, 64 split shards, and stage size is 8, we re-read (re-reprocess) source blocks 8 times, but only 8 output blocks are being populated at any point in time.

The work is not complete and has not been tested. I'd like to discuss the subject before investing more efforts.

/cc @cyriltovena @simonswine

### DIFF
--- a/cmd/pyroscope/help-all.txt.tmpl
+++ b/cmd/pyroscope/help-all.txt.tmpl
@@ -131,10 +131,10 @@ Usage of ./pyroscope:
     	Minimum time to wait for ring stability at startup. 0 to disable.
   -compactor.split-and-merge-shards int
     	The number of shards to use when splitting blocks. 0 to disable splitting.
+  -compactor.split-and-merge-stage-size int
+    	Number of stages split shards will be written to. Number of output split shards is controlled by -compactor.split-and-merge-shards.
   -compactor.split-groups int
     	Number of groups that blocks for splitting should be grouped into. Each group of blocks is then split separately. Number of output split shards is controlled by -compactor.split-and-merge-shards. (default 1)
-  -compactor.stage-size int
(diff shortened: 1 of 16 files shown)


simonswine (commented):
I think it makes sense to go ahead with this. I am not exactly sure how much impact to overall memory consumption this will have, but that's something we can find out putting some real data to it.

knylander-grafana (commented):
Thank you for adding doc.

kolesnikovae:
> Please test this in ops, I think it's a great idea.

@cyriltovena, do you know how long it takes to compact data for last 24 hours there? I deployed it yesterday (32 instances, 32 split shards, stage size 8, group size 16), but it only compacted 2-3 days of data since then.

The intermediate results showed that:
 - It's hard to tell if it really helps reducing memory consumption (I still see peaks up to 32GB), but it does make behaviour predictable. It was running for 24 hours without OOMs.
 - A compaction job takes appx. twice longer than a job with 8 split shards and no staging. Either way, I believe we can significantly speed up compaction process to workaround this.

cyriltovena (approved):
LGTM

Issue #2644, opened 2023-11-06
Panic when resolving symbols


cyriltovena:
Found this in our logs:

```


[...]

PR #2665, opened 2023-11-10
Fix symbols resolver race condition

kolesnikovae:
Fixes https://github.com/grafana/pyroscope/issues/2644: the problem was that, on context cancellation, a partition could be released while still being used by resolver.


cyriltovena (approved):
Nice yes this happens only when closing early

PR #2666, opened 2023-11-10
Allow to configure pyroscope high disk utilization

wilfriedroset:


cmd/pyroscope/help-all.txt.tmpl
```diff
@@ -467,8 +467,14 @@ Usage of ./pyroscope:
     	Minimum time to wait for ring stability at startup, if set to positive value. Set to 0 to disable.
   -pyroscopedb.data-path string
     	Directory used for local storage. (default "./data")
+  -pyroscopedb.enforcement-interval duration
+    	How often to enforce disk retention (default 5m0s)
```
kolesnikovae, line 471:
```
-pyroscopedb.enforcement-interval
```
Should the name be more specific, like `retention-policy-enforcement-interval`? Maybe we also could keep this prefix for all the related settings for better discoverability

docs/sources/configure-server/configure-disk-storage.md
```diff
@@ -29,12 +29,13 @@ uploaded to the object store bucket.
...
-close to running out of disk. This high utilization mode will be active when:
+To avoid losing the most recent data, Pyroscope will remove the oldest blocks
+when it detects that the volume on which the data path is located is close to
+running out of disk. This high utilization mode will be active every
+`-pyroscopedb.enforcement-interval` when:
 
-* less than 5% of the total size of the volume is available and
-* the available disk space smaller then 10GiB.
+* less than `-pyroscopedb.min-disk-available-percentage=0.05` of the total size of the volume is available and
+* the available disk space smaller then `-pyroscopedb.min-free-disk=10GiB`.
```
kolesnikovae, line 38: We probably should not specify value unit here, as the flag actually expects size in bytes – users will likely copy this and change the number, but it won't be a valid value. Should we maybe change the flag unit to GB/MB (e.g. `min-free-disk-gb`)?

pkg/ingester/ingester.go
```diff
@@ -114,7 +114,19 @@ func New(phlarectx context.Context, cfg Config, dbConfig phlaredb.Config, storag
...
+		retentionPolicy.EnforcementInterval = dbConfig.EnforcementInterval
+	}
+	if dbConfig.MinFreeDisk > 0 {
+		retentionPolicy.MinFreeDisk = dbConfig.MinFreeDisk
+	}
+	if dbConfig.MinDiskAvailablePercentage > 0 {
+		retentionPolicy.MinDiskAvailablePercentage = dbConfig.MinDiskAvailablePercentage
+	}
+
+	rpEnforcer := newRetentionPolicyEnforcer(phlarecontext.Logger(phlarectx), i, retentionPolicy, dbConfig)
```
kolesnikovae, line 129: If we make the parameters configurable, do you think it makes sense to allow disabling RP enforcement? I guess in certain cases it may (e.g if the data dir has multiple mount points)

kolesnikovae (approved):
Thank you so much for the contribution @wilfriedroset! I've left a few minor notes, otherwise LGTM 🚀

PR #2669, opened 2023-11-13
Usage stats per language

aleks-p:
Adds metrics for number of profiles and profile size (uncompressed) broken down per language to the existing usage stats. An example of the data structure (note that the drill-down is an array to simplify querying):

```json
{
  "distributor_profiles_received": {
    "total": 25,
    "rate": 2.3,
    "drilldown": [
      {
        "lang": "go",
        "data": {
          "total": 13,
          "rate": 1.3
      },
      {
        "lang": "java",
        "data": {
          "total": 12,
          "rate": 1.2
      }
    ]
  }
}
```

Here is a dashboard showing how we could visualize it (just as an example): 

usage-per-language


pkg/distributor/distributor.go
```diff
@@ -233,6 +235,13 @@ func (d *Distributor) Push(ctx context.Context, grpcReq *connect.Request[pushv1.
 	return d.PushParsed(ctx, req)
 }
 
+func (d *Distributor) GetProfileLanguage(series *distributormodel.ProfileSeries) string {
+	if len(series.Samples) == 0 {
+		return "unknown"
+	}
+	return pprof.GetLanguage(series.Samples[0].Profile, d.logger)
```
aleks-p, line 242: Once the SDKs starts sending the language and more information such as build information this can be adapted to read from the request and fallback to the regex matchers.

pkg/usagestats/stats.go
```diff
@@ -340,6 +346,92 @@ func (s *Statistics) Record(v float64) {
 	}
 }
 
+type MultiStatistics struct {
+	values  map[string]*Statistics
+	keyName string
```
aleks-p, line 351: We can currently place one "key" (label) on MultiStatistics and MultiCounters for simplicity, in the future this can be extended if needed.

cyriltovena (approved):
LGTM

PR #2673, opened 2023-11-13
Merges stacktraces without sorting and pulling labels.

cyriltovena:
This PR adds an alternative ways of merging stacktraces without sorting rows and extracting labels values.

This only possible when there's no need to deduplicate blocks.

I had some trouble along the way with memory management, so I ended up still using a slice iterator at the end. It seems that parquet iterator are not compatible with our tee iterator.


pkg/phlaredb/block_querier.go
```diff
@@ -840,26 +840,8 @@ func MergeProfilesStacktraces(ctx context.Context, stream *connect.BidiStream[in
...
-				if err != nil {
-					return err
-				}
-
-				if len(profiles) == 0 {
-					return nil
-				}
-
 				// TODO(simonswine): Split profiles per row group and run the MergeByStacktraces in parallel.
-				merge, err := querier.MergeByStacktraces(ctx, iter.NewSliceIterator(querier.Sort(profiles)))
```
cyriltovena, line 862: Basically sums up the this PR. No need to order by timestamp than series via slices then back to series then timestamp via querier.Sort.

pkg/phlaredb/block_querier.go
```diff
@@ -1541,6 +1523,134 @@ func (b *singleBlockQuerier) SelectMatchingProfiles(ctx context.Context, params
...
+		b.profiles.columnIter(ctx, "SeriesIndex", query.NewMapPredicate(lblsPerRef), "SeriesIndex"),
+		b.profiles.columnIter(ctx, "TimeNanos", query.NewIntBetweenPredicate(model.Time(params.Start).UnixNano(), model.Time(params.End).UnixNano()), "TimeNanos"),
+	)
+
+	if b.meta.Version >= 2 {
+		it = query.NewBinaryJoinIterator(0,
+			it,
+			b.profiles.columnIter(ctx, "StacktracePartition", nil, "StacktracePartition"),
+		)
+	}
```
simonswine, line 1575: I think it is really easy to miss one case when updating the meta verison in the future, it would be better to share a function call with

https://github.com/grafana/pyroscope/blob/49e1ee75ccad9ef2e5257fcaf67abbbd793bbaab/pkg/phlaredb/block_querier.go#L1469-L1484

simonswine (approved):
LGTM

pkg/phlaredb/block_querier.go
cyriltovena: 100%

cyriltovena:
This will stay there a bit not planning to get it in the next release for now. I think it needs more work.

PR #2677, opened 2023-11-14
fix(ebpf): use correct sys_execve kprobe on arm64

korniltsev:


ebpf/session.go
```diff
@@ -691,10 +692,17 @@ func (s *session) linkKProbes() error {
...
 		required bool
 	}
-	var hooks = []hook{
+	var hooks []hook
+	archSys := ""
+	if "amd64" == runtime.GOARCH {
+		archSys = "__x64_"
+	} else {
+		archSys = "__arm64_"
+	}
```
kolesnikovae, line 701: If we're not going to add support for some other architectures soon 😅

kolesnikovae (approved):
LGTM

ebpf/session.go
korniltsev: i hope no

Issue #2228, opened 2023-08-10
Function metric


kolesnikovae:
It would be useful to visualize the time series for a particular function (a node of the flamegraph). Ideally, grouping by labels should be supported as well. Currently, the metric always shows the "total" value.

![image](https://github.com/grafana/pyroscope/assets/12090599/c063fd72-b2db-4035-97c1-450521f45eec)

cyriltovena:
We should consider extending the current SelectSeries to include a list of function to filter the result with. `[]string`

We should most likely only resolve symbols when the filtering is provided.

PR #2679, opened 2023-11-14
feat: optimized pprof symbolication

kolesnikovae:
The main goal of the change is to make generated pprof proto type the only profile representation we deal with in the backend:
 - Use of [profile.Profile](https://github.com/google/pprof/blob/4ca4178f5c7a/profile/profile.go#L34) does not allow implementing node truncation.
 - Tree representation we currently use for building a flamegraph does not allow including location information. Synthetic benchmarks show that use of pprof proto Profile decreases CPU time needed to built a profile by up to 300% (1.5s -> 350ms), and allocated memory by up to 100%.

Next steps:
 1. https://github.com/grafana/pyroscope/issues/2728
 2. https://github.com/grafana/pyroscope/issues/2729
 3. Switch to pprof in `SelectMergeStacktraces` API

Dependent issues:
 1. #2727 
 2. #2228


knylander-grafana:
@kolesnikovae This looks like an interesting feature. Should we add doc for this?

cyriltovena:
> 3\. Switch to pprof in `SelectMergeStacktraces` API

This is more an performance optimisation so far so not much to document yet.

cyriltovena (approved):
LGTM

kolesnikovae:
> @kolesnikovae This looks like an interesting feature. Should we add doc for this?

The change only affects internals with no impact on the end user in terms of the system behaviour and mechanics. This is a preparation step for some of the features we want to implement (for which we will need docs, indeed)

PR #2684, opened 2023-11-14
fix FQDN to short names

ngnix:
Hello grafana pyroscope team!

This PR fix used FQDN with hardcoded domain `cluster.local`.  
Main idea - don't use FQDN if you deploy resources, use svc short names.

### DIFF
--- a/operations/pyroscope/helm/pyroscope/templates/_helpers.tpl
+++ b/operations/pyroscope/helm/pyroscope/templates/_helpers.tpl
@@ -91,3 +91,16 @@ If no components are defined fall back to single binary statefulset
 all: {kind: "StatefulSet", name: "{{$full_name}}"}
 {{- end }}
 {{- end }}
+
+
+{{- define "cluster.domain" }}
+{{- if .Values.cluster_domain  }}
+{{- if hasSuffix "." .Values.cluster_domain }}
+{{- .Values.cluster_domain | replace " " "" -}}
+{{- else }}
+{{- cat .Values.cluster_domain "." | replace " " "" -}}
+{{- end }}
+{{- else }}
+{{- printf "cluster.local." -}}
+{{- end }}
+{{- end }}
--- a/operations/pyroscope/helm/pyroscope/templates/_helpers.tpl
+++ b/operations/pyroscope/helm/pyroscope/templates/_helpers.tpl
@@ -91,16 +91,3 @@ If no components are defined fall back to single binary statefulset
 all: {kind: "StatefulSet", name: "{{$full_name}}"}
 {{- end }}
 {{- end }}
-
-
-{{- define "cluster.domain" }}
-{{- if .Values.cluster_domain  }}
-{{- if hasSuffix "." .Values.cluster_domain }}
-{{- .Values.cluster_domain | replace " " "" -}}
-{{- else }}
-{{- cat .Values.cluster_domain "." | replace " " "" -}}
-{{- end }}
-{{- else }}
-{{- printf "cluster.local." -}}
-{{- end }}
-{{- end }}
--- a/operations/pyroscope/helm/pyroscope/templates/deployments-statefulsets.yaml
+++ b/operations/pyroscope/helm/pyroscope/templates/deployments-statefulsets.yaml
@@ -64,7 +64,7 @@ spec:
             - "-self-profiling.disable-push=true"
             - "-server.http-listen-port={{ $values.service.port }}"
             - "-memberlist.cluster-label={{ .Release.Namespace }}-{{ include "pyroscope.fullname" .}}"
-            - "-memberlist.join=dns+{{ include "pyroscope.fullname" .}}-memberlist.{{ .Release.Namespace }}.svc:{{ .Values.pyroscope.memberlist.port }}"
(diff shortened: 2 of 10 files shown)


simonswine:
I am not too sure if I would approve this change:

The reason we use the FQDN together with the `.` in the end is so that the DNS resolver is not using the DNS search path at all and can answer the query after the first try.

In your example there will be more request required to resolve the hostname, by having to append the search paths in order.

I would be happy to accept a PR which makes the domain suffix configurable, but with the default `.cluster.local.`, hope that makes sense

ngnix:
On the topic of your comment about the use of short DNS names - it seems to me that this is not a problem since DNS discovery is not performed so often that it would harm DNS servers, but the method itself is universal and does not require explicitly specifying the domain.

It is much more unpleasant to see after redeploy how the pods of any pyroscope continue to access IP addresses that belonged to the cluster components more than an hour ago, and now belong to pods that are not related to the pyroscope. 

This behavior has been observed on all grafana labs products that use the dns discovery mechanism.

simonswine:
> On the topic of your comment about the use of short DNS names - it seems to me that this is not a problem since DNS discovery is not performed so often that it would harm DNS servers, but the method itself is universal and does not require explicitly specifying the domain.

Yes this becomes a problem when you run hundreds of pods

Why is the domain name incorrect in the first place? What is the correct FQDN in your case?

ngnix:
We use k8s cluster mesh topology, and all k8s cluster have uniq cluster domain based on geolocation,infrastructure provider name and cluster project, like `sng-aws-stage.k8s.lcl`. 

Please look at the last commit, maybe you will agree with it.

ngnix:
`cluster_domain` located on top of values because used in both charts - `pyroscope` and `grafana-agent`.

ngnix:
@simonswine Christian Do you have any updates ?

ngnix:
@simonswine Christian Do you have any updates ?

operations/pyroscope/helm/pyroscope/templates/deployments-statefulsets.yaml
```diff
@@ -63,7 +63,7 @@ spec:
             - "-self-profiling.disable-push=true"
             - "-server.http-listen-port={{ $values.service.port }}"
             - "-memberlist.cluster-label={{ .Release.Namespace }}-{{ include "pyroscope.fullname" .}}"
-            - "-memberlist.join=dns+{{ include "pyroscope.fullname" .}}-memberlist.{{ .Release.Namespace }}.svc.cluster.local.:{{ .Values.pyroscope.memberlist.port }}"
+            - "-memberlist.join=dns+{{ include "pyroscope.fullname" .}}-memberlist.{{ .Release.Namespace }}.svc.{{ include "cluster.domain" . }}:{{ .Values.pyroscope.memberlist.port }}"
```
simonswine, line 66: If I understood you correctly you wanted to be able to set things to `pyroscope-memberlist.my-namespace.svc`, we then need to remove the extra dot into the values file:

```suggestion
            - "-memberlist.join=dns+{{ include "pyroscope.fullname" .}}-memberlist.{{ .Release.Namespace }}.svc{{ include "cluster.domain" . }}:{{ .Values.pyroscope.memberlist.port }}"
```

operations/pyroscope/helm/pyroscope/templates/_helpers.tpl
```diff
@@ -91,3 +91,16 @@ If no components are defined fall back to single binary statefulset
 all: {kind: "StatefulSet", name: "{{$full_name}}"}
 {{- end }}
 {{- end }}
+
+
+{{- define "cluster.domain" }}
```
simonswine, line 96: I am not too sure at why we require those replace statements.

In my opinion a normal values.yaml variable should be enough to allow people to override it.

operations/pyroscope/helm/pyroscope/values.yaml
```diff
@@ -2,6 +2,9 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
+# Override default K8S cluster domain for DNS discovery
+# cluster_domain:
```
simonswine, line 6: I think it should life within pyroscope, we can argue that everything within `pyroscope`, should be global yes, but that is now very hard to change.

Also this should set the default:

```suggestion
# Override default K8S cluster domain for DNS discovery
cluster_domain: ".cluster.local."
```

simonswine (commented):
> cluster_domain located on top of values because used in both charts - pyroscope and grafana-agent.

This doesn't work like this, because for the agent it has to be within its scope, we need to have it set again under `agent.cluster_domain: xxx`

julienvey:
Facing the same issue here. We are using a custom domain for our k8s clusters and we had to manually patch the helm chart to make it work. 

@ngnix If you need any help to fix the reviews from Christian, please let me know.

simonswine:
Sorry for the slow progress on this, I had been off work for a bit. I have just pushed the changes, that should be ready to be merged.

Just wondering if it works for you @julienvey @ngnix like this?

simonswine (approved):
LGTM

julienvey:
Hi @simonswine This look goods ! Thank you very much 👍

PR #2685, opened 2023-11-14
Handle block plans correctly when splitToStores

simonswine:
When a plan has been acquired, we need to make sure we do query all underlying stores. If a particular replica from one store is not part of the plan, we will eliminate this later in `forPlannedIngesters|StoreGateways(...)`.


cyriltovena (approved):
LGTM

PR #2690, opened 2023-11-15
fix: stats race condition

kolesnikovae:
I guess a better solution would be using a flat `sync.Map` of atomics (or just pointers to integers) with structured keys (`{dimension string, stat string}`), and upserting new values with `LoadOrUpdate`. However, I don't think the optimisation is warranted at this point


cyriltovena (approved):
LGTM

PR #2691, opened 2023-11-15
fix: flaky aggregation test

kolesnikovae:
Discovered in https://github.com/grafana/pyroscope/actions/runs/6872731032/job/18691976067


cyriltovena (approved):
LGTM

PR #2692, opened 2023-11-15
fix(ReduxQuerySync): Allow unsubscribing from ReduxQuerySync

grafakus:
This prevents unpredictable behaviours when Pyroscope is embedded in a larger React application (e.g some URL query parameters could be reseted to their default value).


public/app/redux/redux-query-sync.ts
```diff
@@ -0,0 +1,109 @@
+import { actions as tracingActions } from '@pyroscope/redux/reducers/tracing';
+import { history } from '@pyroscope/util/history';
+import ReduxQuerySync from 'redux-query-sync';
+import { actions as continuousActions } from './reducers/continuous';
+import { RootState, StoreType } from './store';
+
+let unsubscribeFromReduxQuerySync: unknown;
```
darrenjaneczek, line 7: I am not expecting that this should appear outside of react. Maybe we can make this into a hook, and the useEffect in the hook here can keep the `unsubscribeFromReduxQuerySync` and return it.

Then the `App` just needs to use the hook.

grafakus: Good point, I'll make the changes!

public/app/app.tsx
```diff
@@ -1,26 +1,29 @@
 import React from 'react';
-import ReactDOM from 'react-dom/client';
+import { Sidebar } from '@pyroscope/components/Sidebar';
```
grafakus, line 2: Prettier...

public/app/redux/useReduxQuerySync.ts
```diff
@@ -0,0 +1,104 @@
...
+    },
+    initialTruth: 'location',
+    replaceState: false,
+    history,
+  });
+}
+
+export function useReduxQuerySync() {
+  useEffect(setupReduxQuerySync, []);
+}
```
darrenjaneczek, line 104: Tidy!

public/app/redux/useReduxQuerySync.ts
```diff
@@ -0,0 +1,104 @@
+import { useEffect } from 'react';
+import { actions as tracingActions } from '@pyroscope/redux/reducers/tracing';
+import { history } from '@pyroscope/util/history';
+import ReduxQuerySync from 'redux-query-sync';
+import { actions as continuousActions } from './reducers/continuous';
+import store, { RootState } from './store';
+
+function setupReduxQuerySync() {
```
darrenjaneczek, line 8: I like how this isn't being exported.

darrenjaneczek (approved):
Excellent. Thanks!

PR #2694, opened 2023-11-15
feat: query pprof from store-gateway

kolesnikovae:
The PR adds support for querying pprof from store-gateways. The implementation relies on the physical plan, if available.


pkg/phlaredb/block_querier.go
```diff
@@ -739,6 +739,7 @@ func SelectMatchingProfiles(ctx context.Context, request *ingestv1.SelectProfile
 
 	for i, querier := range queriers {
 		if skipBlock(querier.BlockID()) {
+			iters[i] = iter.NewEmptyIterator[Profile]()
```
cyriltovena, line 742: ^^ smart

cyriltovena (approved):
LGTM

Issue #2541, opened 2023-10-16
Add a block view component


cyriltovena:
It could be useful to have a special component that shows blocks on a timeline scale with their range.

The component would only be used to render `/admin` pages for block listing but could evolve in the future for managing tenant's block.

@simonswine Recently showed me the Thanos block UI

<img width="2172" alt="image" src="https://github.com/grafana/pyroscope/assets/1053421/abdbdd9d-fcb8-4d5d-85e3-3a0315b2379e">

You can try it out using his branch here https://github.com/thanos-io/thanos/compare/main...simonswine:thanos:20231005_read-pyroscope-blocks?expand=1

```
go run ./cmd/thanos  tools bucket web   --log.level=debug --objstore.config='{"type":"GCS", "config": {"bucket": "dev-us-central-0-profiles-dev-001-data"}, "prefix":"1218/phlaredb/"}' --refresh=10m
```

This could be a simple component that uses the `bucketindex` and shows Thanos block UI per tenant.

I'd love to also see the labels on the view when clicking a block.

This would make operations of Pyroscope cluster a bit easier than running a fork locally.

PR #2697, opened 2023-11-15
Add the first iteration of the blocks viewer cli tool

aleks-p:
Addresses #2541

Usage:
1. `profilecli tool blocks web --bucket-name <object_store_bucket_name>`
1. Open `http://localhost:4201/blocks/index` in a browser (the URL is also printed to stdout)

Notes:
- data comes from a object store bucket, though some things are cached on disk (e.g., the meta.json data)
- will use existing user credentials (tested with GCS only)
- blocks are grouped by "min time" to make them easier to navigate
- block groups are sorted by "min time" (descending), blocks within a group are sorted by "upload time" (descending)

Next steps (outside of this PR): 
- porting it to an admin page
- adding a visualization that can show overlapping blocks

https://github.com/grafana/pyroscope/assets/8142643/6ec8fa43-bd52-41a9-b070-61a09f6f9f24


cmd/profilecli/blocks_tool.go
```diff
@@ -0,0 +1,365 @@
...
+	objectStoreType string
+	bucketName      string
+}
+
+func addBlocksWebToolParams(blocksWebToolCmd commander) *blocksWebToolParams {
+	var (
+		params = &blocksWebToolParams{}
+	)
+	blocksWebToolCmd.Flag("object-store-type", "The type of the object storage (e.g., gcs).").Default("gcs").StringVar(&params.objectStoreType)
+	blocksWebToolCmd.Flag("bucket-name", "The name of the object storage bucket.").StringVar(&params.bucketName)
```
simonswine, line 55: I really would prefer if we could mimic flags from pyroscope itself and support all the options by calling `RegisterFlags`.

Unsure that might be also incredibly hard

cmd/profilecli/blocks_tool.go
```diff
@@ -0,0 +1,365 @@
...
+		Stats:           meta.Stats,
+		Labels:          meta.Labels,
+	}
+}
+
+func (t *blocksWebTool) run(ctx context.Context) error {
+	out := output(ctx)
+
+	fmt.Fprintf(out, "The blocks web tool is available at http://localhost:%d/blocks/index\n", t.params.httpListenPort)
+
```
simonswine, line 359: Nice to have: Potentially open the browser automatically: https://stackoverflow.com/questions/39320371/how-start-web-server-to-open-page-in-browser-in-golang

simonswine (commented):
Nice work! LGTM, just a a few comments, I don't think anything is really blocking before merging, but have a look and adapt things that make sense to you 🙂

simonswine (approved):
Ticked the wrong box

pkg/admin/admin.go
```diff
@@ -0,0 +1 @@
+package admin
```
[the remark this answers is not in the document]
aleks-p: No, that is a leftover from something else, good catch!

cmd/profilecli/blocks_tool.go
aleks-p: I tried but it doesn't seem simple enough to do without a bigger refactor, can plan this for later though.

cmd/profilecli/blocks_tool.go
aleks-p: That is a good idea and perhaps it needs to be behind a flag if someone finds getting multiple tabs from multiple runs annoying. One issue with most solutions though is they are not robust enough. For example, to open the browser only when the server is successfully up we need to do something like https://stackoverflow.com/questions/32738188/how-can-i-start-the-browser-after-the-server-started-listening/32742904#32742904 but we can't since we don't start the server this way.

cmd/profilecli/main.go
```diff
@@ -61,6 +61,11 @@ func main() {
 	canaryExporterCmd := app.Command("canary-exporter", "Run the canary exporter.")
 	canaryExporterParams := addCanaryExporterParams(canaryExporterCmd)
 
+	toolCmd := app.Command("tool", "Run one of the available tools.")
```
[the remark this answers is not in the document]
aleks-p: Sounds good, will move things around a bit!

PR #2700, opened 2023-11-16
feat(chart): component level service overrides

kolesnikovae:
The change allows component-specific service overrides, including annotations.


operations/pyroscope/helm/pyroscope/templates/services.yaml
```diff
@@ -10,17 +11,17 @@ metadata:
   labels:
     {{- include "pyroscope.labels" . | nindent 4 }}
     app.kubernetes.io/component: {{ $component | quote }}
-  {{- if $cfg.serviceAnnotations }}
```
JakeCooper, line 13: Question; will this break the chart for those using it in microservice mode? 

If I understand the logical flow it's currently using pyroscope.components.[querier,query-frontend, etc, ...].serviceAnnotations

Changing this will fix it for us but break it for previous consumers

Lmk if I'm wrong here very real possibility. Not a helm expert

kolesnikovae: > If I understand the logical flow it's currently using pyroscope.components.[querier,query-frontend, etc, ...].serviceAnnotations
Changing this will fix it for us but break it for previous consumers
Lmk if I'm wrong here very real possibility. Not a helm expert

Neither I am :) Unfortunately, It does not seem it worked at all (at least it makes no impact if I provide any annotations), therefore it won't make things worse. I tested it locally and can confirm that this is a backwards compatible change

cyriltovena (approved):
LGTM

JakeCooper:
Thank you @kolesnikovae!

JakeCooper:
BTW @kolesnikovae has a new chart been published? I still see it as 1.1.0

kolesnikovae:
Will release it shortly! https://github.com/grafana/pyroscope/pull/2713

PR #2703, opened 2023-11-16
fix(ebpf): python arm64

korniltsev:
- extract dwarfdump to dump multiple things
- generate offsets for arm64 pythons
- generate offsets for arm64 && amd64 glibc
- generate offsets for arm64 && amd64 musl
- add ci job running arm tests in qemu, the VM image is built manually and uploaded to pyroscope dockerhub
- there is one extra difference noticed between python-musl-arm64 and python-glibc-arm64 the size of `_gil_runtime_state` is different because of different size of `pthread_mutex_t`  I currently fix it, by subtracting mutex difference size. we may need to generate separate offsets in the future. https://github.com/grafana/pyroscope/pull/2703/files#diff-27c90d69894a05a2db8d2b9dea5ba7ef89ff2739f0382932c76f1e3beab72bdcR67 ,


korniltsev:
oops, the tests says alpine-arm64 is also broken

ebpf/testutil/testdata.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"os"
+	"os/exec"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func InitGitSubmodule(t *testing.T) {
```
cyriltovena, line 11: 🤣

cyriltovena (approved):
LGTM

ebpf/testutil/testdata.go
korniltsev: everybody hates to use git submodules so I try to pretend it is not there :D

PR #2704, opened 2023-11-16
feat: Improve tracing span tags in the query-frontend.

cyriltovena:
This adds missing informations but specially maxNodes.


kolesnikovae (approved):
LGTM

PR #2705, opened 2023-11-16
Update ebpf.md

AS-Labs:
Added eBPF river config for localhost scraping through pyroscope and grafana-agent-flow


korniltsev (approved):

korniltsev:
This will change in the future when we have process discovery. `targets_only` and `default_target` may go away.

PR #2706, opened 2023-11-16
Propagate store-gateway ring config correct

simonswine:
Note: The replication factor before was always hardcoded to 3. In order
to retain same behaviour, the flag
`-store-gateway.sharding-ring.replication-factor` needs to be set on
querier and store-gateways.


cyriltovena (approved):
LGTM

PR #2707, opened 2023-11-16
Refactor Series to use storeQuery

simonswine:
This is a refactor that aligns how we query series with the MergeStacktraces


kolesnikovae (approved):
LGMT

PR #2711, opened 2023-11-16
feat: Settings API

bryanhuhta:
Closes https://github.com/grafana/pyroscope-app-plugin/issues/187

Creates a settings API. Requests should be handled entirely by the query frontend if it exists, or the querier as a fallback.

Right now I only have an in-memory settings store implemented. As a longer term solution we should have a more persistent data store. Possible options:

- SQLite
  - in S3?
  - on disk?
- Whatever db technology Grafana uses


api/settings/v1/settings.proto
```diff
@@ -0,0 +1,27 @@
...
+message SetSettingsRequest {
+    Setting setting = 1;
+}
+
+message SetSettingsResponse {
+    Setting setting = 1;
+}
+
+message Setting {
+    string name = 1;
```
cyriltovena, line 25: You'll need optimistic concurrency control here https://en.wikipedia.org/wiki/Optimistic_concurrency_control

To do so you'll need to have a modifiedAt for each setting, that you'll use when modifying, if not equal you'll reject the change it means somebody did a change in between and you're about to erase it.

pkg/api/api.go
```diff
@@ -229,11 +230,12 @@ func (a *API) RegisterQuerier(svc querierv1connect.QuerierServiceHandler) {
...
 }
 
-func (a *API) RegisterPyroscopeHandlers(client querierv1connect.QuerierServiceClient) {
-	handlers := querier.NewHTTPHandlers(client)
+func (a *API) RegisterPyroscopeHandlers(qc querierv1connect.QuerierServiceClient, sc settingsv1connect.SettingsServiceClient) {
+	handlers := querier.NewHTTPHandlers(qc, sc)
 	a.RegisterRoute("/pyroscope/render", http.HandlerFunc(handlers.Render), true, true, "GET")
 	a.RegisterRoute("/pyroscope/render-diff", http.HandlerFunc(handlers.RenderDiff), true, true, "GET")
 	a.RegisterRoute("/pyroscope/label-values", http.HandlerFunc(handlers.LabelValues), true, true, "GET")
+	a.RegisterRoute("/pyroscope/settings", http.HandlerFunc(handlers.Settings), true, true, "GET", "POST")
```
cyriltovena, line 238: Can't we use connect-go API only ? those API `render` and such are here for backward compatbility reason.

pkg/frontend/frontend.go
```diff
@@ -139,7 +142,18 @@ func NewFrontend(cfg Config, limits Limits, log log.Logger, reg prometheus.Regis
 		return nil, err
 	}
 
+	settingsStore, err := settings.NewMemoryStore()
```
cyriltovena, line 145: This is actually a query frontend, that is there only for speeding up and applying fair scheduling in a multi-tenant setup.

Happy that we start like this but we might want to move this in another component.

pkg/querier/http.go
```diff
@@ -193,6 +199,49 @@ func (q *QueryHandlers) Render(w http.ResponseWriter, req *http.Request) {
 	}
 }
 
+func (q *QueryHandlers) Settings(w http.ResponseWriter, r *http.Request) {
```
cyriltovena, line 202: We don't need this. We can just register the new connect-go services.

pkg/settings/settings.go
```diff
@@ -0,0 +1,66 @@
...
+	store Store
+}
+
+func (s *settingsService) All(ctx context.Context, req *connect.Request[settingsv1.AllSettingsRequest]) (*connect.Response[settingsv1.AllSettingsResponse], error) {
+	sp, ctx := opentracing.StartSpanFromContext(ctx, "Settings All")
+	defer sp.Finish()
+
+	ctx = connectgrpc.WithProcedure(ctx, settingsv1connect.SettingsServiceAllProcedure)
+
+	tenantIDs, err := tenant.TenantIDs(ctx)
```
cyriltovena, line 32: We should only take one tenant for now

pkg/settings/store.go
```diff
@@ -0,0 +1,12 @@
+package settings
+
+import (
+	"context"
+
+	settingsv1 "github.com/grafana/pyroscope/api/gen/proto/go/settings/v1"
+)
+
+type Store interface {
+	All(ctx context.Context, tenantIDs ...string) ([]*settingsv1.Setting, error)
```
cyriltovena, line 10: Should be only one tenant.

cyriltovena: Get/Set might also be a bette naming.

pkg/api/api.go
bryanhuhta: Yeah good call. I did that. This was the first pass of me hacking around and using the connect-go API only is the way to go.

api/settings/v1/settings.proto
bryanhuhta: Check out this [here](https://github.com/grafana/pyroscope/pull/2711/files#diff-9260ea2e34adbb6eedcaaf2bf8ff746c592f00ccb0fad752e470f85a133890ceR55-R58). Does that seem about right?

pkg/frontend/frontend.go
bryanhuhta: I think how I have it configured as of https://github.com/grafana/pyroscope/pull/2711/commits/600bdb6301754060ead1667bb2128e7dd6d93ce8, it will only run in single binary mode.

Longer term, do we want to deploy a separate component to handle settings? Or do add the settings module to queriers? It might be more effort than its worth to have the settings module running as part of a ring, especially for something that's so small right now.

pkg/querier/http.go
bryanhuhta: 👍

pkg/settings/store.go
bryanhuhta: Yeah Get/Set is much better

api/settings/v1/settings.proto
cyriltovena: Yes that's right, I suggest to make the error more obvious.

pkg/frontend/frontend.go
cyriltovena: That sounds great !

cyriltovena:
> Right now I only have an in-memory settings store implemented. As a longer term solution we should have a more persistent data store. Possible options:
> 
> * SQLite
>   
>   * in S3?
>   * on disk?
> * Whatever db technology Grafana uses


We actually build database :), I suspect we'll store settings in s3, we'll need to make sure only one replica does it. We basically should aim to avoid any external dependencies.

cyriltovena (approved):
LGTM

petethepig:
I think we should add another implementation that stores things in S3 (in the same bucket as the other data). Next steps would be to:
a) expose this API in backend-enterprise https://github.com/grafana/backend-enterprise/blob/e429f85187d922dc78738b712acaee2a6469b536/pkg/gateway/routes.go#L383-L395
b) add a new service in deployment_tools

PR #2723, opened 2023-11-17
fix(ReduxQuerySync): Register the sync as soon as possible

grafakus:
Caused by https://github.com/grafana/pyroscope/pull/2692

This PR fixes a race condition where the apps list is loaded before parsing the values of the `from` and `until` URL query parameters. The effect is that the incorrect app/no app is selected in the dropdown.

|Before | After |
|---|---|
|Screenshot 2023-11-17 at 21 13 01|Screenshot 2023-11-17 at 21 09 16|


public/app/redux/useReduxQuerySync.ts
```diff
@@ -99,6 +99,8 @@ function setupReduxQuerySync() {
   });
 }
 
+const unsubscribe = setupReduxQuerySync();
```
darrenjaneczek, line 102: Won't this cause it to be set up all the time?

public/app/redux/useReduxQuerySync.ts
```diff
@@ -99,6 +99,8 @@ function setupReduxQuerySync() {
...
+
 export function useReduxQuerySync() {
-  useEffect(setupReduxQuerySync, []);
+  useEffect(() => unsubscribe, []);
```
darrenjaneczek, line 105: Won't this cause it to be unsubscribed on the first unmount, but never `setupReduxQuerySync` again?

grafakus: In the case of the app, it's never unmounted right?

public/app/app.tsx
```diff
@@ -15,15 +15,15 @@ import '@szhsin/react-menu/dist/index.css';
...
-import { useReduxQuerySync } from './redux/useReduxQuerySync';
+import { setupReduxQuerySync } from './redux/useReduxQuerySync';
 import './sass/profile.scss';
 
 const container = document.getElementById('reactRoot') as HTMLElement;
 const root = ReactDOM.createRoot(container);
 
-function App() {
-  useReduxQuerySync();
+setupReduxQuerySync();
```
darrenjaneczek, line 24: I really don't like the idea of making this a global effect again. Let's explore a better way to make it occur earlier without

darrenjaneczek: We will explore it later.

darrenjaneczek (approved):
Let's do it.

PR #2732, opened 2023-11-20
Fallback to deduplication when hitting problem block planning

simonswine:
The panics fail a query hard, when one of the assertions is not met.

With e.g changing compactor shards, it panics and fails the query. This changes behaviour, so that it instead falls back to deduplication.

This should be backported into weekly/f41


cyriltovena (approved):
LGTM

PR #2733, opened 2023-11-20
fix(Units): Fix incorrect units in the UI

grafakus:
In the case of (e.g.) `lock_time` profiles, the units were incorrect (count instead of ns). This is because the backend service returns `nanoseconds` as a unit, but the UI doesn't recognize it.

This PR introduces a static JSON file for describing all profile metrics and starts using it in:
- the header of the timeline
- the flamegraph

Note that a lot more work would be needed to ensure that all units/names/descriptions are always properly displayed across the application. 

|Before|After|
|---|---|
| Screenshot 2023-11-20 at 16 30 19|Screenshot 2023-11-20 at 16 29 15|

### DIFF
--- a/public/app/constants/profile-metrics.json
+++ b/public/app/constants/profile-metrics.json
@@ -1,0 +1,135 @@
+{
+  "block:contentions:count:contentions:count": {
+    "id": "block:contentions:count:contentions:count",
+    "description": "Total count of blocking contentions",
+    "type": "contentions",
+    "group": "block",
+    "unit": "short"
+  },
+  "block:delay:nanoseconds:contentions:count": {
+    "id": "block:delay:nanoseconds:contentions:count",
+    "description": "Total nanoseconds spent in blocking delays",
+    "type": "delay",
+    "group": "block",
+    "unit": "ns"
+  },
+  "goroutine:goroutine:count:goroutine:count": {
+    "id": "goroutine:goroutine:count:goroutine:count",
+    "description": "Current number of goroutines",
+    "type": "goroutine",
+    "group": "goroutine",
+    "unit": "short"
+  },
+  "memory:alloc_in_new_tlab_bytes:bytes::": {
+    "id": "memory:alloc_in_new_tlab_bytes:bytes::",
+    "description": "Total bytes allocated inside Thread-Local Allocation Buffers (TLAB)",
+    "type": "alloc_in_new_tlab_bytes",
+    "group": "memory",
+    "unit": "bytes"
+  },
+  "memory:alloc_in_new_tlab_objects:count::": {
+    "id": "memory:alloc_in_new_tlab_objects:count::",
+    "description": "Number of objects allocated inside Thread-Local Allocation Buffers (TLAB)",
+    "type": "alloc_in_new_tlab_objects",
+    "group": "memory",
+    "unit": "short"
+  },
+  "memory:alloc_objects:count:space:bytes": {
+    "id": "memory:alloc_objects:count:space:bytes",
+    "description": "Total objects allocated, represented in bytes",
+    "type": "alloc_objects",
+    "group": "memory",
+    "unit": "count"
+  },
+  "memory:alloc_space:bytes:space:bytes": {
+    "id": "memory:alloc_space:bytes:space:bytes",
+    "description": "Total bytes allocated in the heap",
+    "type": "alloc_space",
+    "group": "memory",
+    "unit": "bytes"
+  },
+  "memory:inuse_objects:count:space:bytes": {
+    "id": "memory:inuse_objects:count:space:bytes",
+    "description": "Count of objects currently in use",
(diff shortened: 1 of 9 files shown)


public/app/constants/profile-metrics.json
```diff
@@ -0,0 +1,135 @@
+{
```
grafakus, line 1: Single source of truth.

public/app/legacy/models/units.ts
```diff
@@ -9,16 +9,13 @@ export const units = [
...
 
-export const UnitsSchema = z.preprocess((u) => {
-  if (typeof u === 'string') {
-    if (units.includes(u)) {
-      return u;
-    }
-  }
-  return 'unknown';
-}, z.enum(['samples', 'objects', 'goroutines', 'bytes', 'lock_samples', 'lock_nanoseconds', 'trace_samples', 'exceptions', 'unknown']));
+export const UnitsSchema = z.preprocess(
```
grafakus, line 15: In the future, we should also refactor this to use the JSON data instead.

public/app/pages/ContinuousComparisonView.tsx
```diff
@@ -180,7 +180,9 @@ function ComparisonApp() {
           title={
             <ChartTitle
               titleKey={
-                isSidesHasSameUnits ? leftSide.metadata.units : undefined
+                isSidesHasSameUnits
+                  ? (leftSide.metadata.name as any)
```
grafakus, line 184: It's an easy escape hatch but more work is needed, so we can tackle this properly in the future.

public/app/util/flamebearer.ts
```diff
@@ -116,6 +116,7 @@ export function flamebearerToDataFrameDTO(
     case 'samples':
     case 'trace_samples':
     case 'lock_nanoseconds':
+    case 'nanoseconds':
```
grafakus, line 119: Same remark as above, it seems we have too many ways to refer to a profile metric unit.

public/app/legacy/flamegraph/FlameGraph/FlameGraphComponent/Header.tsx
```diff
@@ -12,10 +12,9 @@ interface HeaderProps {
   setPalette: (p: FlamegraphPalette) => void;
   toolbarVisible?: boolean;
 }
-export default function Header(props: HeaderProps) {
-  const { format, units, palette, setPalette, toolbarVisible } = props;
 
-  const unitsToFlamegraphTitle = {
+const unitsToFlamegraphTitle = new Map(
```
grafakus, line 16: Changed due to TS issues.

public/app/legacy/flamegraph/Tooltip/Tooltip.tsx
```diff
@@ -208,6 +208,11 @@ const tooltipTitles: Record<
     formattedValue: 'Time',
     total: '% of total seconds',
   },
+  nanoseconds: {
```
grafakus, line 211: Changed due to TS issues.

public/app/pages/ContinuousDiffView.tsx
```diff
@@ -218,7 +220,7 @@ function ComparisonDiffApp() {
             />
           </Panel>
         </div>
-        <Panel isLoading={isLoading} title={<ChartTitle titleKey="diff" />}>
```
grafakus, line 221: Changed due to TS issues. We actually don't really need it in the UI. It saves a bit of vertical real estate (more above the search input):

image

public/app/constants/profile-metrics.json
```diff
@@ -0,0 +1,135 @@
...
+    "unit": "ns"
+  },
+  "process_cpu:alloc_samples:count:cpu:nanoseconds": {
+    "id": "process_cpu:alloc_samples:count:cpu:nanoseconds",
+    "description": "Count of memory allocation samples during CPU time",
+    "type": "alloc_samples",
+    "group": "memory",
+    "unit": "short"
+  },
+  "process_cpu:alloc_size:bytes:cpu:nanoseconds": {
```
cyriltovena, line 86: Didn't know this was a thing.

Rperry2174: yeah its unfortunate that these are so messy but haven't found a good solution yet for standardizing outside of this source-of-truth json file

public/app/legacy/models/units.ts
Rperry2174: can you either leave a comment in the code or create an issue (or both) so we can come back to this at some point?

Rperry2174 (approved):
lgtm!

public/app/legacy/models/units.ts
grafakus: Sure, here it is: https://github.com/grafana/pyroscope/issues/2740

PR #2734, opened 2023-11-20
Improve SelectMatchingProfiles performance

cyriltovena:
This will avoid to grow the `lbls` on every runs and instead copy the slice.

I also removed the safety guard which shouldn't be there and is very expensive to run.


kolesnikovae (approved):
LGTM

Issue #2680, opened 2023-11-14
Add support for Azure Workload Identity


janlonsetteig:
We want to deploy Pyroscope on our AKS clusters along side Loki and Tempo. Storage for all will be Azure blob storage containers.

For Loki and Tempo we use Workload Identity so that we do not have to deal with account keys. We would like to do the same thing with Pyroscope so that we have a consistent configuration and deployment across all the Grafana products we use.

This works as follows:
* We have an existing Service Account that have
```yaml
  annotations:
    azure.workload.identity/client-id: xxxxxx
    azure.workload.identity/tenant-id: yyyyyy
  labels:
    azure.workload.identity/use: "true"
```
* In the deployments we disable creation of a service account and configure it to use the one described above
* The application pods gets this label
```yaml
    labels:
      azure.workload.identity/use: "true"
```
* In the application storage configuration for azure we specify these settings. Is is slightly different for Tempo and Loki, but the principle is the same

Loki:
```yaml
    storage:
      type: azure
      azure:
        useFederatedToken: true
        account_name: zzzzz
        container_name: loki
        use_managed_identity: false
```
Tempo:
```yaml
  storage:
    trace:
      backend: azure
      azure:
        use_federated_token: true
        storage_account_name: zzzzz
        container_name: tempo
        use_v2_sdk: true
```

It would be great if we could also use Workload Identity for the Pyroscope Azure storage

janlonsetteig:
@simonswine I see there are some activity on tasks related to this. Any ETA on support for Azure workload identity?

janlonsetteig:
@simonswine Is this supported now? Can I switch to use workload identity? If yes, what is the configuration that you use for Pyroscope? I see that there is a new connection_string, but I do not see any examples

janlonsetteig:
This actually seems to work now. I only had to remove the account_key and it is now using workload identity. I had already configured the Service Account with workload identity annotations

Thank you so much for adding this :-)

Issue #2412, opened 2023-09-18
Using Azure Blob Storage causes Pyroscope Pod to just hang


elburnetto-intapp:
#### Describe the bug
We've configured Pyroscope to use Azure Blob Storage as an Object Store, by setting the following in it's config map:

account_key: <Storage Account Key>
account_name: <Storage Account Name>
container_name: <Container Name>

When the Pyroscope container then starts up, it just seems to hang, with the following output in the logs:

ts=2023-09-18T09:55:10.230162485Z caller=azure.go:138 level=debug msg="creating new Azure bucket connection" component=storage

The pod never shows as ready/running, due to the health-check failing.

#### To Reproduce

1. Configure Pyroscope's Config Map to use an Azure Storage Account with Name/Key and a Container Name
2. Start Pyroscope Pod in AKS

#### Expected behavior

Pyroscope should connect to the Blob Storage account and start it's processes, or if there's a failure, present it in it's logs.

#### Environment

- Infrastructure: Kubernetes (AKS)
- Deployment tool: Helm

elburnetto-intapp:
This has been resolved by using the endpoint_suffix of 'blob.core.windows.net' (Default Azure Cloud FQDN), so makes me wonder why isn't the default config set to use this value. Also the fact there isn't any error provided back if this value is omitted.

PR #2741, opened 2023-11-21
Update github.com/thanos-io/objstore to latest

simonswine:
This should give us the latest changes and is part of the groundwork for
fixing #2412 #2680.


aleks-p (approved):
LGTM

PR #2742, opened 2023-11-21
Refactor Azure bucket configs

### LINKED ISSUE #2412: Using Azure Blob Storage causes Pyroscope Pod to just hang

simonswine:
This PR fixes #2412 #2680 

It does that by refactoring the config to align with mimir's azure config.

It also avoids retrying for too often and too long, by lowering MaxRetries and making sure retries are done without exponential backoff. 

This depends on #2741

### DIFF
--- a/pkg/objstore/providers/azure/bucket_client.go
+++ b/pkg/objstore/providers/azure/bucket_client.go
@@ -26,6 +26,9 @@ func newBucketClient(cfg Config, name string, logger log.Logger, factory func(lo
 	bucketConfig.MaxRetries = cfg.MaxRetries
 	bucketConfig.UserAssignedID = cfg.UserAssignedID
 
+	// do not delay retries
+	bucketConfig.PipelineConfig.RetryDelay = -1
+
 	if cfg.Endpoint != "" {
 		// azure.DefaultConfig has the default Endpoint, overwrite it only if a different one was explicitly provided.
 		bucketConfig.Endpoint = cfg.Endpoint
--- a/pkg/objstore/providers/azure/bucket_client.go
+++ b/pkg/objstore/providers/azure/bucket_client.go
@@ -9,26 +9,27 @@ import (
 	"github.com/go-kit/log"
 	"github.com/thanos-io/objstore"
 	"github.com/thanos-io/objstore/providers/azure"
-	"gopkg.in/yaml.v3"
 )
 
 func NewBucketClient(cfg Config, name string, logger log.Logger) (objstore.Bucket, error) {
+	return newBucketClient(cfg, name, logger, azure.NewBucketWithConfig)
+}
+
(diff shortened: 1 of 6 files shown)


pkg/objstore/providers/azure/bucket_client.go
```diff
@@ -9,26 +9,30 @@ import (
...
 	bucketConfig.MaxRetries = cfg.MaxRetries
 	bucketConfig.UserAssignedID = cfg.UserAssignedID
 
-	// Thanos currently doesn't support passing the config as is, but expects a YAML,
-	// so we're going to serialize it.
-	serialized, err := yaml.Marshal(bucketConfig)
-	if err != nil {
-		return nil, err
+	// do not delay retries
+	bucketConfig.PipelineConfig.RetryDelay = -1
```
aleks-p, line 30: Perhaps a naive question, why do we want to change the default behavior here? From what I can see Mimir doesn't do this. If we are worried about retries taking too long, there is also MaxRetryDelay that could be set.

aleks-p (approved):
I left one question but it looks good to me overall!

pkg/objstore/providers/azure/bucket_client.go
aleks-p: Without the changes from #2741 I get the hang on startup with those 2 example commands. This goes away when I apply #2741.

Testing this branch, I don't see a difference in behavior between the changes here and latest `main`. Commenting this line out (reenabling retry delay) doesn't seem to make a difference as well for me.

simonswine: This not something I see, I do get a hang on main. I also don't see any big effect on max retries:

```
// main f80352606c8e13a0bcc65805e66899e5e664c275 with #2741, hangs > 1 min

// the max retries change and RetryDelay=-1, fails  after 10s

// the mimir config change, fails after 2-3s

// this pr as of right now, fail after 2-3s
```

I still think all of those changes should be done. MaxRetries=20 is too high, so this expoential back of

PR #2751, opened 2023-11-23
Helm: Handle overriding extraArgs in components correctly

simonswine:
I noticed a bug, when setting extraArgs on a single component (`.values.pyroscope.components.querier.extraArgs`) they would completely overwrite the whole global (`.values.pyroscope.extraArgs`) array.

This change will make sure we do a merge of global and component based `extraArgs`.


cyriltovena (approved):
LGTM

PR #2758, opened 2023-11-24
Add a way to control the aggregation type for the SelectSeries API

aleks-p:
Adds an optional "aggregation" param to the SelectSeries API, allowing to switch between "sum" (the default), "avg" and "first" (used mostly internally). Other aggregation functions can be added in the future. Right now it only impacts the time series, but I will see how much work it would be (and how expensive it is) to apply it to the flamegraph as well. 

A PR in Grafana will follow where we expose this in the Explore view like this:

Screenshot 2023-11-24 at 13 19 18


pkg/model/series.go
```diff
@@ -70,15 +103,141 @@ func (m *SeriesMerger) mergePoints(points []*typesv1.Point) int {
...
 		}
+		m.aggregator.Add(points[i].Timestamp, points[i].Value)
+	}
+	if !m.aggregator.IsEmpty() {
+		points[j] = m.aggregator.GetAndReset()
 	}
 	return j + 1
 }
+
+type TimeSeriesAggregator interface {
```
aleks-p, line 125: the name could be a bit confusing since we also have https://github.com/grafana/pyroscope/blob/9e55dd1235a700bfe8f471090ad7c6429a83db18/pkg/distributor/aggregator/aggregator.go#L11

I think their purpose is slightly different but putting that in the name is hard. Any thoughts @kolesnikovae?

pkg/model/series.go
```diff
@@ -70,15 +103,141 @@ func (m *SeriesMerger) mergePoints(points []*typesv1.Point) int {
...
+	ts    int64
+	sum   float64
+	count int64
+}
+
+func (a *avgTimeSeriesAggregator) Add(ts int64, value float64) {
+	a.ts = ts
+	a.sum += value
+	a.count++
+}
```
kolesnikovae, line 192: I'd like to make sure I fully understand the aggregation and its relation to grouping. I'm sorry if I got it wrong :) 

Given two series (with just one value each):
```
{pod=a, endpoint=x}: { ts_1, 8 }
{pod=a, endpoint=y}: { ts_1, 4 }
```

Given a query `avg (group) by pod`. What would be the result: `{pod=a} 12` or `{pod=a} 6`?

My understanding is that the aggregator gets two values at input (from **[seriesBuilder](https://github.com/grafana/pyroscope/blob/main/pkg/phlaredb/sample_merge.go#L200)**):
```
{pod=a}: { ts_1, 4 }, { ts_1, 8 }
```
Which then will be averaged and the result will be 6. I'm not very sure this is what users expect (despite it is literally the average value)

pkg/model/series.go
```diff
@@ -7,26 +7,59 @@ import (
 	typesv1 "github.com/grafana/pyroscope/api/gen/proto/go/types/v1"
 )
 
+const (
+	AverageAggregationType    = "avg"
+	sumAggregationType        = "sum"
+	firstValueAggregationType = "first"
```
cyriltovena, line 13: I think first is confusing. But I understand why it's there.

Normally in prometheus land `{app="foo"}` returns an instant vector where there will be only one point per unique series per timestamp. This is done using looback period.

Check this one out https://www.linkedin.com/posts/julius-volz_promql-data-selection-explained-selectors-activity-7066421381151768578-Svsm/?trk=public_profile_like_view

So technically `first({app="foo"}) by (pod)` should never be needed, but it seems currently we actually returns a range vector for `{app="foo"}`.


Currently this is a bit invisible because we're doing `sum(sum_over_time({app="foo"}[step])) by (pod)`

I don't think it's super important for now, but if we ever want to have a query lang we most likely should get this resolved.

pkg/model/series.go
aleks-p: It will be `{pod=a} 6` and I am not sure either though summing will often make no sense as well, depending on the profile type. There are two aspects to this aggregation:

1. Reducing samples for the exact same timestamp (happens in [series.go](https://github.com/grafana/pyroscope/blob/94d0a6a03c1cc895a575c36e8adb162891f938f6/pkg/model/series.go#L79-L81), previously either "first" or "sum")
2. Reducing samples within a step interval (happens in [querier.go](https://github.com/grafana/pyroscope/blob/94d0a6a03c1cc895a575c36e8adb162891f938f6/pkg/querier/querier.go#L972), previously always "sum")

Honestly I am not sure we need 1. Leaving duplicate samples in and letting 2 handle the aggregation would produce more reliable results (currently in 2 we will do avg of avg at times).

pkg/model/series.go
aleks-p: I agree. Also, wording the current and new behavior is very nuanced. `{app="foo"}` in Pyroscope doesn't produce a vector at all (instant or range) but a single value at the evaluation time. It does this by first aggregating across series (with "sum" or "first" to avoid dealing with a range vector) and then across time (`sum_over_time(...[step])`).

I would be happy to remove the first aggregation completely. This would let duplicate timestamps in the intermediary result which `sum_over_time` (or `avg_over_time`) will take care off anyway.

pkg/model/series.go
cyriltovena: Totally agree we should only do 2 to avoid avg of avg. May be only when it's not a sum.

PR #2761, opened 2023-11-27
fix(ebpf): prefer pid targets over container targets

korniltsev:
if we have a target specific to process, we should use, even if it is in a container(instead of creating a generic container target)


cyriltovena (approved):
LGTM

PR #2763, opened 2023-11-27
Switch to relative links in block viewer tool

aleks-p:
I noticed that the entrypoint page in https://github.com/grafana/pyroscope/pull/2724 works as expected but all the links stop working when the web application is running with a different base URL (e.g., `/store-gateway/ops/object-store/...` instead of just `/ops/object-store/....`). 

This fixes the problem by resorting to relative URLs. It makes the back URLs uglier in code, but there is also less repetition overall.


kolesnikovae (approved):
LGTM

Issue #91, opened 2021-02-23
DRY actions for leftFrom, leftUntil, rightFrom, and rightUntil


Rperry2174:
In PR #90 I introduced the ability to set three different timelines:
1. The main timeline at that top. this happens on both the "single view" page and the "comparison" page
2. The left timeline -- this only happens on the comparison page (where there are two timelines / flamegraphs)
3. the right timeline -- this only happens on the comparison page ( where there are two timelines / flamegraphs)

In order to update these all independently from each other I created a separate action for each of them. They could probably be merged into one since they are doing essentially the same thing:

https://github.com/pyroscope-io/pyroscope/blob/main/webapp/javascript/redux/actions.js#L22-L56

PR #2765, opened 2023-11-27
Xiyu/css fix dependency update

xiyu95:
## Description
Fixes [#91](https://github.com/grafana/pyroscope-app-plugin/issues/91)

Add overflow to page to avoid out of bound UI.


xiyu95:
Seems like the dependency version won't pass the tests @darrenjaneczek, reverting back to version 5

L2D2Grafana (approved):
LGTM thanks for the video

darrenjaneczek (approved):
Thanks for the fix.

PR #2780, opened 2023-11-29
feat: Cluster API Versioning

cyriltovena:
This adds a new API to returns the current API version of the cluster.

```
curl \
    --header 'Content-Type: application/json' \
    --data '{}' \
    http://localhost:4040/version.v1.Version/Version
{"QuerierAPI":"1"}
```

The versions is guaranteed the minimun version across the cluster to ensure new feature are being used only when the cluster is fully rolled out.

Behind the scene it uses member ist to form a consensus on active members and  their current version. Members update periodically their timestamp and version while they are active, at the same time we're pruning member not active.


pkg/phlare/phlare.go
```diff
@@ -523,6 +529,16 @@ func (f *Phlare) initAPI() (services.Service, error) {
 	return nil, nil
 }
 
+func (f *Phlare) initVersion() (services.Service, error) {
+	var err error
+	f.versions, err = apiversion.New(f.Cfg.Distributor.DistributorRing, f.logger, f.reg)
```
cyriltovena, line 534: not specifically happy that we use the distributor config but it's at least one guaranteed to exists and be valid.

pkg/api/version/version.go
```diff
@@ -0,0 +1,257 @@
...
+				delete(versions.Instances, id)
+			}
+		}
+		return &Versions{
+			Versions: versions,
+		}, true, nil
+	})
+}
+
+func (svc *Service) Version(ctx context.Context, req *connect.Request[versionv1.VersionRequest]) (*connect.Response[versionv1.VersionResponse], error) {
```
cyriltovena, line 231: The service can be passed to other components and allow them to also verify the version of the cluster too.

For example when adding new feature to only store-gateways.

pkg/api/version/version.go
```diff
@@ -0,0 +1,279 @@
...
+	"context"
+	"fmt"
+	"math"
+	"sync"
+	"time"
+
+	"github.com/bufbuild/connect-go"
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/gogo/protobuf/proto"
```
kolesnikovae, line 13: I'm not very sure if we want to use "gogoproto" instead of `google.golang.org/protobuf/proto`.

Also, AFAIK, the official `proto` does not rely on former `proto.Unmarshaler` and `proto.Marshaler` which where deprecated

pkg/api/version/version.go
```diff
@@ -0,0 +1,279 @@
...
+	for _, instance := range versions.Instances {
+		if instance.QuerierAPI < minQuerierVersion {
+			minQuerierVersion = instance.QuerierAPI
+		}
+	}
+	return connect.NewResponse(&versionv1.VersionResponse{
+		QuerierAPI: minQuerierVersion,
+	}), nil
+}
+
```
kolesnikovae, line 273: Does this mean that instances that have not yet reported their versions are not taken into account?

api/version/v1/version.proto
```diff
@@ -0,0 +1,29 @@
...
+  int64 Timestamp = 3;
+
+  // Querier Service API version
+  uint64 QuerierAPI = 4;
+}
+
+// Versions is the top-level type used to model version for all instances, containing information for individual instances.
+message Versions {
+  map<string, InstanceVersion> instances = 1;
+}
```
kolesnikovae, line 29: Should we have a repeated field for sake of simplicity and performance (proto maps in Go are notoriously slow)?

pkg/api/version/version.go
cyriltovena: This is forced by the dskit/memberlist package and the mergeable I had to implement

pkg/api/version/version.go
cyriltovena: Yes and we'll report the lower version so it's safe to query

kolesnikovae: But doesn't this mean that during the first rollout we will report higher version (1) than the lowest deployed (0)? I must be missing something. For example, we add some features that rely on the version reported by this API in v1.3/1.4/1.5. What would be the user experience during the upgrade from the current version (v1.2)?

My guess it that once _the very first_ upgraded instance reports its version (e.g 1), we will report that version because it is literally the lowest _known_ version. As a result, we will try to use new features while not all the instances are upgraded.

pkg/api/version/version.go
kolesnikovae: We could implement own [`Codec`](https://github.com/grafana/dskit/blob/main/kv/codec/codec.go#L9-L16) that uses `google.golang.org/protobuf/proto`. Either way, I don't feel strong about it, we can change this later if we want

kolesnikovae (approved):
LGTM

PR #2788, opened 2023-12-01
fix(ebpf): merge equal samples

korniltsev:
Regular stacktraces are agregated in ebpf program by increasing value in the ebpf maps
Python stacktraces are different, they are sent to userspace as is without any aggregation, so we may end up with multiple equal samples and values=1.

In this PR I aggregate python samples so that pprofs are smaller.
In the future we may try to aggregate python samples in ebpf (this require adding hashing function to ebpf)

### DIFF
--- a/ebpf/pprof/pprof.go
+++ b/ebpf/pprof/pprof.go
@@ -3,9 +3,12 @@ package pprof
 import (
 	"fmt"
 	"io"
+	"reflect"
 	"sync"
 	"time"
+	"unsafe"
 
+	"github.com/cespare/xxhash/v2"
 	"github.com/google/pprof/profile"
 	"github.com/klauspost/compress/gzip"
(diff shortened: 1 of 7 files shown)


ebpf/pprof/pprof.go
```diff
@@ -77,6 +89,42 @@ func (p *ProfileBuilder) AddSample(stacktrace []string, value uint64) {
...
+	if sample != nil {
+		sample.Value[0] += scaledValue
+		return
+	}
+	sample = &profile.Sample{
+		Location: p.tmpLocations,
+		Value:    []int64{scaledValue},
+	}
+	p.sampleHashToSample[h] = sample
+	p.tmpLocations = nil
```
kolesnikovae, line 124: Must be a leftover?

kolesnikovae (approved):
LGTM

PR #2789, opened 2023-12-01
fix(ebpf): #2788 followup

korniltsev:
address #2788  comments

nothing's really changed, just make code little bit cleaner


kolesnikovae:
I'd say that this is a performance optimisation :)

Issue #2062, opened 2023-05-12
Helm ingress doesn't support pyroscope ingress


simonswine:

macmiranda:
Hi @simonswine, do you mean that the Ingress object created by Helm will not support the Pyroscope UI?

I haven't tried it myself but looking at the templates, it does seem like it. I have created an Ingress manually with Prefix `/` and it works fine.

Is that what you have in mind?

rlex:
Ingress seemingly only supports certain paths, but not web ui.

Fix is simple, just add 

```
          - path: /
            pathType: Prefix
            backend:
              service:
                name: pyroscope
                port:
                  number: 4040
```

to list of paths under ingress somewhere here: https://github.com/grafana/pyroscope/blob/main/operations/pyroscope/helm/pyroscope/templates/ingress.yaml#L25

PR #2807, opened 2023-12-05
fix ingress for pyroscope UI

rlex:
Closes https://github.com/grafana/pyroscope/issues/2062 - adding root path for ingress, allowing opening pyroscope UI with ingress


cyriltovena:
Thanks !  I think you'll need to run `make helm/lint` and `make helm/check`

cyriltovena (approved):
LGTM sorry I didn't realize ingress is not on by default !

PR #2812, opened 2023-12-07
fix(ebpf): start profiling after attaching perf reader

korniltsev:
There was  a tiny time window when we already attach ebpf program to perf event (start profiling) and not yet attached a reader for pid info requests.

So this line some times return 2 (ENOENT) instead of 0, causing userspace program miss some processes to profile.
https://github.com/grafana/pyroscope/blob/8dc366aa5027608f62510faaa871a4f5a6397c7f/ebpf/bpf/profile.bpf.c#L50

This PR changes order of enabling profiling - start reading events first, then start profiling

### DIFF
--- a/ebpf/session.go
+++ b/ebpf/session.go
@@ -163,15 +163,15 @@ func (s *session) Start() error {
 
 	btf.FlushKernelSpec() // save some memory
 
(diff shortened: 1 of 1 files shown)


kolesnikovae (approved):
LGTM

PR #2813, opened 2023-12-07
perf: optimize select merge span profile query read path

kolesnikovae:
The change optimizes `SelectMergeSpanProfile` query read path:
* Use physical plan, if available;
* Skip profile filtering, if deduplication is not required.


cyriltovena (approved):
LGTM

PR #2815, opened 2023-12-07
Fixes versions memberlist merging.

cyriltovena:
Uses tombstones and only returns changes in the merge function.


kolesnikovae (approved):
LGTM!

PR #2816, opened 2023-12-07
feat: Clean up uploaded blocks in ingesters

bryanhuhta:
This PR re-works the formerly named `retentionPolicyEnforcer` service to include cleaning up shipped blocks when enforcing disk utilization. This change should substantially reduce the amount of disk ingesters use. Currently in ops we use about 100GB or so of disk to store blocks that have been uploaded and synced with the store gateway.

I made some changes to the components that made up the service that manages disk for ingesters.

### `diskCleaner`

```go
type diskCleaner struct {
	services.Service

	logger        log.Logger
	config        phlaredb.Config
	policy        retentionPolicy
	blockManager  fsBlockManager
	volumeChecker diskutil.VolumeChecker

	stop chan struct{}
	wg   sync.WaitGroup
}
```

This is the primary component which gets registered as a service and handles taking disk clean up every tick. It is composed of three subcomponents:

- `rententionPolicy`
- `fsBlockManager`
- `VolumeChecker`

Between these three components, `diskCleaner` can identify blocks that need to be deleted from disk that have either been uploaded, or (in more aggressive cases) are the most likely candidates for removal in the face of high disk utilization.

### `retentionPolicy`

This hasn't changed. It holds the settings for describing when to trigger a cleanup cycle.

### `fsBlockManager`

```go
type fsBlockManager interface {
	GetTenantIDs(ctx context.Context) ([]string, error)
	GetBlocksForTenant(ctx context.Context, tenantID string) ([]*tenantBlock, error)
	DeleteBlock(ctx context.Context, block *tenantBlock) error
}
```

This has the ability to interact with the file system to read and delete blocks. 

### `VolumeChecker`

This hasn't changed. It collects disk usage information to be used by `diskCleaner` to make decisions about how and when blocks should be deleted.


pkg/ingester/retention.go
```diff
@@ -39,183 +65,322 @@ func defaultRetentionPolicy() retentionPolicy {
...
+	var hasHighDiskUtilization bool
 	for {
-		// Enforce retention policy immediately at start.
-		level.Debug(e.logger).Log("msg", "enforcing retention policy")
-		if err := e.cleanupBlocksWhenHighDiskUtilization(ctx); err != nil {
-			level.Error(e.logger).Log("msg", "failed to enforce retention policy", "err", err)
+		deleted = dc.DeleteUploadedBlocks(ctx)
+		level.Debug(dc.logger).Log("msg", "cleaned uploaded blocks", "count", deleted)
+
+		deleted, bytesDeleted, hasHighDiskUtilization = dc.EnforceHighDiskUtilization(ctx)
```
kolesnikovae, line 103: If I read it right, this can remove blocks that were uploaded but not yet discovered by store-gateway. This means that deleted blocks won't be visible to queriers for some period of time (util store-gateway discovers them). This is fine, if we have to do so due to lack of free space, and the removal is vital. But it seems we remove _all_ uploaded blocks unconditionally.

As of now, when we query blocks, we rely on the `-querier.query-store-after` parameter (but I hope we will get rid of them at some point):

```yaml
# The time after which a metric should be queried from storage and not just
# ingesters. 0 means all queries are sent to store. If this option is enabled,
# the time range of the query sent to the store-gateway will be manipulated to
# ensure the query end is not more recent than 'now - query-store-after'.
# CLI flag: -querier.query-store-after
[query_store_after: <duration> | default = 4h]
```

Theoretically, we could rely here on this parameter as well, or check that at least 2 sync intervals passed since the block has been created, like:
```
now() - block.timestamp > 2 * sg-sync-interval
```
This isn't a reliable approach, and I'd rather try to avoid such speculations. But there is currently no simple (or efficient at least) way to confidently say that store-gateways are aware of the block and it will be read on query.

Another thing that we should keep in mind is that some of the users won't ever use the object store.

This makes me thinking that we, probably, should introduce a new parameter, that would explicitly control for how long uploaded blocks are kept locally (sort of time-based retention policy for local blocks).

pkg/ingester/retention.go
```diff
@@ -39,183 +65,322 @@ func defaultRetentionPolicy() retentionPolicy {
...
+	return bm.Evictor.evictBlock(block.TenantID, block.ID, func() error {
+		err := bm.FS.RemoveAll(block.Path)
+		switch {
 		case os.IsNotExist(err):
-			level.Warn(e.logger).Log("msg", "block not found on disk", "path", b.path)
-		default:
-			return fmt.Errorf("failed to delete block %q: %w", b.path, err)
+			return err
+		case err != nil:
+			return fmt.Errorf("failed to delete block: %q: %w", block.Path, err)
```
kolesnikovae, line 383: I'm not sure I understand why we need to change this logic

kolesnikovae (commented):
Great work 👍🏻

I'm a bit concerned about synchronisation with store-gateways, would like to clarify this moment

pkg/ingester/retention.go
bryanhuhta: I changed it because the block manager no longer has a logger and it seemed relevant to log the fact we were skipping an error. In https://github.com/grafana/pyroscope/pull/2816/commits/312d6c0dfc45f2f6167d721ff34e717e5255a05f I updated the callsite of `DeleteBlock` to log and ignore the error, thus keeping the original logic intact.

pkg/ingester/retention.go
bryanhuhta: Agreed on the configuration parameter. It's unfortunate we have to add such a configuration, but this should allow folks to change how long to keep shipped blocks around. They can make their own trade offs (i.e. if the parameter is set too low, there may be query gaps before store gateways sync shipped blocks, too high and too much disk may be used).

I made this change in https://github.com/grafana/pyroscope/pull/2816/commits/2a763aff9489bd6ea2294d410d977cc45c06191d.

bryanhuhta: Basically as long as a block is older than `now() - expiry`, it will be deleted. By default, the expiry is twice the length of the bucket store sync interval (30 min).

bryanhuhta: That said, the `EnforceHighDiskUtilization` function will not respect this. It will try first delete the blocks marked as shipped (oldest to newest), then move to the unshipped blocks (oldest to newest).

pkg/phlaredb/phlaredb.go
```diff
@@ -49,6 +50,7 @@ type Config struct {
 
 	MinFreeDisk                uint64        `yaml:"min_free_disk_gb"`
 	MinDiskAvailablePercentage float64       `yaml:"min_disk_available_percentage"`
+	BlockExpiry                time.Duration `yaml:"block_expiry"`
```
kolesnikovae, line 53: Ah, just have noticed that our retention policy settings are placed at root. These are especially "vague":

```yaml
pyroscopedb:
  enforcement_interval:
  disable_enforcement:
```

I don't think that there is a nice way to add `retention_policy` prefix and not break compatibility, just something we should keep in mind.

pkg/ingester/retention.go
kolesnikovae: Yeah, that looks good!

pkg/ingester/retention.go
```diff
@@ -39,183 +66,351 @@ func defaultRetentionPolicy() retentionPolicy {
...
+	dc.wg.Wait()
 	return nil
 }
 
-func (e *retentionPolicyEnforcer) localBlocks(dir string) ([]*tenantBlock, error) {
-	blocks := make([]*tenantBlock, 0, 32)
-	tenants, err := fs.ReadDir(e.fileSystem, dir)
+// DeleteUploadedBlocks scans and deletes blocks on all tenants that have
+// already been deleted. It returns the number of blocks deleted.
+func (dc *diskCleaner) DeleteUploadedBlocks(ctx context.Context) int {
```
kolesnikovae, line 131: We should make sure that the operation handles context cancellation and does not block the server shutdown. Potentially, clean up can take quite a while (especially on the first run), therefore I guess we could check context each time we deleted a block. Same for `EnforceHighDiskUtilization`

bryanhuhta: In https://github.com/grafana/pyroscope/pull/2816/commits/0399b5920d5f06f59bf2bf09c2faa1df06baed9c, I added checks for context cancellations before reading or deleting blocks.

pkg/phlaredb/phlaredb.go
bryanhuhta: Yeah, this is a good point. Perhaps it's worth looking into a way to move these keys off the root, but still check for it on the root for legacy purposes.

pkg/ingester/retention.go
```diff
@@ -39,183 +66,363 @@ func defaultRetentionPolicy() retentionPolicy {
...
+	close(dc.stop)
+	dc.wg.Wait()
 	return nil
 }
 
-func (e *retentionPolicyEnforcer) localBlocks(dir string) ([]*tenantBlock, error) {
-	blocks := make([]*tenantBlock, 0, 32)
-	tenants, err := fs.ReadDir(e.fileSystem, dir)
+// DeleteUploadedBlocks scans and deletes blocks on all tenants that have
+// already been deleted. It returns the number of blocks deleted.
```
aleks-p, line 130: ... already been **uploaded**?

pkg/ingester/retention.go
```diff
@@ -39,183 +66,363 @@ func defaultRetentionPolicy() retentionPolicy {
...
 }
 
-func (e *retentionPolicyEnforcer) cleanupBlocksWhenHighDiskUtilization(ctx context.Context) error {
-	var volumeStatsPrev *diskutil.VolumeStats
-	volumeStatsCurrent, err := e.volumeChecker.HasHighDiskUtilization(e.dbConfig.DataPath)
+// EnforceHighDiskUtilization will run more aggressive disk cleaning if high
+// disk utilization is detected. It returns true if high disk utilization was
+// detected, along with the number of files deleted and the estimated bytes
+// recovered. If no high disk utilization was detected, false is returned.
+func (dc *diskCleaner) EnforceHighDiskUtilization(ctx context.Context) (int, int, bool) {
```
aleks-p, line 189: Nitpick, the name suggests to me that the method tries to enforce "high" utilization. Perhaps `diskCleaner.EnforceDiskUtilization` or `diskCleaner.ReduceDiskUtilization` could make it more clear?

pkg/ingester/retention.go
```diff
@@ -39,183 +66,363 @@ func defaultRetentionPolicy() retentionPolicy {
...
+		// Delete a block.
+		err = dc.blockManager.DeleteBlock(ctx, block)
+		switch {
+		case os.IsNotExist(err):
+			level.Warn(dc.logger).Log(
+				"msg", "failed to delete uploaded block, does not exist",
+				"err", err,
+				"path", block.Path,
+			)
+			return filesDeleted, int(volumeStats.BytesAvailable - prevVolumeStats.BytesAvailable), true
```
aleks-p, line 245: could we `continue` here instead or is there a reason to prefer to exit on the first error?

pkg/ingester/retention.go
```diff
@@ -39,183 +66,363 @@ func defaultRetentionPolicy() retentionPolicy {
...
-		// Delete the oldest block.
-		var b *tenantBlock
-		b, blocks = blocks[0], blocks[1:]
-		level.Warn(e.logger).Log("msg", "disk utilization is high, deleting the oldest block", "path", b.path)
-		if err = e.deleteBlock(b); err != nil {
-			return err
+
+		// Check if deletion should stop.
+		if !volumeStats.HighDiskUtilization || prevVolumeStats.BytesAvailable >= volumeStats.BytesAvailable {
+			level.Warn(dc.logger).Log("msg", "disk utilization is not lowered by deletion of a block, pausing until next cycle")
```
aleks-p, line 270: The log message here reads a bit strange, according to the condition - the utilization was lowered or am I reading it wrong?

pkg/ingester/retention.go
```diff
@@ -39,183 +66,363 @@ func defaultRetentionPolicy() retentionPolicy {
...
+			break
 		}
-		volumeStatsPrev = volumeStatsCurrent
-		if volumeStatsCurrent, err = e.volumeChecker.HasHighDiskUtilization(e.dbConfig.DataPath); err != nil {
-			return err
+	}
+
+	return filesDeleted, int(volumeStats.BytesAvailable - prevVolumeStats.BytesAvailable), true
```
aleks-p, line 275: Since we recheck the disk usage (prevVolumeStats gets re-assigned in the loop), is there a chance we return a partial amount for bytesAvailable here?

aleks-p (approved):
Looks good to me overall. Left a few questions / minor remarks. Also, do we want to produce some metrics for the retention flow?

pkg/ingester/retention.go
bryanhuhta: It's possible we could continue. The original implementation stopped on this error, so I mirrored that behavior.

https://github.com/grafana/pyroscope/blob/30f69c48054273649d316c39502a597f7d54e973/pkg/ingester/retention.go#L182-L185

pkg/ingester/retention.go
bryanhuhta: D'oh! Fixed in https://github.com/grafana/pyroscope/pull/2816/commits/683b36d485638eef80b0e89a10e543741d8fef49

pkg/ingester/retention.go
bryanhuhta: I agree. Fixed in https://github.com/grafana/pyroscope/pull/2816/commits/82a1a229196f0c4cb094bf384c930d25d59b6f54

pkg/ingester/retention.go
bryanhuhta: The logic for this case was all messed up. Basically, there are three conditions (besides errors) which cause use to break the loop.

1. No more blocks to delete
2. Disk utilization is no longer reported high
3. Available bytes has not changed after deleting a block -- this can happen if `VolumeChecker` is delayed when reporting the current disk stats, in which case we want to be conservative and stop deleting blocks

In https://github.com/grafana/pyroscope/pull/2816/commits/e61b35a2bd2f8f98c8cba4658506c2603e54175c I did some clean up to break out of the loop in appropriate cases.

pkg/ingester/retention.go
bryanhuhta: You're right. This calculation will tell us the number of bytes freed since the last block deletion, which is incorrect. In https://github.com/grafana/pyroscope/pull/2816/commits/e61b35a2bd2f8f98c8cba4658506c2603e54175c I updated the calculation to reflect the actual number of bytes freed.

pkg/ingester/retention.go
aleks-p: Much easier to read now!

pkg/ingester/retention.go
aleks-p: didn't notice that, lets keep it as it was for now then

docs/sources/configure-server/reference-configuration-parameters/index.md
```diff
@@ -123,6 +123,11 @@ pyroscopedb:
   # CLI flag: -pyroscopedb.retention-policy-min-disk-available-percentage
   [min_disk_available_percentage: <float> | default = 0.05]
 
+  # How long a block that has been uploaded to bucket storage should remain on
+  # disk
+  # CLI flag: -pyroscopedb.retention-policy-block-expiry
+  [block_expiry: <duration> | default = 10m]
```
cyriltovena, line 129: weird the default doesn't match

pkg/ingester/ingester.go
```diff
@@ -79,7 +80,7 @@ func (i *ingesterFlusherCompat) Flush() {
 	}
 }
 
-func New(phlarectx context.Context, cfg Config, dbConfig phlaredb.Config, storageBucket phlareobj.Bucket, limits Limits) (*Ingester, error) {
+func New(phlarectx context.Context, cfg Config, dbConfig phlaredb.Config, storageBucket phlareobj.Bucket, limits Limits, queryStoreAfter time.Duration) (*Ingester, error) {
```
bryanhuhta, line 83: Is this the best way to get access to the `querier.query_store_after` configuration? I can't pass the config in directly here because the `querier` package already depends on the `ingester` package and that would create a cyclic dependency.

docs/sources/configure-server/reference-configuration-parameters/index.md
bryanhuhta: I had to run `make reference-help` https://github.com/grafana/pyroscope/pull/2816/commits/daa1946007fe717df70abe28fb9e604e43a6d36f

pkg/ingester/ingester.go
bryanhuhta: cc @cyriltovena

cyriltovena: May you can do this config validation outside of ingester ? 

```
if queryStoreAfter > 0 {
		retentionPolicy.Expiry = queryStoreAfter
	}
```

In the config validation or the `phlare` package. It's not great but it's also alright.

bryanhuhta:
I tested this in dev and the number of PVCs with less than 3% available capacity dropped to 0 ([query](https://ops.grafana-ops.net/explore?schemaVersion=1&panes=%7B%22-ZF%22:%7B%22datasource%22:%222z9d6ElGk%22,%22queries%22:%5B%7B%22refId%22:%22B%22,%22expr%22:%22sum%20%28%28kubelet_volume_stats_available_bytes%7Bjob%3D%5C%22kube-system%2Fkubelet%5C%22,%20namespace%3D%5C%22fire-dev-001%5C%22,%20persistentvolumeclaim%3D~%5C%22data-fire-ingester-.%2A%5C%22%7D%20%2F%20kubelet_volume_stats_capacity_bytes%7Bjob%3D%5C%22kube-system%2Fkubelet%5C%22,%20namespace%3D%5C%22fire-dev-001%5C%22,%20persistentvolumeclaim%3D~%5C%22data-fire-ingester-.%2A%5C%22%7D%29%20%3C%20bool%200.03%29%22,%22range%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%222z9d6ElGk%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22__auto%22,%22useBackend%22:false,%22disableTextWrap%22:false,%22fullMetaSearch%22:false,%22includeNullMetadata%22:true,%22hide%22:false%7D%5D,%22range%22:%7B%22from%22:%221704235007269%22,%22to%22:%221704237899830%22%7D%7D%7D&orgId=1)).

Screenshot 2024-01-02 at 5 38 28 PM

PR #2822, opened 2023-12-08
fix: span query read path

kolesnikovae:
The recent optimisation (https://github.com/grafana/pyroscope/pull/2813) introduced a bug where a query may result in empty response


cyriltovena (approved):
LGTM

PR #2823, opened 2023-12-08
Improve language detection performance

### LINKED ISSUE #2777: Optimize language detection
We're currently using a set of regular expressions to derive the source code language based on the string table. This works well but the overhead might be quite substantial:

image

We should probably consider optimizing the procedure. Some ideas:
 - Sampling: it makes little sense to inspect the whole table. I assume that sampling 5-15% of strings should be sufficient.
 - Short cuts: in most cases we can confidently determine the language based on the file extensions. We should probably try this first (without regexes, just suffix matching).
 - Skip the procedure if we can get this info from labels (spy_name).

Nevertheless, we should update all our SDKs to send some meta (such as language, version, platform, etc)

aleks-p:
Closes #2777 

1. If the profile has the `pyroscope_spy` label, we use it to determine the language
2. If we do need to look into the string table, we use prefix and suffix matching instead of regular expressions

Here is a summary of the gains with 2:

Regexp:
```
go.cpu.labels.pprof-10         	  330040	     10853 ns/op
heap-10                        	  195819	     18407 ns/op
dotnet.labels.pprof-10         	  143966	     24949 ns/op
profile_java-10                	   61840	     58016 ns/op
profile_nodejs-10              	  109437	     32871 ns/op
profile_python-10              	  391418	      9156 ns/op
profile_ruby-10                	  307896	     11783 ns/op
profile_rust-10                	  636633	      5668 ns/op
```

String matching (prefix or suffix):
```
go.cpu.labels.pprof-10         	 3418387	      1060 ns/op
heap-10                        	 2446440	      1483 ns/op
dotnet.labels.pprof-10         	 1751536	      2057 ns/op
profile_java-10                	 2333244	      1541 ns/op
profile_nodejs-10              	 1722144	      2083 ns/op
profile_python-10              	 4395427	       821 ns/op
profile_ruby-10                	 4546932	       793 ns/op
profile_rust-10                	 2960469	      1221 ns/op
```

benchstat:

```
goos: darwin
goarch: arm64
pkg: github.com/grafana/pyroscope/pkg/pprof
                                                    │    old.txt    │               new.txt               │
                                                    │    sec/op     │   sec/op     vs base                │
_GetProfileLanguage/testdata/go.cpu.labels.pprof-10    10.589µ ± 0%   1.016µ ± 0%  -90.40% (p=0.000 n=10)
_GetProfileLanguage/testdata/heap-10                   18.034µ ± 1%   1.411µ ± 0%  -92.18% (p=0.000 n=10)
_GetProfileLanguage/testdata/dotnet.labels.pprof-10    24.765µ ± 1%   1.981µ ± 1%  -92.00% (p=0.000 n=10)
_GetProfileLanguage/testdata/profile_java-10           57.873µ ± 1%   1.473µ ± 0%  -97.46% (p=0.000 n=10)
_GetProfileLanguage/testdata/profile_nodejs-10         32.770µ ± 0%   2.010µ ± 1%  -93.87% (p=0.000 n=10)
_GetProfileLanguage/testdata/profile_python-10         9125.0n ± 1%   790.8n ± 0%  -91.33% (p=0.000 n=10)
_GetProfileLanguage/testdata/profile_ruby-10          11754.5n ± 1%   772.7n ± 1%  -93.43% (p=0.000 n=10)
_GetProfileLanguage/testdata/profile_rust-10            5.679µ ± 1%   1.184µ ± 0%  -79.16% (p=0.000 n=10)
geomean                                                 16.49µ        1.253µ       -92.40%
```

As before, there is a chance of false detection - though this is only used for statistics for now.

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -239,6 +239,10 @@ func (d *Distributor) GetProfileLanguage(series *distributormodel.ProfileSeries)
 	if len(series.Samples) == 0 {
 		return "unknown"
 	}
+	lang := series.GetLanguage()
+	if lang != "" {
+		return lang
+	}
 	return pprof.GetLanguage(series.Samples[0].Profile, d.logger)
 }
 
--- a/pkg/distributor/model/push.go
+++ b/pkg/distributor/model/push.go
@@ -2,6 +2,7 @@ package model
 
 import (
 	v1 "github.com/grafana/pyroscope/api/gen/proto/go/types/v1"
+	phlaremodel "github.com/grafana/pyroscope/pkg/model"
 	"github.com/grafana/pyroscope/pkg/pprof"
 )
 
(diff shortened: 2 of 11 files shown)


korniltsev:
There may or may be not a new spy label from grafana agent pyroscope.java component

https://github.com/grafana/agent/pull/5985/files#diff-a0c3d4195b76376947e1f9e0059dc3538e97c3fb10ba70d9000ea74b902c1f8dR244

Let me know if you have any preference for its value or its presence

aleks-p:
Thanks @korniltsev, the value makes sense to me. I'll add the mapping for it.

PR #2824, opened 2023-12-08
Use the request context for the admin handlers

### LINKED ISSUE #2821: Admin API should use per request context.
Seems to be using a background context.

https://github.com/grafana/pyroscope/blob/main/pkg/operations/admin.go#L23

better use [Request.Context](https://pkg.go.dev/net/http#Request.Context) for proper task cancellation.

aleks-p:
Closes #2821


cyriltovena (approved):
LGTM

PR #2825, opened 2023-12-11
Avoid loading all symbols for each partition destination

cyriltovena:
Sounds like `partition.init` doesn't care and will always load all symbols for the current partition.

It seems like an oversight.

### DIFF
--- a/pkg/phlaredb/symdb/block_reader.go
+++ b/pkg/phlaredb/symdb/block_reader.go
@@ -31,9 +31,10 @@ type Reader struct {
 
 	chunkFetchBufferSize int
 
-	index         IndexFile
-	partitions    []*partition
-	partitionsMap map[uint64]*partition
+	index            IndexFile
+	partitions       []*partition
+	partitionsMap    map[uint64]*partition
+	partitionsLoaded bool
 
 	locations parquetobj.File
 	mappings  parquetobj.File
@@ -188,8 +189,10 @@ func (r *Reader) partition(ctx context.Context, partition uint64) (*partition, e
 	if !ok {
 		return nil, ErrPartitionNotFound
 	}
-	if err := p.init(ctx); err != nil {
-		return nil, err
+	if !r.partitionsLoaded {
+		if err := p.init(ctx); err != nil {
+			return nil, err
+		}
 	}
 	return p, nil
 }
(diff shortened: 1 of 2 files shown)


cyriltovena:
The tests is showing that `Load` and query path is not friendly. Thinking about how to solve this, basically we shouldn't not use the symbols for the read path if Load is used.

pkg/phlaredb/symdb/block_reader.go
```diff
@@ -183,15 +184,28 @@ func (r *Reader) Partition(ctx context.Context, partition uint64) (PartitionRead
...
+}
+
+// loadedPartition is a wrapper around partition that does not
+// release the resources on Release.
+// This is used in combination with `Load` at compaction time where all partition are loaded in memory once and not released.
+type loadedPartition struct {
+	*partition
+}
+
+func (l *loadedPartition) Release() {
```
kolesnikovae, line 208: Probably, we could avoid `*loadedPartition` as we have access to `Reader` members from the `*partition` methods and can check if it was loaded or not, like:

```go
func (p *partition) init(ctx context.Context) (err error) {
	if !p.reader.loaded {
		err = p.tx().fetch(ctx)
	}
	return err
}

func (p *partition) Release() {
	if !p.reader.loaded {
		p.tx().release()
	}
}

```

kolesnikovae (approved):
Nice catch, thanks for fixing this!

> The tests is showing that Load and query path is not friendly. Thinking about how to solve this, basically we shouldn't not use the symbols for the read path if Load is used.

Agreed: the only use case for `Load` is compaction, and it should never be used in our regular queries (however, loaded blocks should be _queryable_). Moreover, the test itself is not quite useful – it's worth counting object storage client calls to ensure that a partition is only fetched once at Load call. I'll take care of the test when will optimize symbols rewriter (this week hopefully)

pkg/phlaredb/symdb/block_reader.go
cyriltovena: Yeah that sounds good.

cyriltovena: well now I remember I did it this way because I could not get a race solved.

kolesnikovae: Hm, which race?

PR #2826, opened 2023-12-11
Update metric in profile-metrics.json

Rperry2174:
Updates metric in profilemetrics


petethepig (approved):

public/app/constants/profile-metrics.json
```diff
@@ -43,7 +43,7 @@
   },
   "memory:alloc_objects:count:space:bytes": {
     "id": "memory:alloc_objects:count:space:bytes",
```
aleks-p, line 45: Unrelated to the change but still relevant I think, why do we need the `:space:bytes` in the key and id here?

aleks-p (approved):

PR #2828, opened 2023-12-11
Create "Ingest and analyze profile data" section in docs

aleks-p:
Changes:

1. Reorder top level items
2. Add a new top level item "Ingest and analyze profile data"
3. Move "profiling types", "analyzing profiles" and "profile cli" to the new section
4. Minor improvement to the profile CLI docs

Screenshot 2023-12-11 at 16 12 11


docs/sources/ingest-and-analyze-profile-data/profile-cli.md
```diff
@@ -216,19 +241,30 @@ You can use the `profilecli query merge` command to retrieve a merged (aggregate
...
    - Example command:
      ```bash
      export PROFILECLI_URL=https://profiles-prod-001.grafana.net
      export PROFILECLI_USERNAME=my_username
      export PROFILECLI_PASSWORD=my_password
 
      profilecli query merge \
-         --query='{service_name="my_application_name"}' \
          --profile-type=memory:inuse_space:bytes:space:bytes
+         --query='{service_name="my_application_name"}' \
```
Rperry2174, line 256: can we explicitly write out the defaults so that people copy-pasting are aware of how and what format to change the time frame for something like this ?

Rperry2174: also is this really "merging" or just querying? 

for example if I query data from 12:00 to 12:45 and I merge data from 12:00-12:45 isn't that the same thing?

knylander-grafana (approved):
I did some minor updates to this content--mostly just fixing headings and some formatting. Let's get this approved so we can work on the introductory material. Some o f the content on Analyze profiles will move to the introduction content.

docs/sources/ingest-and-analyze-profile-data/profile-cli.md
```diff
@@ -1,111 +1,136 @@
...
+
+## Uploading a profile to a Pyroscope server using `profilecli`
 
 Using `profilecli` streamlines the process of uploading profiles to Pyroscope, making it a convenient alternative to manual HTTP requests.
 
 ### Prerequisites
 
-- Ensure you have `profilecli` installed on your system by following the installation steps above.
-- Have the pprof file you want to upload ready.
+- Ensure you have `profilecli` installed on your system by following the [installation](#installation) steps above.
```
knylander-grafana, line 118:
```suggestion
- Ensure you have `profilecli` installed on your system by following the [installation](#install-profile-cli) steps above.
```

docs/sources/ingest-and-analyze-profile-data/profile-cli.md
aleks-p: > can we explicitly write out the defaults so that people copy-pasting are aware of how and what format to change the time frame for something like this ?

Done!

> also is this really "merging" or just querying?
> 
> for example if I query data from 12:00 to 12:45 and I merge data from 12:00-12:45 isn't that the same thing?

We call it "merge" throughout the codebase because we are (potentially) merging data from multiple profiles into one. I agree that it is a confusing word, merge is often used for mutating operations (e.g., `git merge`).

I don't think we can call it just `query` here since we have two subcommands (`merge` and `series`) but we could call it `query profile` potentially (with the risk of confusion about whether it returns data from one or more profiles). I'll leave it as it is for now though to avoid a breaking change at this time.

PR #2838, opened 2023-12-12
fix: Use a pool of LabelPairs to reduce memory allocs

bryanhuhta:
We see pretty heavy memory usage when `querier.v1.QueryService/Series` calls are made on blocks with large a `index.tsdb`. For a real world example, see [here](https://admin-ops-us-east-0.grafana-ops.net/grafana/a/grafana-pyroscope-app/single?query=memory%3Aalloc_space%3Abytes%3Aspace%3Abytes%7Bservice_name%3D%22profiles-prod-003%2Fstore-gateway%22%7D&from=1702394320000&until=1702395150000).

 In this [flamegraph](https://flamegraph.com/share/7df4e563-993c-11ee-9dd6-fa95dfb0a413/sub/alloc_space), a ~7.7mb `index.tsdb` file is being used, resulting in  using ~680mb of memory.

Screenshot 2023-12-12 at 4 26 51 PM

The culprit is this loop in [`getUniqueLabelsSets`](https://github.com/grafana/pyroscope/blob/e1a13caeb968643ccc5a86bef176c4b35a26277f/pkg/phlaredb/block_querier.go#L1990):

https://github.com/grafana/pyroscope/blob/e1a13caeb968643ccc5a86bef176c4b35a26277f/pkg/phlaredb/block_querier.go#L1992-L2006

The high memory usage comes from us [allocating a slice every iteration](https://github.com/grafana/pyroscope/blob/e1a13caeb968643ccc5a86bef176c4b35a26277f/pkg/phlaredb/block_querier.go#L1993) and even more egregiously, [allocating a new `typesv1.LabelPair` on the heap](https://github.com/grafana/pyroscope/blob/e1a13caeb968643ccc5a86bef176c4b35a26277f/pkg/phlaredb/block_querier.go#L2002-L2005) every time we have a label name/value match.

This PR changes `getUniqueLabelsSets` to use a slice pool strategy, where we allocate a pool of the exact size and fill it with pointers to heap memory. We reuse this pool instead of allocating new memory each iteration of `postings.Next()`. If we need the contents of the pool, we explicitly copy out of the pool. This substantially reduces memory usage.

After the change, the new [flamegraph](https://flamegraph.com/share/5f33361f-993d-11ee-aa0f-1a5f4d469f1a/sub/alloc_space) shows `getUniqueLabelsSets` using ~56mb of memory:

Screenshot 2023-12-12 at 4 27 05 PM

And some benchmarks using the same 7.7mb `index.tsdb`:

| Name | iterations | ns/op    | B/op    | allocs/op |
|------:|------------:|----------:|---------:|-----------:|
| old  | 37         | 27478623 | 9061538 | 188762    |
| new  | 39         | 26336128 | 1681138 | 83927     |
| change |  |  **-4.1%** | **-81.4%** | **-55.5%** | 

Of course, now we have the problem of `index.(*Reader).LabelValueFor` using a lot of memory, but this is a significant and simple enough step that I think it's justifiable to merge on its own.


pkg/phlaredb/block_querier.go
```diff
@@ -1999,22 +2009,32 @@ func (b *singleBlockQuerier) getUniqueLabelsSets(postings index.Postings, names
...
+		labels := &typesv1.Labels{
+			Labels: make([]*typesv1.LabelPair, 0, len(matchedLabelsPool)),
+		}
+		for _, label := range matchedLabelsPool {
+			labels.Labels = append(labels.Labels, &typesv1.LabelPair{
+				Name:  label.Name,
+				Value: label.Value,
+			})
+		}
+		labelsSets = append(labelsSets, labels)
```
cyriltovena, line 2037:
```suggestion
		labelsSets = append(labelsSets, matchedLabelsPool.CloneVT())
```

cyriltovena (approved):
LGTM !!!!!!

pkg/phlaredb/block_querier.go
bryanhuhta: Nice one, fixed in https://github.com/grafana/pyroscope/pull/2838/commits/0f0779b9be17013495bed6b2a6d294d2e22e9290

PR #2842, opened 2023-12-13
Fixes span name for BlockSelect from store-gateway

cyriltovena:
I was looking at a trace and  thought we have a bug selecting Series when asking for Flamegraph.

In fact the span name was incorrect probably from copy-pasta.

image


cyriltovena:
cc @bryanhuhta I was not crazy :P

korniltsev (approved):

bryanhuhta (approved):
😂

PR #2843, opened 2023-12-13
Support discovering docker swarm tasks for ebpf.

cyriltovena:
While reading the ebpf code and documentation I figured we can easily support docker swarm.


korniltsev (approved):
Nice!

PR #2848, opened 2023-12-14
Remove go.mod replace for golang.org/x/exp

ptodev:
Removing a replace directive for `golang.org/x/exp` form the Pyroscope's go.mod files.

This is necessary because we are upgrading Grafana Agent to a new OpenTelemetry version, and it requires a new version of `github.com/grafana/loki/pkg/push` which needs the latest `golang.org/x/exp`. However, the Agent also depends on Pyroscope and it won't build correctly unless the Pyroscope is changed to use the latest `golang.org/x/exp`.

The reason why `golang.org/x/exp` has been problematic is because the return type of `SortFunc` changed from `bool` (in the [old version](https://pkg.go.dev/golang.org/x/exp@v0.0.0-20230711153332-06a737ee72cb/slices#SortFunc)) to `int` (in the [new version](https://pkg.go.dev/golang.org/x/exp@v0.0.0-20231206192017-f3f8817b8deb/slices#SortFunc)). Apparently some packages like to use `golang.org/x/exp` because it's more performant.


korniltsev (approved):
lgtm

PR #2852, opened 2023-12-14
Fix query split logic for LabelNames, LabelValues and ProfileTypes

aleks-p:
While looking at the ingesters memory usage in a high-load environment I noticed that a big portion of it came from `(*singleBlockQuerier) openTSDBIndex`. 

Looking at the `pyroscopedb_blocks_currently_open` metric, many ingesters have a large number of opened blocks (300+ in some cases). Narrowing that further with traces, the large number of open blocks came from requests to `LabelNames` and `LabelValues` with long query intervals. Currently these requests are served entirely by ingesters because until recently such requests didn't contain a `start` and `end` in the body (the client didn't send the range).

With https://github.com/grafana/grafana/pull/78861, clients started sending a `start` and `end` for the LabelNames and LabelValues APIs. This means we can now properly split the queries between ingesters and store gateways which should in theory reduce the memory pressure on ingesters as described above.

Notes:
- The ProfileTypes API is covered here but the behavior for this endpoint won't change until the clients start sending `start` and `end` for those requests (a separate Grafana PR to follow)
- The issue could also be solved with block plans but I've opted for a simple incremental change for now. If this works we could switch entirely to block plans and simplify queriers.
- Ingesters are currently not closing old blocks that they've opened, making this problem worse over time. If needed we could also add periodic eviction of blocks from memory, though I don't think this will make a big impact following the change done here

### DIFF
--- a/pkg/querier/querier.go
+++ b/pkg/querier/querier.go
@@ -155,10 +155,7 @@ func (q *Querier) ProfileTypes(ctx context.Context, req *connect.Request[querier
 
 	if storeQueries.ingester.shouldQuery {
 		group.Go(func() error {
-			ir, err := q.profileTypesFromIngesters(ctx, &ingestv1.ProfileTypesRequest{
-				Start: req.Msg.Start,
-				End:   req.Msg.End,
-			})
+			ir, err := q.profileTypesFromIngesters(ctx, storeQueries.ingester.ProfileTypesRequest(req.Msg))
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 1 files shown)


cyriltovena:
>  If needed we could also add periodic eviction of blocks from memory, though I don't think this will make a big impact following the change done here

We should at some point do that for ingester and store-gateway, I think it's more pressing for store-gateways.

cyriltovena (approved):
LGTM

Issue #2730, opened 2023-11-20
Make Compaction Stage part of the Sharding Key


cyriltovena:
@kolesnikovae  recently introduce the compaction stage size parameter in https://github.com/grafana/pyroscope/pull/2663 to help reduce memory usage during compaction.

We should consider making that new parameter part of the sharding key so that jobs from the same input block and shard but different stage is spread across the workload and run in parallel.

see https://github.com/grafana/pyroscope/blob/main/pkg/compactor/compactor.go#L835

cyriltovena:
This is more complex than I thought and we found better alternative so closing for now.

PR #2858, opened 2023-12-15
feat: Run compactor stages in parallel.

cyriltovena:
The idea is to add the stage index as part of the job shard key.

As suggested in https://github.com/grafana/pyroscope/issues/2730

However now that I did it, I realized that each jobs refers to the same group of blocks and they all will want to delete that group of block under completions which will be problematic.

Not sure how to solve this yet. cc @kolesnikovae  wdyt

### DIFF
--- a/pkg/compactor/bucket_compactor.go
+++ b/pkg/compactor/bucket_compactor.go
@@ -222,7 +222,7 @@ type Compactor interface {
 	// CompactWithSplitting merges and splits the source blocks into shardCount number of compacted blocks,
 	// and returns slice of block IDs.
 	// If given compacted block has no series, corresponding block ID will not be returned.
-	CompactWithSplitting(ctx context.Context, dst string, dirs []string, shardCount, stageSize uint64) (result []ulid.ULID, _ error)
+	CompactWithSplitting(ctx context.Context, dst string, dirs []string, shardCount, stageIndex, stageSize uint64) (result []ulid.ULID, _ error)
 }
 
 const (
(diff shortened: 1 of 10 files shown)


shaleynikov:
Probably wrong person. I didnt work on compactor

cyriltovena:
> Probably wrong person. I didnt work on compactor

Yes sorry this was for @kolesnikovae :)

kolesnikovae:
So, just to confirm, the issue is that when a job completes, we delete blocks locally and mark them for deletion in the object storage, right? https://github.com/grafana/pyroscope/blob/50f11fafe83eaf6db5a25edb78dfe009e224a2ac/pkg/compactor/bucket_compactor.go#L530-L537

As far as I see, there are two problems:
1. We download/remove a local block that could be part of another job that is running concurrently on the same compactor, or is part of the job that will be running after the current one. 

I think we could count references. When we create the list of jobs we should also create a reference counter for each of the blocks:
https://github.com/grafana/pyroscope/blob/50f11fafe83eaf6db5a25edb78dfe009e224a2ac/pkg/compactor/bucket_compactor.go#L799-L802

The initial count of references is the number of jobs referring to that block. And the counter is to be decremented each time such job completes. When the counter reaches 0, we remove the block locally.

2. We mark the source block for deletion albeit it could still be in use by another job (local or not).

TBH, for me it's not clear if we actually can mark blocks at this point, even without staging. It looks like the same source block can potentially be in use by another compactor.

cyriltovena:
Found a better alternative to speed up splitting.

Issue #2859, opened 2023-12-15
[DOC] Create a diagram for sampling example


bryanhuhta:
In our documentation for how to configure the Grafana Agent to sample scrape targets, we should have a diagram illustrating how the configuration works. This will help users better conceptualize the sharding action and selecting a subset of the shards.

See also:
- https://github.com/grafana/pyroscope/pull/2854#discussion_r1427569270
- https://github.com/grafana/pyroscope/pull/2854
- https://github.com/grafana/pyroscope/issues/2767

PR #2861, opened 2023-12-15
doc: Add sampling diagram

bryanhuhta:
closes https://github.com/grafana/pyroscope/issues/2859

Adds a diagram that illustrates the workflow of sampling scrape targets.


knylander-grafana (approved):
Thank you for adding the diagram!

PR #2864, opened 2023-12-18
feat: Improve symbols compaction

cyriltovena:
This improve symbols compaction by using a single output symbols, this means we still merge all symbols, but we don't duplicate them anymore.

While this increase the amount of data, ultimately this is a temporary state when splitting blocks.
When merging blocks without splitting, symbols will still end up fully compacted as the original design.

EDIT: Change to symlink and we're now back same time.

```
❯ benchstat before.txt after.txt
name              old time/op    new time/op    delta
_CompactSplit-16     5.49s ±21%     4.09s ±17%     ~     (p=0.114 n=4+4)

name              old alloc/op   new alloc/op   delta
_CompactSplit-16    6.45GB ± 0%    3.32GB ± 0%  -48.54%  (p=0.029 n=4+4)

name              old allocs/op  new allocs/op  delta
_CompactSplit-16     2.89M ± 0%     0.43M ± 0%  -85.09%  (p=0.029 n=4+4)
```

image

Testing shortly in ops.


kolesnikovae:
I very like the idea! The L2 vs L3 cost trade-off looks reasonable to me. I guess that now L3 requires significantly more memory than it used to be with split symbols, but it is probably still significantly less than the L2 peak. I'm wondering how we can make the merge phase cheaper (but still removing unreferenced symbols from the source blocks) – the first thing I would try is selective partitions loading: basically lazily load partitions _referenced_ in the block.

kolesnikovae (approved):
LGTM

cyriltovena:
> basically lazily load partitions _referenced_ in the block.

Yeah good thinking, I think also in the long run we might want to change how we split to account for symbols to avoid loading too much.

PR #2865, opened 2023-12-18
Add support to set readinessProbe.initialDelaySeconds in Helm

aleks-p:
Makes it possible to set `readinessProbe.initialDelaySeconds` in the Helm chart. For now planned to be used for the `store-gateway` that waits at startup (1 minute by default) for the sharding ring to be stable.


cyriltovena (approved):
LGTM

PR #2868, opened 2023-12-19
fix: inject JFR labels into pprof

kolesnikovae:
The PR changes the way we handle JFR labels: instead of creating a separate pprof profile for each series (a JFR context), we import JFR context labels as pprof sample labels. Without this change tracing integration for Java exports `profile_id`/`span_id` as a regular series lablel.


cyriltovena (approved):
LGTM

PR #2870, opened 2023-12-19
bugfix: Properly cleanup dst folder on compaction

cyriltovena:
The file is hardlinked or fully copied and so the original file is safe to delete.


kolesnikovae (approved):
LGTM

PR #2871, opened 2023-12-20
feat: Only load all symbols when splitting

cyriltovena:
I want to give a try to this in our largest cluster.

load vs fetch

image


kolesnikovae (approved):
LGTM

PR #2873, opened 2023-12-21
chore: bump dskit dependency

kolesnikovae:
`spanprofiler` has been moved to the dskit repository. It's been a while since we updated it last time, and there are a few backward incompatible changes


cyriltovena (approved):
LGTM

PR #2874, opened 2023-12-21
bugfix: Warn only when deleting twice during compaction

cyriltovena:
We often log this error:

```

  | ts=2023-12-21T09:46:56.652203687Z caller=block.go:164 level=warn component=compactor component=compactor tenant=27821 msg="requested to mark for deletion, but file already exists; this should not happen; investigate" err="file 01HJ5FB5R2Z145HJPBYQB0ZBMN/deletion-mark.json already exists in bucket" |  
-- | -- | --
```

This is confusing when it's actually from recurring cleanup this is fine that multiple compactor does try to delete the same file.


kolesnikovae (approved):
LGTM

Issue #2863, opened 2023-12-18
Compactor is not run when using target all


ntimo:
#### Describe the bug
When running pyroscope in the target all mode the compactor is not run and needs to be enabled by using target: „all,compactor“ . This is not the case with Mimir for example there the compactor is run when using target all

#### To Reproduce

1. Start Pyroscope 1.2.1 using target all
2. check that the compactor is not running 

#### Expected behavior

The compactor should be enabled

#### Environment

- Pyroscope docker container

cyriltovena:
Just pushed an image https://hub.docker.com/layers/grafana/pyroscope/main-ba25f23/images/sha256-b85abcf739537bffca81d2dd0e7bbf23f34af498b6046c250a56ba29b7959aff?context=explore if you want to try it out while it's unrelease.

PR #2875, opened 2023-12-21
feat: Adds the compactor for the all target

cyriltovena:
Wasn't there at first cause it was experimental but it seems to be doing good so let's add it to `all` target.

Fixes https://github.com/grafana/pyroscope/issues/2863


kolesnikovae (approved):
LGTM

PR #2878, opened 2023-12-26
feat: add function selector to pprof query

kolesnikovae:
The change extends pprof query interface with an option to scope the result to a specific function call sub-tree


api/types/v1/types.proto
```diff
@@ -71,3 +71,11 @@ enum TimeSeriesAggregationType {
   TIME_SERIES_AGGREGATION_TYPE_SUM = 0;
   TIME_SERIES_AGGREGATION_TYPE_AVERAGE = 1;
 }
+
+message FunctionSelector {
+  repeated Function stack_trace = 1;
```
aleks-p, line 76: A question, would the user select a subtree of a stack trace or can they select functions that are not in the same call chain? Depending on the answer, `functions` could be a more flexible name for the field here.

pkg/phlaredb/symdb/resolver.go
```diff
@@ -52,12 +67,15 @@ type lazyPartition struct {
 	done    chan struct{}
 }
 
-func NewResolver(ctx context.Context, s SymbolsReader) *Resolver {
+func NewResolver(ctx context.Context, s SymbolsReader, opts ...ResolverOption) *Resolver {
```
aleks-p, line 70: that is a nice way to make this more extensible, good stuff!

aleks-p (approved):
Looks good to me!

api/types/v1/types.proto
kolesnikovae: The stack trace points to the subtree root. In the future we will likely extend the selector to cover the case when all the function instances (all the stack traces containing the function) are to be selected. This is the reason why `FunctionSelector` is extendable: we can always add a new field (e.g `function`) that would serve for that purpose.

However, I don't really like names `FunctionSelector` and `Function` here, although they are quite precise (as of now). I'm thinking about renaming it to `LocationSelector` and `Location` respectively

pkg/phlaredb/symdb/resolver.go
kolesnikovae: Thanks to Rob Pike https://commandcenter.blogspot.com/2014/01/self-referential-functions-and-design.html :)

PR #2879, opened 2023-12-26
Helm: update agent configmap creation condition

QuantumEnigmaa:
This PR updates the condition for the `grafana-agent`'s configmap creation in the Helm chart.

Currently the agent's configmap is created whenever the agent is enabled in the values but because of this, it's impossible for a user to provide it's own custom config for the agent, even when using the `agent.agent.configMap.content` field. 
I added another condition to allow the creation of the agent's configmap which is the following : 
```
not .Values.agent.agent.configMap.create
```
This ensures that the agent's configmap in the templates is only created when one doesn't want to provide one's own custom config. Would it be the case, the user would have to set the  `agent.agent.configMap.create` field to `true` and write the wanted config in the `agent.agent.configMap.content` field without having the current default configmap created alongside.

Also, should I bump the chart's version ?


kolesnikovae:
Thank you for the contribution @QuantumEnigmaa! I agree, the way the agent is configured should be refined.

> Currently the agent's configmap is created whenever the agent is enabled in the values but because of this, it's impossible for a user to provide it's own custom config for the agent, even when using the agent.agent.configMap.content field.

Users should be able to provide their own `configMap` by overriding the `agent.configMap.name` field – in this case they are fully responsible for creation/management of the config map resource. Alternatively, if a user want to provide the config content via `agent.configMap.content`, to the best of my knowledge, `agent.configMap.name` should be empty.

However, as you mentioned, the pyroscope chart will generate a new config map resource regardless of the agent configuration.

> I added another condition to allow the creation of the agent's configmap which is the following :

TBH, I'm not 100% if we should rely on this parameter as it introduces a quite implicit dependency which, strictly speaking, isn't necessary in some cases (for example, if a user brings its own `configMap` resource and don't want agent/pyroscope chart to generate any extra resources).

I would consider migrating the default configuration we ship to the agent configuration via the `agent.configMap.content` parameter (+`agent.configMap.create`). In my opinion, this should remove any ambiguity.

That's being said, I have no objections to merging this PR but I would like to hear your thoughts

kolesnikovae (approved):

QuantumEnigmaa:
> I would consider migrating the default configuration we ship to the agent configMap via the agent.configMap.content parameter (+agent.configMap.create). In my opinion, this should remove any ambiguity.

I agree. I'm gonna move the config there and add some comments to explain how to provide another config

QuantumEnigmaa:
@kolesnikovae Can you check to make sure the change is suited to what you expected pls ? :)

kolesnikovae:
@QuantumEnigmaa thanks! I'll take a look by the end of the year ;)

operations/pyroscope/helm/pyroscope/values.yaml
```diff
@@ -186,8 +186,187 @@ agent:
...
+
+        pyroscope.write "pyroscope_write" {
+          endpoint {
+            {{- if hasKey .Values.pyroscope.components "distributor" }}
+            url = "http://{{ include "pyroscope.fullname" . }}-distributor.{{ .Release.Namespace }}.svc.cluster.local.:{{ (.Values.pyroscope.components.distributor.service).port | default .Values.pyroscope.service.port}}"
+            {{- else }}
+            url = "http://{{ include "pyroscope.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local.:{{ .Values.pyroscope.service.port }}"
+            {{- end }}
+          }
+        }
```
kolesnikovae, line 369: Ah, I just have realized that when the template is rendered in the agent chart, it has no access to `.Values.pyroscope`.

I'm sorry @QuantumEnigmaa, you did well, but we probably can't inject the default config to the agent's config map content (in a straight forward way). I think your former solution (check `create`) is the way to go

QuantumEnigmaa: > when the template is rendered in the agent chart, it has no access to .Values.pyroscope

Yeah you're right, I had totally forgot about that ! I reverted this change back to the first one I did :)

kolesnikovae (approved):
LGTM

PR #2881, opened 2023-12-28
feat: use downsampled data in profile queries

kolesnikovae:
The change adds ability to read downsampled profiling data from blocks to profile queries (stacktraces, pprof). Downsampled data is stored in the same block, in parquet tables. When a block is read, querier finds the best matching tables based on the query time range and fetches data for non-overlapping sub-ranges concurrently, reusing tsdb and symdb indices. Downsampled data is only used if `BlockHints` indicate that deduplication is not needed.


kolesnikovae:
As discussed: we want to add an option (tenant limit) to enable/disable it

pkg/util/time_test.go
```diff
@@ -0,0 +1,174 @@
+package util
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_SplitTimeRangeByResolution(t *testing.T) {
```
aleks-p, line 10: Do you think it would make sense to add a few tests for ranges that span over 2 or more days? I know this should work as expected, but it might not be super obvious when looking at the tests only.

aleks-p (approved):
Looks good to me! Left a few minor remarks.

pkg/util/time.go
```diff
@@ -98,3 +99,64 @@ func NewDisableableTicker(interval time.Duration) (func(), <-chan time.Time) {
...
+// SplitTimeRangeByResolution splits the given time range into the
+// minimal number of non-overlapping sub-ranges aligned with resolutions.
+// All ranges have inclusive start and end; one millisecond step.
+func SplitTimeRangeByResolution(start, end time.Time, resolutions []time.Duration, fn func(TimeRange)) {
+	if len(resolutions) == 0 {
+		fn(TimeRange{Start: start, End: end})
+		return
+	}
+
+	sort.Slice(resolutions, func(j, i int) bool {
```
aleks-p, line 118: This could be just me, but the order here (`j` -> `i`) threw me off a bit. If you don't pay attention it looks like resolutions are sorted in an ascending order which is not the case.

kolesnikovae: > This could be just me, but the order here (j -> i) threw me off a bit. If you don't pay attention it looks like resolutions are sorted in an ascending order which is not the case.

I'm pretty sure resolutions are sorted in an ascending order. Swapping `Less`'s args is used e.g. in the standard `sort.Reverse`:
```go
func Reverse(data Interface) Interface {
	return &reverse{data}
}

func (r reverse) Less(i, j int) bool {
	return r.Interface.Less(j, i)
}
```

What caught my eye is that the equality was not strict. Although it should not have any difference in our case, I had no intent – probably I should be more careful with github copilot suggestions :)

Anyways, I changed the order to `i, j` to not confuse anyone

pkg/util/time_test.go
kolesnikovae: Yep, added one

pkg/phlaredb/block_querier.go
```diff
@@ -2140,15 +2167,78 @@ func (q *singleBlockQuerier) openFiles(ctx context.Context) error {
...
+			resolutions = append(resolutions, k.resolution)
+		}
+	}
+	return resolutions
+}
+
+func downSampleAggregation(v typesv1.TimeSeriesAggregationType) string {
+	// TODO: Use constants defined in the `downsample` package.
+	if v == typesv1.TimeSeriesAggregationType_TIME_SERIES_AGGREGATION_TYPE_AVERAGE {
+		return "avg"
```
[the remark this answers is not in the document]
kolesnikovae: Yeah, honestly I doubt we'll add any other aggregation type any soon (avg will likely be handled from the `sum` table). I removed it for now – as you said, if there is no suitable table, we always fallback to the source table

pkg/util/time.go
aleks-p: Oh, I think I see where the ascending / descending confusion came from. The durations are sorted in descending order, but if viewed as resolutions (which they also are) they are sorted in an ascending order (lower resolution first) :)

PR #2882, opened 2024-01-02
Listen on 127.0.0.1 in tests to avoid firewall popups on MacOS

aleks-p:
When the firewall is enabled in MacOS, some tests that involve networking will trigger the firewall popups. Allowing or denying the connections has a temporary effect, once the binary changes the popups are back.

This change makes it so that we listen to 127.0.0.1 in networking tests, which should avoid the popups entirely for most developers.

Screenshot 2024-01-02 at 09 14 35


kolesnikovae (approved):
LGTM!

PR #2885, opened 2024-01-02
Update deployment diagram link

Rperry2174:
we're eventually going to have to update this diagram to add the other languages we support via the grafana agent so doing this so that we can update them in one place instead of multiple


kolesnikovae (approved):
LGTM

PR #2887, opened 2024-01-04
Helm: update ingress template's pathTypes

QuantumEnigmaa:
This PR updates the ingress template's `pathType` field to `ImplementationSpecific` for paths containing `.` since otherwise, when deploying the chart with ingress enabled one gets the following error message : 
```
admission webhook "validate.nginx.ingress.kubernetes.io" denied the request: ingress contains invalid paths: 
path /querier.v1.QuerierService/ cannot be used with pathType Prefix
path /push.v1.PusherService/ cannot be used with pathType Prefix
path /settings.v1.SettingsService/ cannot be used with pathType Prefix
```

This is due to the fact that `pathType: Prefix` only accepts alphanumeric characters and `/`, `_` and `-`.


kolesnikovae (approved):
LGTM

QuantumEnigmaa:
@kolesnikovae is there something else to do before merging ?

kolesnikovae:
@QuantumEnigmaa, thanks for the contribution! Merged

PR #2888, opened 2024-01-04
[WIP] feat: Persist settings in bucket storage

bryanhuhta:
WIP

Related: https://github.com/grafana/pyroscope-app-plugin/issues/43

Leverages bbolt to create an on-disk kv store which we can sync to bucket storage with every write. We can reuse the same bucket profiles are written to. We should refresh the blob in bucket storage every 24 hours or so to make sure the file doesn't expire.


bryanhuhta:
Superseded by https://github.com/grafana/pyroscope/pull/2905. Offline @petethepig and @cyriltovena recommended against using boltdb to keep our implementation as simple as possible.

PR #2894, opened 2024-01-08
Fix compactor metrics registration

aleks-p:
Fixes an issue introduced in #2880 where some compactor metrics are initialized multiple times. Verified in a dev environment.


cyriltovena (approved):
Lgtm

PR #2903, opened 2024-01-09
feat: Improve compaction shutdown

cyriltovena:
I've seen traces where it took 10m to compact. When we rollout a new version  it can get cancelled in the upload and deletion which is usually less than 2m.

I think we should leave a grace period for uploads and marking blocks deleted.


pkg/compactor/bucket_compactor.go
```diff
@@ -539,8 +539,14 @@ func (c *BucketCompactor) runCompactionJob(ctx context.Context, job *Job) (shoul
 	if err = verifyCompactedBlocksTimeRanges(compIDs, toCompactMinTime.UnixMilli(), toCompactMaxTime.UnixMilli(), subDir); err != nil {
 		level.Warn(jobLogger).Log("msg", "compacted blocks verification failed", "err", err)
```
simonswine, line 540: Nit: Should probably be an error:

```suggestion
		level.Error(jobLogger).Log("msg", "compacted blocks verification failed", "err", err)
```

simonswine (approved):
LGTM, I think you also should raise the `terminationGracePeriod` from the default of 30 seconds, for the compactor:

https://github.com/grafana/pyroscope/blob/318f60037c9799afb4819f3b30e6ea818b180412/operations/pyroscope/helm/pyroscope/values-micro-services.yaml#L48

PR #2905, opened 2024-01-09
feat: Use object storage to save tenant settings

bryanhuhta:
Related: https://github.com/grafana/pyroscope-app-plugin/issues/43

Syncs tenant settings to object storage. The settings are not synced when Pyroscope first starts, rather the settings will be synced when a read or write occurs, or after 24 hours during the refresh. Both reads and writes are served from object storage. Every 24 hours the settings will be re-written to keep the timestamp of the object storage file up-to-date, stopping it from being cleaned up by retention policies.

### DIFF
--- a/pkg/phlare/modules.go
+++ b/pkg/phlare/modules.go
@@ -143,6 +143,7 @@ func (f *Phlare) initTenantSettings() (services.Service, error) {
 		store, err = settings.NewBucketStore(f.storageBucket)
 	default:
 		store, err = settings.NewMemoryStore()
+		level.Warn(f.logger).Log("msg", "using in-memory settings store, changes will be lost after shutdown")
 	}
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to init settings store")


pkg/settings/bucket.go
```diff
@@ -0,0 +1,148 @@
...
+	defer s.rw.Unlock()
+
+	return s.unsafeFlush(ctx)
+}
+
+func (s *bucketStore) Close() error {
+	return s.bucket.Close()
+}
+
+// unsafeFlush will flush the store to disk. This is not thread-safe, the
```
petethepig, line 111:
```suggestion
// unsafeFlush will flush the store to object storage. This is not thread-safe, the
```

pkg/settings/bucket.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	err = s.bucket.Upload(ctx, settingsFilename, bytes.NewReader(data))
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// unsafeLoad will sync the store with an on-disk store, if found. This is not
```
petethepig, line 126:
```suggestion
// unsafeLoad will sync the store with an object storage store, if found. This is not
```

petethepig:
Looks good overall. I have 1 suggestion and 2 questions.

Suggestion 1: I think it is slightly better to do an `unsafeLoad` before each Get and Set. It is cheap, and this way you a) don't block service initialization if object storage is slow, b) don't cause a restart if object storage throws an error c) less likely to shoot yourself in the foot if 2 instances of same service are running for whatever reason.  I don't feel super strongly on this though as these are somewhat unlikely corner cases.

Question 1: is it okay for us to put objects into a bucket full of existing blocks? I'm slightly worried that other components might try to read this file and fail. I'd double check that this is okay.

Question 2: are we handling the retention issue? E.g I think currently if you put a random object into an existing bucket it will get deleted at some point because of retention settings. I remember talking about potentially refreshing the object periodically.

cyriltovena:
>Question 1: is it okay for us to put objects into a bucket full of existing blocks? I'm slightly worried that other components might try to read this file and fail. I'd double check that this is okay.

It's ok but not great because if we ever need to list this is might be a lot, I suggest a new folder.

cyriltovena:
>I think currently if you put a random object into an existing bucket it will get deleted at some point because of retention settings. I remember talking about potentially refreshing the object periodically.

The retention policy is on specific file extensions. We should double check not sure it's the case for all provider.

bryanhuhta:
@petethepig 

> Suggestion 1: I think it is slightly better to do an unsafeLoad before each Get and Set.

Your reasoning is valid here. I agree.

> Question 1: is it okay for us to put objects into a bucket full of existing blocks?

There is some risk here, though I think it's small and limited in scope. Even if we create a directory and put our settings.json in the directory, we need to create a directory object in the bucket. Barring provisioning an entirely new bucket, I don't know how we would get around another part of the application incorrectly reading anything we write to the bucket.

I did audit some of the callsites where the application iterates over the contents of the bucket and I've seen those callsites be diligent about making sure it only selected the correct files by applying some form of validation.

Putting this in a directory as suggested [here](https://github.com/grafana/pyroscope/pull/2905#issuecomment-1884345094) seems fair.

> Question 2: are we handling the retention issue?

@cyriltovena and @simonswine had good callouts here. In GCP we can match by extension which should be easy to keep this file around, but other providers don't have a similar mechanism.

bryanhuhta:
@simonswine 

> I think the better option could be to migrate to a more active retention management by the compactor, wdyt?

I like this option, as it will help keep our retention policies consistent across cloud providers and give us flexibility to write retention policies in code, allowing us to have more complex logic if necessary.

In relation to this PR, I would worry that it would push back merging this work too long while we implemented compactor level retention policies.

pkg/settings/bucket.go
bryanhuhta: Fixed in https://github.com/grafana/pyroscope/pull/2905/commits/199ad673e5d4516948982adb2a9a20752d69f91d

pkg/settings/bucket.go
bryanhuhta: Fixed in https://github.com/grafana/pyroscope/pull/2905/commits/199ad673e5d4516948982adb2a9a20752d69f91d

cyriltovena:
Agreed let's not block this PR even if we're not yet there with the retention.

pkg/phlare/modules.go
```diff
@@ -135,14 +135,22 @@ func (f *Phlare) initRuntimeConfig() (services.Service, error) {
...
 func (f *Phlare) initTenantSettings() (services.Service, error) {
-	mem, err := settings.NewMemoryStore()
+	var store settings.Store
+	var err error
+
+	switch {
+	case f.storageBucket != nil:
+		store, err = settings.NewBucketStore(f.storageBucket)
+	default:
+		store, err = settings.NewMemoryStore()
```
simonswine, line 145: A warning that settings are not stored would be nice

simonswine (approved):
Yes let's not block on retention! LGTM

simonswine:
Not blocking in any way, but what I am wondering  is: should we really expose this service to users of open source pyroscope? It feels to me this is something that is Grafana Cloud leaking into the open source product.

https://github.com/grafana/pyroscope/blob/59904f716f63fad310fd76e57cf0c775424d187b/operations/pyroscope/helm/pyroscope/values-micro-services.yaml#L87-L95

pkg/phlare/modules.go
bryanhuhta: Good idea, added in https://github.com/grafana/pyroscope/pull/2905/commits/6274c9877a39ed55190464aa7a7217a039403f35

bryanhuhta:
> should we really expose this service to users of open source pyroscope? It feels to me this is something that is Grafana Cloud leaking into the open source product.

Hmm I agree that this is leaking. I think at the very least it can be removed from the `values-micro-services.yaml`. Enterprising individuals will still be able to turn on the Settings API, but it won't be an "out-of-the-box" feature. I'll create a follow-up PR to do that work.

PR #2909, opened 2024-01-10
Refactor: Upgrade connect-go, protobuf and buf

simonswine:
connect-go switched import paths, hence a fair bit of refactoring required.


ebpf/go.mod
```diff
@@ -25,6 +25,7 @@ require (
 
 require (
 	github.com/beorn7/perks v1.0.1 // indirect
+	github.com/bufbuild/connect-go v1.10.0 // indirect
```
cyriltovena, line 28: ?

cyriltovena (approved):
LGTM

found one old package not sure if this was intentional check it out.

ebpf/go.mod
simonswine: It is complicated, but it is because of that line: https://github.com/grafana/pyroscope/blob/main/ebpf/go.mod#L12

I think I need to merge this first, then tag a new api version and then I can update the ebpf packgae

PR #2910, opened 2024-01-10
Handle profile types call properly without bucket store

simonswine:
Currently store-gateways are panicing if profile types are not found.


kolesnikovae (approved):
LGTM

PR #2911, opened 2024-01-10
Use api/v0.4.0 after releasing it with connect-go

simonswine:
This is a follow-up to #2909

### DIFF
--- a/ebpf/go.mod
+++ b/ebpf/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/cilium/ebpf v0.11.0
 	github.com/go-kit/log v0.2.1
 	github.com/google/pprof v0.0.0-20230926050212-f7f687d19a98
-	github.com/grafana/pyroscope/api v0.2.0
+	github.com/grafana/pyroscope/api v0.4.0
 	github.com/hashicorp/golang-lru/v2 v2.0.5
 	github.com/ianlancetaylor/demangle v0.0.0-20230524184225-eabc099b10ab
 	github.com/klauspost/compress v1.17.3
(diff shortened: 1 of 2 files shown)


go.work.sum
```diff
@@ -973,6 +969,8 @@ github.com/boynux/squid-exporter v1.10.5-0.20230618153315-c1fae094e18e h1:C1vYe7
 github.com/boynux/squid-exporter v1.10.5-0.20230618153315-c1fae094e18e/go.mod h1:8NpZERGK+R9DGuZqqsKfnf2qI/rh7yBT8End29IvgNA=
 github.com/bradfitz/gomemcache v0.0.0-20190913173617-a41fca850d0b h1:L/QXpzIa3pOvUGt1D1lA5KjYhPBAN/3iWdP7xeFS9F0=
 github.com/bradfitz/gomemcache v0.0.0-20190913173617-a41fca850d0b/go.mod h1:H0wQNHz2YrLsuXOZozoeDmnHXkNCRmMW0gwFWDfEZDA=
+github.com/bufbuild/connect-go v1.4.1 h1:6usL3JGjKhxQpvDlizP7u8VfjAr1JkckcAUbrdcbgNY=
+github.com/bufbuild/connect-go v1.4.1/go.mod h1:9iNvh/NOsfhNBUH5CtvXeVUskQO1xsrEviH7ZArwZ3I=
```
simonswine, line 973: I really don't know why this is there

cyriltovena: 🤷

simonswine: I would say it is better than before the PR, the resulting binary is free of bufbuild:

```
go version -m ./pyroscope | grep bufbuild
```

korniltsev: https://github.com/grafana/agent/blob/v0.35.4/go.mod#L21

simonswine: It does, thanks!

korniltsev: I also noticed that there are entries which I can delete manually and `go` command does not add it back :D

cyriltovena (approved):
LGTM

PR #2912, opened 2024-01-11
Use LabelValues for ProfileTypes call in querier

simonswine:
This will allow us to remove the ProfileType endpoints from Ingesters and store-gateways in release v1.4 or later


aleks-p (approved):
A nice simplification, looks good to me!

Issue #2915, opened 2024-01-11
Ingester retention enforcement tries to delete `lost+found`


simonswine:
#### Describe the bug

Ingester retention enforcement accidentally tries to delete the `lost+found` directory from extX filesystems

```
ts=2024-01-11T16:47:26.766656102Z caller=retention.go:147 level=error msg="failed to delete uploaded blocks, could not get blocks for tenant" err="open data/lost+found/local: no such file or directory" tenantID=lost+found
```

#### Expected behavior

`lost+found` is ignored, potentially we should check if we have permission to delete first.

cc @bryanhuhta

PR #2916, opened 2024-01-11
fix: Skip non-tenant directories during block cleanup

bryanhuhta:
Fixes https://github.com/grafana/pyroscope/issues/2915

Some file systems will create a `lost+found` directory. This will never contain data relevant to Pyroscope, so we should skip inspecting this directory.

More generally, this skips any directory that doesn't have a `local` subdirectory.

### DIFF
--- a/pkg/ingester/retention.go
+++ b/pkg/ingester/retention.go
@@ -25,10 +25,6 @@ import (
 const (
 	// TODO(kolesnikovae): Unify with pkg/phlaredb.
 	phlareDBLocalPath = "local"
-
-	// The name of the 'lost+found' directory which can be created on systems
-	// that use fsck (https://en.wikipedia.org/wiki/Fsck).
-	lostAndFoundDirName = "lost+found"
 )
 
 // newDiskCleaner creates a service that will intermittently clean blocks from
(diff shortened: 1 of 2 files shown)


simonswine (approved):
LGTM. Thanks for the fix and the quick turn around 🙂

pkg/ingester/retention.go
```diff
@@ -371,6 +375,11 @@ func (bm *realFSBlockManager) GetTenantIDs(ctx context.Context) ([]string, error
 		if !tenantDir.IsDir() {
 			continue
 		}
+
+		if tenantDir.Name() == lostAndFoundDirName {
```
petethepig, line 379: i'm slightly worried that `lost+found` is not the last directory that might cause issues here. Instead of having a blocklist of directories, that we have to keep up to date, can we go the opposite way and find some heuristic that would tell us that a directory is a "pyroscope data directory"?

simonswine: I guess not having a `local` subfolder might be a good enough heuristic to skip

bryanhuhta: Since we are ultimately going to look into the local folder to delete blocks that would be a good start. Can we also assume tenant ids are going to be integers? That would allow us to also skip directories which aren't integer names.

aleks-p: we do have the `anonymous` tenant as well, though we could skip (or add manually)

petethepig: I think tenant ids can be anything.

> I guess not having a local subfolder might be a good enough heuristic to skip

this sounds like the best one

bryanhuhta: This is fixed in https://github.com/grafana/pyroscope/pull/2916/commits/a952c4d17e14c771acad7afeaad50598fc924484. The `isTenantDir` function holds the logic to determine if a directory is a tenant or not.

petethepig (approved):

PR #2917, opened 2024-01-12
fix: pprof profile builder panics

kolesnikovae:
https://github.com/grafana/pyroscope/pull/2878 introduced a bug where a pprof query with max nodes limit set causes panic. It only affects our experimental AI Assistant where we do use pprof truncation


cyriltovena (approved):
LGTM

I was confused at first on the renaming of the parameter.

kolesnikovae:
> I was confused at first on the renaming of the parameter

Yeah, decided to do it now, before the API is released

PR #2920, opened 2024-01-12
refactor: Removes the delta reserved labels from storage.

cyriltovena:
based @kolesnikovae feedback here https://github.com/grafana/pyroscope/pull/2914#discussion_r1449952138

I think we should remove the `__delta__` reserved labels which is used to disable delta of samples between profiles.


kolesnikovae (approved):
LGTM

PR #2921, opened 2024-01-12
Fix deploying `make deploy-micro-services` after `make deploy`

simonswine:
Make deploy manually creates a service to have the same endpoint for
single-binary and micro-services, which then interferes with the next
deployment of microservces.


cyriltovena (approved):
LGTM

PR #2931, opened 2024-01-17
docs(lambda): AWS Lambda extension SDK support

kolesnikovae:
Presently, we don't distribute Lambda layers tailored to specific runtimes, therefore users have to manage SDK dependency on their own.

While AWS offers comprehensive guidance on this, it might be worthwhile for us to consider packaging our SDKs as layers. Please refer to [AWS Distro for OpenTelemetry](https://aws-otel.github.io/docs/getting-started/lambda) (ADOT) for example. Created a separate issue: https://github.com/grafana/pyroscope-lambda-extension/issues/34

/cc @Rperry2174 @knylander-grafana

### DIFF
--- a/docs/sources/configure-client/aws-lambda.md
+++ b/docs/sources/configure-client/aws-lambda.md
@@ -62,6 +62,18 @@ Configure the extension with the following environment variables:
 
 ### Integrate the Pyroscope SDK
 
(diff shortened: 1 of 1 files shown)


knylander-grafana (approved):
Thank you for updating the docs

PR #2939, opened 2024-01-17
Correct river config in go_pull.md

thomson:
The Pyroscope.write block in the river config didn't next the `basic_auth` block within the endpoint block.


korniltsev (approved):
Thanks

PR #2947, opened 2024-01-18
Improve block grouping in admin block viewer

aleks-p:
The code assumed MinTime is aligned to 0 seconds, but we've found a case where this is not the case. This could happen due to clock-skew on a node, or code-related factors. 

The assumption is now changed to be aligned to either 0 or 30 minutes instead. 

cc. @cyriltovena

### DIFF
--- a/pkg/operations/handlers.go
+++ b/pkg/operations/handlers.go
@@ -90,14 +90,15 @@ func (h *Handlers) filterAndGroupBlocks(index *bucketindex.Index, query *blockQu
 			blkGroup, ok := blockGroupMap[truncatedMinTime]
 			if !ok {
 				blkGroup = &blockGroup{
-					MinTime:          truncatedMinTime,
-					FormattedMinTime: truncatedMinTime.Format(time.RFC3339),
-					Blocks:           make([]*blockDetails, 0),
-					MinTimeAge:       humanize.RelTime(blk.MinTime.Time(), time.Now(), "ago", ""),
+					MinTime:                 truncatedMinTime,
+					FormattedMinTime:        truncatedMinTime.Format(time.RFC3339),
+					Blocks:                  make([]*blockDetails, 0),
+					MinTimeAge:              humanize.RelTime(blk.MinTime.Time(), time.Now(), "ago", ""),
+					MaxBlockDurationMinutes: int(math.Round(blk.MaxTime.Sub(blk.MinTime).Minutes())),
 				}
 				blockGroups = append(blockGroups, blkGroup)
 			}
-			blkGroup.Blocks = append(blkGroup.Blocks, &blockDetails{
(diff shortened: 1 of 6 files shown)


pkg/operations/handlers.go
```diff
@@ -119,7 +119,7 @@ func postProcessBlockGroups(blockGroups []*blockGroup) *blockListResult {
 	maxBlocksPerGroup := 0
 	for i := 0; i < len(blockGroups); i += 1 {
 		blockGroup := blockGroups[i]
-		if i < len(blockGroups)-1 && !strings.Contains(blockGroup.FormattedMinTime, "0:00Z") {
+		if i < len(blockGroups)-1 && shouldAddToNextGroup(blockGroup.FormattedMinTime) {
```
simonswine, line 122: Wouldn't it be better if we group them into equal hours (or half hours), rather than messing with strings?

```
minTime.Truncate(time.Hours)
```

aleks-p: Thanks @simonswine. Switched to using `pyroscopedb.max-block-duration` to avoid having to "guess" the block duration, then truncating and grouping on this instead of strings. Compacted blocks can have a different duration but this still works well enough for now.

simonswine: Yes that's even better, all compactor block ranges will have to be multiples of the max-block-duration.

pkg/operations/handlers.go
```diff
@@ -86,68 +86,53 @@ func filterAndGroupBlocks(index *bucketindex.Index, query *blockQuery) *blockLis
...
-		groupDuration = int(math.Round(groupOne.MinTime.Sub(groupTwo.MinTime).Minutes()))
-	}
+
 	return &blockListResult{
-		BlockGroups:       finalBlockGroups,
-		MaxBlocksPerGroup: maxBlocksPerGroup,
-		GroupDuration:     groupDuration,
+		BlockGroups:          blockGroups,
+		MaxBlocksPerGroup:    maxBlocksPerGroup,
+		GroupDurationMinutes: maxBlockGroupDuration,
```
simonswine, line 135: Nit: Maybe we can make that a time.Duration:

```suggestion
		GroupDurationMinutes: time.Minutes*maxBlockGroupDuration,
```

Currently it shows as `5` or `60`, I would prefer `5m` `1h`

simonswine (approved):
LGTM

pkg/operations/handlers.go
aleks-p: We need the actual numbers for the grid view (we do some calculations there). I can add a human readable duration in addition to this though for the table view.

simonswine: Wasn't realizing, thanks for fixing it

PR #2953, opened 2024-01-19
feat: async profile row iterator

kolesnikovae:
The PR addresses one of the issues with data streaming in the block read path: when we select profile rows for filtering (by profile series ID and the query time range) we synchronously collect the list of filtered profiles, blocking the data flow. This behaviour has been changed so that filtered profiles are passed down to the stream in batches, asynchronously.

It's not expected that the overall query performance will improve in all cases. Usually the delay caused by the described blocking is very small and does not have impact on the overall query latency and/or resource consumption. However, it may be helpful with huge datasets of high cardinality.

### DIFF
--- a/pkg/iter/batch_async.go
+++ b/pkg/iter/batch_async.go
@@ -81,6 +81,9 @@ func (x *AsyncBatchIterator[T, N]) loadBatch() bool {
 	var b batch[N]
 	select {
 	case b = <-x.c:
+		if b.done != nil {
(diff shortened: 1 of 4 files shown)


simonswine (approved):
LGTM

Thanks, this is a great cleanup. I have not tried to benchmark this, but it still should go ahead.

cyriltovena (approved):
100% thank you for doing this.

PR #2961, opened 2024-01-24
feat: flamegraph node time-series

### LINKED ISSUE #2228: Function metric

kolesnikovae:
Resolves #2228


api/types/v1/types.proto
```diff
@@ -74,10 +74,10 @@ enum TimeSeriesAggregationType {
...
 // StackTraceSelector is used for filtering stack traces by locations.
 message StackTraceSelector {
-  // Only stack traces that belong to the subtree (have the prefix provided)
-  // will be selected. If empty, the filter is ignored.
-  // Subtree root location is the last element.
-  repeated Location subtree_root = 1;
+  // Stack trace of the call site. Root at call_site[0].
+  // Only stack traces having the prefix provided will be selected.
+  // If empty, the filter is ignored.
+  repeated Location call_site = 1;
```
kolesnikovae, line 80: Renamed it one more time – as the API it not used anywhere, should not be an issue

cyriltovena: I like it !

cyriltovena: cc @bryanhuhta for the hackathon this is what we're going to use in all API when clicking on a function to reduce size of pprof and zoom on a single function.

pkg/phlaredb/symdb/resolver_stacktraces.go
```diff
@@ -0,0 +1,171 @@
...
+	// Total is the total sum of sample values attributed to the node and
+	// its descendants.
+	Total uint64
+	// LocationFlat is the sum of sample values directly attributed to the
+	// node location, irrespectively of the call chain.
+	LocationFlat uint64
+	// LocationTotal is the total sum of sample values attributed to the
+	// node location and its descendants, irrespectively of the call chain.
+	LocationTotal uint64
+}
```
kolesnikovae, line 23: Please note that at the moment only `Total` is actually used. We can return all, but we need some toggle in UI that switches between them. Otherwise, the timeline view becomes too busy (especially, if `group by` is used)

cyriltovena (approved):
LGTM

PR #2962, opened 2024-01-24
Remove tenant settings from microservices deploy

simonswine:
This fixes a problem when deploying using `make deploy-microservices`:

```
❯ make deploy-micro-services
[...]
Error: UPGRADE FAILED: error validating "": error validating data: kind not set
make: *** [Makefile:403: deploy-micro-services] Error 1

```

A bit shameful but it needed a `git bisect` to find this.


cyriltovena (approved):
LGTM

PR #2963, opened 2024-01-24
Ad Hoc Profiles

aleks-p:
Adds APIs for uploading and retrieving "ad hoc" profiles. Profiles are stored in the tenant object storage, in a new directory (`adhoc/`) next to `phlaredb/`. 

The code for parsing the data and doing conversions is copied (with minor adaptations) from OG Pyroscope to maintain feature parity with the original Ad Hoc implementation. 

> I've also added a new "sidekick" module that is meant to own this feature and perhaps others (tenant settings) in the future, as a means to simplify operations. An alternative could be to create a single deployment with multiple targets (e.g., `-target tenant-settings,ad-hoc-profiles,admin`). cc @cyriltovena, @simonswine 

Edit: As of ad79c09 the module is just `ad-hoc-profiles`. Deployment considerations will be handled elsewhere. 

The only validation done in this PR is to make sure the profile is readable. In the future, we could also add per-tenant limits (e.g., storage usage). 

TODO: 
- tests


pkg/sidekick/adhocprofiles/adhocprofiles.go
```diff
@@ -0,0 +1,247 @@
...
+		return nil, errors.Wrapf(err, "failed to parse profile")
+	}
+
+	bucket, err := a.getBucket(tenantID)
+	if err != nil {
+		return nil, err
+	}
+
+	uid := ulid.MustNew(ulid.Timestamp(adHocProfile.UploadedAt), rand.Reader)
+	id := strings.Join([]string{uid.String(), adHocProfile.Name}, "-")
```
aleks-p, line 83: Keys in the object storage will have the form `01HMXV8BF4EH71NBYZNPPVGJ2X-cpu.pprof`, using `ULID` for the first segment to simplify embedding a timestamp. This is then used in the `/List` endpoint to return meaningful data without querying each individual profile.

pkg/sidekick/sidekick.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/grafana/pyroscope/pkg/frontend"
+	"github.com/grafana/pyroscope/pkg/objstore"
+	"github.com/grafana/pyroscope/pkg/sidekick/adhocprofiles"
+)
+
+type Sidekick struct {
+	services.Service
+
+	AdHocProfiles *adhocprofiles.AdHocProfiles
```
aleks-p, line 17: I could make AdHocProfiles a subservice perhaps, if we decide that having a Sidekick module still makes sense. This could come in in a second PR where we "move" tenant settings here.

pkg/adhocprofiles/adhocprofiles.go
```diff
@@ -0,0 +1,256 @@
...
+	}), nil
+}
+
+func (a *AdHocProfiles) Get(ctx context.Context, c *connect.Request[v1.AdHocProfilesGetRequest]) (*connect.Response[v1.AdHocProfilesGetResponse], error) {
+	tenantID, err := tenant.TenantID(ctx)
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, err)
+	}
+
+	bucket, err := a.getBucket(tenantID)
```
cyriltovena, line 125: Shouldn't you use here ?

```
bucket, err := a.getBucketFromContext(ctx)
	if err != nil {
		return nil, err
	}
```

pkg/adhocprofiles/adhocprofiles.go
```diff
@@ -0,0 +1,256 @@
...
+		return nil, err
+	}
+
+	profiles := make([]*v1.AdHocProfilesProfileMetadata, 0)
+	err = bucket.Iter(ctx, "", func(s string) error {
+		separatorIndex := bytes.IndexByte([]byte(s), '-')
+		id, err := ulid.Parse(s[0:separatorIndex])
+		if err != nil {
+			level.Warn(a.logger).Log("msg", "cannot parse ad hoc profile", "key", s, "err", err)
+			return nil
```
cyriltovena, line 183: nit: may be we can continue here ? Otherwise a tenant can get corrupted and will never be able to list the rest ?

pkg/adhocprofiles/adhocprofiles.go
```diff
@@ -0,0 +1,256 @@
...
+
+func (a *AdHocProfiles) List(ctx context.Context, c *connect.Request[v1.AdHocProfilesListRequest]) (*connect.Response[v1.AdHocProfilesListResponse], error) {
+	bucket, err := a.getBucketFromContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	profiles := make([]*v1.AdHocProfilesProfileMetadata, 0)
+	err = bucket.Iter(ctx, "", func(s string) error {
+		separatorIndex := bytes.IndexByte([]byte(s), '-')
```
cyriltovena, line 179:
```suggestion
		separatorIndex := strings.IndexRune(s, '-')
```

This is to avoid []byte conversion which copies the string, it's irrelevant here but I just wanted to let you know if you ever do that in a hot path be aware of string <-> []byte conversion.

pkg/adhocprofiles/adhocprofiles.go
```diff
@@ -0,0 +1,256 @@
...
+
+func (a *AdHocProfiles) getBucketFromContext(ctx context.Context) (objstore.Bucket, error) {
+	tenantID, err := tenant.TenantID(ctx)
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, err)
+	}
+	return a.getBucket(tenantID)
+}
+
+func (a *AdHocProfiles) getBucket(tenantID string) (objstore.Bucket, error) {
```
cyriltovena, line 207: I don't think you need to cache buckets you could create one child per request and throw it away.

but may be there's a reason ?

cyriltovena (approved):
Amazing LGTM !

PR #2964, opened 2024-01-25
fix: Use correct field in `mappingsKey`

bryanhuhta:
Previously `HasLineNumber` in `mappingsKey` used the value of `HasFunctions`. This appears to not have caused correctness problems, but this PR fixes it anyways to avoid subtle bugs in the future.


kolesnikovae:
Nice catch!

kolesnikovae (approved):

PR #2965, opened 2024-01-25
Experimental: Adds a Github Integration to fetch profiled code

cyriltovena:
This adds a set of API that allows users to connect their Pyroscope to Github and fetch profiling code for future mapping with profiling data.

Right now this is a big focus on Github and go but in the future we can expand to other VCS or language.

For go the API allows to fetch code for go modules (vendor or not), go standard library and main application code.

The API requires to provide:
- Git repository for the main application 
- The git ref
- The local path received from the pprof profile API

The PR contains a html page for testing purpose and will be removed prior merging.

```[tasklist]
### Remaining Tasks
- [x] Figure how to correctly delete a client cookie using connect-go
- [x] Refactor the code into a better structure.
- [x] Better tests for now this is very basic test that allowed me to build everything but we need more mocking instead of calling real http endpoint.
- [x] Documentation and design document on how this works.
``` 

image

cc @bryanhuhta @grafakus


pkg/querier/vcs.go
```diff
@@ -0,0 +1,664 @@
...
+	}
+	if gitURL.GetProvider() != apis.ProviderGitHub.String() {
+		return nil, connect.NewError(connect.CodeInvalidArgument, errors.New("only GitHub repositories are supported"))
+	}
+	cookie, err := (&http.Request{Header: req.Header()}).Cookie("GitSession")
+	if err != nil {
+		return nil, err
+	}
+	// todo: we can support multiple provider: bitbucket, gitlab, etc.
+	client, err := NewGithubClient(ctx, cookie)
```
bryanhuhta, line 94: Why do we create a new client every time?

pkg/querier/vcs.go
```diff
@@ -0,0 +1,664 @@
...
+	case ".go":
+		return findGoFile(ctx, arg)
+	// todo: we can support multiple file types: go, java, python, etc.
+	default:
+		// by default we return the file content at the given path without any processing.
+		content, err := arg.client.GetFile(ctx, arg.repo.GetOwnerName(), arg.repo.GetRepoName(), arg.path, arg.ref)
+		if err != nil {
+			return nil, err
+		}
+		return newFileResponse(content.Content, content.URL)
```
bryanhuhta, line 221: Perfect spot for a cache in the future.

pkg/querier/vcs.go
```diff
@@ -0,0 +1,664 @@
...
+	// For this we should use arbitrary k/v metadata in our request so that we don't need to change the API.
+	// I thought about using go.mod go version but it's a min and doesn't guarantee it hasn't been built with a higher version.
+	// Alternatively we could interpret the build system and use the version of the go compiler.
+	ref := "master"
+	isVendor := strings.HasPrefix(packageName, "vendor/")
+
+	if _, isStd := golang.StandardPackages[packageName]; !isVendor && !isStd {
+		return nil, nil
+	}
+	return fetchGoStd(ctx, path, ref)
```
bryanhuhta, line 639: Probably can avoid a network call here in many cases if we download a few of the most popular go std lib versions to disk.

bryanhuhta (commented):
This will be a killer feature once it gets some polish 🥇

pkg/querier/vcs.go
cyriltovena: Because the credentials are stored in an encrypted AES256 cookie and on every call we need to get that token. This is purely to make this simpler for now so that we don't need to store or cache tokens. In fact tokens are only owned by users.

pkg/querier/vcs.go
cyriltovena: Yeah that's faire but then we need to manage standard runtime version. Right now we default to latest because we can't figure which runtime version the binary has been built with but in the future we want that metadata to exists in pprof and use it to map the correct std go version.

pkg/api/vcs.html
```diff
@@ -0,0 +1,132 @@
...
+        .split("&")
+        .reduce((acc, curr) => {
+          const [key, value] = curr.split("=");
+          acc[key] = value;
+          return acc;
+        }, {});
+      if (!code || !state) {
+        return;
+      }
+      // todo: validate the state is the same.
```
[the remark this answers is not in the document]
cyriltovena: Yes this whole file will go away this is to help integrators to understand the whole dance !

knylander-grafana:
Great idea! Will we need some doc to help people configure this?

bryanhuhta:
> Great idea! Will we need some doc to help people configure this?

@knylander-grafana Eventually we will! Right now we need to focus on polish and some integration details. At this point we have a pretty good idea how it will all work, but let's delay getting docs started until we do more polish and can be more confident on how this feature should be configured by the users.

cyriltovena:
Some update for our community this is how it looks right now.

image

cyriltovena:
This is shaping up nicely 

image

ebpf/go.mod
```diff
@@ -8,30 +8,31 @@ require (
 	github.com/cespare/xxhash/v2 v2.2.0
 	github.com/cilium/ebpf v0.11.0
 	github.com/go-kit/log v0.2.1
-	github.com/google/pprof v0.0.0-20230926050212-f7f687d19a98
-	github.com/grafana/pyroscope/api v0.4.0
```
cyriltovena, line 12: hmm

pkg/querier/vcs/client/github.go
```diff
@@ -0,0 +1,157 @@
...
+	// We could abstract away git provider using git protocol
+	// git clone https://x-access-token:<token>@github.com/owner/repo.git
+	// For now we use the github client.
+	file, _, _, err := gh.client.Repositories.GetContents(ctx, req.Owner, req.Repo, req.Path, &github.RepositoryContentGetOptions{Ref: req.Ref})
+	if err != nil {
+		var githubErr *github.ErrorResponse
+		if ok := errors.As(err, &githubErr); ok && githubErr.Response.StatusCode == http.StatusNotFound {
+			return File{}, fmt.Errorf("%w: %s", ErrNotFound, err)
+		}
+		return File{}, err
```
bryanhuhta, line 133: Will this propagate errors correctly? I.e. if github sends back a 422 do we bubble that back to the frontend?

pkg/querier/vcs/source/find.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		return nil, err
+	}
+	resp, err := arg.httpClient.Do(req) // todo: use a custom client with timeout
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return nil, connect.NewError(connect.CodeNotFound, fmt.Errorf("failed to fetch go lib %s: %s", url, resp.Status))
```
bryanhuhta, line 90: Should this error message be more generic? I assume we will use this function to fetch arbitrary files from other servers, not just Go std lib files.

pkg/querier/vcs/source/find.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	content, err := io.ReadAll(resp.Body)
```
bryanhuhta, line 92: Hopefully we don't get sent to a URL with a 25GB file haha

pkg/querier/vcs/source/find_go.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	if relativePath, ok := golang.VendorRelativePath(ff.path); ok {
+		return ff.fetchRepoFile(ctx, relativePath, ff.ref)
+	}
+
+	modFile, ok := golang.ParseModulePath(ff.path)
+	if ok {
+		mainModule := module.Version{
+			Path:    filepath.Join(ff.repo.GetHostName(), ff.repo.GetOwnerName(), ff.repo.GetRepoName()),
```
bryanhuhta, line 37: Is `filepath.Join` intentional here? This will use `\` on Windows. I'm not sure if we care about Windows or not.

If we always want `/`, then `path.Join` is a better option.

bryanhuhta (approved):
Looks good!  A few small remarks but nothing blocking.

pkg/querier/vcs/source/find.go
```diff
@@ -0,0 +1,108 @@
...
+		logger:     logger,
+		repo:       repo,
+		path:       path,
+		ref:        ref,
+		httpClient: httpClient,
+	}
+}
+
+// Find returns the file content and URL.
+func (arg FileFinder) Find(ctx context.Context) (*vcsv1.GetFileResponse, error) {
```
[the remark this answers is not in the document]
cyriltovena: thanks yep !

pkg/querier/vcs/client/github.go
cyriltovena: No for other errors and I'm not sure we should. TBD

pkg/querier/vcs/source/find.go
cyriltovena: Yeah that's a good point, but the API is different once the file is bigger than 4MB. So it will 404.

pkg/querier/vcs/source/find_go.go
cyriltovena: ho yeah good catch !

pkg/querier/vcs/client/github.go
cyriltovena: The content will though contains more info.

pkg/querier/vcs/source/find.go
cyriltovena: Totally correct, this was a leftover after refactoring.

Issue #2727, opened 2023-11-20
Add locations to the flamegraph


kolesnikovae:
We should add information such as file names/path and line numbers into the flamegraph. Ideally, we should also add information about the application version.

---

Grafana Pyroscope allows users to visualize a profile in various forms, with the most common one being a so-called Flamegraph ([Icicle plot](https://www.arxiv-vanity.com/papers/1908.01277/), to be precise). In a nutshell, a Flamegraph is a directed acyclic graph, where each node refers to a specific spot in the profiled code. The graph itself (the hierarchy of the nodes) is formed by [stack traces](https://en.wikipedia.org/wiki/Stack_trace). Other representations, such as [sandwich view](https://pyroscope.io/blog/introducing-sandwich-view/) and [call graph](https://go.dev/blog/pprof) visualize relationships of the nodes.

Pyroscope backend exposes two API endpoints for profiling data retrieval:
 - **SelectMergeStacktraces** that returns a flamegraph in the [flamebearer format](https://github.com/mapbox/flamebearer). The method is used by frontend. Nodes of the flamegraph refer to [call sites](https://en.wikipedia.org/wiki/Call_site).
 - **SelectMergeProfile** that returns a profile in the [pprof format](https://github.com/google/pprof/blob/main/proto/profile.proto) (gzip protobuf encoded binary file). The method is used by profilecli tool for exporting profiling data, and by pprof exporter available in the UI.

Flamegraph returned by backend does not include information about the exact locations of nodes in code, therefore It is not possible to identify the exact line in the function body where the stack trace samples have been collected. For example, it is not possible to get information about where exactly CPU time has been spent functions `deconstructMemoryProfile` and `parquet-go.Value.Level` from this flamegraph:

<img width="909" alt="image" src="https://github.com/grafana/pyroscope/assets/12090599/9dbf48e2-e8ca-47c0-9a50-20dcda036b94">

The only way to get this information is to use **profilecli** tool and visualize pprof file with 3rd party tools.

knylander-grafana:
Is this work to a point where someone can add line numbers to flame graphs in Pyroscope?

kolesnikovae:
Closing in favor of #3220

PR #2976, opened 2024-01-31
fix: disable Go names decoration

kolesnikovae:
Starting from Go 1.21.0, type parameters are part of the function name in pprof, which often makes them unreadable

image

The change reverts https://github.com/grafana/pyroscope/pull/2343 (replaces Go type parameters with ellipses `[...]`) as it makes it impossible to find the selected node in the source profile in backend (needed for #2228, and #2727 and https://github.com/grafana/pyroscope-app-plugin/issues/207).

The presentation aspects are to be managed on the frontend. Ideally, there should be an option for user to disable all the decorations. Note that a function name change usually causes changes in the tree/flamegraph structure, which we used to [fix](https://github.com/grafana/pyroscope/blob/main/pkg/model/tree.go#L161). I believe we don't have to do this.

### DIFF
--- a/pkg/frontend/frontend_select_merge_span_profile.go
+++ b/pkg/frontend/frontend_select_merge_span_profile.go
@@ -83,7 +83,6 @@ func (f *Frontend) SelectMergeSpanProfile(ctx context.Context,
 	}
 
 	t := m.Tree()
-	t.FormatNodeNames(phlaremodel.DropGoTypeParameters)
 	return connect.NewResponse(&querierv1.SelectMergeSpanProfileResponse{
 		Flamegraph: phlaremodel.NewFlameGraph(t, c.Msg.GetMaxNodes()),
 	}), nil
--- a/pkg/frontend/frontend_select_merge_stacktraces.go
+++ b/pkg/frontend/frontend_select_merge_stacktraces.go
@@ -82,7 +82,6 @@ func (f *Frontend) SelectMergeStacktraces(ctx context.Context,
 	}
 
 	t := m.Tree()
-	t.FormatNodeNames(phlaremodel.DropGoTypeParameters)
 	return connect.NewResponse(&querierv1.SelectMergeStacktracesResponse{
 		Flamegraph: phlaremodel.NewFlameGraph(t, c.Msg.GetMaxNodes()),
 	}), nil


bryanhuhta (approved):

cyriltovena:
cc @aocenas We're going to need some sort of function sanitizer on the frontend but still able to access to original function for programatic need.

aocenas:
@cyriltovena I assume we could just port this function https://github.com/grafana/pyroscope/pull/2343/files#diff-e7b8e2da222a25feb1fdc03174dd0d9783102c532048d0bd2945ab9ef6c87c8bR10 to the frontend right? Is it safe to apply to any code (non Go code)? It seems like @kolesnikovae suggests as much in the comment but wonder if there is some caveat or not.

cyriltovena:
I would not do it for any other language just in case I suggest we found this `[go.shape......]` and replace it with `[...]` that should be good enough and only appear in go.

Again the original must still be accessible on click event.

cyriltovena:
```connectrpc.com/connect.NewUnaryHandler[go.shape.struct { github.com/grafana/pyroscope/api/gen/proto/go/types/v1.state google.golang.org/protobuf/internal/impl.MessageState; github.com/grafana/pyroscope/api/gen/proto/go/types/v1.sizeCache int32; github.com/grafana/pyroscope/api/gen/proto/go/types/v1.unknownFields []uint8; Matchers []string "protobuf:\"bytes,1,rep,name=matchers,proto3\" json:\"matchers,omitempty\""; Start int64 "protobuf:\"varint,2,opt,name=start,proto3\" json:\"start,omitempty\""; End int64 "protobuf:\"varint,3,opt,name=end,proto3\" json:\"end,omitempty\"" },go.shape.struct { github.com/grafana/pyroscope/api/gen/proto/go/types/v1.state google.golang.org/protobuf/internal/impl.MessageState; github.com/grafana/pyroscope/api/gen/proto/go/types/v1.sizeCache int32; github.com/grafana/pyroscope/api/gen/proto/go/types/v1.unknownFields []uint8; Names []string "protobuf:\"bytes,1,rep,name=names,proto3\" json:\"names,omitempty\"" }].func2```

aocenas:
> I would not do it for any other language just in case

Yeah but we still don't have any information about the language from the backend do we?

PR #2977, opened 2024-01-31
ebpf: per target configuration with labels

korniltsev:
Backport some changes from python memory profiling branch:
    - per PID pprof creation instead of per target
    - cmd/playground: split log into stdout and stderr\
    - fix bug when python errored stacktrace is attempted to be used https://github.com/grafana/pyroscope/compare/ebpf-per-target-configuration-with-labels?expand=1#diff-a25b2cb097347420b8c3241b6c0c669ff52be552b1043971d4aec56c82b79653R63 

Allow per target configuration with labels

```
	labelMetaPyroscopeOptionsPrefix = "__meta_pyroscope_options_"

	OptionGoTableFallback    = labelMetaPyroscopeOptionsPrefix + "go_table_fallback"
	OptionPythonFullFilePath = labelMetaPyroscopeOptionsPrefix + "python_full_file_path"
	OptionPythonEnabled      = labelMetaPyroscopeOptionsPrefix + "python_enabled"
	OptionDemangle           = labelMetaPyroscopeOptionsPrefix + "demangle"
```


ebpf/pprof/pprof.go
```diff
@@ -26,21 +27,96 @@ var (
...
+	var periodType *profile.ValueType
+	var period int64
+	if sample.SampleType == SampleTypeCpu {
+		sampleType = []*profile.ValueType{{Type: "cpu", Unit: "nanoseconds"}}
+		periodType = &profile.ValueType{Type: "cpu", Unit: "nanoseconds"}
+		period = time.Second.Nanoseconds() / b.opt.SampleRate
+	} else {
+		sampleType = []*profile.ValueType{{Type: "alloc_objects", Unit: "count"}, {Type: "alloc_space", Unit: "bytes"}}
+		periodType = &profile.ValueType{Type: "space", Unit: "bytes"}
+		period = 512 * 1024 // todo
```
cyriltovena, line 118: ?

cyriltovena: https://github.com/google/pprof/blob/main/proto/profile.proto#L85-L86 Not sure what it should be for memory 🤷 we don't really use that I think anyway.

ebpf/session.go
```diff
@@ -730,7 +738,7 @@ func (s *session) cleanup() {
 	s.symCache.Cleanup()
 
 	for pid := range s.pids.dead {
-		_ = level.Debug(s.logger).Log("msg", "cleanup dead pid", "pid", pid)
+		//_ = level.Debug(s.logger).Log("msg", "cleanup dead pid", "pid", pid)
```
cyriltovena, line 741: may be we should remove it if we don't need it or uncomment if that is useful debug info.

ebpf/symtab/symbols.go
```diff
@@ -49,8 +46,8 @@ func NewSymbolCache(logger log.Logger, options CacheOptions, metrics *metrics.Sy
 		pidCache: cache,
 		kallsyms: nil,
 		elfCache: elfCache,
-		options:  options,
-		metrics:  metrics,
+		//options:  options,
```
cyriltovena, line 49: Sounds like something that should be there no ?

cyriltovena (approved):
LGTM

make sure the cache options is ok to remove, I have some doubt.

PR #2990, opened 2024-02-06
examples: update rideshare examples

korniltsev:
- add `RIDESHARE_LISTEN_PORT` env variable
- update loadgen to accept urls instead of hostnames
- split makefile target to build examples concurrently with `-j`

this is to allow to run multiple rideshare examples in the same network namespace ( pod)


kolesnikovae (approved):
LGTM

PR #2992, opened 2024-02-08
Add workaround for broken span propagation in streaming requests

### LINKED ISSUE #2923: Broken span context propagation for store gateway requests
#### Describe the bug
Some spans emitted by Pyroscope components don't seem to have a parent and are therefore shown at the root level. I've noticed this with the store gateway http client requests but it could be present elsewhere as well:

Screenshot 2024-01-12 at 13 49 59

#### To Reproduce
1. Make a query involving the store gateway(s)
2. Inspect the trace for the query

aleks-p:
Fixes #2923 

**Background:**
The current instrumentation for the HTTP client is based on https://github.com/opentracing-contrib/go-stdlib. This works well enough for standard HTTP requests but does not support streaming. This instrumentation library "finishes" the opened trace spans only when the response body gets closed, which doesn't happen for streaming requests where the initial response has an empty body and is never read / closed. The result is below. Note the disconnected spans at the bottom and the empty "HTTP Client" spans:

span-propagation-before

**Proper fix:**
Switching to an instrumentation library that has proper support for streaming is the ideal way to address this. We will eventually switch to Open Telemetry which should offer a good path forward. This is a bigger change and in the interest of improving trace analysis I am proposing a workaround.

**Workaround:**
When we detect an empty response body we finish the span directly. This means that the span will report an incorrect time, but this is still an improvement over the current state where this span was completely removed which broke the span tree. In addition, I've merged the "HTTP Client" and "HTTP <METHOD>" spans into one for simplicity. All this is done in a new component inspired by https://github.com/opentracing-contrib/go-stdlib. 

**The result:**
Screenshot 2024-02-07 at 15 55 40


kolesnikovae (approved):
LGTM

Thank you for fixing this! I like the workaround – I don't think there is a better way to to handle client spans. And I agree with your conclusions: we should migrate to OTel SDK

cyriltovena:
Thank you so much for doing this !

PR #2996, opened 2024-02-09
ebpf: aggregate python stacks in ebpf program instead of userspace

korniltsev:
We used to stream each sample once it is captured, now we increment counter in ebpf world, and dump it once every profiling interval.

This seem to consume less cpu

```
Before:
         PM   UID       PID    %usr %system  %guest   %wait    %CPU   CPU  Command
Average:        0    382723    4.57    0.88    0.00    0.00    5.45     -  ___ebpf_python_

After
         PM   UID       PID    %usr %system  %guest   %wait    %CPU   CPU  Command
Average:        0    803223    2.43    0.18    0.00    0.00    2.62     -  ___ebpf_python_
```


ebpf/cmd/playground/main.go
```diff
@@ -35,6 +35,14 @@ import (
 
 var configFile = flag.String("config", "", "config file path")
 var server = flag.String("server", "http://localhost:4040", "")
+var discoverFreq = flag.Duration("discover.freq",
+	//5 * time.Second,
+	5*time.Hour, //todo remove
```
simonswine, line 40: TODO: Remove

korniltsev: removed, thanks. I was benchmarking and wanted to exclude discovery

simonswine (approved):
LGTM

I noticed a small diff comparing the old python profiles with the new ones. Can't really tell if this is something significant wrong. Maybe just have a quick look:

https://flamegraph.com/share/2cd304bb-cf01-11ee-a059-7a6c3496f108

Green is the profile collected using #2996 and red is before

korniltsev:
> I noticed a small diff comparing the old python profiles with the new ones. Can't really tell if this is something significant wrong. Maybe just have a quick look:
> 
> https://flamegraph.com/share/2cd304bb-cf01-11ee-a059-7a6c3496f108
> 
> Green is the profile collected using https://github.com/grafana/pyroscope/pull/2996 and red is before


I've also added kernel traces in this PR and it looks like the only difference is that new version has kernel stacks in places of 
`fibonacci.py _get_from_memoization` etc.

Does it look reasonable?

simonswine:
> Does it look reasonable?

Yes makes total sense, missed the kernel stacks change tbf 🙂

PR #2997, opened 2024-02-09
Fix downsampler debug log message

aleks-p:
Addresses "unsupported value type" in 

`ts=2024-02-09T13:13:02.072918921Z caller=downsample.go:162 level=debug component=compactor component=compactor msg="flushing downsampled profile" interval="unsupported value type" aggregation="unsupported value type" sourceProfileCount=20 sampleCount=250`


cyriltovena (approved):
LGTM

PR #3010, opened 2024-02-19
fix: drop type params from Go function names

kolesnikovae:
This change involves the removal of type parameters from Go function names during the ingestion process.

The primary rationale behind this modification is that, in Go, function type parameters within profiles are often encountered in unexpected code paths. Allegedly, the Go profiler does not consider the call site when constructing the function name and may include parameters from irrelevant function instantiations. This behavior significantly increases the cardinality of stack traces. For more details, please refer to https://github.com/grafana/pyroscope/issues/2241, specifically [this](https://github.com/grafana/pyroscope/issues/2241#issuecomment-1689804566) comment.

The elimination of type parameters results in a reduction of cardinality by up to 50% (ingester's heap, 1 hour):

```
	fix_test.go:24:  * Sample:     6785 -> 3797
	fix_test.go:25:  * Location:   4848 -> 4680
	fix_test.go:26:  * Function:   2801 -> 2724
	fix_test.go:27:  * Strings:    3536 -> 3458
```

Regarding the impact on PGO: given that we cannot rely on the parameters, no negative impact is expected.

We probably should be more specific about the Go version, but we do not have this information yet. Theoretically, we only have to do this for go < 1.22 / 1.21.6.

Related:
 - https://github.com/grafana/pyroscope/issues/2241
 - https://github.com/grafana/pyroscope/pull/2343
 - https://github.com/grafana/pyroscope/pull/2976 (revert of 2343)

Go:
- https://github.com/golang/go/issues/64528 (fix)
- https://github.com/golang/go/issues/64641
- https://github.com/golang/go/issues/65147


pkg/pprof/fix.go
```diff
@@ -0,0 +1,39 @@
...
+// This leads to incorrect profiles and incorrect flame graphs,
+// and hugely increases cardinality of stack traces.
+//
+// FixGoProfile will change x[T1].foo and x[T2].foo to x[...].foo
+// and normalize the profile, if type parameters are present in
+// the profile. Otherwise, the profile returned unchanged.
+func FixGoProfile(p *profilev1.Profile) *profilev1.Profile {
+	var n int
+	for i, s := range p.StringTable {
+		c := DropGoTypeParameters(p.StringTable[i])
```
kolesnikovae, line 22:
```suggestion
		c := DropGoTypeParameters(s)
```

cyriltovena (approved):
LGTM

Can't wait to try this out in our biggest cluster. Do you know if there's a report of that bug somewhere in the go community ?

simonswine:
@cyriltovena it is linked in the top comment under `Go:`

cyriltovena:
I'm blind today ! sorry.

PR #3025, opened 2024-02-26
ebpf: run tests on many kernels

korniltsev:
Run ebpf tests on multiple kernels 5.4 5.10 5.15 6.1 5.10-amazon
4.19 does not work  https://github.com/grafana/pyroscope/issues/3033

In the future: consider periodically building a kernel from the Linus master branch and from bpf-next branch


simonswine (approved):
LGTM ❤️ 

This will allow to catch future ebpf problems early.

I do think we shouldn't hide that many qemu errors, something might be a bit more transparent: https://github.com/pyroscope-io/pyroscope-ebpf-testdata/pull/2/files

korniltsev:
Great!
In the future It would be nice to have kernel output if CI fails, just in case

korniltsev:
Interesting, qemu tests have not finished but it allowed me to merge :D

PR #3026, opened 2024-02-26
docs: Reorganize examples folder to match docs structure

Rperry2174:
This reorganizes our examples to match the "configure the client" part of our docs. I'm not sure if this will break things in places where we use these examples.

i.e. 
- I believe in the `/tracing/tempo` folder we use relative references and this might mess that up cc @kolesnikovae
- I also don't know how sedemo and play.grafana.net use these examples and if this will break those @korniltsev 

image

We will also probably need to go through docs and make sure that we didn't break any links there where we link to the examples. cc @knylander-grafana :
image


knylander-grafana (commented):
New structure looks good. Link updates also look good. I can only approve the doc changes.

kolesnikovae:
@Rperry2174 thanks for pinging me – I've pushed changed straight to the branch

Issue #2845, opened 2023-12-13
Panic in store-gateways


cyriltovena:
<img width="2143" alt="image" src="https://github.com/grafana/pyroscope/assets/1053421/649067a0-4b71-41e0-b641-a41238efd0f5">

Found this one while on-call. I suspect the index is not open. May be it was closed just before querying ? 

see [logs](https://admin-ops-us-east-0.grafana-ops.net/grafana/explore?schemaVersion=1&panes=%7B%22bD3%22:%7B%22datasource%22:%22loki-ops%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bnamespace%3D%5C%22profiles-ops-001%5C%22%7D%20%7C%3D%20%5C%22panic:%5C%22%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki-ops%22%7D,%22editorMode%22:%22code%22%7D%5D,%22range%22:%7B%22from%22:%221702464637769%22,%22to%22:%221702464915315%22%7D%7D%7D&orgId=1)

cc @bryanhuhta

PR #3030, opened 2024-02-27
fix: wait for ongoing queries to finish at close

kolesnikovae:
This change forces block querier to wait until all outgoing queries to finish at `Close` call. Resolves https://github.com/grafana/pyroscope/issues/2845

The solution does not completely address the issue of block ownership. Since our queries involve stateful processes (block iteration + series iteration + deduplication + merge), and our replicas (ingesters and store gateways) lack awareness of the query context, there is a possibility that a block may be closed while the query is still running. This occurs because replicas may not retain any references to the block in-between query stages.

For instance, if a block is evicted immediately after the list of replica blocks has been returned to the querier, attempting to retrieve profiling data from the block will result in an empty response. This is to be solved in a separate PR. I plan to introduce a graceful period for block eviction. Instead of immediate eviction, the plan is to mark the block and initiate eviction only after a reasonable grace period, such as 5 minutes, has expired.

### DIFF
--- a/pkg/phlaredb/block_querier.go
+++ b/pkg/phlaredb/block_querier.go
@@ -299,6 +299,7 @@ type singleBlockQuerier struct {
 
 	tables []tableReader
 
+	queries  sync.WaitGroup
 	openLock sync.Mutex
 	opened   bool
 	index    *index.Reader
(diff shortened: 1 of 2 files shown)


pkg/phlaredb/block_querier.go
```diff
@@ -1627,6 +1639,9 @@ func (b *singleBlockQuerier) SelectMergeByLabels(
 	if err := b.Open(ctx); err != nil {
 		return nil, err
 	}
+	b.queries.Add(1)
```
cyriltovena, line 1642: I feel like we should update Open to be a function with a call back like `Query(fn ...) error` and all of this is handled there.

cyriltovena (approved):
LGTM

PR #3031, opened 2024-02-27
Fixes merge of empty samples in pprof

cyriltovena:
Introduced in https://github.com/grafana/pyroscope/pull/3010/files#diff-1716fbe09b95462c58b96e3abdc3b60585159cad6f026e731598ca62859ea5e5R49

When we insert an empty profile we still would consider the next one as new and rewrite string from the original twice.

This has caused the pprof API to fail recently.

cc @bryanhuhta @grafakus 

Sneaky one !


kolesnikovae (approved):
LGTM

Nice catch! Thank for fixing this

PR #3035, opened 2024-02-27
ebpf(python): update offsets for new pythons

korniltsev:


kolesnikovae (approved):

ebpf/python/python_offsets_gen_amd64.go
```diff
@@ -3219,8 +3374,8 @@ var pyVersions = map[Version]*UserOffsets{
 		PyInterpreterFrame_previous:       8,
 		PyInterpreterFrame_localsplus:     72,
 		PyInterpreterFrame_owner:          70,
-		PyRuntimeState_gilstate:           3112,
-		PyRuntimeState_autoTSSkey:         1848,
```
simonswine, line 3223: I am assuming this is because it is not a "released" version?

simonswine (approved):
LGTM

ebpf/python/python_offsets_gen_amd64.go
korniltsev: I am assuming the same :)

korniltsev: Actually I am surprised my integration tests did not catch it, It should fail now, because test image is based on old release candidate. I will look into it.

korniltsev: `PyRuntimeState_gilstate` is not used on >= 3.13
And `PyRuntimeState_autoTSSkey` is used to read `pthread_key` which is an int and a counter and almost always it is zero. I am not 100% sure, but I think the test did not fail because the offset is wrong but it still read some garbage uint32=0.

PR #3058, opened 2024-02-29
fix: Panics when passing a malformed GitSession

bryanhuhta:
If a carefully chosen `GitSession` is passed to the backend, it causes an out of range panic and crash. This isn't ideal as attackers will have a way to continually crash queriers.

If `GitSession` is a valid base64 encoding which, when decoded, results in a cipher text less than the chose GCM Cipher nonce size, the decryption subroutine will try split the cipher text into two parts, a "nonce" part and a "rest" part. if the cipher text is shorter than the "nonce" part, it panics.

See: https://github.com/oauth2-proxy/oauth2-proxy/blob/e3dc927e570700ae598425344cf07073965bd19c/pkg/encryption/cipher.go#L124-L125

As far as I can tell, the nonce size is always 12, though I don't know how reliably we can depend on this.


kolesnikovae (approved):
IV size (nonce) size is determined [in the oauth2-proxy](https://github.com/oauth2-proxy/oauth2-proxy/blob/e3dc927e570700ae598425344cf07073965bd19c/pkg/encryption/cipher.go#L103) (uses the standard size 12) and ideally should be verified at decryption inside the library (e.g., [here](https://github.com/oauth2-proxy/oauth2-proxy/blob/e3dc927e570700ae598425344cf07073965bd19c/pkg/encryption/cipher.go#L125)),  – it's very unlikely that it will change, so I think it's ok to do it on our side. OTOH, I'm not sure if we benefit from use of oauth2-proxy dependency

pkg/querier/vcs/client/encryption.go
```diff
@@ -24,11 +25,19 @@ func encryptToken(token *oauth2.Token, key []byte) (string, error) {
...
+// TODO(bryan) this is kind of a hack and I found it through debugging.
+const gcmNonceSize = 12
+
 func decryptToken(encodedText string, key []byte) (*oauth2.Token, error) {
 	encryptedData, err := base64.StdEncoding.DecodeString(encodedText)
 	if err != nil {
 		return nil, err
 	}
+
+	if len(encryptedData) < gcmNonceSize {
```
cyriltovena, line 37: I suspect we need to recover panic in this function or above too just in case ?

cyriltovena (approved):
LGTM

pkg/querier/vcs/client/encryption.go
```diff
@@ -24,11 +25,19 @@ func encryptToken(token *oauth2.Token, key []byte) (string, error) {
...
+		return nil, fmt.Errorf("malformed token")
```
simonswine, line 38:
```suggestion
		return nil, errors.New("malformed token")
```

simonswine (approved):
LGTM

pkg/querier/vcs/client/encryption.go
bryanhuhta: Yeah, we should. Shouldn't all recoverable panics get caught by the grpc handler? Why doesn't this happen today?

PR #3073, opened 2024-03-01
[DOC] Set up profiling and tracing intro for sharing

knylander-grafana:
Changes the profiling and traces into content to be shared from grafana/grafana/docs/sources/shared/datasources and added a "Enriching profiling data" section.

Related: https://github.com/grafana/grafana/pull/83739


Rperry2174 (approved):
lgtm

docs/sources/configure-client/_index.md
```diff
@@ -109,4 +109,12 @@ To get started, choose one of the integrations below:
    </tr>
 </table>
 
+## Enriching profile data
+
+You can add tags to your profiles to help correlate them with your other telemetry signals. Some common tags that are used are version, region, environment, request types, etc. You have the ability to add tags using both the SDK and the agent.
+
+Valid tag formats may contain ASCII letters and digits, as well as underscores. It must match the regex `[a-zA-Z_][a-zA-Z0-9_]`.
```
Rperry2174, line 116: can we explicitly say that `.` is not a valid character inside tags/labels in pyroscope?

knylander-grafana: Of course. How is this? 
```suggestion
Valid tag formats may contain ASCII letters and digits, as well as underscores. It must match the regex `[a-zA-Z_][a-zA-Z0-9_]`. A period (`.`) is not a valid character inside of tags and labels in Pyrorscope. 
```

PR #3090, opened 2024-03-11
Add QuerierService.GetProfileStats

### LINKED ISSUE #3003: API to check whether there is any data for the current tenant
To provide a better onboarding experience in Pyroscope frontends, we would benefit from an API that can tell us whether we have any data for a tenant. An example usage would be showing instructions on how to set up profiling rather than an empty UI. 

API Contract Proposal:
```
service QuerierService {
  ...
  rpc GetProfileStats(GetProfileStatsRequest) returns (GetProfileStatsResponse) {}
}

message GetProfileStatsRequest {}

message GetProfileStatsResponse {
  // Whether we received any data.
  bool has_never_ingested_data = 1;
  // Milliseconds since epoch.
  int64 oldest_profile_time = 2;
  // Milliseconds since epoch.
  int64 newest_profile_time = 3;
}
```

aleks-p:
Adds a new API for returning simple profile statistics. Right now limited to a flag indicating if profiling data exists and the timestamp of the oldest and newest profile if available. More stats could be added in the future (e.g., number of unique series, number of profiles, samples, etc.).

Closes #3003 

cc @grafakus


pkg/phlaredb/phlaredb.go
```diff
@@ -524,3 +525,46 @@ func (f *PhlareDB) BlockMetadata(ctx context.Context, req *connect.Request[inges
...
+	metas := make([]*block.Meta, 0)
+
+	f.headLock.RLock()
+	for _, h := range f.heads {
+		metas = append(metas, h.meta)
+	}
+	for _, h := range f.flushing {
+		metas = append(metas, h.meta)
+	}
+	f.headLock.RUnlock()
```
kolesnikovae, line 542: I'm not 100% sure but access to the `h.meta.MaxTime` and `MinTime` should be only done with the mutex acquired, so there will probably be a race when you latter access meta in `getProfileStatsFromMetas`. It's better to call `h.Bounds()` here

pkg/phlare/modules.go
```diff
@@ -245,20 +246,18 @@ func (f *Phlare) setupWorkerTimeout() {
...
+		StoreGatewayCfg: f.Cfg.StoreGateway,
+		Overrides:       f.Overrides,
+		CfgProvider:     f.Overrides,
+		StorageBucket:   f.storageBucket,
+		IngestersRing:   f.ring,
+		Reg:             f.reg,
+		Logger:          log.With(f.logger, "component", "querier"),
+		ClientOptions:   []connect.ClientOption{f.auth},
+	}
+	querierSvc, err := querier.New(newQuerierParams)
```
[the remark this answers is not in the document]
aleks-p: There is no perfect solution for this. You mentioned most of the pros and cons of both approaches. 

Personally I find the code snippet you highlighted more readable compared to

`
querierSvc, err := querier.New(f.Cfg.Querier, f.ring, nil, storeGatewayQuerier, f.reg, log.With(f.logger, "component", "querier"), f.auth)
`

or:

```
querierSvc, err := querier.New(
    f.Cfg.Querier,
    f.ring,
    nil,
    storeGatewayQuerier,
    f.reg,
    log.With(f.logger, "component", "querier"),
    f.auth
)
```

I find the struct approach more readable because I can see what is being passed on without switching contexts and without paying attention to the argument order. The readability gains are higher when the number of arguments is higher.

As for safety, I expect constructor functions like this to perform validation on the input. E.g., we are already sending a `nil` in my last example which is maybe fine in this context but not in another and the function itself has to enforce this.

I am also open to any approach and would be curious to hear what others think.

@simonswine @bryanhuhta @korniltsev @petethepig any thoughts?

pkg/phlaredb/phlaredb.go
aleks-p: Good point, I copied this from `BlockMetadata` but didn't realize that one is using `Bounds()`.

kolesnikovae (approved):

PR #3100, opened 2024-03-14
fix: query plan optimisation

### LINKED ISSUE #3088: Ingester and store-gateway time ranges overlap
I noticed that we have an overlap between the data retrieved from ingesters and store-gateways: in the screenshot below you can see a 2x "surge" that starts at `now().Truncate(queryStoreGatewayAfter)` and lasts till `now - queryStoreGatewayAfter`

image

kolesnikovae:
Fixes #3088.

The problem manifests when compactor does not keep up with the ingestion rate, and store-gateways load sharded L2 blocks (with duplicates).

Normally, L3 blocks supersede all the parent blocks in the query plan, and presence of L2 blocks does not pose any issues. However, when an L3 shard is incomplete, all its blocks are removed from the plan before this step, preserving L2 shards in the query plan (partial shards are ignored in the plan). This happens, when L3 compaction delay exceeds `blocks-storage.bucket-store.ignore-blocks-within`. The way query plan is built expects that L2 blocks do not contain duplicate series, which is not true if blocks are shared. As a result, query returns duplicated data.

### DIFF
--- a/pkg/ingester/retention.go
+++ b/pkg/ingester/retention.go
@@ -290,6 +290,9 @@ func (dc *diskCleaner) CleanupBlocksWhenHighDiskUtilization(ctx context.Context)
 
 // isBlockDeletable returns true if this block can be deleted.
 func (dc *diskCleaner) isBlockDeletable(block *tenantBlock) bool {
+	// TODO(kolesnikovae):
+	//  Expiry defaults to -querier.query-store-after which should be deprecated,
+	//  blocks-storage.bucket-store.ignore-blocks-within can be used instead.
 	expiryTs := time.Now().Add(-dc.policy.Expiry)
 	return block.Uploaded && ulid.Time(block.ID.Time()).Before(expiryTs)
 }
--- a/pkg/querier/replication.go
+++ b/pkg/querier/replication.go
@@ -200,10 +200,10 @@ func (r *replicasPerBlockID) removeBlock(ulid string) {
 }
 
 // this step removes sharded blocks that don't have all the shards present for a time window
-func (r *replicasPerBlockID) pruneIncompleteShardedBlocks() error {
+func (r *replicasPerBlockID) pruneIncompleteShardedBlocks() (bool, error) {
 	type compactionKey struct {
-		level int32
-		minT  int64
+		level   int32
+		minTime int64
 	}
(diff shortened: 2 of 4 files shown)


pkg/ingester/retention.go
```diff
@@ -290,6 +290,9 @@ func (dc *diskCleaner) CleanupBlocksWhenHighDiskUtilization(ctx context.Context)
 
 // isBlockDeletable returns true if this block can be deleted.
 func (dc *diskCleaner) isBlockDeletable(block *tenantBlock) bool {
+	// TODO(kolesnikovae):
+	//  Expiry defaults to -querier.query-store-after which should be deprecated,
+	//  blocks-storage.bucket-store.ignore-blocks-within can be used instead.
 	expiryTs := time.Now().Add(-dc.policy.Expiry)
 	return block.Uploaded && ulid.Time(block.ID.Time()).Before(expiryTs)
```
kolesnikovae, line 297: However, we also need to take into account that a new block does not become available to the store gateway immediately. I'd say that it probably makes sense to delay till expiration of `blocks-storage.bucket-store.ignore-blocks-within` + 2*(block sync interval)

aleks-p (approved):
Thanks for fixing this, looks good to me!

pkg/querier/replication.go
```diff
@@ -331,11 +338,21 @@ func (r *replicasPerBlockID) blockPlan(ctx context.Context) map[string]*ingestv1
...
+	sharded, err := r.pruneIncompleteShardedBlocks()
+	if err != nil {
 		level.Warn(r.logger).Log("msg", "block planning failed to prune incomplete sharded blocks", "err", err)
 		return nil
 	}
-	if err := r.pruneSupersededBlocks(); err != nil {
+
+	// Depending on whether split sharding is used, the compaction level at
+	// which we the data gets deduplicated differs: if split sharding is
+	// enabled, we deduplicate at level 3, and at level 2 otherwise.
```
kolesnikovae, line 349:
```suggestion
	// Depending on whether split sharding is used, the compaction level at
	// which the data gets deduplicated differs: if split sharding is enabled,
	// we deduplicate at level 3, and at level 2 otherwise.
```

PR #3103, opened 2024-03-15
fix: head data duplication

### LINKED ISSUE #3092: Data duplication in head queries
In the screenshot you can see a "splash" starting at 11:45 – start of the most recent head block. The increase is artificial and remains in the same place over time.

image

I can only reproduce the issue with series extracted from pprof labels: apparently, we have label/series order mismatch, that is fixed when a head block if flushed.

kolesnikovae:
Resolves #3092


aleks-p (approved):
LGTM

PR #3105, opened 2024-03-15
perf: optimize tree node insertion

kolesnikovae:
This is a pretty obvious optimization that we should have implemented much earlier. This function may contribute up to 25% of the total query CPU time in rare cases. The larger the tree, the more significant the benefit.

image

### DIFF
--- a/pkg/model/tree.go
+++ b/pkg/model/tree.go
@@ -63,9 +63,28 @@ func (t *Tree) InsertStack(v int64, stack ...string) {
 	}
 	r := &node{children: t.root}
 	n := r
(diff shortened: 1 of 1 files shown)


aleks-p (approved):
Good find! Out of curiosity, what prevents inlining to happen here at compile time? Would using PGO have made a difference?

kolesnikovae:
I haven't looked too deep into the inlining budget, but I AFAIK that Go never inlines functions that use **range**. Plus the `sort.Search` function argument – I'm not sure if this is possible to inline it at all

PR #3106, opened 2024-03-16
chore: remove repetitive words

sunxunle:


kolesnikovae (approved):
Thank you for the fixes @sunxunle!

@knylander-grafana do we need to do anything special for v1-1 release notes to update (if we want to, of course)?

knylander-grafana (approved):
Thank you for removing the duplicate words.

@kolesnikovae I don't think so. Version 1.1 didn't have release notes. We only just recently added release notes to the docs.

PR #3115, opened 2024-03-19
chore: use pyroscope:latest in tracing integration demo

kolesnikovae:
The tracing integration feature has been released a long time ago, but in the demo we're still building the pyroscope image locally, which might be challenging for users


aleks-p (approved):
LGTM

PR #3130, opened 2024-03-22
Add integration testing of microservices/replication for label queries

simonswine:
I am currently trying to understand some behaviour while running with an unhealthy ingester/store-gateway each.

For this purpose I've  create this testing helper that can spin up multiple microservices locally.

### DIFF
--- a/pkg/phlare/phlare.go
+++ b/pkg/phlare/phlare.go
@@ -220,17 +220,18 @@ type Phlare struct {
 	serviceMap    map[string]services.Service
 	deps          map[string][]string
 
-	API           *api.API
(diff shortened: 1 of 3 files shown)


simonswine:
I think the current state contains something that is already mergable, I would like to see how stable the test is and then potentially extend the coverage.

kolesnikovae (approved):
Thank you for adding this! This will definitely make development and testing easier, and will improve test coverage in general 👍🏻 I'll add cases for `select stacktraces` and `select pprof` queries soon. Also, I'll see if `testify/suite` can be useful here

PR #3135, opened 2024-03-26
Stop Disk cleaner when storage bucket is not defined

### LINKED ISSUE #2974: Don't try delete old ingester blocks without bucket store configured
When running Pyroscope without bucket storage configured (this typically occurs when running locally in monolith mode), the ingester block cleaner tries to clean blocks off disk. However, since a shipper isn't running, no `shipper.json` file is created and the cleaner logs the following error:

```
ts=2024-01-29T12:54:11.499628801Z caller=retention.go:147 level=error msg="failed to delete uploaded blocks, could not get blocks for tenant" err="open data/anonymous/local/shipper.json: no such file or directory" tenantID=anonymous
```

We should account for `shipper.json` not existing by turning off the block cleaner. We should also log a warn level message (or no message at all) when this occurs instead of an error, since it's not exactly an error scenario.

weastel:
Fixes #2974 

As per issue, we have to activate disk cleaner when only storage is defined, otherwise we are getting error of `shipper.json`, does not exist, as shipper only runs when storage bucket is defined, as per code `pkg/ingester/instance.go:53`

### DIFF
--- a/pkg/ingester/retention.go
+++ b/pkg/ingester/retention.go
@@ -153,7 +153,7 @@ func (dc *diskCleaner) DeleteUploadedBlocks(ctx context.Context) int {
 		}
 
 		for _, block := range blocks {
-			if !dc.isBlockDeletable(block, true) {
+			if !block.Uploaded || !dc.isExpired(block) {
 				continue
 			}
 
@@ -233,7 +233,7 @@ func (dc *diskCleaner) CleanupBlocksWhenHighDiskUtilization(ctx context.Context)
 	prevVolumeStats := &diskutil.VolumeStats{}
 	filesDeleted := 0
 	for _, block := range blocks {
-		if !dc.isBlockDeletable(block, false) {
+		if !dc.isExpired(block) {
 			continue
 		}
 
@@ -289,19 +289,12 @@ func (dc *diskCleaner) CleanupBlocksWhenHighDiskUtilization(ctx context.Context)
 }
 
 // isBlockDeletable returns true if this block can be deleted.
-func (dc *diskCleaner) isBlockDeletable(block *tenantBlock, deleteOnlyUploaded bool) bool {
+func (dc *diskCleaner) isExpired(block *tenantBlock) bool {
 	// TODO(kolesnikovae):
 	//  Expiry defaults to -querier.query-store-after which should be deprecated,
 	//  blocks-storage.bucket-store.ignore-blocks-within can be used instead.
 	expiryTs := time.Now().Add(-dc.policy.Expiry)
-
-	isExpired := ulid.Time(block.ID.Time()).Before(expiryTs)
-
-	if deleteOnlyUploaded {
-		return block.Uploaded && isExpired
-	}
-
-	return isExpired
+	return ulid.Time(block.ID.Time()).Before(expiryTs)
 }
 
 // blocksByUploadAndAge implements sorting tenantBlock by uploaded then by age
--- a/pkg/ingester/retention_test.go
+++ b/pkg/ingester/retention_test.go
@@ -323,7 +323,7 @@ func TestDiskCleaner_EnforceHighDiskUtilization(t *testing.T) {
 	})
 }
 
-func TestDiskCleaner_isBlockDeletable(t *testing.T) {
+func TestDiskCleaner_isBlockDeletableForUploadedBlocks(t *testing.T) {
 	tests := []struct {
 		Name   string
 		Expiry time.Duration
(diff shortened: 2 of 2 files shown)


weastel:
@kolesnikovae, As discussed, new changes will delete blocks when high disk utilization

kolesnikovae (commented):
The error the original issue refers to is caused by the fact that we do not handle the case when there is no shipper metadata file: https://github.com/grafana/pyroscope/blob/26b27357682530445808d5b901fff29f572b67ab/pkg/ingester/retention.go#L395

To me it looks like we should silently ignore missing shipper metadata file: there's no guaranty it is present at the time we access it (e.g., disk cleaner may fire before the first run of shipper). That should solve the problem

kolesnikovae:
Thank you for investigating this @weastel! I totally agree
> As per me, It should be deleted whether shipper.json exist or not, GetBlocksForTenant should be fixed for this scenario, default behaviour should be defined when shipper.json does not exist.

I was too optimistic, the change will likely be more invasive. First, just as you said, we need to be able to list tenants and blocks without the shipper metadata file. We used to do it in this way before:
https://github.com/grafana/pyroscope/blob/d32dbce2fb03031583cda0445e9be796d4cad885/pkg/ingester/retention.go#L128
We can revive this approach: we read directories / blocks on our own, and only set `uploaded` attribute checking the shipper index file. This is an optional flag. If no shipper index exists, we ignore it.

Also, on high utilization, the flag should be ignored: 
https://github.com/grafana/pyroscope/blob/main/pkg/ingester/retention.go#L236-L238

Thank you for your help once more! If you need any assistance – please let me know

pkg/ingester/retention.go
```diff
@@ -359,6 +359,32 @@ type realFSBlockManager struct {
...
+		return nil, err
+	}
+
+	uploadedBlockIDs := make(map[ulid.ULID]struct{}, len(meta.Uploaded))
+	for _, id := range meta.Uploaded {
+		uploadedBlockIDs[id] = struct{}{}
+	}
+
+	return uploadedBlockIDs, nil
+}
```
weastel, line 386: @kolesnikovae, Extracted out abstraction for getting list of uploaded blocks ids, and when `shipper metafile` does not exist, it ignores and return empty map. This mostly solves problem, as without any information of uploaded, it will work fine only the `uploaded` flag will be unset and it will not affect any flows, `CleanupBlocksWhenHighDiskUtilization` will work fine as it sorts by both age and uploaded(which will be false for given scenario)

pkg/ingester/retention.go
```diff
@@ -289,12 +289,19 @@ func (dc *diskCleaner) CleanupBlocksWhenHighDiskUtilization(ctx context.Context)
 }
 
 // isBlockDeletable returns true if this block can be deleted.
-func (dc *diskCleaner) isBlockDeletable(block *tenantBlock) bool {
+func (dc *diskCleaner) isBlockDeletable(block *tenantBlock, deleteOnlyUploaded bool) bool {
```
kolesnikovae, line 292: I see that `deleteOnlyUploaded` is always `true`. The only exception is the test `TestDiskCleaner_isBlockDeletable`. I think we should remove the argument and update the test

kolesnikovae (commented):
Nice work @weastel! Could you please update the `TestDiskCleaner_isBlockDeletable test`? Now it tests the former logic, and `isBlockDeletable` has a special case for this, which we actually do not need

pkg/ingester/retention.go
weastel: I have updated some code, `deleteOnlyUploaded` helps function to detect if function should ignore upload or not. In terms of function usage in  `DeleteUploadedBlocks` should delete only uploaded blocks and for `CleanupBlocksWhenHighDiskUtilization` should delete any block(Prefarably uploaded, which is handled by sorting blocks), This has been updated in latest commits. In terms of test, some test are failing, I am investigating on that

kolesnikovae, line 292: Super minor suggestion: what do you think about splitting the function?
```go
func (dc *diskCleaner) isExpired(block *tenantBlock) bool
```

Caller can check `block.Unloaded` on its own, we do not have to hide it in the function

kolesnikovae (commented):
Thank you @weastel, the PR looks good! I've left a couple of minor notes, otherwise LGTM!

pkg/ingester/retention.go
weastel: Make sense, Updating PR

weastel: Updated PR, splitting function makes code more readable now. Thanks for reviewing

kolesnikovae (approved):

PR #3137, opened 2024-03-27
fix: remove redundant tag override that was not updated to 1.5.0

toscott:
Fixes #3121


simonswine (approved):
Thank you for letting us know (and fixing it)!

LGTM

PR #3141, opened 2024-03-27
Fix example links in README.md

simonswine:
cc @Rperry2174 there might be more of those problem looming somewhere 😂


Rperry2174 (approved):
Maybe easier to update the Alia's on the pages themselves to catch any more of these. I'll take a look tho thanks for catching!

simonswine:
@Rperry2174 don't think github itself has redirects, as the outdated bit is a link to a github folder.

In the meantime I also used this tool on the main README.md, which reports no broken links:

```
$ docker run -v ${PWD}:/tmp:ro -w /tmp --rm -i ghcr.io/tcort/markdown-link-check:stable README.md
```

PR #3156, opened 2024-04-02
feat(s3Storage): add BucketLookupType options for s3 storage

cao2358:
why: s3 storage need options for direct set the lookup type

### DIFF
--- a/pkg/objstore/providers/s3/bucket_client.go
+++ b/pkg/objstore/providers/s3/bucket_client.go
@@ -58,6 +58,19 @@ func newS3Config(cfg Config) (s3.Config, error) {
 			Transport:             cfg.HTTP.Transport,
 		},
 		// Enforce signature version 2 if CLI flag is set
-		SignatureV2: cfg.SignatureVersion == SignatureVersionV2,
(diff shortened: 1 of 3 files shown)


simonswine:
@cao2358 do you think that would work for you too? Do you mind describing the problem or error you get from it? I guess 
`auto` chooses virtual host and you want to force it to use path style?

Btw: Loki calls this [`-s3.force-path-style` ](https://grafana.com/docs/loki/latest/configure/#aws_storage_config) (So does [Tempo](https://grafana.com/docs/tempo/latest/configuration/#storage-block-configuration-example))

cao2358:
> @cao2358 do you think that would work for you too? Do you mind describing the problem or error you get from it? I guess `auto` chooses virtual host and you want to force it to use path style?
> 
> Btw: Loki calls this [`-s3.force-path-style` ](https://grafana.com/docs/loki/latest/configure/#aws_storage_config) (So does [Tempo](https://grafana.com/docs/tempo/latest/configuration/#storage-block-configuration-example))

I can close this pull request. Can you submit this option to the repository?

simonswine:
@cao2358  I have just created #3158, let me know if this works for you

PR #3160, opened 2024-04-03
docs: Add link to use pprof and go sdk simultaneously

Rperry2174:
This adds a link that shows people an example of how to use pprof + pyroscope sdk simultaneously. This is mostly so that we have easy access to a link to direct people to. I would say in most cases this is an edge case.


simonswine (approved):
LGTM

PR #3168, opened 2024-04-05
ci: Add test reports

simonswine:


simonswine:
Not really worth the effort

PR #3179, opened 2024-04-08
fix: clear profile buffer at flush

kolesnikovae:
When writing a chunk of profiles to disk, we allocate a buffer to hold the affected profiles, which we reuse throughout the lifetime of a mutable block in memory. Each entry in this buffer refers to a profile, including its metadata and samples.

The problem is that we do not clear this buffer after use, preventing the written profiles from being garbage collected, and thus consuming memory unnecessarily.

To compound the problem, creating a new parquet row group may require a considerable amount of memory, as evidenced by the spike in the graph. This makes the ingesters susceptible to OOM kills, even in relatively small deployments.

Notable changes proposed in the PR:
 - Parquet writer is not retained in-between chunk flushes.
 - Iterators are not buffers when row groups are read from the disk at block flush.
 - Flush buffers are cleaned out explicitly.

image

No noticeable increase in CPU consumption was identified. From the OS view, the working set size of the process will likely not change substantially, because of the FS cache.

The impact of the change is expected to be larger in large-scale deployments, where the amount of erroneously retained object and buffers is larger.

### DIFF
--- a/pkg/phlaredb/profiles.go
+++ b/pkg/phlaredb/profiles.go
@@ -493,7 +493,6 @@ func (pi *profilesIndex) cutRowGroup(rgProfiles []*schemav1.InMemoryProfile) err
 		n := copy(ps.profiles, ps.profiles[count:])
 		slices.Clear(ps.profiles[n:])
 		ps.profiles = ps.profiles[:n]
-		// attach rowGroup and rowNum information
 		ps.profilesOnDisk = append(
 			ps.profilesOnDisk,
 			rowRangePerFP[ps.fp],
--- a/pkg/phlaredb/profiles.go
+++ b/pkg/phlaredb/profiles.go
@@ -462,7 +462,7 @@ func (pi *profilesIndex) writeTo(ctx context.Context, path string) ([][]rowRange
 	return rangesPerRG, writer.Close()
 }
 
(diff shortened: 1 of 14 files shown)


kolesnikovae:
I noticed that the change causes notable increase working set size, which is not expected; debugging

image

pkg/phlaredb/profiles.go
```diff
@@ -489,21 +490,14 @@ func (pi *profilesIndex) cutRowGroup(rgProfiles []schemav1.InMemoryProfile) erro
...
-		for i := range ps.profiles[:count] {
-			// Allow GC to evict the object.
-			ps.profiles[i] = nil
-		}
-		ps.profiles = ps.profiles[count:]
-
+		n := copy(ps.profiles, ps.profiles[count:])
+		slices.Clear(ps.profiles[n:])
+		ps.profiles = ps.profiles[:n]
 		// attach rowGroup and rowNum information
```
simonswine, line 496: Don't think that comment is helpful anymore

simonswine (approved):
LGTM

kolesnikovae:
Closing it for now: we do have some issues with how buffers are handled / retained, however, this fix does not solve them completely, and the benefit is barely measurable at scale. Instead, it makes sense to check how the row groups are merged at flush (we could probably avoid this altogether)

Issue #3148, opened 2024-04-01
ebpf: allow configuring maps size 


korniltsev:
- allow resizing `pids` map
- Maybe increase default value from 1024
- allow resizing `stacks` map
- allow resizing `symbols` map

PR #3181, opened 2024-04-08
ebpf: allow resizing maps

korniltsev:
address https://github.com/grafana/pyroscope/issues/3148
- allow resizing maps
- cleanup the `pids` map faster 
- increase default `pids` map size
- do not use arch specific hook name( cillium handles it)


korniltsev:
vm image broken down :(

simonswine (approved):
Tested this! Seems to work well.

LGTM

korniltsev:
> Just to make sure my understanding is correct, that this will double the memory footprint of the map, from 8K (2 * 4bytes key/value * 1024) to 16k.

 I think it is a bit more complicated
 
 At the map creation it will preallocate max_entries (2048) buckets  [here](https://github.com/torvalds/linux/blob/6c3eba1c5e283fd2bb1c076dbfcb47f569c3bfde/kernel/bpf/hashtab.c#L514-L513)
And each bucket is [probably ](https://github.com/torvalds/linux/blob/6c3eba1c5e283fd2bb1c076dbfcb47f569c3bfde/kernel/bpf/hashtab.c#L79-L78) `16 bytes`
So it will preallocate 32k memory for buckets.

Then for each map update it will allocate an [element of size](https://github.com/torvalds/linux/blob/6c3eba1c5e283fd2bb1c076dbfcb47f569c3bfde/kernel/bpf/hashtab.c#L500-L505) `sizeof(htab_elem) + round_up(htab->map.key_size, 8) + round_up(htab->map.value_size, 8)` which is `48+8+8=64` and if we insert 2048 elements into the map, it will take `64*2048=128k`

In total `160k`

I am not 100% sure in the math, just looked into the source of hashmap for the first time

towolf:
@korniltsev just for my information, will this fix help with this error message from grafana-agent?

`ts=2024-04-17T13:09:31.61514799Z level=error msg="updating pids map" component=pyroscope.ebpf.instance err="update: key too big for map: argument list too long"`

korniltsev:
> @korniltsev just for my information, will this fix help with this error message from grafana-agent?
> 
> ts=2024-04-17T13:09:31.61514799Z level=error msg="updating pids map" component=pyroscope.ebpf.instance err="update: key too big for map: argument list too long"

Yes, it may help, especially if you have a lot of short lived processes.

I think this commit is not included in any public release of the grafana-agent/alloy yet though

PR #3185, opened 2024-04-09
feat: Auto-refresh github token

bryanhuhta:
related: https://github.com/grafana/pyroscope-squad/issues/129
related: https://github.com/grafana/pyroscope-app-plugin/pull/392

When we go through the OAuth login flow with GitHub, GitHub sends us a use token and a refresh token. The user token expires in 8 hours and the refresh token expires in 6 months. Both tokens are encrypted into cookie the `GitHubLogin` endpoint send back to the UI.

This PR introduces mechanisms for the UI to request a new user token. To do so, I created a new endpoint `/vcs.v1.VCSService/GithubRefresh` which will use the refresh token to get a new user token.


pkg/querier/vcs/client/github.go
```diff
@@ -5,102 +5,22 @@ import (
...
-	}
-	if githubAppClientSecret == "" {
-		return nil, errors.New("missing GITHUB_CLIENT_SECRET environment variable")
-	}
-	return &oauth2.Config{
-		ClientID:     GithubAppClientID,
-		ClientSecret: githubAppClientSecret,
-		Endpoint:     o2endpoints.GitHub,
-	}, nil
-}
```
bryanhuhta, line 43: A lot of this OAuth stuff got abstracted to `token.go`

pkg/querier/vcs/service.go
```diff
@@ -31,44 +33,96 @@ func New(logger log.Logger) *Service {
...
 }
 
 func (q *Service) GithubLogin(ctx context.Context, req *connect.Request[vcsv1.GithubLoginRequest]) (*connect.Response[vcsv1.GithubLoginResponse], error) {
-	resp := connect.NewResponse(&vcsv1.GithubLoginResponse{})
-	cookie, err := client.AuthorizeGithub(ctx, req.Msg.AuthorizationCode)
+	cfg, err := githubOAuthConfig()
 	if err != nil {
-		return nil, fmt.Errorf("failed to authorize github: %w", err)
+		q.logger.Log("err", err, "msg", "failed to get GitHub OAuth config")
+		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("failed to authorize with GitHub"))
```
bryanhuhta, line 44: This pattern of "log the real error, return a generic message" is so we don't accidentally leak anything sensitive to the UI. While errors won't include tokens or any really sensitive info, better to be on the safe side with what we send back to the client.

pkg/querier/vcs/service.go
```diff
@@ -31,44 +33,96 @@ func New(logger log.Logger) *Service {
...
+	token, err := tokenFromRequest(req)
+	if err != nil {
+		q.logger.Log("err", err, "msg", "failed to extract token from request")
+		return nil, connect.NewError(connect.CodeUnauthenticated, fmt.Errorf("invalid token"))
+	}
+
+	err = rejectExpiredToken(token)
+	if err != nil {
+		return nil, err
+	}
```
bryanhuhta, line 100: We can check if the token is expired without having to talk to GitHub.

cyriltovena (approved):
LGTM

PR #3188, opened 2024-04-10
Fix panic when dealing with missing mapping ID

### LINKED ISSUE #3164: Panic: Index out of range error when merging profiles
Encountered index out of range errors in the read path (ingesters):

https://github.com/grafana/pyroscope/blob/4fdd876d8144494cb483ef74089d4809021ad5a8/pkg/pprof/merge.go#L257-L261

called from:

https://github.com/grafana/pyroscope/blob/4fdd876d8144494cb483ef74089d4809021ad5a8/pkg/pprof/merge.go#L66-L68

```
golang.org/x/sync@v0.5.0/errgroup/errgroup.go:72 +0x96
created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 93207350
golang.org/x/sync@v0.5.0/errgroup/errgroup.go:75 +0x56
golang.org/x/sync/errgroup.(*Group).Go.func1()
github.com/grafana/pyroscope/pkg/phlaredb/symdb/resolver.go:263 +0x138
github.com/grafana/pyroscope/pkg/phlaredb/symdb.(*Resolver).withSymbols.func1()
github.com/grafana/pyroscope/pkg/phlaredb/symdb/resolver.go:248 +0x178
github.com/grafana/pyroscope/pkg/phlaredb/symdb.(*Resolver).Pprof.func1(0x2258460?, {{0xc18379d420, 0x1, 0x1}, {0xc18379d428, 0x1, 0x1}, {0x0, 0x0, 0x0}})
github.com/grafana/pyroscope/pkg/pprof/merge.go:26
github.com/grafana/pyroscope/pkg/pprof.(*ProfileMerge).Merge(...)
github.com/grafana/pyroscope/pkg/pprof/merge.go:68 +0x62d
github.com/grafana/pyroscope/pkg/pprof.(*ProfileMerge).merge(0xc02ecaa7e0, 0xc0aee65350, 0x1?)
github.com/grafana/pyroscope/pkg/pprof/merge.go:259
github.com/grafana/pyroscope/pkg/pprof.RewriteMappings(...)
goroutine 93207317 [running]:

panic: runtime error: index out of range [0] with length 0
```

At a glance it looks like we are unable to merge profiles that have no mappings.

simonswine:
I have not gotten to the bottom of why the mapping ID is missing when
maxNodes!=0.

This is presenting a workaround until we full understand the issue.

Fixes #3164

### DIFF
--- a/pkg/phlaredb/symdb/resolver_pprof_truncate.go
+++ b/pkg/phlaredb/symdb/resolver_pprof_truncate.go
@@ -262,6 +262,10 @@ func createLocationStub(profile *googlev1.Profile) {
 		SystemName: stubNodeNameIdx,
 	}
 	profile.Function = append(profile.Function, stubFn)
+	// in the case there is no mapping, we need to create one
+	if len(profile.Mapping) == 0 {
+		profile.Mapping = append(profile.Mapping, &googlev1.Mapping{Id: 1})
+	}
 	stubLoc := &googlev1.Location{
 		Id:        uint64(len(profile.Location) + 1),
 		Line:      []*googlev1.Line{{FunctionId: stubFn.Id}},


simonswine:
@kolesnikovae I was able to reproduce the panic consistently with and this PR fixes it.


Have a look at this code, the block is in a shared bucket that you should be able to access after being logged in: https://github.com/simonswine/pyroscope/commit/3cf401507ce5a23d3cc399cf7d02d35a5d09cb86

All boils down to a single series

aleks-p:
How easy would it be to add a test that covers this specific case?

kolesnikovae (approved):
Great job!

I realize that the root cause of the problem lies a bit deeper – the profile that causes the issue is malformed: it has a single sample with the sentinel stack trace ID (0). Temporary fix: https://github.com/grafana/pyroscope/pull/3193. I'll investigate this further: https://github.com/grafana/pyroscope/issues/3195. 

This specific issue is caused by the fact that we do not expect that a profile has samples and does not have a mapping, therefore we reference the default one. Now that we rely on the mappings (that keep version and git ref), I think we should always create a stub mapping – to be fixed separately.

Apart from the read path fixes, I think we should make sure such profiles do get into the block: https://github.com/grafana/pyroscope/issues/3194

PR #3193, opened 2024-04-11
fix: discard invalid samples at read

kolesnikovae:
It's been discovered that some of the persisted profiles may contain invalid samples (stack trace ID == 0). As an immediate fix, such samples should be discarded, as further processing does not expect "empty" stack traces. This has led to numerous cryptic failures (#3164 and #2377, for example).

We should also ensure that such profiles are discarded before they gets writtien to the block (see #3194)


simonswine (approved):
LGTM

I have also found this path, where it probably would make sense to add a check too.:

https://github.com/grafana/pyroscope/blob/032c28bf73474287a53ea36fb3577e75d6611e64/pkg/phlaredb/schemas/v1/profiles.go#L339

PR #3201, opened 2024-04-11
[DOC] Add note for Agent to Alloy to docs

knylander-grafana:
With the release of Grafana Alloy, we're adding an admonition to the Agent-related documenation notifying people about Alloy and the deprecation of Agent.

This admonition is contained in a shared file that is in the Alloy repository. When used, the admonition looks like this:
image


simonswine (approved):
LGTM

PR #3204, opened 2024-04-12
fix: ensure pprof strings[0] is always ""

### LINKED ISSUE #3199: Exported pprof strings
Per pprof specification:
```
string_table[0] must always be "".
```

In practice, this is not guaranteed: an exported pprof file may have non-empty string at zero index.

kolesnikovae:
Resolves #3199


simonswine (approved):
LGTM

PR #3205, opened 2024-04-12
Upgrade prometheus to v2.51.2

simonswine:
Noticed when doing some refactor on `pkg/phlaredb/tsdb/index`, that we target a quite strange prometheus version.

Note: the ebpf test failing is unrelated, that will be fixed via #3181


korniltsev (approved):

korniltsev:
haha, now if I try to update ebpf in grafana agent - it does not compile :smiling_imp:

PR #3200, opened 2024-04-11
Improve wording in docs / examples

aleks-p:
- Fixes various typos I've encountered
- Replaces "flamegraph" with "flame graph" in most places for consistency
- Switches to official names for frameworks / tools (e.g., Node.js, GitHub)

cc @knylander-grafana 

Note: if we don't want to do the "flamegraph" -> "flame graph" change now, I can revert this part of the PR

PR #3208, opened 2024-04-12
doc: More spelling errors

simonswine:
see #3206

@dgzlopes more of those lurking 😂


aleks-p:
I think these were fixed in #3200

PR #3218, opened 2024-04-15
fix: sanitize pprof references

### LINKED ISSUE #3203: Handle invalid profiles gracefully
Ingestion handling of pprof is currently quite unsafe, meaning that a panic is possible on any unexpected profile structure. For example, elements with ID 0 (not allowed in pprof) may cause panic like this:

image

Instead, only affected elements should be discarded from the profile, and the valid part should be preserved.

kolesnikovae:
Resolves #3203


pkg/pprof/merge_test.go
```diff
@@ -33,51 +35,6 @@ func Test_Merge_Self(t *testing.T) {
 	testhelper.EqualProto(t, p.Profile, m.Profile())
 }
 
-func Test_Merge_ZeroReferences(t *testing.T) {
```
kolesnikovae, line 36: I had to remove the test as it's no longer valid: invalid elements are discarded at normalization/merge

pkg/pprof/pprof.go
```diff
@@ -419,29 +422,6 @@ func (p *Profile) clearAddresses() {
 	}
 }
 
-// ensureHasMapping ensures all locations have at least a mapping.
-func (p *Profile) ensureHasMapping() {
```
kolesnikovae, line 423: Now we do it at reference sanitisation

pkg/pprof/merge.go
```diff
@@ -36,7 +36,7 @@ func (m *ProfileMerge) merge(p *profilev1.Profile, clone bool) error {
 	if p == nil || len(p.StringTable) < 2 {
 		return nil
 	}
-	ConvertIDsToIndices(p)
+	sanitizeReferences(p)
```
kolesnikovae, line 39: The only reason we do it here is that `Merge` might be called without `Normalize`. It is a fairly expensive operation, but it's better than a failure

pkg/pprof/pprof.go
```diff
@@ -1195,3 +1175,135 @@ func Unmarshal(data []byte, p *profilev1.Profile) error {
...
+	// invalid references (>= len(string_table)) are set to 0.
+	// references to empty string are set to 0.
+	str := func(i int64) int64 {
+		if i == 0 && z > 0 {
+			// z > 0 indicates that "" is not at string_table[0].
+			// This means that element that used to be at 0 has
+			// been moved to z.
+			return z
+		}
+		if i == z || i >= ms {
```
simonswine, line 1208: We probably should also cover this case: 

```suggestion
		if i == z || i >= ms || i < 0 {
```

simonswine (commented):
Generally LGTM.

A few things that a simple fuzzing session uncovered (not originating from this PR):

```
--- a/pkg/pprof/merge.go
+++ b/pkg/pprof/merge.go
@@ -196,6 +196,9 @@ func combineHeaders(a, b *profilev1.Profile) error {
 // returns nil if the profiles are compatible; otherwise an error with
 // details on the incompatibility.
 func compatible(a, b *profilev1.Profile) error {
+       if a == nil || b == nil {
+               return fmt.Errorf("a nil profile was specified")
+       }
        if !equalValueType(a.PeriodType, b.PeriodType) {
                return fmt.Errorf("incompatible period types %v and %v", a.PeriodType, b.PeriodType)
        }
@@ -213,6 +216,12 @@ func compatible(a, b *profilev1.Profile) error {
 // equalValueType returns true if the two value types are semantically
 // equal. It ignores the internal fields used during encode/decode.
 func equalValueType(st1, st2 *profilev1.ValueType) bool {
+       if st1 == nil && st2 == nil {
+               return true
+       }
+       if st1 == nil || st2 == nil {
+               return false
+       }
        return st1.Type == st2.Type && st1.Unit == st2.Unit
 }

@@ -242,11 +251,13 @@ func RewriteStrings(p *profilev1.Profile, n []uint32) {
        }
        p.DropFrames = int64(n[p.DropFrames])
        p.KeepFrames = int64(n[p.KeepFrames])
-       if p.PeriodType.Type != 0 {
-               p.PeriodType.Type = int64(n[p.PeriodType.Type])
-       }
-       if p.PeriodType.Unit != 0 {
-               p.PeriodType.Unit = int64(n[p.PeriodType.Unit])
+       if p.PeriodType != nil {
+               if p.PeriodType.Type != 0 {
+                       p.PeriodType.Type = int64(n[p.PeriodType.Type])
+               }
+               if p.PeriodType.Unit != 0 {
+                       p.PeriodType.Unit = int64(n[p.PeriodType.Unit])
+               }
        }
        for i, x := range p.Comment {
                p.Comment[i] = int64(n[x])
```
```

simonswine (approved):

kolesnikovae:
I see that we actually do not check if an element is nil (e.g. nil sample) at all – to be added in this PR

PR #3221, opened 2024-04-15
Remove tenant-settings from Ingress

nasa9084:
The backend Service is no longer available as it was removed in #2922.
Honestly, I don't know what it was in detail but the PR said the service is not relevant when running outside of Grafana Cloud so I believe this ingress path setting can be removed as well.


kolesnikovae (approved):
Thank you for removing this – you're right, the route is not needed here

PR #3222, opened 2024-04-15
Implement fuzzer for Merge_Single

simonswine:
Not too sure if there is a better/more performant way of doing that. But this also uncovered a few cases already 🙂

### DIFF
--- a/pkg/pprof/merge_test.go
+++ b/pkg/pprof/merge_test.go
@@ -1,8 +1,6 @@
 package pprof
 
 import (
-	"bytes"
-	"encoding/gob"
 	"os"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


pkg/pprof/merge_test.go
```diff
@@ -17,6 +20,29 @@ func Test_Merge_Single(t *testing.T) {
...
+		}
+
+		var m ProfileMerge
+		err = m.Merge(p.Profile)
+		if err != nil {
+			return
+		}
+	})
+}
+
```
kolesnikovae, line 45: What do you think about using use of proto encoding? It's way faster and seems more natural

```go
func Fuzz_Merge_Single(f *testing.F) {
	raw, err := OpenFile("testdata/go.cpu.labels.pprof")
	require.NoError(f, err)
	data, err := raw.MarshalVT()
	require.NoError(f, err)
	f.Add(data)

	f.Fuzz(func(t *testing.T, data []byte) {
		var p profilev1.Profile
		if err = p.UnmarshalVT(data); err != nil {
			var m ProfileMerge
			_ = m.Merge(&p)
		}
	})
}
```

simonswine: I think I was having trouble because I forgot about gzip, when doing an approach like that 🙂

kolesnikovae:
I've been running it on main for 2 hours with no failures 🎉

pkg/pprof/merge_test.go
```diff
@@ -19,6 +19,30 @@ func Test_Merge_Single(t *testing.T) {
...
+		raw, err := OpenFile(file)
+		require.NoError(f, err)
+		data, err := raw.MarshalVT()
+		require.NoError(f, err)
+		f.Add(data)
+	}
+
+	f.Fuzz(func(t *testing.T, data []byte) {
+		var p profilev1.Profile
+		if err := p.UnmarshalVT(data); err != nil {
```
simonswine, line 39: As @aleks-p has pointed out yesterday, we exit early.

```suggestion
		if err := p.UnmarshalVT(data); err == nil {
```

kolesnikovae: Oops, sorry about that! 😅 I made sure I was testing the correct version. I ran it again for 2 hours and didn't spot any errors. Without the recent changes, it fails very quickly

simonswine:
@aleks-p I have looekd into how we can actually run those Fuzz tests regularly. Not too sure if it will work, but at least I tried to have a valid config around 🙂 

@kolesnikovae I was curious how often we get post decoding and merge:

```
postDecode: 5325/522287 (last 3s, total) postMerge 5325/520733 (last 3s, total)

# Also for completeness when encoding/decoding was gob based:
postDecode: 5/39149 (last 3s, total) postMerge 5/39149 (last 3s, total)
```

kolesnikovae (approved):

PR #3243, opened 2024-04-23
[DOC] Add changes from Cloud Profiles UI to OSS docs

knylander-grafana:
Add doc changes from the Cloud Profiles UI docs back to the Pyroscope OSS equivalent. Based on https://github.com/grafana/website/pull/19509


docs/sources/view-and-analyze-profile-data/pyroscope-ui.md
```diff
@@ -87,8 +87,13 @@ In this case, you can use the sandwich view to see that a logging function calle
 
 ## Comparison page
```
knylander-grafana, line 88:
```suggestion
## Comparison view
```

docs/sources/view-and-analyze-profile-data/pyroscope-ui.md
```diff
@@ -74,7 +74,7 @@ The Single View page in Pyroscope's UI is built for in-depth profile analysis. H
 <!-- Visual Placeholder:** *Screenshots demonstrating each view option in the Single View page.* -->
 
 Let's say that your app has a spike in memory usage.
-Without profiling, you would go from a memory spike to digging through code or guessing the cause.
+Without profiling, you would go from a spike CPU usage metric to digging through code or guessing the cause.
```
aleks-p, line 77: This seems off, as we talk about memory in the previous line, do we want to reuse the entire paragraph from https://github.com/grafana/website/pull/19509/?

knylander-grafana: Here's the entire text from the website PR: 

>This screenshot shows a spike in CPU usage.
>Without profiling, you would go from a spike in CPU usage metric to digging through code or guessing what the cause.
>However, with profiling, you can use the flame graph and table to see exactly which function is most responsible for the spike.
>Often this shows up as a single node taking up a noticeably disproportionate width in the flame graph as seen below with the `checkDriverAvailability` function.

From what you're saying, it looks like we should replace "Let's say that your app has a spike in memory usage" with "This screenshot shows a spike in CPU usage." 

Is there additional text you would like to see included?

aleks-p: Yes, that replacement makes sense since the screenshot we show after this text (https://grafana.com/docs/pyroscope/latest/view-and-analyze-profile-data/pyroscope-ui/#single-view) has a CPU profile. 

> Is there additional text you would like to see included?
Not really, the "fix" above should be enough.

I do however think we have a minor phrasing issue in this line (also present in https://github.com/grafana/website/pull/19509)

Here we say 
> Without profiling, you would go from a spike CPU usage metric to digging through code or guessing the cause.

In https://github.com/grafana/website/pull/19509 we say
> Without profiling, you would go from a spike in CPU usage metric to digging through code or guessing what the cause.

This could work:
> Without profiling, you would go from the CPU usage metric spike to digging through code or guessing the cause.

aleks-p (approved):
I found a small phrasing issue, otherwise it looks good!

docs/sources/view-and-analyze-profile-data/pyroscope-ui.md
knylander-grafana: I think the bottom sentence is what it's supposed to be and missed deleting "what".

docs/sources/view-and-analyze-profile-data/pyroscope-ui.md
```diff
@@ -74,7 +74,7 @@ The Single View page in Pyroscope's UI is built for in-depth profile analysis. H
 <!-- Visual Placeholder:** *Screenshots demonstrating each view option in the Single View page.* -->
 
 Let's say that your app has a spike in memory usage.
```
knylander-grafana, line 76:
```suggestion
This screenshot shows a spike in CPU usage.
```

docs/sources/view-and-analyze-profile-data/pyroscope-ui.md
knylander-grafana: Fixed: https://github.com/grafana/website/pull/19597

PR #3246, opened 2024-04-24
fix: panic recovery interceptor

kolesnikovae:
We have a metric that counts recovered panics – `pyroscope_panic_total`, however it does not include panics recovered in gRPC handlers


simonswine (approved):
LGTM

Thanks for looking into that

PR #3251, opened 2024-04-25
fix: deprecate pprof MergeNoClone

kolesnikovae:
`pprof.MergeNoClone` has particularly tricky semantics which is very easy to misuse: specifically, it may cause duplication of samples. Although the function is significantly cheaper than `pprof.Merge` and can be fixed, I think it's better to deprecate it in favour of safety

### DIFF
--- a/pkg/pprof/merge.go
+++ b/pkg/pprof/merge.go
@@ -9,6 +9,19 @@ import (
 	"github.com/grafana/pyroscope/pkg/slices"
 )
 
+// TODO(kolesnikovae):
+//   Add Clean function that incorporates Merge and Normalize.
+//   Both functions perform some sanity checks but none of them
+//   is enough to "vet" the profile completely.
+//   Specifically:
+//    - it's possible that unreferenced objects will remain in the
+//      profile, and therefore will be written to the storage.
+//    - Normalize does not remove duplicates and unreferenced objects
+//      except samples.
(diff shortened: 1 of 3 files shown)


pkg/pprof/merge.go
```diff
@@ -9,6 +9,19 @@ import (
 	"github.com/grafana/pyroscope/pkg/slices"
 )
 
+// TODO(kolesnikovae):
+//   Add Clean function that incorporates Merge and Normalize.
```
kolesnikovae, line 13:
```suggestion
//   Add a function that incorporates Merge and Normalize.
```

pkg/pprof/merge_test.go
```diff
@@ -23,10 +24,17 @@ func Test_Merge_Single(t *testing.T) {
...
+	sortLabels(p.Profile)
 	testhelper.EqualProto(t, p.Profile, m.Profile())
 }
 
+func sortLabels(p *profilev1.Profile) {
+	for _, s := range p.Sample {
+		sort.Sort(LabelsByKeyValue(s.Label))
+	}
+}
+
```
kolesnikovae, line 37: Note that `Merge` modifies profiles in place. Therefore, some tests were incorrect. `MergeNoClone` in addition retains the objects; it's quite hard to use the function correctly. I'm almost sure this is causing issues.

korniltsev (approved):

PR #3254, opened 2024-04-25
Update ebpf testdata

simonswine:
Pulls in https://github.com/pyroscope-io/pyroscope-ebpf-testdata/pull/4


aleks-p:
Last time I looked at this, these files were not duplicate - their sizes are different. If you google some of these kernel files you'll find that typically both sets are together and they likely have a different purpose.

Also, if someone runs these instructions, the files will be brought in again: 
https://github.com/pyroscope-io/pyroscope-ebpf-testdata/blob/main/qemu_img/readme.txt

I am ok with removing them (I find dealing with the dirty Git repo annoying as well) but perhaps @korniltsev can share his thoughts?

korniltsev:
Thanks  Aleks  I think you are right 
https://github.com/torvalds/linux/blob/master/net/netfilter/xt_DSCP.c
https://github.com/torvalds/linux/blob/master/net/netfilter/xt_dscp.c
I think we need to revert submodule changes

I guess I do not have such issues because I have CaseSensitive APFS on my mac

simonswine:
> Thanks Aleks I think you are right https://github.com/torvalds/linux/blob/master/net/netfilter/xt_DSCP.c https://github.com/torvalds/linux/blob/master/net/netfilter/xt_dscp.c I think we need to revert submodule changes
> 
> I guess I do not have such issues because I have CaseSensitive APFS on my mac

aleks-p:
It would still be nice to "fix" this in some way, perhaps by ignoring these files and pulling them in with a script when needed. I don't know how easy that is though.

korniltsev:
1. you can create a case sensitive volume and put repo there.
2. we can move the files needed for VM building into yet another repo (https://github.com/pyroscope-io/pyroscope-ebpf-testdata/pull/4)
3. maybe you can try to apply 'git update-index --assume-unchanged` on those files and git would be happy to ignore them, but i am not sure

PR #3255, opened 2024-04-26
[DOC] Clarify required config options for Java span profiles

knylander-grafana:
Added the two required configuration options for Java SDK to product span profiles.


simonswine (commented):
I don't think that is correct:

The two variables allow to enable lock and heap profiling, which by default is of. 

There is no support for span profiles for lock and heap profiles at all.

knylander-grafana:
Thank you for closing this PR. 

I opened a new doc issue to capture some updates to clarify which instrumentation methods support which profiler types. https://github.com/grafana/pyroscope/issues/3259

PR #3258, opened 2024-04-26
Fix debug info attribute

simonswine:
This is something that would be nice to test, but I guess we don't want to get into circular dependecies with alloy :cry:

I also migrated the targets to be fully structured, so they display better than:

image


bryanhuhta (approved):
LGTM 👍

PR #3273, opened 2024-05-02
feat: Track metrics around usage of the GitHub API

bryanhuhta:
Related: https://github.com/grafana/pyroscope-squad/issues/138

We have metrics surrounding how the VCS service behaves (latency, errors, etc) but we have limited insight into our usage of the GitHub API. This PR tracks our GitHub API usage in two ways:

- duration of requests
- count of how close we are to rate limiting


pkg/querier/vcs/client/metrics.go
```diff
@@ -0,0 +1,103 @@
...
+		APIDuration: promauto.With(reg).NewHistogramVec(
+			prometheus.HistogramOpts{
+				Namespace: "pyroscope",
+				Name:      "vcs_github_request_duration",
+				Help:      "Duration of GitHub API requests in seconds",
+				Buckets:   prometheus.ExponentialBucketsRange(0.1, 10, 8),
+			},
+			[]string{"path", "status"},
+		),
+		APIRateLimit: promauto.With(reg).NewGauge(prometheus.GaugeOpts{
```
simonswine, line 32: I worry that this metric might not be too helpful/accurate:

My understanding is that we are subject to rate limits of the github user that authenticated via oauth2. So if we have two github users using the service at the same time we would constantly have a jumping value, between those two values of remaining requests.

I do think this is something important to watch. Maybe this could be a histogram with fitting buckets, that would show us how close to being rate limited we got.

pkg/querier/vcs/client/metrics.go
```diff
@@ -0,0 +1,103 @@
...
+			Namespace: "pyroscope",
+			Name:      "vcs_github_rate_limit",
```
simonswine, line 34:
```suggestion
			Name:      "vcs_github_remaining_request_quota",
```

pkg/querier/vcs/client/github.go
```diff
@@ -15,21 +15,26 @@ import (
 )
 
 // GithubClient returns a github client.
-func GithubClient(ctx context.Context, token *oauth2.Token) (*githubClient, error) {
+func GithubClient(ctx context.Context, token *oauth2.Token, metrics Metrics) (*githubClient, error) {
 	return &githubClient{
-		client: github.NewClient(nil).WithAuthToken(token.AccessToken),
+		client:  github.NewClient(nil).WithAuthToken(token.AccessToken),
```
simonswine, line 20: My feeling is that a more straightforward way of getting the actual github status code instrumented would be by using an instrumented `http.Client` (for NewClient's parameter) with a middleware that could collect the same metrics.

The prometheus client already comes with ones e.g. this one for duration:

https://github.com/prometheus/client_golang/blob/77d4003c72f054ac435df1223deac17b1f8858ea/prometheus/promhttp/instrument_client.go#L87

We do something for tracing already here (we don't use it for the github): https://github.com/grafana/pyroscope/blob/09f0787e8a0f049b36ffa105161aa66bdb3c9e89/pkg/util/http.go#L53

korniltsev (approved):

simonswine (commented):
Thanks for looking into this. 

I do think it might be easier instrumenting on the `*http.Client` level with a rounder tripper, as this would be a lot less subject to refactoring and breakages and has access to a lot more data.

simonswine (approved):

pkg/querier/vcs/client/metrics.go
bryanhuhta: You're absolutely right here. The rate limit remaining here is _per-user_ ([doc](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-authenticated-users)) since we're using a user token. If we had an installation token, the rate limit would be per app.

How I'm using the metric here is not helpful at all. Like you said, it's going to jump to whatever remaining rate limit the last token had. I'm going to rethink how to track this, as a gauge is clearly not the right approach.

pkg/querier/vcs/client/github.go
bryanhuhta: I did think about this. I agree with you that the current approach is very brittle and doesn't provide very good access to other information surrounding the request. I think adding `RoundTripper` middleware is an excellent solution.

The main drawback to that approach is I'm not sure how to label the metrics with the API path template. Using middleware we can get the complete path, but that includes ids, shas/ filenames, etc. E.g. I want

```
/repos/{owner}/{repo}/commits/{ref}
```

but the middleware would provide

```
/repos/grafana/pyroscope/commits/58f66fc6e6806101fa3eb9cc01f73a8c50b8aed5
```

simonswine: I think something to adopt from dskit: https://github.com/grafana/dskit/blob/c41c58734f46e8d4b5f1f407a6a8c01eb4184d17/middleware/instrument.go#L131

simonswine (approved):
LGTM

simonswine:
Thanks for this new revision, I have not fully tested it myself, but it looks all correct code wise

PR #3275, opened 2024-05-03
docs: improves S3 example config

petethepig:
Making sure it's less likely that people use `grafana-pyroscope-data` as their bucket name.


bryanhuhta (approved):
LGTM

PR #3279, opened 2024-05-06
fix(helm): kubernetes_pods_*_default_name keep only targets(container…

korniltsev:
…s) with needed port

Without this change If I have a single pod with two containers and just one port from one container  and `profiles.grafana.com/${type}.port: "${port}"` annotation, the agent will attempt to scrape the port twice, which is unnecessary and also does not work for cpu .


simonswine (approved):
Good catch. LGTM

PR #3280, opened 2024-05-07
fix(ebpf): fix issue when a cls arg is a cell

korniltsev:
Fixes "invalid utf string" issue in our dev cluster.

The issue was caused by ebpf profiler type confusing - trying to use `PyCellObject` as `PyTypeObject`

In addition fixes issue when first argument name is incorrectly detected by zeroing the arg name buffer. (previously if could contain garbage from previous frame)


simonswine (approved):
LGTM

Couldn't find a regression, with a quick happy path check. And I also spotted this for the first time 🙂 

https://github.com/grafana/pyroscope/blob/a189d5750cb205c30491a04722ba7f5aaf4f1ac8/ebpf/bpf/pyperf.bpf.c#L295-L300

korniltsev:
I preproduced the issue here https://github.com/pyroscope-io/pyroscope-ebpf-testdata/blob/fc63fce7e3e3995d305cdfab0efc239e5632c164/rideshare-flask-no-pip/lib/server.py#L28-L43

korniltsev:
> And I also spotted this for the first time 🙂

Yesterday I found another issue with the way the code detects class name.

`f_code->co_varnames` may contain same name multiple times. It may look like `('self', 'foo', 'bar', 'self')` and `f_localsplus[0] == NULL` and `f_localsplus[3] == 0x7f....`

PR #3284, opened 2024-05-08
fix(ebpf): handle case when self is put in cell

korniltsev:
Reproducer:

 https://github.com/pyroscope-io/pyroscope-ebpf-testdata/blob/e567550674d0312c3a2f2d07ea3bb395a13a085b/rideshare-flask-no-pip/lib/server.py#L45-L59


simonswine (approved):
LGTM

PR #3285, opened 2024-05-08
Pyroscope: Provide an example for ebpf sidecar collection

simonswine:
This shows how Alloy can be used to collect profiles as a sidecar. I am not too sure if this is something we should advertise too heavily, but valuable if you would only like to collect a very specific workloads, that is supported by eBPF.

Maybe we also want to have something like `pyroscope exec` was for this usecase.

The test workload launches some short-lived sub processes of varying length, the problem is that collection requires to have a quite low `discovery.process.refresh_interval`

### DIFF
--- a/examples/grafana-agent-auto-instrumentation/ebpf/sidecar/deploy.yaml
+++ b/examples/grafana-agent-auto-instrumentation/ebpf/sidecar/deploy.yaml
@@ -1,0 +1,118 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: collect-sidecar
+  labels:
+    app: example-ebpf-sidecar
+spec:
+  replicas: 2
+  selector:
+    matchLabels:
+      app: example-ebpf-sidecar
+  template:
+    metadata:
+      labels:
+        app: example-ebpf-sidecar
+    spec:
+      shareProcessNamespace: true
+      containers:
+      - name: collect
+        image: docker.io/simonswine/collect-ebpf-sidecar:latest@sha256:40b10016e90324dadabf5141e10b164bdbd05cd35de0e277609a1b29f44e3fda
+      - name: alloy
+        image: grafana/alloy
+        securityContext:
+          runAsUser: 0
+          allowPrivilegeEscalation: true
+          capabilities:
+            add:
+              - BPF
+              - SYS_ADMIN
+        command:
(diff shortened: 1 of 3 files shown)


examples/grafana-agent-auto-instrumentation/ebpf/sidecar/deploy.yaml
```diff
@@ -0,0 +1,118 @@
...
+  replicas: 2
+  selector:
+    matchLabels:
+      app: example-ebpf-sidecar
+  template:
+    metadata:
+      labels:
+        app: example-ebpf-sidecar
+    spec:
+      shareProcessNamespace: true
```
simonswine, line 17: This is a hard requirement for this to work

examples/grafana-agent-auto-instrumentation/ebpf/sidecar/deploy.yaml
```diff
@@ -0,0 +1,118 @@
...
+        command:
+          - /bin/bash
+          - -c
+          - |
+            set -euo pipefail
+            set -x
+
+            cat > /tmp/config.alloy <<"EOF"
+            discovery.process "all" {
+              refresh_interval = "200ms"
```
simonswine, line 39: With this I see the 100ms long running processes very rarely. I guess something that we loose by not consuming the PID events from the ebpf.Session in alloy directly.

simonswine:
@korniltsev if you got a moment, would like to understand how you feel about this

examples/grafana-agent-auto-instrumentation/ebpf/sidecar/deploy.yaml
korniltsev: I dont like the idea to run discovery.process every 200ms, but idk how bad it is.
I also don't like using alloy as sidecar at all. The binary is huge, consuming tons of space and memory.

1. As you said we can notify `discovery.process` from `discovery.ebpf` about new  PID
2. Another option could be to somehow revive (but maybe with a different naming), the `targets_only = false && default_target={"service_name": "foo239"}`

This could look like this
```
sidecar {
  enabled=true

  discover_config {
    cwd = true
    exe = true
...
  }
relabel {
  rule {
...
  }
... 
}
}


```

Can we get k8s labels somehow?

PR #3293, opened 2024-05-10
vcs: Derive per tenant key

simonswine:
Currently we use the same global session encryption secret, for each tenant. In order to ensure tenant isolation, this change will derive a custom secret per tenant.

By using sha256 we also increase the secret used for encryption from 128bit to 256bit, while allowing to get an arbitrary secret specified. 

Note: This change will require all users to re-authenticate, as the the previous GitSession won't be decrypted by this. It is possible to implement this without this breaking change, but given session length is 8 hours, I rather would re-authenticate instead.


bryanhuhta (approved):
Nice, this looks good!

PR #3308, opened 2024-05-20
examples: Update pyroscope-nodejs integration

simonswine:
This was required to support node v22 and to get rid of upstream pprof in favour of datadog/pprof


korniltsev:
It may be worth to update tools/update_examples.go yo teach it updating node examples

PR #3310, opened 2024-05-24
Connect: Use custom codec for vtproto

simonswine:
This implements a custom codec for use with connect-go. It uses the vtproto provided methods UnmarhsalVT rather than the upstream protobuf decoder/encoder.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -268,7 +268,7 @@ $(BIN)/protoc-gen-go: Makefile go.mod
 
 $(BIN)/protoc-gen-connect-go: Makefile go.mod
 	@mkdir -p $(@D)
-	GOBIN=$(abspath $(@D)) $(GO) install connectrpc.com/connect/cmd/protoc-gen-connect-go@v1.14.0
+	GOBIN=$(abspath $(@D)) $(GO) install connectrpc.com/connect/cmd/protoc-gen-connect-go@v1.16.2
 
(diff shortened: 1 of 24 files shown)


simonswine:
While I don't think this lead to significant shift in performance, it brings makes us use the benefits of vtprotobuf for the proto that are marshalled/unmarshalled from/for requests.

```
func (p *YourProto) MarshalVT() ([]byte, error): this function behaves identically to calling proto.Marshal(p), except the actual marshalling has been fully unrolled and does not use reflection or allocate memory. This function simply allocates a properly sized buffer by calling SizeVT on the message and then uses MarshalToSizedBufferVT to marshal to it.
```

```
unmarshal: generates a func (p *YourProto) UnmarshalVT(data []byte) that behaves similarly to calling proto.Unmarshal(data, p) on the message, except the unmarshalling is performed by unrolled codegen without using reflection and allocating as little memory as possible. If the receiver p is not fully zeroed-out, the unmarshal call will actually behave like proto.Merge(data, p). This is because the proto.Unmarshal in the ProtoBuf API is implemented by resetting the destination message and then calling proto.Merge on it. To ensure proper Unmarshal semantics, ensure you've called proto.Reset on your message before calling UnmarshalVT, or that your message has been newly allocated.
```

korniltsev (approved):
lgtm

did you run any benchmark locally?

simonswine:
No, I only deployed it to the distributors in ops. 

image

PR #3314, opened 2024-05-24
connect: Use klauspost/compress/gzip decompression

simonswine:
Now that we have shared options between all connect Handlers/Clients it is easy to swap out compress/gzip with klauspost/compress/gzip.

At the same time this makes sure that all handlers and clients use the same compressor pool. Otherwise every new handler and client maintains its own pool (And we create a lot of clients dynamically).

Relies on the work of #3310


simonswine:
image

bryanhuhta (approved):
LGTM!

PR #3315, opened 2024-05-24
Fix bug in query frontend diff handling

simonswine:
This fixes a panic, that could happen when one of the query results in the diff are empty.

Also adds some more coverage to the Diff handler in the frontend.

Note: We should pull this into f68.


aleks-p (approved):
Looks good to me, thank you for adding the tests!

Issue #3320, opened 2024-05-28
Panic during Ad Hoc profile upload


aleks-p:
I've observed the following panic today:

```
connectrpc.com/connect.NewUnaryHandler[...].func1({0x2f22c70, 0xc0001e2100})
	github.com/grafana/pyroscope/pkg/adhocprofiles/adhocprofiles.go:78 +0x2a5
github.com/grafana/pyroscope/pkg/adhocprofiles.(*AdHocProfiles).Upload(0xc000afc480, {0x2f14e78, 0xc000036000}, 0xc0001e2100)
	github.com/grafana/pyroscope/pkg/adhocprofiles/adhocprofiles.go:224 +0x105
github.com/grafana/pyroscope/pkg/adhocprofiles.parse(0xc000d3ef28, 0x0, 0xc000d3eee0?)
	github.com/grafana/pyroscope/pkg/og/structs/flamebearer/convert/convert.go:63 +0x8a
github.com/grafana/pyroscope/pkg/og/structs/flamebearer/convert.FlamebearerFromFile({{0xc000156708, 0x11}, {0x0, 0x0}, {{0x0, 0x0}, {0x0, 0x0}}, {0xc00138c000, 0x38d, ...}}, ...)
	github.com/grafana/pyroscope/pkg/og/structs/flamebearer/convert/convert.go:74 +0x39
github.com/grafana/pyroscope/pkg/og/structs/flamebearer/convert.Converter.func1({0xc00138c000?, 0x11?, 0x0?}, {0xc000156708?, 0x0?}, 0x0?)
	github.com/grafana/pyroscope/pkg/og/structs/flamebearer/convert/convert.go:169 +0x178
github.com/grafana/pyroscope/pkg/og/structs/flamebearer/convert.JSONToProfile({0xc00138c000, 0x38d, 0x38d}, {0x0?, 0x0?}, 0x2000)
	github.com/grafana/pyroscope/pkg/og/structs/flamebearer/flamebearer.go:347 +0x11e
github.com/grafana/pyroscope/pkg/og/structs/flamebearer.ProfileToTree({0x1, {{{0xc0001e2180, 0x6, 0x8}, {0xc000f9c000, 0x3, 0x4}, 0x64, 0x3c}, {{0xc000a98100, ...}, ...}, ...}, ...})
	github.com/grafana/pyroscope/pkg/og/structs/flamebearer/flamebearer.go:361 +0x1c5
[...]

PR #3324, opened 2024-05-29
Improve validation for Ad Hoc profile uploads

aleks-p:
Tries to address #3320. Without the actual input, it is only a guess that this is what caused it.


simonswine (approved):
LGTM! Input validation ftw.

Thanks for fixing this 🤞

PR #3325, opened 2024-05-29
Improve block plan creation for standalone binary mode

aleks-p:
In standalone mode the ingester and store-gateway are represented with the same replica address. This doesn't cause real issues in the query path (the same physical block plan is re-used for both query paths), however it does impact query analysis because we end up including one of the paths only.

Not sure if this would break something else, but seems fairly safe to me.


simonswine (approved):
LGTM

Thanks for addressing this long standing TODO. ❤️

PR #3331, opened 2024-05-30
Add initial load tests

bryanhuhta:
This is the initial pass at adding load tests with k6. Right now we only are testing read APIs. To run a load test against a local Pyroscope, run:

```
./tools/k6/run.sh reads.js
```

To run a load test against a remote Pyroscope, configure your `.env` file correctly (see `tools/k6/README.md` for more details) and run:

```
./tools/k6/run.sh -c reads.js
```

To view all available load tests, run

```
./tools/k6/run.sh -h
```

### DIFF
--- a/tools/k6/reads.js
+++ b/tools/k6/reads.js
@@ -17,8 +17,16 @@ export const options = {
     },
   },
 
-  vus: 1,
-  duration: '30s',
+  scenarios: {
+    even_reads: {
+      executor: 'constant-arrival-rate',
+      duration: '15m',
+      rate: 5,
+      timeUnit: '1m',
+      preAllocatedVUs: 3,
+      maxVUs: 10,
+    },
+  },
 
   thresholds: {
     checks: ['rate>0.9'],
--- a/tools/k6/reads.js
+++ b/tools/k6/reads.js
@@ -1,0 +1,129 @@
+import { group } from 'k6';
+
+import {
+  doLabelNamesRequest,
+  doRenderDiffRequest,
+  doRenderRequest,
+  doSelectMergeProfileRequest,
+  doSelectMergeStacktracesRequest,
+  doSeriesRequest,
+} from './lib/request.js';
+
+export const options = {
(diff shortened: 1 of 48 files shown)


tools/k6/reads.js
```diff
@@ -0,0 +1,137 @@
...
+export default function() {
+  group('reads last 1h', function() {
+    const { start, end } = newRelativeTimeRange(1, 'h');
+    doAllQueryRequests(start, end);
+  });
+
+  group('reads last 24h', function() {
+    const { start, end } = newRelativeTimeRange(24, 'h');
+    doAllQueryRequests(start, end);
+  });
```
simonswine, line 67: I think an environment variable for those durations would be nice

export K6_QUERY_DURATIONS=1h,24h

tools/k6/reads.js
```diff
@@ -0,0 +1,137 @@
...
+}
+
+function doAllQueryRequests(start, end) {
+  doSelectMergeProfileRequest({
+    start,
+    end,
+    profile_typeID: 'process_cpu:cpu:nanoseconds:cpu:nanoseconds',
+    label_selector: '{service_name="fire-dev-001/ingester"}',
```
simonswine, line 75: I also think that would be useful to customize without code change:

```suggestion
    label_selector: '{service_name="%s"}' % (K6_QUERY_SERVICE_NAME),
```

simonswine (approved):
LGTM

tools/k6/reads.js
bryanhuhta: Added in https://github.com/grafana/pyroscope/pull/3331/commits/e00e5fd306de2847c028a665c75c367b134d8e86.

tools/k6/reads.js
bryanhuhta: Added in https://github.com/grafana/pyroscope/pull/3331/commits/e00e5fd306de2847c028a665c75c367b134d8e86.

PR #3332, opened 2024-05-31
chore: Fix the helm-ci workflow

aleks-p:
Currently the helm-ci workflow is broken due to a syntax error, resulting in "PR run failed at startup: helm-ci " emails from GitHub for every new PR (and subsequent push).

Once the syntax error is gone, the workflow itself fails because of a deadlock in the CI setup:
- we use `ct` (chart testing), which in turn runs `helm install ... --wait`
- when we use minio in our chart (`minio.enabled: true`), a Kubernetes job is generated to create a bucket in minio
- the job has a helm "post-install" hook, so it only gets created **after** `helm install` completes
- `helm install --wait` can only complete after all pods are `ready`, but the store-gateway needs a bucket which never gets created

The same scenario would happen outside the CI, but our current docs instruct users to call `helm install` without `--wait`. This way the bucket creation job gets created and store-gateways eventually start. 

Related: https://github.com/grafana/pyroscope/issues/3385


.github/workflows/helm-ci.yml
```diff
@@ -14,4 +14,3 @@ jobs:
       ct_configfile: operations/pyroscope/helm/ct.yaml
       ct_check_version_increment: false
       helm_version: v3.14.3
-      kubeVersion: "1.23"
```
simonswine, line 17: I wonder why this is a problem, I added it with this PR: https://github.com/grafana/pyroscope/pull/3159 to support the horizontal pod auto-scaler mode.

It is also not aligning with the version set here: https://github.com/grafana/pyroscope/blob/main/operations/pyroscope/helm/ct.yaml#L4

simonswine (approved):
LGTM

.github/workflows/helm-ci.yml
aleks-p: https://github.com/grafana/pyroscope/actions/runs/9778263318

> [Invalid workflow file: .github/workflows/helm-ci.yml#L17](https://github.com/grafana/pyroscope/actions/runs/9778263318/workflow)
The workflow is not valid. .github/workflows/helm-ci.yml (Line: 17, Col: 20): Invalid input, kubeVersion is not defined in the referenced workflow.

`kubeVersion` is not a valid field, see https://github.com/grafana/helm-charts/blob/32ad8e1a78724cf1595090857b30d6032880d47a/.github/workflows/lint-test.yaml#L5-L28

PR #3335, opened 2024-06-04
chore: support dots in label names

kolesnikovae:
We're currently using [prometheus](https://prometheus.io/docs/concepts/data_model/) data model for labels (including those that are set as pprof tags) for consistency with other products:
> Labels may contain ASCII letters, numbers, as well as underscores. They must match the regex [a-zA-Z_][a-zA-Z0-9_]*.

Although we can not easily loosen the requirement, I'm not very sure if discarding profiles that violate it is the best strategy.

There is ongoing work on support for UTF-8 labels in Prometheus: 
https://github.com/prometheus/prometheus/issues/13095:

> OTel allows UTF-8 in label names while Prometheus has a much [more restrictive set](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels). This is causing friction for users when using Prometheus. In particular, . (dot) is a very common character in OTel and we convert that to _ when converting to Prometheus. For example, service.version becomes service_version.

Until the support is fully implemented, I propose to do the same in Pyroscope – convert `.` to `_` in label names to make interoperability easier. Theoretically, we could replace any disallowed characters, but that leads to ambiguity in certain cases and may confuse users.

### DIFF
--- a/pkg/validation/validate.go
+++ b/pkg/validation/validate.go
@@ -123,19 +123,54 @@ func ValidateLabels(limits LabelValidationLimits, tenantID string, ls []*typesv1
 	for _, l := range ls {
 		if len(l.Name) > limits.MaxLabelNameLength(tenantID) {
 			return NewErrorf(LabelNameTooLong, LabelNameTooLongErrorMsg, phlaremodel.LabelPairsString(ls), l.Name)
-		} else if len(l.Value) > limits.MaxLabelValueLength(tenantID) {
+		}
+		if len(l.Value) > limits.MaxLabelValueLength(tenantID) {
 			return NewErrorf(LabelValueTooLong, LabelValueTooLongErrorMsg, phlaremodel.LabelPairsString(ls), l.Value)
-		} else if !model.LabelName(l.Name).IsValid() {
+		}
+		var ok bool
+		if l.Name, ok = SanitizeLabelName(l.Name); !ok {
 			return NewErrorf(InvalidLabels, InvalidLabelsErrorMsg, phlaremodel.LabelPairsString(ls), "invalid label name '"+l.Name+"'")
-		} else if !model.LabelValue(l.Value).IsValid() {
+		}
+		if !model.LabelValue(l.Value).IsValid() {
 			return NewErrorf(InvalidLabels, InvalidLabelsErrorMsg, phlaremodel.LabelPairsString(ls), "invalid label value '"+l.Value+"'")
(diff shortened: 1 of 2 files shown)


pkg/validation/validate_test.go
```diff
@@ -400,3 +401,36 @@ func TestValidateFlamegraphMaxNodes(t *testing.T) {
...
+		{"..", "__", true},
+		{"..a", "__a", true},
+		{"a..", "a__", true},
+		{"a.a", "a_a", true},
+		{".a.", "_a_", true},
+		{"..a..", "__a__", true},
+		{"世界", "世界", false},
+		{"界世_a", "界世_a", false},
+		{"界世__a", "界世__a", false},
+		{"a_世界", "a_世界", false},
```
korniltsev, line 424:
```suggestion
		{"a_世界", "a_世界", false},
		{"0.a", "0.a", false},
		{"0a", "0a", false},
```

pkg/validation/validate.go
```diff
@@ -123,13 +123,18 @@ func ValidateLabels(limits LabelValidationLimits, tenantID string, ls []*typesv1
...
 		if len(l.Name) > limits.MaxLabelNameLength(tenantID) {
 			return NewErrorf(LabelNameTooLong, LabelNameTooLongErrorMsg, phlaremodel.LabelPairsString(ls), l.Name)
-		} else if len(l.Value) > limits.MaxLabelValueLength(tenantID) {
+		}
+		if len(l.Value) > limits.MaxLabelValueLength(tenantID) {
 			return NewErrorf(LabelValueTooLong, LabelValueTooLongErrorMsg, phlaremodel.LabelPairsString(ls), l.Value)
-		} else if !model.LabelName(l.Name).IsValid() {
+		}
+		var ok bool
+		if l.Name, ok = SanitizeLabelName(l.Name); !ok {
```
simonswine, line 131: The duplicates error message might be a bit confusing, let's say.

With input `hello_world=og` and `hello.world=the-one-with-dots`, you would get something along the lines of:

```
invalid_argument: profile with labels '{..., hello_world="the-one-with-dots", hello_world="og", ...}' has duplicate label name: 'hello_world'
```

Maybe we want to show original `phlaremodel.LabelPairsString(ls)` before modification.

simonswine (approved):
Thanks for fixing into this. LGTM

A small concern not necessarily blocking, mabye we slightly adapt the output, or that we modified label names with `.` to `_`, if we did. Just wonder about your opinion.

PR #3338, opened 2024-06-05
Fix error handling in filterProfiles

simonswine:
We seem to ignore errors during the filterProfiles deduplication, at the stage when we received the decisions if something should be kept or not.


kolesnikovae (approved):
LGTM

Nice catch!

PR #3339, opened 2024-06-05
Update outdated java span profile information

simonswine:
I had been going through those docs with one of our customers, I think the code blocks are very confusing and I think settings should be set using environment variables.

I do hope I did not misunderstand anything.

### DIFF
--- a/docs/sources/configure-client/trace-span-profiles/java-span-profiles.md
+++ b/docs/sources/configure-client/trace-span-profiles/java-span-profiles.md
@@ -52,7 +52,7 @@ EXPOSE 5000
 CMD ["java", "-Dserver.port=5000", "-jar", "./my-app.jar" ]
 ```
 
-Now by adding the OTEL Java agent and our Pyroscope Otel Java Agent extensions, you can enrich your profiles with span IDs which makes it possible to query for span-specific profiling data in Grafana Tempo UI:
+By adding the OTel Java agent and the Pyroscope OTel Java Agent extensions, you can enrich your profiles with span IDs. This makes it possible to query for span-specific profiling data in Grafana Tempo:
 
 ```Dockerfile
 # [...]
--- a/docs/sources/configure-client/trace-span-profiles/java-span-profiles.md
+++ b/docs/sources/configure-client/trace-span-profiles/java-span-profiles.md
@@ -42,7 +42,7 @@ Your applications must be instrumented for profiling and tracing before you can 
 
 To start collecting Span Profiles for your Java application, you need to include [otel-profiling-java](https://github.com/pyroscope-io/otel-profiling-java) as an extension to your application.
 
-Let's assume you have this sample application docker image:
+Assuming you have this sample application Docker image:
 
 ```Dockerfile
 # [...]
--- a/docs/sources/configure-client/trace-span-profiles/java-span-profiles.md
+++ b/docs/sources/configure-client/trace-span-profiles/java-span-profiles.md
@@ -40,49 +40,52 @@ Your applications must be instrumented for profiling and tracing before you can 
 
 ## Configure the otel-profiling-java package
 
-To start collecting Span Profiles for your Java application, you need to include [otel-profiling-java](https://github.com/pyroscope-io/otel-profiling-java) in your code.
+To start collecting Span Profiles for your Java application, you need to include [otel-profiling-java](https://github.com/pyroscope-io/otel-profiling-java) as an extension to your application.
 
-This package is a `TracerProvider` implementation, that labels profiling data with span IDs which makes it possible to query for span-specific profiling data in Grafana Tempo UI.
+Let's assume you have this sample application docker image:
 
-```shell
-java -jar ./build/libs/rideshare-1.0-SNAPSHOT.jar \
-    -javaagent:./opentelemetry-javaagent.jar \
-    -Dotel.javaagent.extensions=./pyroscope-otel.jar \
-    -Dotel.pyroscope.start.profiling=true \
-    -Dpyroscope.application.name=ride-sharing-app-java-instrumentation  \
-    -Dpyroscope.format=jfr \
-    -Dpyroscope.profiler.event=itimer \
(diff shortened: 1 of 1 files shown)


docs/sources/configure-client/trace-span-profiles/java-span-profiles.md
```diff
@@ -40,49 +40,52 @@ Your applications must be instrumented for profiling and tracing before you can
...
-    span.end();
-}
+## Those environment variables need to be overwritten at runtime, if you are using Grafana Cloud 
+ENV PYROSCOPE_SERVER_ADDRESS=http://localhost:4040
+# ENV PYROSCOPE_BASIC_AUTH_USER=123     ## Grafana Cloud Username
+# ENV PYROSCOPE_BASIC_AUTH_PASSWORD=glc_secret ## Grafana Cloud Password / API Token
 
-// Your code goes here.
+## Add the pyroscope and the opentelemetry java-agents
+CMD ["java", "-Dserver.port=5000", "-javaagent:./opentelemetry-javaagent.jar", "-javaagent:pyroscope.jar", "-jar", "./my-app.jar" ]
```
kolesnikovae, line 88: I can remember there was a discussion on whether we need to `-javaagent:pyroscope.jar` in the case of OTEL extension, or if it is optional. @korniltsev could you please clarify this?

simonswine: Yes I remember the discussion but not the outcome. Also @aleks-p might know. I will also give it an e2e try.

docs/sources/configure-client/trace-span-profiles/java-span-profiles.md
```diff
@@ -40,49 +40,52 @@ Your applications must be instrumented for profiling and tracing before you can
...
 
+EXPOSE 5000
+
+CMD ["java", "-Dserver.port=5000", "-jar", "./my-app.jar" ]
 ```
 
-Next, you need to create and configure the tracer provider:
-```java
-implementation("io.pyroscope:otel:0.10.1.11")
+Now by adding the OTEL Java agent and our Pyroscope Otel Java Agent extensions, you can enrich your profiles with span IDs which makes it possible to query for span-specific profiling data in Grafana Tempo UI:
```
knylander-grafana, line 55: Some consistent suggestions. I removed the UI reference for Tempo since Tempo doesn't have a UI per se, not the way that Pyroscope does. 
```suggestion
By adding the OTel Java agent and the Pyroscope OTel Java Agent extensions, you can enrich your profiles with span IDs which makes it possible to query for span-specific profiling data in Grafana Tempo:
```

knylander-grafana (approved):
One small question, but otherwise, looks good. Thank you for updating the docs!

docs/sources/configure-client/trace-span-profiles/java-span-profiles.md
aleks-p: The removal of the second agent was attempted with #3169, but in my tests this was fine at startup but as soon as we invoke an endpoint we got an error. Could be worth re-testing anyway.

docs/sources/configure-client/trace-span-profiles/java-span-profiles.md
```diff
@@ -40,49 +40,52 @@ Your applications must be instrumented for profiling and tracing before you can
 
 ## Configure the otel-profiling-java package
 
-To start collecting Span Profiles for your Java application, you need to include [otel-profiling-java](https://github.com/pyroscope-io/otel-profiling-java) in your code.
+To start collecting Span Profiles for your Java application, you need to include [otel-profiling-java](https://github.com/pyroscope-io/otel-profiling-java) as an extension to your application.
 
-This package is a `TracerProvider` implementation, that labels profiling data with span IDs which makes it possible to query for span-specific profiling data in Grafana Tempo UI.
+Let's assume you have this sample application docker image:
```
knylander-grafana, line 45: Docker website uses a capital D when referring to a Docker image. Also, we [avoid using first person](https://grafana.com/docs/writers-toolkit/contribute/release-notes/) (i, me, we, us) in documentation, except for release notes. However, there are cases where things like "Let's" sounds better than rewording it. 
```suggestion
Let's assume you have this sample application Docker image:
```

docs/sources/configure-client/trace-span-profiles/java-span-profiles.md
simonswine: Yes verified, removing it causes this:

```
2024-06-21 13:42:18.421 ERROR 1 --- [nio-5000-exec-7] o.a.c.c.C.[.[.[/].[dispatcherServlet]    : Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Handler dispatch failed; nested exception is java.lang.UnsatisfiedLinkError: no asyncProfiler in java.library.path: /usr/java/packages/lib:/usr/lib64:/lib64:/lib:/usr/lib] with root cause

java.lang.UnsatisfiedLinkError: no asyncProfiler in java.library.path: /usr/java/packages/lib:/usr/lib64:/lib64:/lib:/usr/lib
        at java.base/java.lang.ClassLoader.loadLibrary(ClassLoader.java:2429) ~[na:na]
        at java.base/java.lang.Runtime.loadLibrary0(Runtime.java:818) ~[na:na]
        at java.base/java.lang.System.loadLibrary(System.java:1989) ~[na:na]
        at io.pyroscope.one.profiler.AsyncProfiler.getInstance(AsyncProfiler.java:50) ~[agent-0.13.1.jar!/:na]
        at io.pyroscope.one.profiler.AsyncProfiler.getInstance(AsyncProfiler.java:26) ~[agent-0.13.1.jar!/:na]
[...]

```

docs/sources/configure-client/trace-span-profiles/java-span-profiles.md
simonswine:
```suggestion
Assuming you have this sample application Docker image:
```

simonswine: This what I went with. Thanks for the feedback ❤️

Issue #3340, opened 2024-06-06
The sample count displayed is wrong


simonswine:
#### Describe the bug

Our currently displayed sample count in the UI is wrong. It seems to represent the sum of values multiplied by 10e9. In order to correct this we need to signal the sampling rate from the collection point (SDK/alloy/agent) and persist it in Pyroscope.

The OG Pyroscope endpoint has this metadata, which is unused by the adapter:

https://github.com/grafana/pyroscope/blob/ff0426278f16120b951cb4466b103349b0097c9d/pkg/og/ingestion/ingestion.go#L50

We then need to calculate the average sample rate for the profiles in a flamegraph (sample rate is a flamegraph wide parameter in flamebearer) and update our read path accordingly.

#### To Reproduce
Steps to reproduce:

1. Query e.g. a CPU profile:

<img width="405" alt="image" src="https://github.com/grafana/pyroscope/assets/223048/0896ce5a-b779-468b-82ee-d8f931bfcde9">

Both the value in the tool tip and and the total value, are wrong and just derived by 

#### Expected behavior

This is the hard part to define the right count of samples in a flameGraph.

For time unit profiles: 

* Sum of values / sampleRate, while the sampleRate is only valid per profile

For memory/object unit profiles:

* Unsure what we can do here, go for example scales up the sampled in-use space to represent the actual memory usage

#### Next steps

Variant A:

- [ ] Update the default value for CPU to be more correct for defaults: 

Variant B:

- [ ] Persist sample rate per profile
- [ ] Adapt query path to use avg. sampleRate for profiles in a flamegraph
- [ ] Adapt query path display indivual sample counts per flamegraph box

zaaklhm:
Same issue:

<img width="657" alt="Image" src="https://github.com/user-attachments/assets/3070799f-7250-4203-8e8c-a59424556dbe" />

PR #3342, opened 2024-06-06
Use default sample rate of 100 per second for CPU

simonswine:
This fixes the default sample rate for profiling and clarfies the field value. Relates to #3340.

### DIFF
--- a/pkg/og/structs/flamebearer/flamebearer.go
+++ b/pkg/og/structs/flamebearer/flamebearer.go
@@ -82,7 +82,7 @@ type FlamebearerMetadataV1 struct {
 	Format string `json:"format"`
 	// Name of the spy / profiler used to generate the profile, if any.
 	SpyName string `json:"spyName"`
-	// Sample rate at which the profiler was operating.
+	// Sample rate at which the profiler was operating. Note: This unit is not related to the Units field and is measured in samples per second.
 	SampleRate uint32 `json:"sampleRate"`
 	// The unit of measurement for the profiled data.
 	Units metadata.Units `json:"units"`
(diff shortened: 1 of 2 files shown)


pkg/og/structs/flamebearer/flamebearer.go
```diff
@@ -82,7 +82,7 @@ type FlamebearerMetadataV1 struct {
 	Format string `json:"format"`
 	// Name of the spy / profiler used to generate the profile, if any.
 	SpyName string `json:"spyName"`
-	// Sample rate at which the profiler was operating.
+	// Sample rate at which the profiler was operating. Note: This unit is not related to the Units field and is measured in samples per second.
 	SampleRate uint32 `json:"sampleRate"`
```
simonswine, line 86: I am not too sure if this is the right type and definition. I think it would be better if we would signal the sampleRate per SampleUnit. But for this we would need to change the field to be a float.

The default `100 Hz`, would become `0.00001 per nanosecond`. Any insights that you have on this @korniltsev

aleks-p:
This seems to revert #2202 which I don't fully understand, do you have more context @korniltsev-grafanista ?

korniltsev-grafanista:
I don't think its a revert - it was `100_000_000` then `1_000_000_000`  with #2202 and now 100 with #3342

The fix in https://github.com/grafana/pyroscope/pull/2202 "fixed" the duration formatting for `process_cpu:cpu:nanoseconds:cpu:nanoseconds` sample type.
Before the fix, the duration was 10 times bigger.

I look at the js code and this is how the formatter used to work back then
```
    const n = samples / sampleRate / this.divider;
    let nStr = n.toFixed(2);
```
which will accept for example 60_00_000_000 as the samples and `sampleRate` value is 1 second in nanoseconds and the divier is 60 - to get minutes, so the `nStr` is 1.00

I admit I don't fully understand the frontend code, not now, not back then when I "fixed" #2202 

I also don't know how the current formatter works

korniltsev-grafanista:
Screenshot_2025-10-20_16-06-19
Screenshot_2025-10-20_16-09-02

aleks-p:
Closing this as it hasn't been merged in 2 years, if it is still valid and needed - feel free to reopen.

PR #3344, opened 2024-06-06
Fix truncated Go CPU profiles

simonswine:
This extends the truncation fixing of heap profiles to also cover CPU profiles.

I am also adding an mimir-querier profile, but that seems to be not fixable, as it is doens't contain the connecting stacks.

image


pkg/pprof/fix_go_truncated.go
```diff
@@ -23,11 +23,20 @@ const (
...
+		minDepth = 64
+	} else {
 		return false
 	}
-	// Some truncated stacks have depth less than the depth limit (32).
-	const minDepth = 28
+
+	// Some truncated stacks have depth less than the depth limit.
+	minDepth -= 3
+
```
kolesnikovae, line 39: I'm wondering why we set it to 3. It was set to max-4 because in Go mem profiler they do cut 4 frames:
https://github.com/golang/go/blob/f7c330eac7777612574d8a1652fd415391f6095e/src/runtime/mprof.go#L446

pkg/pprof/fix_go_truncated.go
```diff
@@ -23,11 +23,20 @@ const (
...
 func MayHaveGoHeapTruncatedStacktraces(p *profilev1.Profile) bool {
-	if !hasGoHeapSampleTypes(p) {
+	var minDepth int
+
+	if hasGoHeapSampleTypes(p) {
+		// TODO: Go heap profiles in Go 1.23 have a depth limit of 128 frames. At some point we can probably increase/remove this
+		minDepth = 32
+	} else if hasGoCPUSampleTypes(p) {
+		minDepth = 64
+	} else {
```
kolesnikovae, line 33: Should we do it now? :) The function could report `false`, if the profile contains at least one stack trace strictly longer/taller that the limit by some margin (not accounting for skipped frames, probably – not sure). Like, we do not want to try to repair a heap profile, if it has a stack trace with 33 frames

WDYT?

pkg/pprof/fix_go_truncated.go
```diff
@@ -23,11 +23,20 @@ const (
 // MayHaveGoHeapTruncatedStacktraces reports whether there are
 // any chances that the profile may have truncated stack traces.
 func MayHaveGoHeapTruncatedStacktraces(p *profilev1.Profile) bool {
-	if !hasGoHeapSampleTypes(p) {
+	var minDepth int
```
kolesnikovae, line 26: Can't make a suggestion but we should rename the function:
```go
func MayHaveGoTruncatedStacktraces(p *profilev1.Profile) bool {
```

kolesnikovae:
Found a good sample – CloudWatch exporter. AWS Smithy Go code has always been very "deep" because of the countless middlewares, callbacks, hooks, etc.

image

[flamegraph_2024-06-07_1220-to-2024-06-07_1320.pb.gz](https://github.com/user-attachments/files/15710566/flamegraph_2024-06-07_1220-to-2024-06-07_1320.pb.gz)

kolesnikovae:
There's one thing that may require an adjustment: a very conservative recursion check that may cause a situation where part of the profiles are not repaired, which makes the whole idea somewhat ineffective

https://github.com/grafana/pyroscope/blob/4837be642ac3521fa30a9faaf7bc7f4a357f853d/pkg/pprof/fix_go_heap_truncated.go#L171-L174

I suggest that we double it regardless of the profile/sample type to relax the restriction

pkg/pprof/fix_go_truncated.go
simonswine: Yeah sorry this was me testing something and accidentally comitting it.

pkg/pprof/fix_go_truncated.go
simonswine: Maybe `PotentialTruncatedGoStacktraces`?

simonswine:
> RE: connecting / overlapping frames – the easiest way to examine truncated stack traces I found is to switch to the ?
> sandwich view, where the "ladder" at the top indicates presence of the issue (this is mimir querier in the screenshot, btw)

My strategy is using pprof's new sandwich view, it just takes a bit more than doing it in Pyroscope itself.

> There's one thing that may require an adjustment: a very conservative recursion check that may cause a situation where part of the profiles are not repaired, which makes the whole idea somewhat ineffective

https://github.com/grafana/pyroscope/blob/4837be642ac3521fa30a9faaf7bc7f4a357f853d/pkg/pprof/fix_go_heap_truncated.go#L171-L174

The first time I see a difference in cloud-watch-exporter's profiles is if I am using values fairly high 56 or higher.

kolesnikovae:
> The first time I see a difference in cloud-watch-exporter's profiles is if I am using values fairly high 56 or higher.

Yeah, this is more like a safety check. I also haven't seen too many examples. I'm suggesting increasing it because I see that the repair does not happen in some cases where it should have worked; and this limit is the only meaningful explanation I can find. Actually, I added this limit after testing it in one of our internal deployments, due to the observed CPU burn

kolesnikovae (approved):
LGTM 🚀

PR #3346, opened 2024-06-07
Omit replicas for components with HPA enabled

lpetrazickisupgrade:
ArgoCD cannot handle a disagreement in replica count between a Deployment/Statefulset definition and HorizontalPodAutoscaler definition. When an HPA is enabled, it is better to omit the replicas field on the Deployment/StatefulSet and let the HPA manage it instead


simonswine (approved):
LGTM! Thanks for contributing that makes perfect sense. If you would like a release with this change, also increment the version: https://github.com/grafana/pyroscope/blob/4837be642ac3521fa30a9faaf7bc7f4a357f853d/operations/pyroscope/helm/pyroscope/Chart.yaml#L5

PR #3349, opened 2024-06-10
perf: eliminate flamegraph merge

kolesnikovae:
It has been observed that in a distributed scenario with high parallelism of sub-range queries, the query frontend quickly becomes a bottleneck. In the screenshot below, you can see that approximately 5.3 seconds out of a total of 7 were spent on flame graph merges:

image

The problem is that all sub-range query results are merged synchronously in a single thread. A proper solution to this problem would be enabling DAG query execution, so that merges are distributed among queriers, which is planned for the near future.

However, this process could be vastly optimized by using trees instead of flame graphs (which are particularly inefficient to merge) in the query-frontend to queries requests.


examples/language-sdk-instrumentation/golang-push/simple/main.go
```diff
@@ -25,7 +25,7 @@ func fastFunction(c context.Context) {
 
 func slowFunction(c context.Context) {
 	// standard pprof.Do wrappers work as well
-	pprof.Do(c, pprof.Labels("function", "slow"), func(c context.Context) {
+	pprof.Do(c, pprof.Labels("function", "slow", "A_label", "a_value"), func(c context.Context) {
```
kolesnikovae, line 28:
```suggestion
	pprof.Do(c, pprof.Labels("function", "slow"), func(c context.Context) {
```

korniltsev (approved):
lgtm

PR #3345, opened 2024-06-07
feat: optimize series order

kolesnikovae:
Series order determines the physical placement of profiling data in blocks. It has been observed that the standard prometheus label and series order is not aligned with our access patterns, leading to an increase in the read amplification factor and query latency.

The PR proposes an option to enforce the optimized order. The main access pattern is a query that targets a specific profile/sample type of a specific service, therefore we want all the profiles matching such query to be stored sequentially. To achieve this, we ensure that the first labels of any series are `__profile_type__` and `__service_name__`.

The change should be applied with great caution. Since we can only enforce the order in ingesters (`__profile_type__` is not available in distributors), we need to ensure that all ingester instances either apply the optimization for a given profile, or not. Otherwise, there's possibility that the profile replicas will be handled differently, and the same series will have two representations: one with the standard order, and another one optimized; this will make it challenging or impossible to resolve the conflicts at query time and compaction. This is solved by "marking" profiles with the new private label `__order__` in distributors, which indicates whether we need to enforce the new order, and is removed before ingestion. This will result in doubling the active series temporarily but will allow for the normal deduplication process.

N.B.: No changes are required in the query path, because the order does not affect "user" labels. For the `Series` call that lists profile types and services, the order does not change.

Thus, the change should be applied in two steps:
 1. Rollout the new version: all ingester instances _must_ run the new version.
 2. Enable the `enforce_labels_order` limit.

Eventually, this optimization should become the default behaviour.

The optimization is only recommended if you are using user-defined labels (static or dynamic) that begin with `_` or with a capital letter and are experiencing very poor query performance.

PR #3352, opened 2024-06-11
fix: group by with enforced labels order

kolesnikovae:
The recently added series order optimization (#3345) has a bug, leading to invalid `group by` results. The cause is that `BytesWithLabels` assumed the standard order. The requirement has been relaxed: this makes the function less efficient, but not to the extent that it is a concern.

Also cleaned up the package a bit.


simonswine (approved):
LGTM. Good to catch this before a release.

PR #3356, opened 2024-06-14
fix: infinite loop in index writer

kolesnikovae:
Manual backporting of https://github.com/prometheus/prometheus/pull/12505. Thanks @pracucci!

I only wonder why it took so long to surface.

### DIFF
--- a/pkg/model/labels_test.go
+++ b/pkg/model/labels_test.go
@@ -68,6 +68,42 @@ func TestLabelsUnique(t *testing.T) {
 	}
 }
 
+func Test_LabelsBuilder_Unique(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    Labels
+		add      Labels
+		expected Labels
+	}{
+		{
+			name: "duplicates in input are preserved",
+			input: Labels{
+				{Name: "unique", Value: "yes"},
+				{Name: "unique", Value: "no"},
+			},
+			add: Labels{
+				{Name: "foo", Value: "bar"},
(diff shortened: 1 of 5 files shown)


pkg/model/labels_test.go
```diff
@@ -68,6 +68,42 @@ func TestLabelsUnique(t *testing.T) {
...
+		t.Run(test.name, func(t *testing.T) {
+			builder := NewLabelsBuilder(test.input)
+			for _, l := range test.add {
+				builder.Set(l.Name, l.Value)
+			}
+			assert.Equal(t, test.expected, builder.Labels())
+		})
+	}
+}
+
```
kolesnikovae, line 106: I do not see any practical sense in fixing `LabelsBuilder` we borrowed from prometheus. Instead, in our case, a reusable map would work way better (and I'm 99.(9)% sure, more efficiently).

To be addressed in a separate PR.

image

pkg/phlaredb/labels/labels.go
```diff
@@ -60,6 +60,7 @@ func CreateProfileLabels(enforceOrder bool, p *profilev1.Profile, externalLabels
 		} else {
 			sort.Sort(lbs)
 		}
+		lbs = lbs.Unique()
```
kolesnikovae, line 63: Just in case: `Unique` is order-agnostic

kolesnikovae:
I'm going to add an e2e test

korniltsev (approved):

PR #3360, opened 2024-06-19
feat: pprof export for Go PGO

kolesnikovae:
The PR adds a method to handle pprof queries for Go PGO purposes. The output file is up to 10-100 times smaller in size compared to the full pprof profile (without truncation). Based on [the discussion](https://groups.google.com/g/golang-nuts/c/ST4CyqOyBgk) and [the snippet](https://go-review.googlesource.com/c/tools/+/449500/1/trim/trim.go#67) by Michael Pratt (Go team).

The PR also extends `profilecli`: users may query the profile by running:
```
profilecli query go-pgo --query='{service_name="my-service"}'
```

I'm going to experiment with it a little more before merging. In the meantime, I'm curious what you think about this.


simonswine (approved):
LGTM

kolesnikovae:
I compared the number of cases using the PGO debug (`-gcflags="-d=pgodebug=2"`). The results are identical for both trimmed and full profiles. So I think it should be used instead of the export via UI

docs/sources/view-and-analyze-profile-data/profile-cli.md
```diff
@@ -271,3 +271,37 @@ By default it looks for samples within the last hour, though this can be control
        115366240: 107 13 14 15 16 17 1 2 3
      ...
      ```
+
+### Exporting a profile for Go PGO
+
+You can use the `profilecli query go-pgo` command to retrieve an aggregated profile from a Pyroscope server for use with Go PGO.
+Profiles retrieved with `profilecli query merge` include all samples found in the profile store, resulting in a large profile size, which may cause issues with network transfer and slow down the PGO process.
```
knylander-grafana, line 278:
```suggestion
Profiles retrieved with `profilecli query merge` include all samples found in the profile store, resulting in a large profile size.
The profile size may cause issues with network transfer and slow down the PGO process.
```

docs/sources/view-and-analyze-profile-data/profile-cli.md
```diff
@@ -271,3 +271,37 @@ By default it looks for samples within the last hour, though this can be control
...
+In contrast, profiles retrieved with `profilecli query go-pgo` include only the information used in Go PGO, making them significantly smaller and more efficient to handle.
+By default, it looks for samples within the last hour, though this can be controlled with the `--from` and `--to` flags. The source data can be narrowed down with the `--query` flag in the same way as with the `query` command.
+
+1. Specify optional flags.
+
+    - You can provide a label selector using the `--query` flag, for example, `--query='{service_name="my_application_name"}'`.
```
knylander-grafana, line 284: We should check if the four spaces renders correctly in the preview build.

kolesnikovae: I confirm it's rendered okay

simonswine:
@kolesnikovae I think we should also expose this via the UI. Ideally I would like to have every feature in the UI and only where it makes sense provide support for `profilecli`. Do you mind creating an issue in the plugin repo?

kolesnikovae:
Yeah, that makes sense: https://github.com/grafana/explore-profiles/issues/457. I'm not sure what the UI should look like, though.

I think this specific feature will be primarily used via CLI since it's supposed to be part of the build pipeline.

PR #3362, opened 2024-06-19
UI: Only mount UI index at known paths

simonswine:
This will ensure that the backend reports 404s for wrong paths.


kolesnikovae (approved):
LGTM

PR #3363, opened 2024-06-19
Fix frontend header handling

simonswine:
The query-frontend accidentally leaked content-type and content-encoding
information from the underlying connectgrpc implementation, which is
used to schedule queries on to the workers (queriers).

This will filter those out, so they won't be passed on through the
frontend.

This could be observed by errors like this:

```
$ profilecli query series
level=info msg="query series from querier" url=http://localhost:8080 from=2024-06-18T15:46:19.932471+01:00 to=2024-06-18T16:46:19.932472+01:00 labelNames=[]
error: failed to query: unknown: invalid content-type: "application/grpc,application/proto"; expecting "application/grpc"
exit status 1
```

Which themselves were produced only because envoy (involved on our GC
ingress), converted headers from this:

```
content-type: application/grpc
content-type: application/proto
```

to

```
content-type: application/grpc,application/proto
```

This then triggered a validation recently added to connect-go:

https://github.com/connectrpc/connect-go/pull/679


aleks-p (approved):
Looks good to me. Thanks for adding the test as well!

PR #3368, opened 2024-06-19
feat(profilecli): Allow to use different protocols

simonswine:
This allows to switch out the connect protocol for either json or grpc-web.


aleks-p:
Looks good. Do we want the upload command to also respect the selected protocol?

simonswine:
> Looks good. Do we want the upload command to also respect the selected protocol?

Good catch see 7fe0410

aleks-p (approved):

PR #3369, opened 2024-06-19
feat(distributor): Relabel profiles at ingest

simonswine:
This is an important operational tool: It allows us to deal with problems in the ingested profile labels.

The relabeling rules cover both series label and also sample labes.

By default we supply relabeling rules, which do remove the `godeltaprof_` prefix from profiles generated by pyroscope-go.

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -39,6 +39,7 @@ import (
 	"github.com/grafana/pyroscope/pkg/distributor/aggregator"
 	distributormodel "github.com/grafana/pyroscope/pkg/distributor/model"
 	phlaremodel "github.com/grafana/pyroscope/pkg/model"
+	"github.com/grafana/pyroscope/pkg/model/relabel"
 	"github.com/grafana/pyroscope/pkg/pprof"
 	"github.com/grafana/pyroscope/pkg/slices"
 	"github.com/grafana/pyroscope/pkg/tenant"
@@ -123,6 +124,7 @@ type Limits interface {
 	MaxProfileSymbolValueLength(tenantID string) int
 	MaxSessionsPerSeries(tenantID string) int
 	EnforceLabelsOrder(tenantID string) bool
+	IngestionRelabelingRules(tenantID string) []*relabel.Config
 	validation.ProfileValidationLimits
 	aggregator.Limits
 }
@@ -399,8 +401,11 @@ func (d *Distributor) sendRequests(ctx context.Context, req *distributormodel.Pu
 		series.Labels = d.limitMaxSessionsPerSeries(maxSessionsPerSeries, series.Labels)
 	}
 
-	// Next we split profiles by labels.
-	profileSeries := extractSampleSeries(req)
+	// Next we split profiles by labels and apply relabel rules.
+	profileSeries, bytesRelabelDropped, profilesRelabelDropped := extractSampleSeries(req, d.limits.IngestionRelabelingRules(tenantID))
+	validation.DiscardedBytes.WithLabelValues(string(validation.RelabelRules), tenantID).Add(bytesRelabelDropped)
+	validation.DiscardedProfiles.WithLabelValues(string(validation.RelabelRules), tenantID).Add(profilesRelabelDropped)
+
 	// Filter our series and profiles without samples.
 	for _, series := range profileSeries {
 		series.Samples = slices.RemoveInPlace(series.Samples, func(sample *distributormodel.ProfileSample, _ int) bool {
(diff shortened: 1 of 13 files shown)


pkg/distributor/distributor.go
```diff
@@ -401,6 +403,20 @@ func (d *Distributor) sendRequests(ctx context.Context, req *distributormodel.Pu
 
 	// Next we split profiles by labels.
 	profileSeries := extractSampleSeries(req)
+
+	// Apply relabeling rules to the series.
+	relabelingRules := d.limits.IngestionRelabelingRules(tenantID)
+	builder := phlaremodel.NewLabelsBuilder(phlaremodel.EmptyLabels())
+	for _, series := range profileSeries {
```
kolesnikovae, line 410: I think there's a problem with relabeling after `extractSampleSeries`: say, we have two series formed by pprof labels:
```
{foo=bar, baz=qux}
{foo=bar, baz=zoo}
```

Then, we drop the `baz` label:
```
{foo=bar}
{foo=bar}
```

The problem is that both profiles have the same timestamp (and even profile ID, IIRC), which means that one will "replace" another at ingestion (or will be deduplicated at reads). Probably same applies to the explicit series labels.

A simple solution for this would be merging such "matching" profiles, but merges are relatively expensive. It would be great if we could do all the relabeling stuff during the series extraction from pprof samples.

Also, not sure if it's feasible (rules must be idempotent) but I'd consider first applying them on the explicit series labels

pkg/distributor/distributor.go
```diff
@@ -633,7 +638,46 @@ func (d *Distributor) HealthyInstancesCount() int {
...
+
+func (g *groupsWithFingerprints) add(lbls phlaremodel.Labels, group pprof.SampleGroup) {
+	fp := lbls.Hash()
+	idxs, ok := g.fingerprints[fp]
+	if ok {
+		// fingerprint matches, check if the labels arethe same
+		for _, idx := range idxs {
+			if phlaremodel.CompareLabelPairs(g.labelSets[idx], lbls) == 0 {
+				// append samples to the group
+				g.groups[idx].Samples = append(g.groups[idx].Samples, group.Samples...)
```
kolesnikovae, line 663: I'm afraid this may result in samples with identical stack traces, which is not expected in further processing (including the ingestion process). I can't foresee all the consequences, and the issue might be very obscure. I think it's safer to deduplicate them.

```go
type lazyGroup struct {
  sampleGroup pprof.SampleGroup
  // The map is only initialized when the group is being modified.
  // Key is the string representation (unsafe) of the sample stack trace.
  sampleMap map[string]*pprof.Sample
  labels    phlaremodel.Labels
}

// For reference (defined in pkg/pprof)
type SampleGroup struct {
	Labels  []*profilev1.Label
	Samples []*profilev1.Sample
}
```

Then, `groupsWithFingerprints` could be simplified to:
```go
type groupByFingerprint map[uint64][]*lazyGroup
```
That would be easier to read and (likely) a tiny bit faster

pkg/distributor/distributor.go
```diff
@@ -643,24 +687,64 @@ func extractSampleSeries(req *distributormodel.PushRequest) []*distributormodel.
...
+			// iterate through groups relabel them and find relevant overlapping labelsets
+			groupsKept := newGroupsWithFingerprints(len(groups))
 			for _, group := range groups {
+				lblbuilder.Reset(series.Labels)
+				addSampleLabelsToLabelsBuilder(lblbuilder, raw.Profile.Profile, group.Labels)
+				if len(relabelRules) > 0 {
+					keep := relabel.ProcessBuilder(lblbuilder, relabelRules...)
+					if !keep {
+						// not too sure if this is worth doing, as it might be quite expensive
+						bytesRelabelDropped += float64(exportSamples(e, group.Samples).SizeVT())
```
kolesnikovae, line 725: Agreed, it's absolutely not worth it. I think it's sufficient to account for only the size of samples. exportSamples will pull referenced symbols, which will result in over-accounting due to overlaps between groups.

pkg/distributor/distributor.go
```diff
@@ -633,7 +638,46 @@ func (d *Distributor) HealthyInstancesCount() int {
...
+	// add the labels to the list
+	g.fingerprints[fp] = append(g.fingerprints[fp], len(g.labelSets))
+	g.labelSets = append(g.labelSets, lbls)
+	g.groups = append(g.groups, group)
+
+}
+
+func extractSampleSeries(req *distributormodel.PushRequest, relabelRules []*relabel.Config) (result []*distributormodel.ProfileSeries, bytesRelabelDropped, profilesRelabelDropped float64) {
+	var (
+		lblbuilder = phlaremodel.NewLabelsBuilder(phlaremodel.EmptyLabels())
```
kolesnikovae, line 678: I recently learned the `LabelBuilder` does not always guarantee uniqueness the hard way: https://github.com/grafana/pyroscope/pull/3356#discussion_r1639478243. Its performance also raises some concerns. It might be absolutely irrelevant, mentioning this just in case.

pkg/validation/relabeling_test.go
```diff
@@ -0,0 +1,147 @@
...
+      - action: drop
+  custom-rule-start:
+    ingestion_relabeling_default_rules_position: last
+    ingestion_relabeling_rules:
+      - action: drop
+  custom-rule-only:
+    ingestion_relabeling_default_rules_position: disabled
+    ingestion_relabeling_rules:
+      - action: drop
+
```
kolesnikovae, line 50: Ah, it will take some time to get used to this notation

kolesnikovae (commented):
Nice job! I added a few comments

pkg/distributor/distributor.go
simonswine: For `groupByFingerprint` i also added a slice to keep the order intact.

See edb981689

pkg/distributor/distributor.go
simonswine: edb981689

With this there might be some overaccounting for interned strings.

pkg/validation/relabeling_test.go
kolesnikovae: Yeah, what I meant is overrides within overrides. Although we don't have to use it, a simple plain order will work well for us

kolesnikovae (approved):
LGTM

PR #3371, opened 2024-06-20
fix: remove unused stacktracesPartition.header field

korniltsev:
The field seems to be unused


kolesnikovae (approved):
LGTM: it's a leftover

PR #3383, opened 2024-06-24
chore: remove Persister[].SortingColumns method

korniltsev:
- remove `Persister[].SortingColumns` which is ued only in tests
- remove index argument from `Desconstruct`
- remove index return value from `Reconstruct`
- remove sorting from `ReaderWriter` test helper 

The PR does not change behavior or data, just tiny cleanup


aleks-p (approved):
LGTM

Issue #3192, opened 2024-04-10
Prototype ingestor-less replication-less architecture


petethepig:
#### Is your feature request related to a problem? Please describe.

Currently at large scale Pyroscope struggles with a few issues:
* compaction
  * is using a lot of resources, particularly ram
    * this is mostly because of the need for deduplication of data, which is caused by replication
* replication (x3)
  * uses a lot of resources (cup, disk, ram)
  * is hard to maintain
  * leads to complex bugs
  * reduces read performance due to the need for deduplication of data

#### Describe the solution you'd like

We could do the following:
* remove ingesters
* make it so that distributors create small blocks in memory and flush them to object storage
* remove deduplication code from read path and compaction
* tweak compaction to work with increased number of blocks
* tweak read path to work with increased number of blocks

#### Concerns / Risks

These are not blockers, but rather a list of things that might derail the project, so we should make sure we keep these concerns in mind and address these early:

* increased object storage write costs
* reduced query performance due to too many small blocks
* running into some unforeseen performance limitations of underlying object storage

#### Acceptance Criteria

* [ ] it should work with traffic in ops
* [ ] costs should not go up. it is fine if we have to exchange reduced ingesters costs with increased queriers cost though
* [ ] performance should not go down

#### Timeline / Staffing

Great news is that we already have all of the components built for this and so this project becomes a lot more about moving things around and tweaking the system rather than building new stuff.

I think if we split up the task we could get a good working prototype in about 4 weeks and 3 people. I think @kolesnikovae should lead this project.

Assuming the project succeeds we could then spend another 4 weeks doing migrations / further tweaking algorithms to work with the new system.

#### Outcome

The main thing is that the system becomes simpler. To elaborate on that, these changes would:
* improve maintainability of the system
* reduce toil
* significantly reduce tco
* improve read performance

#### Additional context

All credit for this idea should go to @kolesnikovae — I'm just trying to document the proposed solution. Also, this is somewhat of a meta-issue. I imagine a lot of other existing issues could be steps towards implementation of this project.

PR #3388, opened 2024-06-26
New architecture prototype

kolesnikovae:
The PR is the main development branch for the work on the new architecture (https://github.com/grafana/pyroscope/issues/3192). ~The PR will never be merged; it was created only for visibility. Once the prototype is finished, we will publish appropriate documentation based on the experiment results.~

---

Update:

The experiment is over. The results are very promising.

The new code is very well isolated and does not affect behaviour of the current version, therefore I'd like to merge the changes: we will continue working on the new components in the main branch.


pkg/metastore/metastore_compaction_queue.go
```diff
@@ -84,50 +78,45 @@ func (c *jobQueueEntry) load(job *compactionpb.CompactionJob) {
...
-		deadline: 0, // Deadline and token will be assigned
-		token:    0, // when the job is "dequeued" (transferred).
-		proto:    job,
+		status: js,
+		index:  0,
+		proto:  job,
 	}
+	// Deadline will be assigned when the job is "dequeued" (transferred).
+	job.RaftLogIndex = 0
+	job.LeaseExpiresAt = 0
```
kolesnikovae, line 87: This is only true for job in the initial state. If the job already has index we do want to restore this information

pkg/phlaredb/tsdb/index/index.go
```diff
@@ -279,14 +283,14 @@ type FileWriter struct {
 	name string
 }
 
-func NewFileWriter(name string) (*FileWriter, error) {
+func NewFileWriter(name string, bufferSize int) (*FileWriter, error) {
```
korniltsev, line 286: I will revert these changes in tsdb/index/index.go - it is not used anymore

api/metastore/v1/metastore.proto
```diff
@@ -0,0 +1,71 @@
+syntax = "proto3";
+
+package metastore.v1;
```
korniltsev, line 3: nit: maybe we should put it in v2 or v2-experimental package

pkg/phlaredb/tsdb/index/index.go
kolesnikovae: It's currently used in compactor. I see no need to change it now

api/metastore/v1/metastore.proto
kolesnikovae: This is the first version of metastore API.

As long as the API is not public, I see no practical benefits of separating the API

korniltsev (approved):
lgtm

simonswine (approved):
Thank you so much for cleaning that up and make it mergable it main @kolesnikovae .

Also kudos to everyone else working on this experiment 👏

PR #3389, opened 2024-06-26
Pyroscope: Add model for collection rules

simonswine:
This adds the message model for managing collection rules.

Like to get this merged ahead of time, so I can use it in external projects


simonswine:
Replaced by #3588

PR #3395, opened 2024-07-02
Reject queries without a time range

bryanhuhta:
In https://github.com/grafana/explore-profiles/pull/15, I ran across a bug where the UI was sending queries without a time range. It was a little tricky to track down because Pyroscope would respond with a success status code, but an empty payload. This made me wonder if there was a bug in Pyroscope's query logic.

I think we should classify queries with no time range an error. I don't know of any use cases in which we want to have a time range, but allow it to be optionally provided.

### DIFF
--- a/pkg/validation/validate.go
+++ b/pkg/validation/validate.go
@@ -329,7 +329,7 @@ type ValidatedRangeRequest struct {
 }
 
 func ValidateRangeRequest(limits RangeRequestLimits, tenantIDs []string, req model.Interval, now model.Time) (ValidatedRangeRequest, error) {
-	if req.Start == 0 && req.End == 0 {
+	if req.Start == 0 || req.End == 0 {
 		return ValidatedRangeRequest{}, NewErrorf(QueryMissingTimeRange, QueryMissingTimeRangeErrorMsg)
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/validation/validate.go
```diff
@@ -327,6 +329,10 @@ type ValidatedRangeRequest struct {
 }
 
 func ValidateRangeRequest(limits RangeRequestLimits, tenantIDs []string, req model.Interval, now model.Time) (ValidatedRangeRequest, error) {
+	if req.Start == 0 && req.End == 0 {
+		return ValidatedRangeRequest{}, NewErrorf(QueryMissingTimeRange, QueryMissingTimeRangeErrorMsg)
+	}
```
kolesnikovae, line 334: Should we require both `Start` and `End`?
```suggestion
	if req.Start == 0 || req.End == 0 {
		return ValidatedRangeRequest{}, NewErrorf(QueryMissingTimeRange, QueryMissingTimeRangeErrorMsg)
	}
```

kolesnikovae (approved):
LGTM. Thanks for looking into this!

> The Unix time 0 is exactly midnight UTC on 1 January 1970

Although 0 is a valid Unix time, I do think we should report an error in this case.

pkg/validation/validate.go
bryanhuhta: Yeah, you are right. I think this is a good suggestion 👍

PR #3398, opened 2024-07-02
fix: don't compute delta on relabeled godeltaprof_memory->memory profiles

korniltsev:


simonswine (approved):
LGTM

There is still a test, that requires updating otherwise good.

Suggest we backport this into weekly/f73

PR #3400, opened 2024-07-02
WIP: Distribute profiles based on tenant id, service name and labels

aleks-p:
- turns replication off
- splits incoming requests into `ingester_count * tokens_per_ingester` shards
- distributes requests based on tenant id, service name and labels
- passes the shard number to ingesters as part of the push request
- a single retry on failures between distributor and ingester

Verified to not break things in local / standalone mode. Tests in microservice mode pending.


pkg/distributor/distributor.go
```diff
@@ -481,16 +495,79 @@ func (d *Distributor) sendRequests(ctx context.Context, req *distributormodel.Pu
...
+		i = i + 1
+	}
+	if i >= len(instance.Tokens) {
+		i = 0
+	}
+	shard := uint32(instanceIdNum*len(instance.Tokens) + i)
+	level.Debug(logger).Log("msg", "shard found", "shard", shard)
+	return shard
+}
+
```
kolesnikovae, line 570: I'm wondering if we could get it via:
```
shard_id = hash(tenant_id+service_name) % (instances_total * shards_per_instance)
```

Shard placement does not have to match the ring distribution (maybe event should not)

kolesnikovae (approved):
Looks great!

It would be nice to have metrics that show how the data (size in bytes and count of requests/profiles) is distributed over the shards and ingesters

pkg/distributor/distributor.go
aleks-p: The current logic will result in more even block sizes when we have e.e., a single large tenant service because it'll be spread across multiple shards as opposed to a single one. We could solve that by spreading it to more ingesters though in the future, so I can simplify things.

kolesnikovae: Sorry, didn't mean to confuse you. I'm just thinking out loud: the current implementation is fine – we can rely on the instance names in the prototype. In real life, we want to have something more abstract, though.

You're right about shard distribution. I'm just wondering if this is bad: this allows us to scope the app to a single shard (which drastically simplifies deduplication), while distributing the load among ingesters. The block size is determined by the compaction strategy, so we have control over this. The method is bad for another reason: really bad remapping when adding or removing instances. There are workarounds, but I don't think they are any better than rendezvous hashing (HRW) or jump hashing – something we will experiment with later on.

BTW, you can also determine the shard based on the _first_ instance: `hash(instance) % shards_total` :) Although this is also prone to issues with remapping.

aleks-p: Will leave it as is (`hash(tenant_id+service_name) % shards_total`), we can provide alternative sharding strategies later on.

PR #3401, opened 2024-07-03
experiment: new architecture deployment

kolesnikovae:
The PR enables deployment of the experimental [experiment/new-architecture](https://github.com/grafana/pyroscope/tree/experiment/new-architecture) to a dev cluster. Note that I haven't disabled checks (lint, test, etc), so the deployment will only trigger if those are green. I suggest that we mark failing tests with `Skip()`, if they are irrelevant


simonswine (approved):
LGTM

PR #3403, opened 2024-07-03
chore: Update kubeVersion and branch to main

simonswine:
Small follow-up to #3332 

This fixes the kubeVersion to match the one from the makefile and compares to the correct branch.


aleks-p (approved):
LGTM

PR #3405, opened 2024-07-05
wip: segments writer

korniltsev:


pkg/ingester/ingester.go
```diff
@@ -161,110 +143,29 @@ func (i *Ingester) running(ctx context.Context) error {
...
-		return fn()
+		return nil, connect.NewError(connect.CodeInvalidArgument, err)
 	}
-	return nil
-}
 
-func (i *Ingester) Push(ctx context.Context, req *connect.Request[pushv1.PushRequest]) (*connect.Response[pushv1.PushResponse], error) {
-	return forInstanceUnary(ctx, i, func(instance *instance) (*connect.Response[pushv1.PushResponse], error) {
+	wait, err := i.segmentWriter.ingest(shardKey(0), func(segment segmentIngest) error {
 		for _, series := range req.Msg.Series {
```
aleks-p, line 157: I had a quick look at the shard key here and currently a single `pushv1.PushRequest` request can have multiple shard keys because they are passed at the `RawProfileSeries` level (for each `req.Msg.Series`). This is because the distributor groups profiles by ingester and sends a batch and in theory profiles in a batch could be assigned to different shards. 

We will need to consume into separate shards here and respond when all segments are flushed?

korniltsev: yep, will do

kolesnikovae (approved):
LGTM. Looking forward to deploying it

Issue #3406, opened 2024-07-05
symdb: Panic during ResolveStacktraceLocations


simonswine:
Internally at GL, we observed those panics on at least two ingesters that gotten querried:

```
{"from":"2024-07-05 08:22:16","to":"2024-07-05 08:53:12"}
```

```
  |   | 2024-07-05 08:47:07.580 |   | created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 223338 |  
  |   | 2024-07-05 08:47:07.580 |   | golang.org/x/sync@v0.7.0/errgroup/errgroup.go:78 +0x56 |  
  |   | 2024-07-05 08:47:07.580 |   | golang.org/x/sync/errgroup.(*Group).Go.func1() |  
  |   | 2024-07-05 08:47:07.580 |   | github.com/grafana/pyroscope/pkg/util/recovery.go:59 +0x58 |  
  |   | 2024-07-05 08:47:07.580 |   | github.com/grafana/pyroscope/pkg/phlaredb/symdb.(*Resolver).withSymbols.RecoverPanic.func2() |  
  |   | 2024-07-05 08:47:07.580 |   | github.com/grafana/pyroscope/pkg/phlaredb/symdb/resolver.go:279 +0x5e |  
  |   | 2024-07-05 08:47:07.580 |   | github.com/grafana/pyroscope/pkg/phlaredb/symdb.(*Resolver).withSymbols.func1() |  
  |   | 2024-07-05 08:47:07.580 |   | github.com/grafana/pyroscope/pkg/phlaredb/symdb/resolver.go:238 +0x45 |  
  |   | 2024-07-05 08:47:07.580 |   | github.com/grafana/pyroscope/pkg/phlaredb/symdb.(*Resolver).Tree.func1(0xc00409ba40?, 0x2f5f598?) |  
  |   | 2024-07-05 08:47:07.580 |   | github.com/grafana/pyroscope/pkg/phlaredb/symdb/resolver.go:299 |  
  |   | 2024-07-05 08:47:07.580 |   | github.com/grafana/pyroscope/pkg/phlaredb/symdb.(*Symbols).Tree(...) |  
  |   | 2024-07-05 08:47:07.580 |   | github.com/grafana/pyroscope/pkg/phlaredb/symdb/resolver_tree.go:40 +0x442 |  
[...]

Issue #3262, opened 2024-05-01
Improve symbols partitioning heuristics


kolesnikovae:
Pyroscope stores symbolic information and stack traces in a dedicated storage called symdb. The storage is partitioned by key, which is derived from the `service_name` label, or the main binary name, if it's present.

However, this approach leads to "shared" partitions, when we store symbols of multiple services together. In turn, this increases the read amplification factor; often, we fetch data that's not referenced by samples of a particular query. Analysis shows that more than 50% of the fetched symbols might be irrelevant to the query.

For example, consider the following series being ingested:
 * `cpu nanoseconds` `service_name="foo"` `main_binary="main"`
 * `heap allocs` `service_name="foo"` `main_binary="main"`
 * `heap inuse` `service_name="foo"` `main_binary="main"`
 * `cpu nanoseconds` `service_name="bar"` `main_binary="main"`

All them will be stored in the same partition, but in 99.9% cases, only one of them is accessed at query time.

Basically, there are two issues:
 - We do not take sample types into account. The problem is that different sample type categories often reference non-overlapping elements (especially stack traces).
 - We rely on the binary name. Ideally, there should be a way to distinguish binaries by build ID, but in most cases, we lack this information.

I propose to change the heuristic in the way that the partition key is built based on the service name and the sample type set the profile has. Thus, for example, `service="foo"` `cpu nanoseconds` won't be stored together with `heap allocs`, but `service="foo"` `heap allocs` will be stored together with `heap inuse`.

The caveat is that this will cause an increase in memory consumption in ingesters because some symbols will be duplicated among partitions. A notable case is when the same binary is run in multiple instances with different service_name label values (Pyroscope itself is an example). Either way, this duplication allows us to make the data more independent and decrease the read amplification factor.

kolesnikovae:
Another issue is that:

1. The main binary name (the name of the first mapping) might be malformed or non-deterministic.
2. We determine the partition after the pprof split, which may result in profiles of the same service being split across multiple partitions.

PR #3408, opened 2024-07-08
fix: honor stacktrace partitions at downsampling

kolesnikovae:
Fixes https://github.com/grafana/pyroscope/issues/3406.

Currently, during compaction, when the downsampler aggregates profiles, the stack trace partition is ignored. The initial assumption was that profiles of the same series (label set) would always have the same stack trace partition. This is almost always the case; however, this is not guaranteed as the partitioning heuristic also depends on the main mapping name (see [#3262](https://github.com/grafana/pyroscope/issues/3262)).

We've recently discovered a few cases where this led to data inconsistencies. Specifically, profiles that refer to the wrong partition, and therefore can't be read: their samples refer to stack traces that can't be found in the partition. Another issue is that such occurrences are not handled properly: a wrong stack trace may be resolved to wrong locations (the stack trace buffer is not cleared if the stack trace is not found), which will likely cause a panic when attempting to resolve it.

I also increased the symbol rewriter LRU cache size from 2 to 8, as I've seen that the interlacing of partitions might be quite heavy.

### DIFF
--- a/pkg/phlaredb/symdb/stacktrace_tree.go
+++ b/pkg/phlaredb/symdb/stacktrace_tree.go
@@ -153,10 +153,10 @@ func newParentPointerTree(size uint32) *parentPointerTree {
 }
 
 func (t *parentPointerTree) resolve(dst []int32, id uint32) []int32 {
+	dst = dst[:0]
 	if id >= uint32(len(t.nodes)) {
 		return dst
 	}
-	dst = dst[:0]
 	n := t.nodes[id]
 	for n.p >= 0 {
 		dst = append(dst, n.r)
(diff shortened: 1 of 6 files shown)


pkg/phlaredb/symdb/stacktrace_tree.go
```diff
@@ -153,10 +153,10 @@ func newParentPointerTree(size uint32) *parentPointerTree {
 }
 
 func (t *parentPointerTree) resolve(dst []int32, id uint32) []int32 {
+	dst = dst[:0]
```
simonswine, line 156: Apart from that It might also make sense to empty slices before putting them into the pool:

https://github.com/grafana/pyroscope/blob/b6816e31eb0c8418e74d9f7d57783fb983cdeb89/pkg/phlaredb/symdb/partition_memory.go#L205

simonswine (approved):
LGTM.

Wdyt should be cherry pick this into the weekly?

pkg/phlaredb/symdb/stacktrace_tree.go
kolesnikovae: I've been thinking about this, but in practice, it doesn't improve anything: the slice is reused many times before it is put back. So I decided to keep it as is. To be honest, the pool is barely helpful; I should have removed it

kolesnikovae:
> Wdyt should be cherry pick this into the weekly?

Yeah, I'll cherry pick it

PR #3413, opened 2024-07-11
fix: revert rewriter LRU cache size

kolesnikovae:
Increasing the LRU cache size from 2 to 8 turned out to be quite a problem for large deployments and might be causing OOMs, so I'm reverting it

See #3408


korniltsev (approved):

kolesnikovae:
After some testing on a large cluster I can't confirm that this solves the issue

PR #3415, opened 2024-07-11
chore: Build weeklies with debug information

simonswine:
This will allow to have more meaningful ways of debugging core dump and live processes in all environments.

I tested this quickly:

```
# Without any env we still strip the debug info
❯ goreleaser build --clean --snapshot --single-target
[...]
❯ dlv exec dist/profilecli_darwin_arm64/profilecli
could not launch process: decoding dwarf section info at offset 0x0: too short - debuggee must not be built with 'go run' or -ldflags='-s -w', which strip debug info

# With environment set
❯ GORELEASER_STRIP_DEBUG_INFO=false goreleaser build --clean --snapshot --single-target
[...]
❯ dlv exec dist/profilecli_darwin_arm64/profilecli
Type 'help' for list of commands.
(dlv)
```

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -3,7 +3,10 @@ version: 2
 before:
   hooks:
     # This hook ensures that goreleaser uses the correct go version for a Pyroscope release
-    - sh -euc "go version | grep "go version go1.21.11 " || { echo "Unexpected go version"; exit 1; }"
+    - sh -euc 'go version | grep "go version go1.21.11 " || { echo "Unexpected go version"; exit 1; }'
+env:
+  # Strip debug information from the binary by default, weekly builds will have debug information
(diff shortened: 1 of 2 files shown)


.goreleaser.yaml
```diff
@@ -3,7 +3,10 @@ version: 2
 before:
   hooks:
     # This hook ensures that goreleaser uses the correct go version for a Pyroscope release
-    - sh -euc "go version | grep "go version go1.21.11 " || { echo "Unexpected go version"; exit 1; }"
+    - sh -euc 'go version | grep "go version go1.21.11 " || { echo "Unexpected go version"; exit 1; }'
```
simonswine, line 6: This check actually never worked, because of broken quoting

korniltsev (approved):
I hope dwarf info is not loaded into memory in any way :crossed_fingers:

PR #3417, opened 2024-07-11
chore: Deploy tenant-settings and adhoc-profiles microservices

simonswine:
They were so far not exposed in the values.yaml

This also adds them to ingress object.


operations/pyroscope/helm/pyroscope/values-micro-services.yaml
```diff
@@ -84,5 +84,27 @@ pyroscope:
...
+      persistence:
+        enabled: false
+      resources:
+        limits:
+          memory: 4Gi
+        requests:
+          memory: 16Mi
+          cpu: 0.1
+    ad-hoc-profiles:
+      kind: StatefulSet
```
aleks-p, line 99: should these be deployments instead?

aleks-p (approved):

operations/pyroscope/helm/pyroscope/values-micro-services.yaml
simonswine: Yes you are right, 9b0e996f612bd939917c6d896058efe5597b9a8c

PR #3420, opened 2024-07-12
profiles parquet encoding: fix profile column count

korniltsev:
noticed last columns appends are allocating

image


aleks-p (approved):
That is a great find!

PR #3421, opened 2024-07-12
profileStore: add err check, return err if phlareparquet.CopyAsRowGroups returns err

korniltsev:
this likely never happened, but still


aleks-p (approved):
LGTM

PR #3422, opened 2024-07-12
profileStore: fix unlocking rowsLock

korniltsev:
There is an error check of err value from cutRowGroup call. If cutRowGroup returns err, the rowsLock seem to be never released


aleks-p (approved):
LGTM

PR #3427, opened 2024-07-16
WIP: Compaction planning in metastore

aleks-p:


pkg/metastore/metastore_state.go
```diff
@@ -114,3 +155,18 @@ func (s *metastoreShard) loadSegments(b *bbolt.Bucket) error {
...
+	p.jobsMutex.Lock()
+	defer p.jobsMutex.Unlock()
+	c := b.Cursor()
+	for k, v := c.First(); k != nil; k, v = c.Next() {
+		var job compactorv1.CompactionJob
+		if err := job.UnmarshalVT(v); err != nil {
+			return fmt.Errorf("failed to unmarshal job %q: %w", string(k), err)
+		}
+		p.jobsByName[job.Name] = &job
+		// TODO aleks: restoring from a snapshot will lose "partial" jobs
```
aleks-p, line 169: @kolesnikovae I now remember the full reason why the earlier iteration persisted partial jobs (queued blocks) to the DB.

It was for the case when we restore the state from a snapshot. The snapshot will contain committed blocks and compaction jobs, but not the blocks queued for compaction (future jobs). In this case, the `AddBlock` command will not be re-run in my understanding, resulting in blocks that won't get compacted.

I'll leave it as is for now, but would like to hear your thoughts.

kolesnikovae: This is basically how Raft works:

1. You have a state that mutates with commands.
2. All the commands go through the log that is being replicated. In fact, a command written locally might be applied on followers before it is applied on the leader.
3. At a certain point, when the log of commands is large, a snapshot of the state is taken.
4. Then, upon recovery, the state is restored from the snapshot, and the log entries added after the snapshot are applied on top of it.

Thus, we won't lose the state.

In our case, we only use boltdb to make snapshots more easily without locking all operations on the state. In practice, we could drop it altogether and serialize the snapshot of the state on our own.

aleks-p: My point is that at step 3, the snapshot of the state will not contain the compaction block queues. They are only stored in memory and the snapshot taking logic ignores them. Then they won't get restored in step 4. What am I missing?

kolesnikovae: I see your point now – yeah, everything we rely on should be persisted first. Queued blocks should be written to the DB as well (just the key, without values)

PR #3429, opened 2024-07-16
feat: Per-app usage metrics

bryanhuhta:
related https://github.com/grafana/pyroscope-squad/issues/162

This implements per-app usage for bytes ingested/dropped (both metrics required to calculate billable bytes). I took the following metrics

- `distributor_received_decompressed_bytes`
- `discarded_bytes_total`

and added the `service_name` label to them. This label value is typically blank, but when the tenant overrides have the following property set

```yaml
overrides:
  "1234":
    distributor_usage_groups:
      - service_a
      - service_b
```

any series which has a `service_name` label that matches one of `distributor_usage_groups` will have `service_name` set on the ingest/dropped metrics. This should allow us to create recording rules to send to the billing cortex and let customers view per-app usage breakdowns.

Note that there is a cost to adding an app to the allowlist. We should be prudent when adding services, especially for large customers.

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -126,6 +126,7 @@ type Limits interface {
 	MaxSessionsPerSeries(tenantID string) int
 	EnforceLabelsOrder(tenantID string) bool
 	IngestionRelabelingRules(tenantID string) []*relabel.Config
+	DistributorUsageGroups(tenantID string) *validation.TenantUsageGroups
 	validation.ProfileValidationLimits
 	aggregator.Limits
 }
@@ -284,9 +285,13 @@ func (d *Distributor) PushParsed(ctx context.Context, req *distributormodel.Push
 		return nil, err
 	}
 
+	usageGroups := d.limits.DistributorUsageGroups(tenantID)
+
 	for _, series := range req.Series {
 		profName := phlaremodel.Labels(series.Labels).Get(ProfileName)
+		serviceName := usageGroups.GetServiceName(phlaremodel.Labels(series.Labels))
 		profLanguage := d.GetProfileLanguage(series)
+
 		for _, raw := range series.Samples {
 			usagestats.NewCounter(fmt.Sprintf("distributor_profile_type_%s_received", profName)).Inc(1)
 			d.profileReceivedStats.Inc(1, profLanguage)
@@ -295,14 +300,14 @@ func (d *Distributor) PushParsed(ctx context.Context, req *distributormodel.Push
 			}
 			p := raw.Profile
 			decompressedSize := p.SizeVT()
-			d.metrics.receivedDecompressedBytes.WithLabelValues(profName, tenantID).Observe(float64(decompressedSize))
+			d.metrics.receivedDecompressedBytes.WithLabelValues(profName, tenantID, serviceName).Observe(float64(decompressedSize))
 			d.metrics.receivedSamples.WithLabelValues(profName, tenantID).Observe(float64(len(p.Sample)))
 			d.profileSizeStats.Record(float64(decompressedSize), profLanguage)
 
 			if err = validation.ValidateProfile(d.limits, tenantID, p.Profile, decompressedSize, series.Labels, now); err != nil {
 				_ = level.Debug(d.logger).Log("msg", "invalid profile", "err", err)
 				validation.DiscardedProfiles.WithLabelValues(string(validation.ReasonOf(err)), tenantID).Add(float64(req.TotalProfiles))
-				validation.DiscardedBytes.WithLabelValues(string(validation.ReasonOf(err)), tenantID).Add(float64(req.TotalBytesUncompressed))
+				validation.DiscardedBytes.WithLabelValues(string(validation.ReasonOf(err)), tenantID, serviceName).Add(float64(req.TotalBytesUncompressed))
 				return nil, connect.NewError(connect.CodeInvalidArgument, err)
 			}
 
@@ -402,10 +407,10 @@ func (d *Distributor) sendRequests(ctx context.Context, req *distributormodel.Pu
 		series.Labels = d.limitMaxSessionsPerSeries(maxSessionsPerSeries, series.Labels)
 	}
 
+	usageGroups := d.limits.DistributorUsageGroups(tenantID)
+
 	// Next we split profiles by labels and apply relabel rules.
-	profileSeries, bytesRelabelDropped, profilesRelabelDropped := extractSampleSeries(req, d.limits.IngestionRelabelingRules(tenantID))
-	validation.DiscardedBytes.WithLabelValues(string(validation.RelabelRules), tenantID).Add(bytesRelabelDropped)
-	validation.DiscardedProfiles.WithLabelValues(string(validation.RelabelRules), tenantID).Add(profilesRelabelDropped)
+	profileSeries, _, _ := extractSampleSeries(req, usageGroups, d.limits.IngestionRelabelingRules(tenantID))
 
 	// Filter our series and profiles without samples.
 	for _, series := range profileSeries {
@@ -427,9 +432,12 @@ func (d *Distributor) sendRequests(ctx context.Context, req *distributormodel.Pu
 		if enforceLabelsOrder {
 			series.Labels = phlaremodel.Labels(series.Labels).InsertSorted(phlaremodel.LabelNameOrder, phlaremodel.LabelOrderEnforced)
 		}
+
+		serviceName := usageGroups.GetServiceName(phlaremodel.Labels(series.Labels))
+
 		if err = validation.ValidateLabels(d.limits, tenantID, series.Labels); err != nil {
 			validation.DiscardedProfiles.WithLabelValues(string(validation.ReasonOf(err)), tenantID).Add(float64(req.TotalProfiles))
-			validation.DiscardedBytes.WithLabelValues(string(validation.ReasonOf(err)), tenantID).Add(float64(req.TotalBytesUncompressed))
+			validation.DiscardedBytes.WithLabelValues(string(validation.ReasonOf(err)), tenantID, serviceName).Add(float64(req.TotalBytesUncompressed))
 			return nil, connect.NewError(connect.CodeInvalidArgument, err)
 		}
 		keys[i] = TokenFor(tenantID, phlaremodel.LabelPairsString(series.Labels))
@@ -746,7 +754,7 @@ func (g *groupsWithFingerprints) add(stringTable []string, lbls phlaremodel.Labe
 	})
 }
 
-func extractSampleSeries(req *distributormodel.PushRequest, relabelRules []*relabel.Config) (result []*distributormodel.ProfileSeries, bytesRelabelDropped, profilesRelabelDropped float64) {
+func extractSampleSeries(req *distributormodel.PushRequest, usageGroups *validation.TenantUsageGroups, relabelRules []*relabel.Config) (result []*distributormodel.ProfileSeries, bytesRelabelDropped, profilesRelabelDropped float64) {
 	var (
 		lblbuilder = phlaremodel.NewLabelsBuilder(phlaremodel.EmptyLabels())
 	)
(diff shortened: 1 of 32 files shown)


pkg/distributor/distributor.go
```diff
@@ -746,7 +754,7 @@ func (g *groupsWithFingerprints) add(stringTable []string, lbls phlaremodel.Labe
 	})
 }
 
-func extractSampleSeries(req *distributormodel.PushRequest, relabelRules []*relabel.Config) (result []*distributormodel.ProfileSeries, bytesRelabelDropped, profilesRelabelDropped float64) {
+func extractSampleSeries(req *distributormodel.PushRequest, usageGroups *validation.TenantUsageGroups, relabelRules []*relabel.Config) (result []*distributormodel.ProfileSeries, bytesRelabelDropped, profilesRelabelDropped float64) {
```
bryanhuhta, line 757: It's unfortunate this function signature had to change. This function previously returned all dropped bytes, so I couldn't use the return value as there may be series present that got dropped which need to specially labeled.

pkg/distributor/distributor.go
```diff
@@ -862,6 +880,9 @@ func (d *Distributor) rateLimit(tenantID string, req *distributormodel.PushReque
 	// rate limit the request
 	if !d.ingestionRateLimiter.AllowN(time.Now(), tenantID, int(req.TotalBytesUncompressed)) {
 		validation.DiscardedProfiles.WithLabelValues(string(validation.RateLimited), tenantID).Add(float64(req.TotalProfiles))
+
+		// note(bryanhuhta): We don't need to label this metric with service
+		// name as rate limited requests don't count towards billable bytes.
 		validation.DiscardedBytes.WithLabelValues(string(validation.RateLimited), tenantID).Add(float64(req.TotalBytesUncompressed))
```
bryanhuhta, line 886: I think this is a valid assumption. When calculating billable bytes, the recording rules seem to subtract out any bytes that were rate limited.

https://github.com/grafana/deployment_tools/blob/6b6e9832e7abb0fa2b8aeafc3aa6ef75d69880f6/ksonnet/lib/billing-mixin/recording_rules/profiles.libsonnet#L48-L50

pkg/distributor/distributor_test.go
```diff
@@ -997,8 +997,12 @@ func Test_SampleLabels(t *testing.T) {
 
 	for _, tc := range testCases {
 		tc := tc
+		ug := &validation.TenantUsageGroups{
+			TenantID: "",
+		}
```
bryanhuhta, line 1002: This is only here because the `extractSampleSeries` signature changed.

pkg/validation/usage_groups.go
bryanhuhta, line 1: The implementation of the service name allowlist.

petethepig (approved):

petethepig:
Looks good. Few random thoughts:

1. We could copy the config format from Cloud metrics. Theirs is more flexible because you can set different selectors (not just service_name), but more complex
2. I might be overly paranoid, but I would maybe create a new metric instead of using the existing one. This way if something goes wrong (e.g cardinality explosion) it would be easier to fix.
3. I would avoid empty strings in label values. I'd call it "unknown" or "other" or something like that. This also has a benefit of showing up in the dashboard later so that it's clear for users what's going on.

I don't feel very strongly about 1 and 2, I do feel strongly on the third point.

bryanhuhta:
Here's it working in fire-dev-001: https://ops.grafana-ops.net/goto/yu16UVXIg?orgId=1

Screenshot 2024-07-18 at 5 52 27 PM

Using the following usage group definitions:

```yaml
distributor_usage_groups:
  - pyroscope: '{service_name=~"fire-dev-001/.*"}'
  - cortex-dev-01/ingester: '{service_name="cortex-dev-01/ingester"}'
```

This groups all the `fire-dev-001/*` apps under the `pyroscope` usage group and puts `cortex-dev-01/ingester` into its own usage group.

pkg/validation/usage_groups.go
```diff
@@ -0,0 +1,160 @@
...
+
+			if name == "" {
+				return nil, fmt.Errorf("empty service name in usage group for tenant %q", tenantID)
+			}
+
+			if matchersString == "" {
+				return nil, fmt.Errorf("no matchers for usage group %q and tenant %q", name, tenantID)
+			}
+
+			amMatchers, err := amlabels.ParseMatchers(matchersString)
```
simonswine, line 122: Maybe I am missing something, but we probably should parse those the same way we do in the query path: https://github.com/simonswine/pyroscope/blob/ed0b5643f48792ac988d975e17e285b807d5c1f9/pkg/phlaredb/head.go#L440

pkg/validation/usage_groups.go
```diff
@@ -0,0 +1,160 @@
...
+	usageGroupReceivedDecompressedBytes.WithLabelValues(profileType, u.tenantID, u.name).Add(float64(n))
+}
+
+func (u UsageGroup) CountDiscardedBytes(reason string, n int64) {
+	UsageGroupDiscardedBytes.WithLabelValues(reason, u.tenantID, u.name).Add(float64(n))
+}
+
+// DistributorUsageGroups returns the usage groups that are enabled for this
+// tenant.
+func (o *Overrides) DistributorUsageGroups(tenantID string) (*UsageGroupConfig, error) {
```
simonswine, line 90: Any returned error in this method will lead to profiling traffic being dropped. This is not ideal.
This seems quite a severe reaction to "I have wrong accounting usage groups". 

I do suggest we should move all those errors into the parsing part of the overrides.

simonswine: Separately this method does a lot and is in the hot path of ingestion. We should also move as much as we can into the "loading the overrides" part of the code.

E.g. the parsing of label matchers should only happen only when the override is parsed and not for every profile ingested.

pkg/validation/limits.go
```diff
@@ -50,6 +50,9 @@ type Limits struct {
 	MaxProfileStacktraceDepth        int `yaml:"max_profile_stacktrace_depth" json:"max_profile_stacktrace_depth"`
 	MaxProfileSymbolValueLength      int `yaml:"max_profile_symbol_value_length" json:"max_profile_symbol_value_length"`
 
+	// Distributor per-app usage breakdown.
+	DistributorUsageGroups []map[string]string `yaml:"distributor_usage_groups" json:"distributor_usage_groups"`
```
simonswine, line 54: Do you mind explaining why this is a slice of maps, rather than just a map, which I would expect?

Here the yaml "explanation of my question":

```yaml
current_usage_groups:
- my_team: '{namespace="cool-stuff"'
- other_team: '{namespace="boring-stuff"}'

why_not_usage_groups:
 my_team: '{namespace="cool-stuff"}'
 other_team: '{namespace="boring-stuff"}'	
```

simonswine (commented):
have a few concerns which I outlined in the line by line comments. Happy to dive into them a bit more if stuff is unclear.

It is also worth taking a look at how Mimir does it, because their implementaiton is obviously a bit more tested: https://github.com/grafana/mimir/blob/main/pkg/ingester/activeseries/custom_trackers_config.go

pkg/validation/usage_groups.go
bryanhuhta: Good point, a lot of this file was copied from how Mimir implemented this. So no, you aren't missing anything, I just happened to use Mimir's parse metrics approach instead of our own. I'll switch this.

pkg/validation/limits.go
simonswine: Totally get the order and deterministic result problem (now  that you mentioned it😆 ), but I think the map inside the slice will still be non-deterministic in the current implementation.

I think maybe a slice of two string might be a better solution:

```[]struct{
name string
selector string
}
```

or just sorting by the map key (name of the rule group).

Probably it is best to aim for consistency with other Grafana Products and do what Mimir does.

pkg/validation/usage_groups.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	for name, matchersText := range m {
+		name = strings.TrimSpace(name)
+		if name == "" {
+			return UsageGroupConfig{}, fmt.Errorf("usage group name cannot be empty")
+		}
+
+		// TODO(bryanhuhta): We should probably validate the usage group name
+		// is a valid label value.
```
bryanhuhta, line 145: I couldn't figure out where prometheus does its validation of label values, but wherever it is, we should copy it.

pkg/validation/usage_groups.go
```diff
@@ -0,0 +1,187 @@
...
+	return config, nil
+}
+
+func (o *Overrides) DistributorUsageGroups(tenantID string) *UsageGroupConfig {
+	config := o.getOverridesForTenant(tenantID).DistributorUsageGroups
+
+	// It should never be nil, but check just in case!
+	if config == nil {
+		config = &UsageGroupConfig{}
+	}
```
bryanhuhta, line 164: With the current code, the config will never be nil. However, I think we should leave this check here so we treat a nil config like a empty config, in case the config could be nil at this point.

bryanhuhta (commented):
This implementation has had a few iterations and has changed significantly since I opened this PR. This is a recap of the implementation to-date.

### Design

The idea of the usage group at this point is identical (or nearly so) to Mimir's approach. A usage group has a name and matcher(s). When a profile is ingested, it's labels are parsed and matched against each usage group. Any usage group it matches will be collected and the profile size will be counted by the `pyroscope_usage_group_received_decompressed_total` metric. Later, if that profile is dropped for any reason, it the drop will be counted by the `pyroscope_usage_group_discarded_bytes_total` metric.

As noted before, a profile can match 0 or more usage groups.

- If 0, then it's bucketed into a default usage group called "other"
- If 1 or more, the `pyroscope_usage_group_received_decompressed_total` will be counted with each matching usage group

As an example, consider the following labels attached to a profile:

```json
[
  { "service_name": "foo" }
  { "namespace": "barbaz" }
]
```

along with the following usage groups:

```yaml
distributor_usage_groups:
  app/foo: '{service_name="foo"}'
  namespace/bar: '{namespace=~"bar.*"}'
```

the `pyroscope_usage_group_received_decompressed_total` metric will be counted once for `app/foo` and once for `namespace/bar`:

```
pyroscope_usage_group_received_decompressed_total{usage_group="app/foo"}
pyroscope_usage_group_received_decompressed_total{usage_group="namespace/bar"}
```

### Implementation

I created a struct `UsageGroupConfig` which is responsible for

- unmarshalling the usage group config (from both yaml and json)
- validating the group count (limit 50 for now)
- normalizing usage group names
- parsing and validating the matchers

This struct is built once at start up and any errors it generates will cause the app to fail to run. I believe this is reasonable.

`UsageGroupConfig` has a method `GetUsageGroups(phlaremodel.Labels)` which accepts a label set. It will match the label set against all the usage groups and return a list of all that match via the proxy object `UsageGroupMatch`. `UsageGroupMatch` has methods that can report a number as "bytes received" or "bytes dropped". This is where the logic of emitting a metric label for each usage group occurs.

### Remarks

@simonswine identified one of the former approaches as being detrimental to the hot path both in aggressive failure cases and lots of new computation. Both these concerns are addressed by doing the bulk of the error-prone parsing and validation at startup, leaving simple accounting to the hot path.

Additionally, there was an ask to make this look and feel more like Mimir's approach. Whilst it isn't exactly the same, I'm hoping with the most recent changes we move much, much closer to their implementation. I took a lot of inspiration from how they implemented this feature, so the user-facing look-and-feel and the code of this feature should feel very similar to Mimir's.

Lastly, It's important to note that this approach isn't designed to be accurate down to the byte and penny. @petethepig and I agreed to take this approach because it should be approximately accurate and provide _proportionally_ accurate results. So even if this feature doesn't report penny-exact numbers, the error should be proportionally the same across all usage groups. Big apps will produce big numbers, small apps will produce small numbers.

pkg/validation/usage_groups.go
simonswine: It is any valid utf8 string: https://github.com/prometheus/client_golang/blob/9f203a098ec630d179ccef4efbaa8c3341291d00/prometheus/labels.go#L158


I suggest we also disallow "other" to avoid undefined behaviour.

pkg/validation/usage_groups.go
```diff
@@ -0,0 +1,187 @@
...
+
+	// Make sure the usage group config has the tenant ID set. It is not known
+	// when it is unmarshaled from the configuration.
+	if config.TenantID == "" {
+		config.TenantID = tenantID
+	}
```
simonswine, line 170: This feels a bit sketchy to me, we despite not registering it as through "RegisterFlag" in theory you could still set in the pyroscope.yaml config a default usage groups for all tenants.

This bit would then expose the same *UsageGroupConfig to all tenant that don't override it, with a race to that bit which sets the tenant.

I think it would be better to do the tenant field in a wrapped type or even simpler put it to a parameter to the `Count*()` methods

```
type UsageGroupConfig // without TenantID

type UsageGroupInstance{
*UsageGroupConfig
tenantID string
}
```

simonswine (approved):
Thanks for looking into this and also for providing so much detailed comments and ensure this is properly tested. ❤️ 

I would like you to take a look at my two suggestions regarding validation and `tenant_id` flow. But I think this is already for a LGTM.

PR #3432, opened 2024-07-17
docs: Link to supported languages

simonswine:
I think is better to link to docs, in order to stay up-to-date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,17 +116,7 @@ For more information on how to use Pyroscope with other programming languages, i
 
 You can download the latest version of pyroscope for macOS, linux and Docker from our [Releases page](https://github.com/grafana/pyroscope/releases).
 
-## Supported Languages
-
-* [x] Go (via `pprof`)
-* [x] Python (via `py-spy`)
-* [x] Ruby (via `rbspy`)
-* [x] Linux eBPF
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -116,17 +116,7 @@ For more information on how to use Pyroscope with other programming languages, i
...
-* [x] Go (via `pprof`)
-* [x] Python (via `py-spy`)
-* [x] Ruby (via `rbspy`)
-* [x] Linux eBPF
-* [x] Java (via `async-profiler`)
-* [x] Rust (via `pprof-rs`)
-* [x] .NET
-* [x] PHP (via `phpspy`)
-* [x] Node
+## [Supported Languages](https://grafana.com/docs/pyroscope/latest/configure-client/)
```
simonswine, line 119: I was in two minds if this link would even be better:

https://grafana.com/docs/pyroscope/latest/view-and-analyze-profile-data/profiling-types/#available-profiling-types

bryanhuhta: We could do both. We could have a subsection under this header that links to support profiles types.

bryanhuhta (approved):

knylander-grafana (approved):
Thank you for updating the docs.

PR #3435, opened 2024-07-19
fix: readiness check for ingesters and frontend

korniltsev:
This will likely slowdown (already slow) ingesters rollouts due to `  [min_ready_duration: <duration> | default = 15s]`

As a side quest I made `make frontend/build` use docker.


kolesnikovae (approved):
LGTM!

PR #3460, opened 2024-08-01
chore: Move paths from docs/phlare to docs/pyroscope

simonswine:
I had noticed a couple of paths, which still point to the old path /docs/phlare instead of /docs/pyroscope


aleks-p (approved):
LGTM

PR #3462, opened 2024-08-01
chore: Add relabeling rules flags

simonswine:
This adds flags for the relabeling rules and also makes sure they will show up in the documentation properly.


docs/sources/configure-server/reference-configuration-parameters/index.md
```diff
@@ -1843,6 +1843,8 @@ The `limits` block configures default and per-tenant limits imposed by component
 # CLI flag: -validation.max-profile-symbol-value-length
 [max_profile_symbol_value_length: <int> | default = 65535]
 
+distributor_usage_groups:
```
simonswine, line 1846: @bryanhuhta you probably want to provide a similar PR for usage groups.

bryanhuhta (approved):

Issue #3459, opened 2024-08-01
Support for EKS Pod Identity


mogopz:
#### Is your feature request related to a problem? Please describe.

Hello! It would be great if Pyroscope worked with the new(ish) EKS feature [Pod Identity](https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html). I've just tried to use it but I _think_ the AWS SDK version being used is [too old](https://docs.aws.amazon.com/eks/latest/userguide/pod-id-minimum-sdk.html) so I was getting Access Denied errors and had to revert to using IRSA instead.

#### Describe the solution you'd like

The ability to map roles to Pyroscope pods using Pod Identity instead of IRSA.

#### Describe alternatives you've considered

We're currently using IRSA instead which works fine, but Pod Identity is preferred 😁

simonswine:
This should be working after #3465 has merged.

PR #3465, opened 2024-08-05
chore: Refactor block listing

simonswine:
We stopped using the prefix based listing a while ago, so this brings us back to upstream of thanos-io/objstore.

It also addresses #3459, by updating the minio library past 7.070 which contains https://github.com/thanos-io/thanos/issues/7157


kolesnikovae (approved):
LGTM

PR #3466, opened 2024-08-05
chore: Apparently github broke docker compose:

simonswine:
Related: https://github.com/actions/runner-images/issues/9692


korniltsev (approved):

simonswine:
Pull this into the weekly of this week already.

Should be backported everywhere, to not block a bugfix

PR #3477, opened 2024-08-13
fix(ci): remove reviewdog hack

kolesnikovae:
In #3388, I added a workaround to bypass some bugs in reviewdog that were breaking CI on large diffs. The workaround relies on the PR head commit SHA, which is not available when merging to main and causes CI to fail. I could have added a separate doc-validation action or modified the script to work in both cases. Instead, I'm removing the workaround to fix the CI pipeline for now. Large diffs (20K+ lines, 300+ files) will break CI.

/cc @jdbaldry

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -67,30 +67,6 @@ jobs:
     steps:
       - name: "Checkout code"
         uses: "actions/checkout@v4"
-      # reviewdog is having issues with large diffs.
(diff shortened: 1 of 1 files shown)


Rperry2174 (approved):

jdbaldry:
Thanks for bringing this to my attention. Sorry for the lack of engagement on this over the last week, I've been on holiday.

I'm thinking of wrapping up this workflow in a composite action and I'll make sure that this is included when I do.

PR #3479, opened 2024-08-13
fix(flameql): allow dots in tag name

korniltsev:
The dot will be replaced with underscore later anyways.

https://github.com/grafana/pyroscope/pull/3335


kolesnikovae (approved):
LGTM

PR #3480, opened 2024-08-13
chore: add experimental modules

kolesnikovae:


operations/pyroscope/helm/pyroscope/templates/deployments-statefulsets.yaml
```diff
@@ -81,6 +81,13 @@ spec:
               {{- end }}
           {{- end }}
           {{- end }}
+          env:
+            - name: POD_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
+            - name: NAMESPACE_FQDN
+              value: "{{ .Release.Namespace }}.svc{{ .Values.pyroscope.cluster_domain }}"
```
kolesnikovae: I'm not fond of doing this, but I couldn't find a simpler solution. I don't want to involve DNS here

pkg/experiment/metastore/client/client.go
```diff
@@ -72,16 +52,18 @@ func dial(cfg Config, logger log.Logger) (*grpc.ClientConn, error) {
...
+	//   Consider: https://github.com/sercand/kuberesolver
+	//	if os.Getenv("KUBERNETES_SERVICE_HOST") != "" {
+	//		builder, err := NewGrpcResolverBuilder(logger, address)
+	//		if err != nil {
+	//			return nil, fmt.Errorf("failed to create grpc resolver builder: %w", err)
+	//		}
+	//		options = append(options, grpc.WithResolvers(builder))
+	//		return grpc.Dial(builder.resolverAddrStub(), options...)
+	//	}
+	return grpc.Dial(address, options...)
```
kolesnikovae, line 66: Disabled it for now because of permissions issue in the local deployment – nothing bad, just didn't have time to solve. Also, stumbled upon https://github.com/sercand/kuberesolver – looks pretty good

tools/dev/experiment/values-micro-services-experiment.yaml
```diff
@@ -0,0 +1,71 @@
...
+  extraEnvVars:
+    PYROSCOPE_V2_EXPERIMENT: 1
+  extraArgs:
+    query-backend.address: "dns:///_grpc._tcp.pyroscope-query-worker-headless.$(NAMESPACE_FQDN):9095"
+    metastore.address: "dns:///_grpc._tcp.pyroscope-metastore-headless.$(NAMESPACE_FQDN):9095"
+    metastore.raft.bind-address: ":9099"
+    metastore.raft.server-id: "$(POD_NAME).pyroscope-metastore-headless.$(NAMESPACE_FQDN)"
+    metastore.raft.advertise-address: "$(POD_NAME).pyroscope-metastore-headless.$(NAMESPACE_FQDN):9099"
+    metastore.raft.bootstrap-peers: "dnssrvnoa+_raft._tcp.pyroscope-metastore-headless.$(NAMESPACE_FQDN):9099"
+    metastore.raft.bootstrap-expect-peers: "3"
```
kolesnikovae, line 11: This is why I added those env variables. Obviously, we can craft the addresses in the helm templates, but I decided not to modify them yet. I think that we may want to add component-specific templates instead of reusing the same one

tools/dev/experiment/values-micro-services-experiment.yaml
```diff
@@ -0,0 +1,71 @@
...
+
+    compaction-worker:
+      kind: StatefulSet
+      replicaCount: 3
+      persistence:
+        enabled: false
+      extraVolumeMounts:
+        - name: data
+          mountPath: /data-compaction-worker
+          subPath: compaction-worker
```
kolesnikovae, line 68: Note that `compaction-worker` is a statefulset: it will be using own persistent disks to avoid impacting the host

pkg/phlare/phlare.go
```diff
@@ -149,6 +166,14 @@ func (c *Config) RegisterFlagsWithContext(ctx context.Context, f *flag.FlagSet)
 	c.LimitsConfig.RegisterFlags(f)
 	c.Compactor.RegisterFlags(f, log.NewLogfmtLogger(os.Stderr))
 	c.API.RegisterFlags(f)
+
+	c.v2Experiment = os.Getenv("PYROSCOPE_V2_EXPERIMENT") != ""
```
kolesnikovae, line 170: This is the environment variable that enables the new components

korniltsev (approved):

PR #3494, opened 2024-08-16
docs: Add notice about heap profiling

simonswine:
This was a profiler in the dotnet sdk I noticed we are missing.

### DIFF
--- a/docs/sources/configure-client/language-sdks/dotnet.md
+++ b/docs/sources/configure-client/language-sdks/dotnet.md
@@ -24,6 +24,7 @@ The .NET Profiler supports the following profiling types:
 * Allocations
 * Lock contention
 * Exceptions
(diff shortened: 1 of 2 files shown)


korniltsev (approved):
I am curtain there was a reason we removed live heap from the docs. 
I don't remember and could not find the exact reason though.

knylander-grafana (approved):
Thank you for updating the docs and the table!

knylander-grafana:
> I am certain there was a reason we removed live heap from the docs. I don't remember and could not find the exact reason though.

Are there any caveats that go along with this option?

simonswine:
> Are there any caveats that go along with this option?

We are not fully sure, at least it requiers .Net 7.0+. Let's document the flag it will be still default off.

PR #3496, opened 2024-08-16
Build profilecli and pyroscope separately

bryanhuhta:
I use a local debug binary of Pyroscope when testing various changes locally. Right now, the only way to get a debug binary of Pyroscope is to use

```
make go/bin-debug
```

which builds  both `profilecli` and `pyroscope`. It's a minor annoyance to have to wait for `profilecli` to get built every time. This creates two separate rules to build Pyroscope or `profilecli`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,10 +77,15 @@ build: frontend/build go/bin ## Do a production build (requiring the frontend bu
 build-dev: ## Do a dev build (without requiring the frontend)
 	$(MAKE) EMBEDASSETS="" go/bin
 
-
 .PHONY: frontend/build
 frontend/build:
-	docker build  -f cmd/pyroscope/frontend.Dockerfile --output=public/build .
+	docker build -f cmd/pyroscope/frontend.Dockerfile --output=public/build .
+
+.PHONY: profilecli/build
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -77,10 +77,15 @@ build: frontend/build go/bin ## Do a production build (requiring the frontend bu
 build-dev: ## Do a dev build (without requiring the frontend)
 	$(MAKE) EMBEDASSETS="" go/bin
 
-
```
bryanhuhta, line 80: Just whitespace formatting

Makefile
```diff
@@ -180,7 +198,7 @@ define deploy
 		--set pyroscope.podAnnotations."profiles\.grafana\.com\/goroutine\.port_name"=http-metrics \
 		--set pyroscope.extraEnvVars.JAEGER_AGENT_HOST=jaeger.monitoring.svc.cluster.local. \
 		--set pyroscope.extraArgs."pyroscopedb\.max-block-duration"=5m
-  endef
+endef
```
bryanhuhta, line 201: More whitespace formatting

korniltsev (approved):

PR #3498, opened 2024-08-18
segment writer service

korniltsev:
Bring back segment writer service.
Add push protobuf api for segment writer.

The service is still detached - nobody is pushing to it yet.
The service is not as optimized as in POC.

It will be addressed in followups.


pkg/experiment/ingester/service.go
```diff
@@ -0,0 +1,237 @@
...
+					case validation.SeriesLimit:
+						return connect.NewError(connect.CodeResourceExhausted, err)
+					}
+				}
+			}
+			return nil
+		})
+		if err != nil {
+			return err
+		}
```
kolesnikovae, line 210: As we won't have `SeriesLimit` in segment writer, we can simplify this piece

pkg/experiment/ingester/service.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+	if i.cfg.Async {
+		return connect.NewResponse(&segmentWriterV1.PushResponse{}), nil
+	}
+	t1 := time.Now()
+	for wait := range waits {
+		if err = wait.waitFlushed(ctx); err != nil {
+			i.segmentWriter.metrics.segmentFlushTimeouts.WithLabelValues(tenantID).Inc()
+			i.segmentWriter.metrics.segmentFlushWaitDuration.WithLabelValues(tenantID).Observe(time.Since(t1).Seconds())
+			level.Error(i.logger).Log("msg", "flush timeout", "err", err)
```
kolesnikovae, line 179: We assume that the error indicates a timeout. We probably want to check the error type here (or `context.Err()`)

pkg/experiment/ingester/service.go
```diff
@@ -0,0 +1,237 @@
...
+	for _, series := range req.Msg.Series {
+		var shard = shardKey(series.Shard)
+		wait, err := i.segmentWriter.ingest(shard, func(segment segmentIngest) error {
+			return i.ingestToSegment(ctx, segment, series, tenantID)
+		})
+		if err != nil {
+			return nil, err
+		}
+		waits[wait] = struct{}{}
+	}
```
kolesnikovae, line 170: NB: If we moved pprof split from distributors to segment writers and restricted requests to a single profile, we would not need to wait multiple segments to flush (which may result in 2 * segment_duration latency)

api/segmentwriter/v1/push.proto
```diff
@@ -0,0 +1,36 @@
...
+  rpc Push(PushRequest) returns (PushResponse) {}
+}
+
+message PushResponse {}
+
+// WriteRawRequest writes a pprof profile
+message PushRequest {
+  // series is a set raw pprof profiles and accompanying labels
+  repeated RawProfileSeries series = 1;
+}
```
kolesnikovae, line 17: We discussed this internally at some point, and I recall the consensus was that batching does not benefit us here. On the contrary, it introduces several issues:
1. Callers have to wait all the affected shard segment writers to flush, which badly impacts latency, and may also impact resource usage on the distributor side.
2. It complicates error handling. I'm not 100% sure that partial success is handled properly.  
3. It complicates retries on the distributor end.

I hope we'll amend the API and implementation accordingly in follow-up PRs.

kolesnikovae (approved):
LGTM

api/segmentwriter/v1/push.proto
korniltsev: I removed repeated series, but kept repeated samples

PR #3503, opened 2024-08-19
symdb: add WritePartitionV3 function

korniltsev:
`writerV3` depends on `fileWriter` which is an actual os file

For segment writer we only need a single partition and don't need os file.

```go
func WritePartitionV3(p *PartitionWriter, dst io.Writer) error
```
Adding this function seem to allow to not use os file system and also to not touch old/existing write path

@kolesnikovae wdyt?


pkg/phlaredb/symdb/block_writer_v3.go
```diff
@@ -120,18 +112,63 @@ func writePartitionV3(w *writerV3, p *PartitionWriter) (err error) {
...
+	h.Size = uint32(w.offset) - uint32(h.Offset)
 	h.CRC = crc.Sum32()
 	h.Length = uint32(len(s))
 	h.BlockSize = uint32(e.blockSize)
 	h.BlockHeaderSize = uint16(e.blockEncoder.headerSize())
 	h.Format = e.blockEncoder.format()
 	return h, nil
 }
+
+func WritePartitionV3(p *PartitionWriter, dst io.Writer) error {
```
kolesnikovae, line 131: I think we may omit the version, as we always want v3 here (or the latest version):

```suggestion
func WritePartition(p *PartitionWriter, dst io.Writer) error {
```

kolesnikovae (approved):

kolesnikovae:
The approach looks good to me 👍🏻

Issue #185, opened 2021-05-14
Implement TLS support for the server


petethepig:
Currently if you want to use TLS with pyroscope you'd have to put a load balancer in front of it. We should make it so that you can provide a TLS certificate + private key for the server to be able to support TLS encrypted traffic.

petethepig:
Another similar feature @alonlong mentioned is that we might want to provide an option for users to have 2 ports:
* one for web ui
* one for ingestion api

^ this might simplify installation in some environments

PR #3505, opened 2024-08-20
feat: add segment writer client

kolesnikovae:
Depends on https://github.com/grafana/pyroscope/pull/3506
Closes: [#199](https://github.com/grafana/pyroscope-squad/issues/199), [#185](https://github.com/grafana/pyroscope-squad/issues/185) (private)

The change introduces the segment writer client and the write-path router components. A basic distribution algorithm based on jump consistent hashing was also implemented: dynamic sizing is to be implemented later.

Note that this also moves pprof-split out of the distributor – to be implemented in the segment writer in a follow-up PR.


korniltsev (approved):

kolesnikovae:
Depends on #3506

pkg/util/circuitbreaker/cirquit_breaker.go
```diff
@@ -0,0 +1,27 @@
...
+
+func UnaryClientInterceptor(cb *gobreaker.CircuitBreaker[any]) grpc.UnaryClientInterceptor {
+	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		_, err := cb.Execute(func() (interface{}, error) {
+			return nil, invoker(ctx, method, req, reply, cc, opts...)
+		})
+		// gobreaker returns unwrapped errors.
+		switch err {
+		case nil:
+		case gobreaker.ErrOpenState,
```
korniltsev, line 21: nit: should this be checked with error.Is  / errors.As or is it guaranteed the library does not return wrapped errors?

korniltsev (approved):
I love have the routing is separated from the distributor

korniltsev:
and sorry for merge conflicts :shrug:

pkg/util/circuitbreaker/cirquit_breaker.go
kolesnikovae: Thanks for asking! It is guaranteed that the errors are not wrapped. I was on the fence but finally settled on this. I think if there are questions, it's better to go another way

kolesnikovae: Addressed in [clarify gobreaker error handling](https://github.com/grafana/pyroscope/pull/3505/commits/c31c9c8b679291a0a663f947f8fd1d9054c384f9)

PR #3509, opened 2024-08-21
chore: Update rideshare example docker builds and makefile

aleks-p:
- restore files that were deleted by accident in #3026
- fix paths in Makefile.examples


simonswine (approved):
LGTM

PR #3512, opened 2024-08-23
docs: Bring back update-contributors but in go

simonswine:
I think it is a nice to have this appreciation, but we haven't updated in a while, as the nodejs of the old tools no longer maintained.

Just threw some go together quickly.


aleks-p (approved):
LGTM

PR #3516, opened 2024-08-23
Update credentials: Use git directly

simonswine:
Apart from using a shell script to determine when to update this also debugs problems with the token better (using `gh auth status`). Also disabling the persisting of credentials, avoids using wrong credentials.

### DIFF
--- a/.github/workflows/update-contributors.yml
+++ b/.github/workflows/update-contributors.yml
@@ -2,7 +2,7 @@ name: Update Contributors in README
 
 on:
   push:
-    branches: [main]
+    branches: [main, maintest]
 
 jobs:
   update-contributors:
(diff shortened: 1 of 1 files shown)


aleks-p (approved):

.github/workflows/update-contributors.yml
```diff
@@ -2,30 +2,31 @@ name: Update Contributors in README
 
 on:
   push:
-    branches: [main]
+    branches: [main, maintest]
```
bryanhuhta, line 5:
```suggestion
    branches: [main]
```

Is this needed?

bryanhuhta (approved):

.github/workflows/update-contributors.yml
simonswine: No, good catch 9031a41

PR #3519, opened 2024-08-26
feat: improve compaction job state management

aleks-p:
The main change here is how we update the state of compaction jobs, in particular when workers are polling with state updates and ask for jobs.

The current implementation intermixes the persistence (boltdb) and in-memory state updates. This causes a few cases where an error could leave the 2 storage layers in an inconsistent state.

The new implementation does everything in memory first and constructs a list of items that need to be durably stored. If we fail to durably store something or otherwise end up in an unexpected state while persisting, the application will panic.

Bonus:
- handle failed compaction jobs (with max retries)
- prioritize jobs on compaction level before lease expiry
- unit tests for compaction job creation and state management
- convert some flags to config variables
- rename "job pre queue" to "job block queue"


pkg/experiment/metastore/metastore_state_poll_compaction_jobs.go
```diff
@@ -27,119 +26,122 @@ func (m *Metastore) PollCompactionJobs(_ context.Context, req *compactorv1.PollC
...
+			for _, b := range job.Blocks {
+				m.getOrCreateShard(job.Shard).deleteSegment(b)
+				stateUpdate.deletedBlocks[job.Shard] = append(stateUpdate.deletedBlocks[job.Shard], b)
+				m.compactionMetrics.deletedBlocks.WithLabelValues(
+					fmt.Sprint(job.Shard), job.TenantId, fmt.Sprint(job.CompactionLevel)).Inc()
+			}
+		case compactorv1.CompactionStatus_COMPACTION_STATUS_IN_PROGRESS:
+			m.compactionJobQueue.update(statusUpdate.JobName, raftAppendedAtNanos, statusUpdate.RaftLogIndex)
+			stateUpdate.updatedJobs = append(stateUpdate.updatedJobs, job.Name)
+		case compactorv1.CompactionStatus_COMPACTION_STATUS_FAILURE:
```
aleks-p, line 105: This is still missing a failure reason from the worker, I am considering adding it so that we can persist it and show it in tooling. However, after reaching max failures we currently delete the job to avoid accumulating jobs.

We could implement a job retention policy for this (we might need it anyway) but I would do that separately.

pkg/experiment/metastore/metastore_state_poll_compaction_jobs.go
```diff
@@ -340,3 +217,79 @@ func (m *metastoreState) findJobsToAssign(jobCapacity int, raftLogIndex uint64,
 
 	return jobsToAssign
 }
+
+func (m *metastoreState) writeToDb(sTable *pollStateUpdate) error {
+	return m.db.boltdb.Update(func(tx *bbolt.Tx) error {
+		for shard, blocks := range sTable.newBlocks {
```
kolesnikovae, line 223: We need to ensure atomicity of operations within a shard. Consider a case:
- We remove source blocks.
- We handle query that targets the blocks we just have removed.  
- We add compacted blocks.

For me it's not very apparent how we're handling this. I suspect that the query will return incomplete results

kolesnikovae (approved):

pkg/experiment/metastore/metastore_state_poll_compaction_jobs.go
aleks-p: Good point!

Currently from what I can see in the read path we read data directly from memory. 9442699 locks the shards mutex while replacing source blocks with compacted ones so things should be a bit better. Ideally we should lock a single shard but this makes the code more complex and swapping of blocks should anyway be quite fast.

PR #3520, opened 2024-08-26
chore: replace handwriten mocks with mockery

korniltsev:


pkg/experiment/ingester/segment_test.go
```diff
@@ -227,20 +221,18 @@ func TestBusyIngestLoop(t *testing.T) {
...
+	bucket.On("Upload", mock.Anything, mock.MatchedBy(func(name string) bool {
+		return strings.HasSuffix(name, pathBlock)
+	}), mock.Anything).Return(nil)
+	bucket.On("Upload", mock.Anything, mock.MatchedBy(func(name string) bool {
+		return strings.Contains(name, pathDLQ)
+	}), mock.Anything).Return(fmt.Errorf("mock upload DLQ error"))
+	client := mockmetastorev1.NewMockMetastoreServiceClient(t)
+	client.On("AddBlock", mock.Anything, mock.Anything, mock.Anything).
+		Return(nil, fmt.Errorf("mock add block error"))
+
```
kolesnikovae, line 235: You may also want to specify whether each method (and how many times) was called and `AssertExpectations` in the end. Otherwise, if a method has not been called, the test will pass (unless you assert side-effects, which you do)

kolesnikovae (approved):

PR #3529, opened 2024-08-29
feat: add GetCommits endpoint for efficient multiple commits fetching

marcsanmi:
This PR introduces a significant enhancement to the VCS service by adding a new endpoint for parallel fetching of multiple commits. This change improves performance and efficiency when retrieving commit information for multiple refs while maintaining full backwards compatibility

### Key changes
- Added a new `GetCommits` endpoint in the proto definition to handle multiple refs in a single request
- Implemented a new `getCommits` function that fetches commits in parallel using goroutines and channels
- Updated the service layer to handle partial successes and provide detailed logging for the new endpoint
- Added comprehensive tests for the new functionality
- Maintained the existing `GetCommit` endpoint for full backwards compatibility

### Main benefits
1. Reduced network overhead: The frontend can now request multiple commits in a single API call, significantly reducing the number of HTTP requests.
2. Improved performance: Parallel fetching of commits leads to faster overall response times, especially when dealing with multiple refs.
3. Better error handling: The service now provides detailed information about both successful and failed commit fetches.
4. Backwards compatibility: Existing integrations using the `GetCommit` endpoint remain fully functional and unchanged.

### Frontend impact
- No breaking changes for existing frontend code
- Frontend can update their code to use the new `GetCommits` endpoint to fetch multiple commits at once.


api/vcs/v1/vcs.proto
```diff
@@ -48,11 +48,16 @@ message GetFileResponse {
 message GetCommitRequest {
   // the full path to the repository
   string repositoryURL = 1;
-  // the vcs ref to get the file from
-  string ref = 2;
+  // the vcs refs to get the file from
+  repeated string refs = 2;
```
kolesnikovae, line 52: I would likely extend the API by introducing a new method: `ListCommits`

[The standard approach](https://protobuf.dev/programming-guides/proto3/#updating) is to extend the API and preserve the former behavior for the old schema.

Please consider implementing the changes in a backward-compatible way; otherwise, we can't avoid service interruptions during the rollout, as the client (frontend) is updated independently of the backend

kolesnikovae (approved):
Awesome job, it looks great!

pkg/querier/vcs/commit.go
```diff
@@ -0,0 +1,83 @@
...
+// 3. An overall error if no commits were successfully fetched
+// This function provides partial success behavior, returning any commits
+// that were successfully fetched along with errors for those that failed.
+func getCommits(ctx context.Context, client gitHubCommitGetter, owner, repo string, refs []string) ([]*vcsv1.CommitInfo, []error, error) {
+	type result struct {
+		commit *vcsv1.CommitInfo
+		err    error
+	}
+
+	var wg sync.WaitGroup
```
bryanhuhta, line 28: I think ideally we'd use [errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup) here so we can set a limit on the number of concurrent requests (maybe like 10).

errgroup also hides away a lot of the channel nonsense, which really limit the number of "read from closed channel" and other similar mistakes (not saying any were made here, but it just takes one careless update!).

marcsanmi: Nice suggestion to leverage `errgroup` to limit concurrent requests. However, I think we should keep our custom channel error handling to return all errors, since `errgroup` only returns the first error encountered.

PR #3533, opened 2024-08-30
chore(vcs): refactor getCommits to use errgroup

marcsanmi:
Follow-up PR for https://github.com/grafana/pyroscope/pull/3529#discussion_r1737536976. 

Refactor the `getCommits` function in the VCS service to use `errgroup` instead of `sync.WaitGroup`.

### Changes

- Implement `errgroup` for better concurrency management
- Set a limit (10) on concurrent requests to GitHub API

### Note on error handling

I kept the existing error handling approach, collecting errors explicitly through the channel rather than relying on errgroup's built-in error handling, since it will only return the first encountered error.

### DIFF
--- a/pkg/querier/vcs/commit.go
+++ b/pkg/querier/vcs/commit.go
@@ -3,10 +3,13 @@ package vcs
 import (
 	"context"
 	"fmt"
-	"sync"
+
+	"golang.org/x/sync/errgroup"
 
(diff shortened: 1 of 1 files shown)


bryanhuhta (approved):
This looks good, I forgot that `errgroup` only returns the first error. That's pretty lame. Anyway, at least we have some concurrency limits here so we don't gobble up our entire ratelimit budget in one RPC 😁

PR #3541, opened 2024-09-04
fix (v2): add support for ProfileTypes requests

aleks-p:
ProfileTypes is still used for datasource health checks (see [this](https://github.com/grafana/grafana/blob/c00a3bed42db25a3d88b01edd5c00b9625f770d0/pkg/tsdb/grafana-pyroscope-datasource/instance.go#L258)), therefore we need to have it supported in the v2 read path as well. Adding a simple implementation that collects profile types from the block metadata datasets.

TODO: add unit test


pkg/experiment/query_backend/backend.go
```diff
@@ -138,9 +138,9 @@ func (q *QueryBackend) read(
 }
 
 func (q *QueryBackend) withThrottling(fn func() (*queryv1.InvokeResponse, error)) (*queryv1.InvokeResponse, error) {
+	defer q.running.Dec()
```
aleks-p, line 141: this is unrelated, but a previous query-backend based implementation hit an issue where the number of running queries kept going up once the limit was hit

pkg/frontend/read_path/query_service_handler.go
```diff
@@ -203,10 +204,16 @@ func (r *Router) GetProfileStats(
...
 	ctx context.Context,
-	req *connect.Request[querierv1.ProfileTypesRequest],
+	c *connect.Request[querierv1.ProfileTypesRequest],
 ) (*connect.Response[querierv1.ProfileTypesResponse], error) {
-	if r.frontend != nil {
-		return r.frontend.ProfileTypes(ctx, req)
-	}
-	return connect.NewResponse(&querierv1.ProfileTypesResponse{}), nil
+	return Query[querierv1.ProfileTypesRequest, querierv1.ProfileTypesResponse](ctx, r, c,
+		func(a, b *querierv1.ProfileTypesResponse) (*querierv1.ProfileTypesResponse, error) {
```
aleks-p, line 210: this merges data coming from the v1 and the v2 read paths, when the time range includes the transition between the two

korniltsev (approved):
I wonder if we should ask frontend to switch to some other method for healthcheck. I susupect this is the only usage of the method. I had the same issue during POC

aleks-p:
> I wonder if we should ask frontend to switch to some other method for healthcheck. I susupect this is the only usage of the method. I had the same issue during POC

I agree and I can make the switch to a different method in Grafana. Since it is a public API though and would be used in older Grafana versions we can only mark it as deprecated for now but not remove it.

PR #3543, opened 2024-09-05
fix (v2): remove outdated compaction jobs from storage

aleks-p:
When we encounter a job that has no valid blocks (e.g., because they were deleted by the truncation logic) we currently remove it from the in-memory store but not from bolt DB. This causes a burst of new (invalid) jobs every time we restore from a snapshot.

This change ensures that such jobs are removed. There are a few other (mostly cosmetic) changes as well.

### DIFF
--- a/pkg/experiment/metastore/metastore_state_poll_compaction_jobs.go
+++ b/pkg/experiment/metastore/metastore_state_poll_compaction_jobs.go
@@ -169,11 +169,20 @@ func (m *metastoreState) pollCompactionJobs(request *compactorv1.PollCompactionJ
 	resp = &compactorv1.PollCompactionJobsResponse{}
 	if request.JobCapacity > 0 {
 		newJobs := m.findJobsToAssign(int(request.JobCapacity), raftIndex, raftAppendedAtNanos)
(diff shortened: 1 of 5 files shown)


pkg/experiment/metastore/metastore_state_poll_compaction_jobs.go
```diff
@@ -169,12 +169,21 @@ func (m *metastoreState) pollCompactionJobs(request *compactorv1.PollCompactionJ
...
 			m.compactionMetrics.assignedJobs.WithLabelValues(
 				fmt.Sprint(j.Shard), j.TenantId, fmt.Sprint(j.CompactionLevel)).Inc()
 		}
+		for _, j := range invalidJobs {
+			key := tenantShard{
+				tenant: j.TenantId,
+				shard:  j.Shard,
+			}
+			m.compactionJobQueue.evict(j.Name, math.MaxInt64)
+			stateUpdate.deletedJobs[key] = append(stateUpdate.deletedJobs[key], j.Name)
```
aleks-p, line 185: this is the actual bug fix, the rest is minor improvements (mostly logging)

korniltsev (approved):

PR #3555, opened 2024-09-12
Update nodejs examples

bryanhuhta:
With the release of https://github.com/grafana/pyroscope-nodejs/releases/tag/v0.4.0, we need to update all the nodejs sdk examples. This PR does that and also refactors all of them to make them easier to start.


examples/language-sdk-instrumentation/load.js
```diff
@@ -0,0 +1,18 @@
+import { check } from 'k6';
```
aleks-p, line 1: is this referenced anywhere or just a helper to verify things? I see we have one in the `nodejs` folder as well

aleks-p (approved):
LGTM. The tinyhttp example is a great addition!

examples/language-sdk-instrumentation/load.js
bryanhuhta: I accidentally committed two of these scripts. I removed one in https://github.com/grafana/pyroscope/pull/3555/commits/316666d3e083e1a8ef03c672330e30a15f3e8ce4. I used this to quickly apply load to a locally running example.

The Python load generator is nice for the Docker Compose examples because it runs its load continuously, but this k6 script is nice for testing an example that is running locally without having to spin up the Python load generator. Instead, a quick

```
k6 run load.js
```

command will start sending load to a local instance.

I've wanted to move all our examples over to using k6 as a load generator but haven't found time to experiment.

aleks-p: Makes sense and yeah, replacing them with k6 eventually would be nice.

PR #3573, opened 2024-09-19
refactor: deprecating  and start sending  and

alsoba13:
Cookie format apparently introduces a bit of a unwanted complexity. In this PR we deprecate current cookie field in favor of just sending encrypted token and expiry


pkg/querier/vcs/token.go
```diff
@@ -102,14 +107,14 @@ func tokenFromRequest(ctx context.Context, req connect.AnyRequest) (*oauth2.Toke
 	return token, nil
 }
 
-// encodeToken encrypts then base64 encodes an OAuth token.
-func encodeToken(token *oauth2.Token, key []byte) (*http.Cookie, error) {
+// Deprecated: encodeTokenInCookie creates a cookie by encrypting then base64 encoding an OAuth token.
```
aleks-p, line 110: perhaps we should explain why this is deprecated and if we have any plans on removing this soon

aleks-p (approved):
Looks good to me. Any thoughts @bryanhuhta ?

pkg/querier/vcs/token.go
bryanhuhta: Yeah, agree. In the past I've seen people leave target dates/releases of when something can get removed. I.e.
> Remove this after the v1.11 release

I've found this to be useful because it helps signal to other readers (or myself later) that this bit of code is living on borrowed time.

pkg/querier/vcs/token.go
```diff
@@ -140,18 +145,34 @@ func decodeToken(value string, key []byte) (*oauth2.Token, error) {
...
 	sessionToken := gitSessionTokenCookie{}
 	err = json.Unmarshal(decoded, &sessionToken)
-	if err != nil {
-		// This may be a legacy cookie. Legacy cookies aren't base64 encoded
-		// JSON objects, but rather a base64 encoded crypto hash.
-		var innerErr error
-		token, innerErr = decryptToken(value, key)
+	if err != nil || sessionToken.Token == nil {
+		// This may be a legacy cookie. Legacy cookies are base64 encoded legacyGitSessionTokenCookie objects.
+		token, innerErr := decodeLegacyToken(decoded, key)
```
bryanhuhta, line 150: Not a blocker, but I think we can finally move away from this. It's safe to say that no one is using the OG session cookie and if they are, they're gonna have to re-auth. We also want to reduce the number "legacy" versions of this cookie we have floating around haha.

bryanhuhta (approved):
LGTM, a few remarks but nothing blocking.

pkg/querier/vcs/token.go
alsoba13: In fact, I get rid of the legacy OG session data. 

Legacy here refers to the recently deprecated cookie that the PR address. This decode function now works with new format and has this backup plan for deprecated cookies. I will rename the function and rewrite the comment to be more clear!

api/vcs/v1/vcs.proto
```diff
@@ -21,13 +21,27 @@ message GithubLoginRequest {
 }
 
 message GithubLoginResponse {
+  // Deprecated
```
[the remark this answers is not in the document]
alsoba13: done

pkg/querier/vcs/token.go
alsoba13: done by linking the frontend issue that does that. It's difficult to estimate a  release version where this is not needed and can be deleted, as I have no idea on the priority of the frontend part to be completed

PR #3578, opened 2024-09-20
feat(v2): adaptive sharding

kolesnikovae:
Please refer to [the documentation](https://github.com/grafana/pyroscope/blob/feat/dynamic-sharding/pkg/experiment/distributor/README.md).


pkg/experiment/metastore/metastore_state_add_block.go
```diff
@@ -23,6 +25,9 @@ func (m *Metastore) AddBlock(_ context.Context, req *metastorev1.AddBlockRequest
 	_ = level.Info(l).Log("msg", "adding block")
 	t1 := time.Now()
 	defer func() {
+		if err == nil {
+			m.placementMgr.RecordStats(statSamplesFromMeta(req.Block))
```
aleks-p, line 29: Nit, perhaps we want to abstract this call here, we probably don't want "add blocks" to be able to start/stop the placement manager.

aleks-p (approved):
Looks good to me. Thanks for providing the background with the readme file. The test generating plots is also a great addition!

pkg/experiment/distributor/placement/adaptive_placement/store.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+var (
+	ErrRulesNotFound = errors.New("placement rules not found")
+	ErrStatsNotFound = errors.New("placement stats not found")
+)
+
+type StoreReader interface {
+	LoadRules(context.Context) (*adaptive_placementpb.PlacementRules, error)
+	LoadStats(context.Context) (*adaptive_placementpb.DistributionStats, error)
```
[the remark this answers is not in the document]
kolesnikovae: I find it very helpful for debugging purposes. Stats contain lots of interesting information on data distribution of datasets over shards and nodes, which we can't export as prom metrics. There will be an admin interface that provides access to the data.

pkg/experiment/distributor/placement/adaptive_placement/placement_manager.go
```diff
@@ -0,0 +1,210 @@
...
+
+// The function is only needed to satisfy the services.OneIteration
+// signature: there's no case when the service stops on its own:
+// it's better to serve outdated rules than to not serve at all.
+func (m *Manager) updateRulesNoError(ctx context.Context) error {
+	util.Recover(func() { m.updateRules(ctx) })
+	return nil
+}
+
+func (m *Manager) updateRules(ctx context.Context) {
```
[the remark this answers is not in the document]
kolesnikovae: I doubt Go benchmarks make sense here. What part is concerning you the most?

Unless we're talking about dozens of millions of entries, I have no concerns about performance and scalability. Nevertheless, to get millions of entries here, a malicious user would need to send millions of requests with unique service names within the rules retention period; by that time, it will be blocked because of the rate limit.

However, we will limit the number of services per tenant to prevent DDoS in distributors because of other reasons – mainly because of the impact on the metastore; and this will be done based on the placement rules.

I don't see how data isolation at the tenant level might be beneficial.

pkg/experiment/ingester/client/client.go
```diff
@@ -203,45 +202,53 @@ func (c *Client) stopping(_ error) error {
...
-		if !ok {
-			_ = level.Error(c.logger).Log(
-				"msg", "no segment writer instances available for the request",
-				"tenant", req.TenantId)
-			return nil, status.Error(codes.Unavailable, errServiceUnavailableMsg)
-		}
-
-		resp, err := c.pushToInstance(ctx, req, instance.Addr)
-		// Happy path.
+	for attempts := 5; attempts >= 0 && instances.Next() && ctx.Err() == nil; attempts-- {
```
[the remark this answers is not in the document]
kolesnikovae: Unlikely. Therefore, I hardcoded it here. I can't find a reason why anyone would need to change this.

The surrounding code for clarity:
```go
	// In case of a failure, the request is sent to another instance.
	// At most 5 attempts to push the data to the segment writer.
	instances := placement.ActiveInstances(p.Instances)
	req.Shard = p.Shard
	for attempts := 5; attempts >= 0 && instances.Next() && ctx.Err() == nil; attempts-- {
		instance := instances.At()
```

We have something very similar in the current version: https://github.com/grafana/pyroscope/blob/main/pkg/distributor/distributor.go#L523

pkg/experiment/metastore/metastore_state_add_block.go
kolesnikovae: Could you please elaborate? I'd like to avoid abstractions introduced for sake of abstraction.

> we probably don't want "add blocks" to be able to start/stop the placement manager.

I don't think this is the case: `RecordStats` does not control the placement manager's state.

pkg/experiment/distributor/placement/adaptive_placement/placement_manager.go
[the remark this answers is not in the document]
kolesnikovae: BTW, the most expensive part here (`updateRules`) is sorting, which we don't need, strictly speaking. Therefore, If you insist, I'll remove it to save another ~0.25% of CPU time :)

image

I was more concerned with `RecordStats`, but in practice, it utilizes less than logging here
https://github.com/grafana/pyroscope/blob/596641bfb9ae75748a828c9eded29aae3190a864/pkg/experiment/metastore/metastore_state_add_block.go#L16-L32

image

pkg/experiment/metastore/metastore_state_add_block.go
kolesnikovae: However, I think I understand what you mean here.

The thing is that our metastore is quite cumbersome. I was going to split it into multiple services:

```proto

service MetastoreService {
  rpc AddBlock(AddBlockRequest) returns (AddBlockResponse) {}
}

service MetadataQueryService {
  rpc QueryMetadata(QueryMetadataRequest) returns (QueryMetadataResponse) {}
}

service TenantService {
  // Same as GetProfileStats
  rpc GetTenant(GetTenantRequest) returns (GetTenantResponse) {}
  rpc DeleteTenant(DeleteTenantRequest) returns (DeleteTenantResponse) {}
}

// Methods needed for the operator interface.
service RaftNodeService {
  rpc GetCommitIndex(GetCommitIndexRequest) returns (GetCommitIndexResponse) {}
  rpc ListPeers(ListPeersRequest) returns (ListPeersResponse) {}
}
```

In this way, `AddBlock` can be easily abstracted from the underlying metastore implementation:

```go
type StatsRecorder interface {
	RecordStats(*metastorev1.BlockMeta)
}

type RaftLog[T, R any] interface {
	Apply(context.Context, T) (R, error)
}

type MetastoreService struct {
	StatsRecorder
	RaftLog[*metastorev1.AddBlockRequest, *metastorev1.AddBlockResponse]
}
```

Same for other methods:
```go
type MetadataIndex interface {
	FindBlocksInRange(start, end int64, tenants []string) ([]*metastorev1.BlockMeta, error)
}

// To ensure consistent reads.
type RaftLeader interface {
	CommitIndex(context.Context) (uint64, error)
}

type MetadataQueryService struct {
	RaftLeader
	MetadataIndex
}
```

aleks-p: Yeah, the underlying "issue" is that the add block command implementation is part of the `Metastore`, so it has access to many things it shouldn't have access to (including the entire `PlacementManager` interface). I don't like too many abstractions either, but a few (like the ones you suggested) can make things cleaner.

kolesnikovae: We'll decompose the metastore; that will require quite a massive refactoring, though

kolesnikovae:
Thanks for the review @aleks-p!

PR #3582, opened 2024-09-23
Fix issues when porting alloy/pyroscope to android

zmj64351508:
I'm porting alloy/pyroscope to the mobile phone running Android 14. And I found some issues:

1. The config `CONFIG_PID_NS` is not set in standard Android kernel, which means /proc/self/ns/pid is not exist and alloy will exit with failure. The first commit fixed this issue by ignoring error when /proc/self/ns/pid doesn't exist.

2. When kernel modules or bpf programs are dynamic loaded from userspace after kernel boot, the address in /proc/kallsyms may be unordered, which causes the origin code find wrong symbol. The second commit fixed this issue by sorting the symbol by address.
Here's an example for such /proc/kallsyms:
```
# cat /proc/kallsyms
...
ffffffd6b60e2140 t cleanup_module       [bootprof]
ffffffd6b60e53c0 d __this_module        [bootprof]  
ffffffc0145a52cc t bpf_prog_98e4c662c49a1bca_xdp_drop_ipv4_u    [bpf]
ffffffc0145ad4a8 t bpf_prog_a06399a1ec0a769e_schedcls_tether    [bpf]
ffffffc012b4b218 t bpf_prog_e814348da2c722a4_schedcls_tether    [bpf]
...
```

### DIFF
--- a/ebpf/session.go
+++ b/ebpf/session.go
@@ -167,16 +167,19 @@ func (s *session) Start() error {
 	}
 
 	_, nsIno, err := getPIDNamespace()
-	if err != nil {
-		return fmt.Errorf("unable to get pid namespace %w", err)
-	}
-	err = spec.RewriteConstants(map[string]interface{}{
-		"global_config": pyrobpf.ProfileGlobalConfigT{
-			NsPidIno: nsIno,
-		},
-	})
-	if err != nil {
-		return fmt.Errorf("pyrobpf rewrite constants %w", err)
+	// if the file does not exist, CONFIG_PID_NS is not supported, so we just ignore the error
+	if !os.IsNotExist(err) {
+		if err != nil {
+			return fmt.Errorf("unable to get pid namespace %w", err)
+		}
+		err = spec.RewriteConstants(map[string]interface{}{
+			"global_config": pyrobpf.ProfileGlobalConfigT{
+				NsPidIno: nsIno,
+			},
+		})
+		if err != nil {
+			return fmt.Errorf("pyrobpf rewrite constants %w", err)
+		}
 	}
 	err = spec.LoadAndAssign(&s.bpf, opts)
 	if err != nil {
--- a/ebpf/symtab/kallsyms.go
+++ b/ebpf/symtab/kallsyms.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"sort"
 	"strconv"
 )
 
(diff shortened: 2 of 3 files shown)


ebpf/session.go
```diff
@@ -167,16 +167,19 @@ func (s *session) Start() error {
...
-	}
-	err = spec.RewriteConstants(map[string]interface{}{
-		"global_config": pyrobpf.ProfileGlobalConfigT{
-			NsPidIno: nsIno,
-		},
-	})
-	if err != nil {
-		return fmt.Errorf("pyrobpf rewrite constants %w", err)
+	// if the file does not exist, CONFIG_PID_NS is not supported, so we just ignore the error
+	if os.IsExist(err) {
```
korniltsev, line 171: Something looks off here. I would expect this condition to be true if error is ErrExist and I do not expect this error to be returned during pid ns check ( we do not create any files/objects). 
The next error check looks suspicious as well. It looks like we rewrite global_config constant here only in case if ErrExist and err==nil which looks impossible to my brain 
Did you test this change with PD_NS configured and inside a pid namespace ( a docker container for example)?

We need to change something here.
Let me know if I am missing/misunderstanding something.

ebpf/symtab/kallsyms.go
```diff
@@ -94,5 +95,9 @@ func NewKallsymsFromData(kallsyms []byte) (*SymbolTab, error) {
 	if allZeros {
 		return NewSymbolTab(nil), nil
 	}
+	// kallsyms maybe unsorted when bpf/modules are loaded from userspace after kernel boot.
+	sort.Slice(syms, func(i, j int) bool {
```
korniltsev, line 99: nit: It would be nice to add a test, but it is is not required

korniltsev (changes requested):

korniltsev:
Thank you for the contribution. I left some comments.

ebpf/session.go
zmj64351508: PR Updated. I'm new to golang and it's my misunderstanding of `os.IsExist`. It's now replaced by `!os.IsNotExist()`.

zmj64351508: I tested the new patch on WSL and Android Phone, both work fine.

ebpf/symtab/kallsyms.go
zmj64351508: PR Updated. I add 2 unordered symbols in testdata. I tested it and it will fail without this patch and success with this patch.

korniltsev (approved):

korniltsev:
Thank you for the contribution!
I've created an issue to add an integration tests for these cases. Feel free to submit a PR if you want to
https://github.com/grafana/pyroscope/issues/3588

PR #3586, opened 2024-09-23
feat(v2): metastore index

aleks-p:
Introduces the `metastore/index` package which adds a time-partitioned block index and integrates it with the existing flows for adding, compacting and querying blocks. This is a draft with a few parts missing (some of which are marked with FIXME and TODO comments).

This is a breaking change meant for the v2 work stream. Once merged existing blocks will not be reachable.


pkg/experiment/metastore/index/models.go
```diff
@@ -0,0 +1,72 @@
...
+	"sync"
+	"time"
+
+	"github.com/go-kit/log"
+
+	metastorev1 "github.com/grafana/pyroscope/api/gen/proto/go/metastore/v1"
+)
+
+var DefaultConfig = Config{
+	PartitionDuration: time.Hour,
```
kolesnikovae, line 14: I understand that this is for testing, but just to make sure we're on the same page: I believe partition duration should be substantially longer than the expected block duration (after compaction)

pkg/experiment/metastore/index/index.go
```diff
@@ -0,0 +1,511 @@
...
+		i.allPartitions = append(i.allPartitions, pMeta)
+		if pMeta.Ts.Add(i.Config.PartitionTTL).Before(time.Now()) {
+			// too old, will load on demand
+			continue
+		}
+		_, _ = i.getOrLoadPartition(pMeta)
+	}
+
+	i.sortPartitions()
+}
```
kolesnikovae, line 64: Is it safe to ignore errors here?

pkg/experiment/metastore/index/index.go
```diff
@@ -0,0 +1,511 @@
...
+	for _, meta := range i.allPartitions {
+		g.Go(func() error {
+			fn(meta)
+			return nil
+		})
+	}
+	err := g.Wait()
+	if err != nil {
+		level.Error(i.logger).Log("msg", "error during partition iteration", "err", err)
+	}
```
kolesnikovae, line 82: Shouldn't we return the error? I think it's up to the caller to decide how to handle it

pkg/experiment/metastore/index/index.go
```diff
@@ -0,0 +1,511 @@
...
+// partition or create a new partition altogether. Meant to be used only in the context of new blocks.
+func (i *Index) GetOrCreatePartitionMeta(b *metastorev1.BlockMeta) (*PartitionMeta, error) {
+	key := i.CreatePartitionKey(b.Id)
+	meta := i.FindPartitionMeta(key)
+
+	if meta == nil {
+		ts, duration, err := key.Parse()
+		if err != nil {
+			return nil, err
+		}
```
kolesnikovae, line 228: `CreatePartitionKey` should not produce malformed keys. Block ID should be validated before any attempt to access the index

pkg/experiment/metastore/index/models.go
```diff
@@ -0,0 +1,72 @@
...
+
+type indexShard struct {
+	tenantsMu sync.Mutex
+	tenants   map[string]*indexTenant
+}
+
+type indexTenant struct {
+	blocksMu sync.Mutex
+	blocks   map[string]*metastorev1.BlockMeta
+}
```
kolesnikovae, line 58: For me it's not clear why we need these mutexes if we do everything with `partitionMu` acquired.

Usually, such hierarchy of locks is used when you want to delegate access to the object to the caller instead of proxy access. For example, that would make sense if we had something like:

```go
index.Partition(k).Shard(s).Tenant(t).Block(b)
```

But I can't find places where we use/need this.

pkg/experiment/metastore/index/index.go
```diff
@@ -0,0 +1,511 @@
...
+			if err != nil {
+				level.Error(i.logger).Log("msg", "error loading partition", "key", meta.Key, "err", err)
+				return nil, err
+			}
+			tenantBlocks := i.collectTenantBlocks(p, tenants)
+			blocks = append(blocks, tenantBlocks...)
+		} else if firstPartitionIdx != -1 {
+			lastPartitionIdx = idx - 1
+		}
+	}
```
kolesnikovae, line 340: Use of the mutexes is inconsistent: we do not lock access to shards and tenants in `getOrLoadPartition` (it is safe because we took global `partitionMu`) but do acquire locks in `collectTenantBlocks`; note that the locks are redundant there

pkg/experiment/metastore/index/index.go
```diff
@@ -0,0 +1,511 @@
...
+	if b.TenantId != "" {
+		meta.AddTenant(b.TenantId)
+	} else {
+		for _, ds := range b.Datasets {
+			meta.AddTenant(ds.TenantId)
+		}
+	}
+
+	return meta, nil
+}
```
kolesnikovae, line 249: It looks like we need to take the global index mutex here. Despite the fact that `GetOrCreatePartitionMeta` is only called in the context of a raft command, i.e. synchronously and without a risk that someone else will add the partition right after `FindPartitionMeta` returning a nil, it may race with access to `allPartitions` from the read path functions.

pkg/experiment/metastore/metastore_block_store.go
```diff
@@ -0,0 +1,217 @@
...
+				tenants = append(tenants, "")
+			} else {
+				tenants = append(tenants, string(name))
+			}
+			return nil
+		})
+	})
+	if err != nil {
+		panic(err)
+	}
```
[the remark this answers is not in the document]
kolesnikovae: Should be fine

pkg/experiment/metastore/index/index.go
```diff
@@ -0,0 +1,511 @@
...
+	if lastPartitionIdx > -1 && lastPartitionIdx < len(i.allPartitions)-1 {
+		meta := i.allPartitions[lastPartitionIdx+1]
+		p, err := i.getOrLoadPartition(meta)
+		if err != nil {
+			level.Error(i.logger).Log("msg", "error loading next partition", "key", meta.Key, "err", err)
+			return nil, err
+		}
+		tenantBlocks := i.collectTenantBlocks(p, tenants)
+		blocks = append(blocks, tenantBlocks...)
+	}
```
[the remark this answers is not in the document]
aleks-p: Added a single look around period that is configured separately to the ingestion window (we can't use it directly because it is defined per tenant). Will optimize with binary search in a second pass.

pkg/experiment/metastore/metastore_block_store.go
```diff
@@ -0,0 +1,217 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "error creating partition bucket for %s", partitionMeta.Key)
+	}
+
+	var buf bytes.Buffer
+	enc := gob.NewEncoder(&buf)
+	err = enc.Encode(partitionMeta)
+	if err != nil {
+		return errors.Wrapf(err, "could not encode partition meta for %s", partitionMeta.Key)
+	}
```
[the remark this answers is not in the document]
kolesnikovae: > We would be iterating through <retention_days>*<shards>*<num_tenants> buckets, which could go into a few million buckets (30 days, 64 shards, 1000 tenants).

Only limited number of tenants hit each shard. More realistic equation: 30 * 64 * 5 (where 5 is _an average_ number of shards allocated to each tenant)

Note that iterating a million of keys should take a few seconds (amortised 1-10us per key is a fair estimate, I believe). Therefore, even without parallelisation, this should be affordable

However, it clearly makes sense to "seal" old partitions and prebuild metas for them – this does not have to be in the first version

pkg/experiment/metastore/index/index.go
```diff
@@ -0,0 +1,462 @@
...
+
+			for _, t := range i.store.ListTenants(meta.Key, s) {
+				te := &indexTenant{
+					blocks: make(map[string]*metastorev1.BlockMeta),
+				}
+				for _, b := range i.store.ListBlocks(meta.Key, s, t) {
+					te.blocks[b.Id] = b
+				}
+				sh.tenants[t] = te
+			}
```
kolesnikovae, line 181: I believe we should avoid loading all tenant for the partition (I just can't find a use case)

pkg/experiment/metastore/index/index.go
```diff
@@ -0,0 +1,462 @@
...
+}
+
+func (i *Index) tryDelete(key PartitionKey, shard uint32, tenant string, blockId string) (*metastorev1.BlockMeta, *PartitionMeta, bool) {
+	meta := i.findPartitionMeta(key)
+	if meta == nil {
+		return nil, nil, false
+	}
+
+	p := i.getPartition(meta)
+
```
kolesnikovae, line 444: I think we shouldn't load partition from disk here:
1. If partition (its part you want to modify) is memory – update it.
2. Delete the block from store.

pkg/experiment/metastore/index/index.go
```diff
@@ -0,0 +1,462 @@
...
+	"golang.org/x/sync/errgroup"
+
+	metastorev1 "github.com/grafana/pyroscope/api/gen/proto/go/metastore/v1"
+)
+
+type Index struct {
+	Config *Config
+
+	partitionMu      sync.Mutex
+	loadedPartitions *lru.Cache[PartitionKey, *indexPartition]
```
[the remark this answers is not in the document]
aleks-p: If you mean using partition+tenant as the key, that will also have interference (busy tenants would push others out of the cache). For separation we could do separate caches per tenant (complex, costly) or go back to the previous solution (an unbounded TTL-based cache, uneven memory usage).

As discussed separately, this cache is also not suitable because we can get the active (for writes) partition unloaded with a large query. This again can be solved by switching to a different caching strategy (ARC, 2Q, etc.).

Personally, given our usage patterns (frequent writes, infrequent reads) I lean towards a custom solution similar to what we had before, with an upper bound on how many items stay in memory and explicit checks that prevent the write partition to be unloaded.

pkg/experiment/metastore/index/index.go
aleks-p: Agreed, we still need to find partition for the block in order to delete it but there is no need to load the partition.

pkg/experiment/metastore/index/index.go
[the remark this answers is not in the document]
aleks-p: The cache is now per partition+tenant, though I've switched to a custom caching solution where each tenant gets `Config.PartitionCacheSize` slots in the cache.

pkg/experiment/metastore/index/index.go
aleks-p: This is now done, together with the cache key change.

pkg/experiment/metastore/index/index.go
aleks-p: This is now removed and the delete contract is simplified. We no longer return back information about the partition the block was removed from. As a consequence, we now try to remove the block from several partitions in the store, however this only happens when the partition duration changes.

kolesnikovae (approved):
Great work, @aleks-p! Thanks for taking care of this!

PR #3589, opened 2024-09-24
feat: make service_name configurable through environment variable

ldvc:
As it is [done on Mimir](https://github.com/grafana/mimir/commit/b4028c9a2b01df2ead2020c75780ee667af880fb), this would allow administrators to override the default `service_name` using the `JAEGER_SERVICE_NAME` environment variable.


kolesnikovae (approved):
LGTM

Thanks for the contribution @ldvc!

PR #3591, opened 2024-09-25
chore: cleanup frontend dependencies, remove storybook

aleks-p:
A few cleanup changes for the frontend:
- Removes storybook entirely. It doesn't build or run and it causes this security alert: https://github.com/grafana/pyroscope/security/dependabot/612.
- Bumps a few dependencies and adapts to minor breaking changes caused by them. This should address https://github.com/grafana/pyroscope/security/dependabot/617 and get us closer to addressing https://github.com/grafana/pyroscope/security/dependabot/637

Tested with:
- yarn test
- yarn backend:dev

cc @grafakus


aleks-p:
@Rperry2174 any thoughts on removing storybook from the frontend project here? As I wrote above it seems to be broken atm and it causes a security alert that can't be easily addressed.

package.json
```diff
@@ -75,7 +66,6 @@
     "conventional-changelog-cli": "^2.1.1",
     "copy-webpack-plugin": "^11.0.0",
     "cypress": "^12.11.0",
-    "cypress-image-snapshot": "^4.0.1",
```
grafakus, line 78: It seems to be used https://github.com/grafana/pyroscope/blob/main/og/cypress/support/commands.js#L26

grafakus (approved):
Removing code always looks good to me! 🥳

I'm just wondering if @Rperry2174 has some opinion on removing StoryBook.

Rperry2174 (approved):
LGTM

PR #3595, opened 2024-09-27
feat(v2): dlq recovery

korniltsev:


pkg/experiment/metastore/metastore.go
```diff
@@ -178,6 +191,7 @@ func (m *Metastore) starting(context.Context) error {
 func (m *Metastore) stopping(_ error) error {
 	close(m.done)
 	m.wg.Wait()
+	m.dlq.Stop()
 	return m.Shutdown()
 }
```
kolesnikovae, line 196: Stop is called twice: here and in `m.Shutdown`, where I believe it should live. Let's remove if from here

NB: I guess we should stop DLQ before raft just to minimize chances of concurrent recovery

pkg/experiment/metastore/metastore.go
```diff
@@ -233,7 +247,13 @@ func (m *Metastore) initRaft() (err error) {
...
 	}
 
-	m.leaderhealth.Register(m.raft, metastoreRaftLeaderHealthServiceName)
+	m.leaderhealth.Register(m.raft, func(st raft.RaftState) {
+		if st == raft.Leader {
+			m.dlq.Start()
+		} else {
+			m.dlq.Stop()
+		}
+	})
```
kolesnikovae, line 256: I'm not sure we want to use `leaderhealth` now that we do not use grpc health thing for leader discovery. I'd suggest removing it altogether.

On a side note: this assumes that DLQ recovery may run on a follower (as this is a weak check), which should be fine as long as we ensure idempotence of the `AddBlock` call. Is this correct?

pkg/experiment/metastore/dlq/recovery.go
```diff
@@ -0,0 +1,152 @@
...
+	r.cancel()
+	r.wg.Wait()
+	r.started = false
+	r.l.Log("msg", "recovery stopped")
+}
+
+const pathAnon = tenant.DefaultTenantID
+const pathBlock = "block.bin"
+const pathMetaPB = "meta.pb"
+const pathDLQ = "dlq"
```
kolesnikovae, line 78: Let's find a place for the constants (we already have them defined in other places)

pkg/experiment/metastore/metastore.go
```diff
@@ -133,11 +140,12 @@ type Metastore struct {
 	readySince time.Time
 
 	dnsProvider *dns.Provider
+	dlq         *dlq.Recovery
 }
 
 type Limits interface{}
 
-func New(config Config, limits Limits, logger log.Logger, reg prometheus.Registerer, client *metastoreclient.Client) (*Metastore, error) {
+func New(config Config, limits Limits, logger log.Logger, reg prometheus.Registerer, client *metastoreclient.Client, bucket objstore.Bucket) (*Metastore, error) {
```
kolesnikovae, line 148: I'm almost 100% sure we need to update dependencies for metastore

pkg/experiment/metastore/dlq/recovery.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		level.Error(r.l).Log("msg", "failed to add block", "err", err, "path", metaPath)
+		return
+	}
+	err = r.bucket.Delete(ctx, metaPath)
+	if err != nil {
+		level.Error(r.l).Log("msg", "failed to delete block meta", "err", err, "path", metaPath)
+	}
+	return
+}
```
kolesnikovae, line 135: Should we drop the file if we can't handle it?

pkg/experiment/metastore/dlq/recovery.go
```diff
@@ -0,0 +1,152 @@
...
+	"time"
+)
+
+type RecoveryConfig struct {
+	Period time.Duration
+}
+
+type LocalServer interface {
+	AddBlock(context.Context, *metastorev1.AddBlockRequest) (*metastorev1.AddBlockResponse, error)
+}
```
kolesnikovae, line 25: I think it may make sense to distinguish blocks added directly and through the recovery mechanism (e.g.: distinct metrics, online stream statistics, etc.)

```go
type LocalMetastore interface {
    AddRecoveredBlock(context.Context, *metastorev1.AddBlockRequest) (*metastorev1.AddBlockResponse, error)
}
```

pkg/experiment/metastore/dlq/recovery.go
kolesnikovae: Also, we probably want to handle raft errors explicitly: if we discover that we are not the leader, the loop should stop immediately:

``` go
func (r *Recovery) recover(ctx context.Context, path string) bool
```

pkg/experiment/metastore/dlq/recovery.go
```diff
@@ -0,0 +1,152 @@
...
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
+		r.recover(ctx, metaPath)
+		return nil
+	}, objstore.WithRecursiveIter)
+	if err != nil {
+		level.Error(r.l).Log("msg", "failed to iterate over dlq", "err", err)
+	}
+}
```
kolesnikovae, line 103: NB: One danger I see is that we mix old and new blocks in compaction jobs, which may increase the read amplification factor. I think we need to implement some logic to prevent this – just one more rule for the compaction planner. It’s not really critical, as we don't expect this to happen frequently.

/cc @aleks-p

pkg/experiment/metastore/metastore.go
korniltsev: :+1: applied

pkg/experiment/metastore/dlq/recovery.go
korniltsev: created a separate tiny package

pkg/experiment/metastore/metastore.go
korniltsev: thanks, added `Storage` depdendency

pkg/experiment/metastore/dlq/recovery.go
korniltsev: > Should we drop the file if we can't handle it?

I would prefer to remove them manually, or have a configuration option to delete/keep them

> if we discover that we are not the leader, the loop should stop immediately

done

pkg/experiment/metastore/dlq/recovery.go
korniltsev: :+1: changed

pkg/experiment/metastore/metastore.go
korniltsev: > I'd suggest removing it altogether.

The health related bits were removed and I removed the word health from type names. Let me know if this is sufficient. We still need to observe leadership status to start/stop recovery routine

pkg/scheduler/queue/queue_test.go
```diff
@@ -294,7 +294,7 @@ func assertChanReceived(t *testing.T, c chan struct{}, timeout time.Duration, ms
 	select {
 	case <-c:
 	case <-time.After(timeout):
-		t.Fatalf(msg)
+		t.Fatalf("%s", msg)
```
kolesnikovae, line 297: nit: I'm curious if a simple `t.Fatal(msg)` would do the trick

kolesnikovae (approved):
LGTM

pkg/experiment/metastore/dlq/recovery.go
kolesnikovae: I'll add s little more context here.

Basically, when we try to add the same md entry twice there are three cases:
1. We still hold the entry in memory in the metastore. That's simple: we just reject it.
2. We've already compacted the block and we still have tombstone for it. That's simple: we just reject it.
3. We've already compacted the block and removed it from object storage and don't have tombstone for it. Somewhat tricky – I propose the following:
   * Handle the md entry as usual: add it to the index, add it to the compaction job, create a tombstone afterwards, etc. The entry is visible to readers.
   * Ignore "object not found" in compactors and queriers (reflect it in logs and metrics). Clearly, even if the object is missing due to any other reason, it's probably the best strategy. Better than failing the whole query/job at last.
   * Ensure that we only remove tombstone if the object is actually deleted from object storage.

PR #3600, opened 2024-10-01
fix: fix slice init length

huochexizhan:
The intention here should be to initialize a slice with a capacity of len(prof.Location)  rather than initializing the length of this slice.


kolesnikovae (approved):
Thank you for fixing this!

We inherited the issue from pprof, where the bug was fixed a couple of month ago https://github.com/google/pprof/pull/867

Issue #3593, opened 2024-09-26
Add an option to limit the number of series returned in `SelectSeries` API


kolesnikovae:
`SelectSeries` always returns all time series in a single response. In many cases, only a subset of the series is actually used (for example, in Explore Profiles, we only show the top 10).

As a quick optimization, we could add a parameter that limits the number of series returned, something like:

```proto
message Limit {
   int32 top_k_total = 1;
}
```

Note that the limit should be applied in the query frontend, after all the series have been collected. Later on, we can implement a fully-fledged time-series query engine capable of handling PromQL syntax (or a subset)

PR #3602, opened 2024-10-02
feat: add limit to SelectSeries API

kolesnikovae:
Resolves [#2758 ](https://github.com/grafana/pyroscope/issues/3593) 

cc @grafakus 

Note that the change affects the order of series: even if the limit is not specified, series are ordered by value first (descending) and then by labels (ascending). AFAIK, we do not rely on the order anywhere.


grafakus:
Looks great, thank you!

pkg/model/time_series_merger.go
```diff
@@ -20,6 +22,40 @@ func MergeSeries(aggregation *typesv1.TimeSeriesAggregationType, series ...[]*ty
...
+			sum += p.Value
+		}
+		aggregated[i] = series{Series: x, sum: sum}
+	}
+	slices.SortFunc(aggregated, func(a, b series) int {
+		c := cmp.Compare(a.sum, b.sum)
+		if c == 0 {
+			return CompareLabelPairs(a.Labels, b.Labels)
+		}
+		return c * -1 // Invert to sort in descending order.
```
bryanhuhta, line 48: This is strictly an observation but to invert a number you can use the negative sign as well.

```suggestion
		return -c // Invert to sort in descending order.
```

bryanhuhta (approved):
LGTM 👍

PR #3599, opened 2024-10-01
docs: add Linux capabilities config for pyroscope java

marcsanmi:
As per https://github.com/grafana/alloy/pull/1788#issuecomment-2385901229

PR #3603, opened 2024-10-02
Example for Java profiling using Grafana Alloy in Kubernetes

aleks-p:
- Moves the existing example from `examples/grafana-agent-auto-instrumentation/java` to `examples/grafana-agent-auto-instrumentation/java/docker`
- Creates a new example under `examples/grafana-agent-auto-instrumentation/java/kubernetes`


korniltsev (approved):

marcsanmi (approved):
Do you think it makes sense to add this https://github.com/grafana/pyroscope/pull/3599 (alloy helm chart config) into the new Kubernetes `README.md`?

aleks-p:
> Do you think it makes sense to add this #3599 (alloy helm chart config) into the new Kubernetes `README.md`?

That is a good point. I've added the capabilities to the yaml file directly and this worked in my local setup. I think the doc section makes sense in our public documentation, but the README.md can be kept more simple (and it links to the docs for more info).

PR #3619, opened 2024-10-10
feat(v2): implement ReadIndex for linearizable reads

marcsanmi:
This PR implements the ReadIndex in the Metastore to ensure linearizable reads, as described in the Raft consensus algorithm. The implementation leverages the Hashicorp Raft library's existing features while adding necessary checks to guarantee read consistency.

Key changes:
1. Update `ReadIndex` method to verify leadership and return current commit index.
2. Added `readIndex` method to ensure the node is up-to-date before serving reads.
3. Modified `QueryMetadata` and `GetProfileStats` to use `readIndex` before querying data.
4. Updated `CheckReady` method to use `readIndex` for consistency checks.

Raft Algorithm Considerations:
- Hashicorp's Raft implementation [already sends](https://github.com/hashicorp/raft/blob/main/raft.go#L565-L572) a no-op log entry when `runLeader` is called for the leader to be up to the latest commit index, which aligns with the Raft paper's recommendation for ensuring a committed entry from the current term (thanks @kolesnikovae for the heads up).

Testing Considerations:
- Unit tests were not added for these changes as they primarily interact with the Raft package itself.
- Integration tests would be more suitable for verifying the correct interaction with Raft and ensuring linearizable reads.
- The current metastore lacks integration tests, so adding them would be a significant change that deserves its own focused effort.

### DIFF
--- a/pkg/experiment/metastore/metastore_readindex.go
+++ b/pkg/experiment/metastore/metastore_readindex.go
@@ -5,128 +5,65 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/go-kit/log"
-	"github.com/google/uuid"
-
 	metastorev1 "github.com/grafana/pyroscope/api/gen/proto/go/metastore/v1"
 )
 
-var tcheckFreq = 10 * time.Millisecond
+const tCheckFreq = 10 * time.Millisecond
 
-func (m *Metastore) ReadIndex(ctx context.Context, req *metastorev1.ReadIndexRequest) (*metastorev1.ReadIndexResponse, error) {
-	//todo
-	//If the leader has not yet marked an entry from its current term committed, it waits until it
-	//has done so. The Leader Completeness Property guarantees that a leader has all committed
-	//entries, but at the start of its term, it may not know which those are. To find out, it needs to
-	//commit an entry from its term. Raft handles this by having each leader commit a blank no-op
-	//entry into the log at the start of its term. As soon as this no-op entry is committed, the leader’s
-	//commit index will be at least as large as any other servers’ during its term.
-	t := time.Now()
-	readIndex := m.raft.CommitIndex()
-	raftLogger := func() log.Logger {
-		return log.With(m.logger, "component", "raft_debug",
-			"request_id", req.DebugRequestId,
-			"op", "ReadIndex",
-			"read_index", readIndex,
-			"applied_index", m.raft.AppliedIndex(),
-			"commit_index", m.raft.CommitIndex(),
-			"last_index", m.raft.LastIndex(),
-			"duration", time.Since(t),
-		)
-	}
-
-	raftLogger().Log("msg", "verify_leader")
+// ReadIndex returns the current commit index and verifies leadership.
+func (m *Metastore) ReadIndex(_ context.Context, req *metastorev1.ReadIndexRequest) (*metastorev1.ReadIndexResponse, error) {
+	commitIndex := m.raft.CommitIndex()
 	if err := m.raft.VerifyLeader().Error(); err != nil {
 		return nil, wrapRetryableErrorWithRaftDetails(err, m.raft)
 	}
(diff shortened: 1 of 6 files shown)


pkg/experiment/metastore/metastore_readindex.go
```diff
@@ -5,128 +5,65 @@ import (
...
 		return nil, wrapRetryableErrorWithRaftDetails(err, m.raft)
 	}
+	return &metastorev1.ReadIndexResponse{ReadIndex: commitIndex}, nil
+}
+
+// readIndex ensures the node is up-to-date for read operations,
+// providing linearizable read semantics. It calls the ReadIndex RPC
+// and waits for the local applied index to catch up to the returned read index.
+// This method should be used before performing local reads to ensure consistency.
+func (m *Metastore) readIndex(ctx context.Context) error {
```
kolesnikovae, line 26: I may sound silly, but I propose naming this function in an absurdly verbose way:
```go
func (m *Metastore) waitLeaderCommitIndexAppliedLocally(context.Context) error
```

pkg/experiment/metastore/metastore_readindex.go
```diff
@@ -5,128 +5,65 @@ import (
...
-		case <-ctx.Done():
-			raftLogger().Log(status, notReady, "err", "context canceled")
-			return fmt.Errorf("metastore check context canceled %w", ctx.Err())
-		}
+	minReadyTime := m.config.MinReadyDuration
+	if time.Since(m.readySince) < minReadyTime {
+		return fmt.Errorf("waiting for %v after being ready", minReadyTime)
 	}
+
+	return nil
```
kolesnikovae, line 68: It does look like we're replicating k8s behaviour: https://kubernetes.io/blog/2021/08/27/minreadyseconds-statefulsets. I'm wondering if this is really necessary

pkg/experiment/metastore/metastore_state_get_profile_stats.go
```diff
@@ -14,7 +18,9 @@ func (m *Metastore) GetProfileStats(
 	ctx context.Context,
 	r *metastorev1.GetProfileStatsRequest,
 ) (*typesv1.GetProfileStatsResponse, error) {
-	// TODO(kolesnikovae): ReadIndex
+	if err := m.readIndex(ctx); err != nil {
+		return nil, status.Error(codes.FailedPrecondition, fmt.Sprintf("failed to read index: %v", err))
```
kolesnikovae, line 22: `FailedPrecondition` might be not the best choice here:

```
	//  (c) Use FailedPrecondition if the client should not retry until
	//      the system state has been explicitly fixed. E.g., if an "rmdir"
	//      fails because the directory is non-empty, FailedPrecondition
	//      should be returned since the client should not retry unless
	//      they have first fixed up the directory by deleting files from it.
	//  (d) Use FailedPrecondition if the client performs conditional
	//      REST Get/Update/Delete on a resource and the resource on the
	//      server does not match the condition. E.g., conflicting
	//      read-modify-write on the same resource.
```

Let's pass the error as is and let gRPC to deduce the error code (in almost all cases it will be `canceled` or `deadline exceeded`). It's worth logging the error, though

pkg/experiment/metastore/metastore_readindex.go
marcsanmi: You're right.. so, I'd say we can remove it from the "generic" metastore config since we're duplicating behavioiur https://github.com/grafana/pyroscope/blob/main/pkg/experiment/metastore/metastore.go#L53

kolesnikovae (approved):
Thanks for taking care of this @marcsanmi!

@korniltsev could you please take a look as well? Just to make 100% sure the approach is valid

kolesnikovae:
I've reverted the min ready check:
1. This is not quite the same as k8s `minReadySeconds`
2. It might be useful in deployments outside of k8s

PR #3623, opened 2024-10-14
refactor: unify query and query-blocks

alsoba13:
In this PR:
* We unify `query-blocks` into `query`. Now, in the presence of `--block` flag, the query will be performed directly on a local/remote block instead of using a pyroscope server.
* We deprecate `profilecli query merge` in favor of `profilecli query profile`. Old `merge` gets hidden from doc (although still runnable) and logs a deprecation notice.
* Other minor code refactors, mostly moving lines from one file to another.


cmd/profilecli/main.go
```diff
@@ -65,11 +65,11 @@ func main() {
 	tsdbSeriesFiles := tsdbSeriesCmd.Arg("file", "tsdb file path").Required().ExistingFiles()
 
 	queryCmd := app.Command("query", "Query profile store.")
-	queryMergeCmd := queryCmd.Command("merge", "Request merged profile.")
-	queryMergeOutput := queryMergeCmd.Flag("output", "How to output the result, examples: console, raw, pprof=./my.pprof").Default("console").String()
```
alsoba13, line 69: Output flag is used in both `merge` and `go-pgo`. We can move it to `addQueryMergeParams` which is also called by `addQueryGoPGOParams`.

PR #3630, opened 2024-10-15
refactor: renaming `merge` commands to `profile`

alsoba13:
Simple refactor here: just renaming `profilecli query merge` to `profilecli query profile` (same for `profilecli admin blocks query merge`), while keeping the old `merge` subcommand as a valid alias.

Doc updated accordingly

### DIFF
--- a/docs/sources/view-and-analyze-profile-data/profile-cli.md
+++ b/docs/sources/view-and-analyze-profile-data/profile-cli.md
@@ -236,7 +236,7 @@ By default it looks for samples within the last hour, though this can be control
    - You can provide a custom time range using the `--from` and `--to` flags, for example, `--from="now-3h" --to="now"`.
    - You can specify the profile type via the `--profile-type` flag. The available profile types are listed in the output of the `profilecli query series` command.
 
-2. Construct and execute the Query Profile command.
+2. Construct and execute the `query profile` command.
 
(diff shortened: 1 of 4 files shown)


docs/sources/view-and-analyze-profile-data/profile-cli.md
```diff
@@ -224,27 +224,27 @@ You can narrow the results down with the `--query` flag. See `profilecli help qu
...
+#### Query profile steps
 
 1. Specify optional flags.
 
    - You can provide a label selector using the `--query` flag, for example, `--query='{service_name="my_application_name"}'`.
    - You can provide a custom time range using the `--from` and `--to` flags, for example, `--from="now-3h" --to="now"`.
    - You can specify the profile type via the `--profile-type` flag. The available profile types are listed in the output of the `profilecli query series` command.
 
-2. Construct and execute the Query Merge command.
+2. Construct and execute the Query Profile command.
```
knylander-grafana, line 239: Because we're referencing the command specifically, should we format it as a command?
```suggestion
2. Construct and execute the `query profile` command.
```

knylander-grafana (commented):
Thank you for updating the docs! The updates look good.

kolesnikovae (approved):
Thank you @alsoba13! I personally like the new CLI much better

PR #3632, opened 2024-10-16
chore: make the update-contributors workflow run monthly

### LINKED ISSUE #3574: Improve update contributors CI
#3556 addressed an issue when we need to release a new version and the latest commit to main was made via the "update contributors" CI action. The side effect of the fix is that we now try to push and deploy images twice for each push to main. This results in cancelled jobs and docker image tags pointing to uninteresting commits.

It would be nice if we can skip unnecessary job runs while still having the release pipeline work properly. 

cc @korniltsev @marcsanmi

aleks-p:
Fixes #3574 by scheduling the workflow to run on the first ~~day~~ Sunday of every month and removing the [skip ci] flag in the commit. 

We don't really need frequent updates imo and if we run it monthly we don't have to worry about it being wasteful. 

~~There is a small chance of jobs getting cancelled if someone merges a PR right around when this runs (midnight UTC, on the first day of the month), I don't think this is a real problem but we could update the concurrency config if it becomes an issue.~~
This probability is now reduced by having it run on the first Sunday of the month, at 00:00 UTC.

### DIFF
--- a/.github/workflows/update-contributors.yml
+++ b/.github/workflows/update-contributors.yml
@@ -2,7 +2,7 @@ name: Update Contributors in README
 
 on:
   schedule:
-    - cron: '0 0 1 * *'
+    - cron: '0 0 1-7 * SUN'
 
 jobs:
   update-contributors:
--- a/.github/workflows/update-contributors.yml
+++ b/.github/workflows/update-contributors.yml
@@ -1,8 +1,8 @@
 name: Update Contributors in README
 
 on:
-  push:
-    branches: [main]
+  schedule:
+    - cron: '0 0 1 * *'
 
 jobs:
   update-contributors:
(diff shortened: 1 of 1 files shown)


.github/workflows/update-contributors.yml
```diff
@@ -24,7 +24,7 @@ jobs:
           git config --local user.email 'dmitry+bot@pyroscope.io'
           if ! git diff --exit-code README.md; then
               git add README.md
-              git commit -m 'docs: updates the list of contributors in README [skip ci]'
```
marcsanmi, line 27: Another option would be to run it monthly but keep the `skip ci` flag to avoid having an image tied to this commit. The only necessary check would be to schedule it for Wednesday(?), to allow some buffer after our release day.

marcsanmi (approved):
LGTM 👍

kolesnikovae (approved):

.github/workflows/update-contributors.yml
aleks-p: I am unsure about bringing back the `skip ci` flag. If we have a slow week (not many PRs merged), we can get in the same situation again where we can't cut a release. If there are no PRs merged between that Wednesday and Monday, the weekly release will fail and if we decide to make the official release on the same Wednesday it can break too.

PR #3637, opened 2024-10-17
feat(v2): block cleaner

aleks-p:
Adds a block cleaner package for tracking deleted blocks. The purpose is:
- preventing deleted blocks from re-entering the system because of retries / DLQ recovery
- remove deleted blocks from the bucket after a grace period

Some other details:
- block tombstones are stored in memory and in bolt db
- the Raft leader applies a Raft command on an interval
- when applying the command, each metastore instance removes expired tombstones from memory and disk
- the Raft leader also removes the block from the bucket when it applies the command

TODO: 
- improve test coverage
- Analyze performance


pkg/experiment/metastore/metastore_state_clean_blocks.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+func (m *Metastore) ApplyCommand(req *raftlogpb.CleanBlocksRequest) (resp *anypb.Any, err error) {
+	_, resp, err = applyCommand[*raftlogpb.CleanBlocksRequest, *anypb.Any](m.raft, req, m.config.Raft.ApplyTimeout)
+	return resp, err
+}
+
+func (m *metastoreState) applyCleanBlocks(log *raft.Log, request *raftlogpb.CleanBlocksRequest) (*anypb.Any, error) {
+	return nil, m.blockCleaner.RemoveExpiredBlocks(log.AppendedAt.UnixMilli())
+}
```
kolesnikovae, line 17: I think we shouldn't make this method public.

We probably don't want to reserve `ApplyCommand` for `CleanBlocksRequest`.

pkg/experiment/metastore/blockcleaner/block_cleaner.go
```diff
@@ -0,0 +1,417 @@
...
+		})
+	})
+	if err != nil {
+		return nil, err
+	}
+	return blocks, nil
+}
+
+func (c *blockCleaner) removeBlock(blockId string, shard uint32, removalContext *blockRemovalContext) error {
+	err := c.db().Update(func(tx *bbolt.Tx) error {
```
kolesnikovae, line 367: This comment is not specific to this specific call, but concerns use of transactions in general: due to performance considerations we want to minimize the number of them and batch operations as much as we can

kolesnikovae:
Looks good overall, however, I think we need to make sure that only the current leader removes blocks from the object storage. I might be mistaken, but this is not the case in the current version

pkg/experiment/metastore/metastore_state_clean_blocks.go
aleks-p: I think you meant the previous method, but I agree.

pkg/experiment/metastore/blockcleaner/block_cleaner.go
```diff
@@ -0,0 +1,417 @@
...
+	IsMarked(blockId string) bool
+
+	RemoveExpiredBlocks(now int64) error
+}
+
+type RaftLog[Req, Resp proto.Message] interface {
+	ApplyCommand(req Req) (resp Resp, err error)
+}
+
+type RaftLogCleanBlocks RaftLog[*raftlogpb.CleanBlocksRequest, *anypb.Any]
```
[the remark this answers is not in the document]
aleks-p: > We could benefit from delegating managing the raftlogpb.CleanBlocksRequest to a separate component

Agreed.

> // Note that tombstones is a widely used term: https://en.wikipedia.org/wiki/Tombstone_(data_store). // Here and below I use word "mark", but I'd use tombstone, if I were to implement this.

I know what a tombstone is and its usage. It is more nuanced than that because this component, for the better or worse, does more than managing tombstones. A large part of its responsibility is doing garbage collection which the word "tombstone" doesn't communicate. If it were to remain one component (which doesn't have to be the case), "cleaner" encapsulates the two responsibilities more accurately. 

> type DeletionMarkLoader interface {

This part is actually missing currently which will cause a few bugs, good catch!

> `DeletionMarkChecker`, `DeletionMarkAdder`, `BlockRemover`

These are part of a single interface because the "user" is essentially the same (`metastoreState`). We can separate them but it wouldn't make a big difference unless we decompose the commands themselves.

> `RaftLeaderRoutine`

Agreed that we need this. We now have a third component that wants to be run only on the leader. We could benefit from a construct for facilitating this.

aleks-p:
> Looks good overall, however, I think we need to make sure that only the current leader removes blocks from the object storage. I might be mistaken, but this is not the case in the current version

I've reworked things a bit, the `blockcleaner` package now acts mainly as a store for the markers. The Raft command and handling is moved to the `metastore` package. Instead of a leader check we now check that the instance that submitted the command against the instance running the command (using a generated id for now, but could be switched to use raft server ids).

I've also reduced the number of transactions dramatically, deletion is done in a single update transaction. Still lacking extensive test coverage but looks more robust in a dev deployment.

kolesnikovae (approved):
LGTM

PR #3643, opened 2024-10-21
feat(v2): revive grpc health checks

kolesnikovae:
Due to how the grpc server and services are managed, we can't rely on the (de-)initialization order, which causes major issues. Moreover, many of the components are accessed directly through various discovery mechanisms, which provide different (and, generally, week) guarantees on the instance readiness.

Health checks at the connection level help to partially eliminate the problem: a well behaving client won't experience any significant issues when server instances go offline.

### DIFF
--- a/pkg/experiment/metastore/client/methods.go
+++ b/pkg/experiment/metastore/client/methods.go
@@ -3,15 +3,17 @@ package metastoreclient
 import (
 	"context"
 	"fmt"
-	compactorv1 "github.com/grafana/pyroscope/api/gen/proto/go/compactor/v1"
-	metastorev1 "github.com/grafana/pyroscope/api/gen/proto/go/metastore/v1"
-	typesv1 "github.com/grafana/pyroscope/api/gen/proto/go/types/v1"
+	"math/rand"
+	"time"
+
 	"github.com/hashicorp/raft"
 	"google.golang.org/grpc"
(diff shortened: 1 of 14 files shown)


pkg/experiment/metastore/client/methods.go
```diff
@@ -3,23 +3,25 @@ package metastoreclient
...
 )
 
 func invoke[R any](ctx context.Context, cl *Client,
 	f func(ctx context.Context, instance instance) (*R, error),
 ) (*R, error) {
 	const (
 		n        = 50
-		backoff  = 11 * time.Millisecond
+		backoff  = 51 * time.Millisecond
 		deadline = 500000000 * time.Millisecond
```
korniltsev, line 25: Not related to your changes, but I think I forgot to change this value after debugging, it should be smaller, like 5s or 50*51=2550ms

korniltsev (approved):

pkg/experiment/metastore/client/methods.go
kolesnikovae: We should have an exponent here, but even 50ms is not too harmful. I tested various failure cases and it did work well in most of them

kolesnikovae: I have some distant plans to delegate this retries to gRPC conn implementation – I'm thinking about custom resolver and "picker"

PR #3656, opened 2024-10-29
Use uint64 for `proc` offset

### LINKED ISSUE #3650: eBPF profiler `/proc/{id}/maps` offset conversion using small bit size
#### Describe the bug
We noticed the following errors from eBPF profiler when profiling Python processes:

```
ts=2024-10-25T12:27:37.474994587Z level=error msg="pyperf get python process data failed" component_path=/ component_id=pyroscope.ebpf.default err="GetPythonProcInfo error /proc/342561/maps: strconv.ParseInt: parsing \"ffffa6f1ce560000\": value out of range" pid=342561 target="{__name__=\"process_cpu\", __process_pid__=\"342561\", service_name=\"390048\"}"
```

Clearly the hex string is being converted to `int` type [here](https://github.com/grafana/pyroscope/blob/3f6deba3943063acf81de4c9f5fcdac70ed2019a/ebpf/symtab/procmap.go#L249) which exceeds the bounds. Shouldn't it be converted to `unit64` type to ensure we capture offsets for bigger files?

#### To Reproduce

We noticed this using Grafana Alloy 1.9.1. I think the processes need to map very huge files to reproduce this behaviour. 

#### Expected behavior

File offsets converted correctly to int using a bit size big enough to accommodate the offset.

#### Environment

- Bare-metal
- Grafana Alloy 1.9.1 deployed using pre-compiled binary downloaded from repo

#### Additional Context

We are testing this on a HPC platform that has huge file systems and probably that is why we could trigger this bug. If developers agree that it needs to be a `uint64`, would be happy to put up a PR.

mahendrapaipuri:
Fixes #3650

@korniltsev I think this should be fine?

### DIFF
--- a/ebpf/symtab/proc.go
+++ b/ebpf/symtab/proc.go
@@ -82,7 +82,6 @@ func (p *ProcTable) Refresh() {
 	p.err = p.refreshProcMap(procMaps)
 	if p.err != nil {
 		_ = level.Error(p.logger).Log("err", p.err)
-	} else {
 	}
 }
 
--- a/ebpf/symtab/procmap.go
+++ b/ebpf/symtab/procmap.go
@@ -63,7 +63,7 @@ type ProcMap struct {
 	// The permissions for this mapping
 	Perms *ProcMapPermissions
 	// The current offset into the file/fd (e.g., shared libs)
-	Offset int64
+	Offset uint64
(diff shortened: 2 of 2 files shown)


korniltsev (approved):

korniltsev:
Thanks for the contribution. A test would be nice as well.

PR #3674, opened 2024-11-12
docs: Add step to pull latest grafana

Rperry2174:
I often run into issues of grafana being on a different version and not having some of the features of pyroscope so adding a step to the readmes to explicitly update the grafana image


knylander-grafana (approved):
Thank you for updating the READMEs!

knylander-grafana:
I'll go ahead and merge this.

PR #3675, opened 2024-11-12
feat(v2): read path improvements

aleks-p:
Addresses a few issues encountered in the read path for v2:

1. Simplifies the query plan representation. The current approach encodes the graph for the query plan as a int32 array which makes it difficult to reason about and maintain. We now have a straightforward node graph in place.
2. Replaces the static concurrency limit in the query backend with the gradient2 implementation from [go-concurrency-limits](https://github.com/platinummonkey/go-concurrency-limits). This adds some elasticity to the query backend under high load and should allow for fine tuning in the future. The limiter parameters are internal for now, but might be exposed in the future.
3. Updates the query backend client config to reduce aggressive retries. This gives large queries a better chance of succeeding by slowing them down. Further tuning will likely be needed here in the future.


pkg/experiment/query_backend/query_plan/query_plan.go
```diff
@@ -77,270 +22,74 @@ func Open(p *queryv1.QueryPlan) *QueryPlan {
...
-			// the first node are _always_ placed at the very
-			// beginning (for both read and merge nodes).
-			nodes[i].off = 0
-			o[c.typ] = int(c.len)
-		} else {
-			nodes[i].off = uint32(o[c.typ])
-			o[c.typ] += int(c.len)
+	// create leaf nodes and spread the blocks in a uniform way
+	var leafNodes []*queryv1.QueryNode
+	for i := 0; i < len(blocks); i += maxReads {
```
kolesnikovae, line 41: This is an optimisation we can make later:

The breakdown is not entirely uniform, as is also the case in the source version. If we have `maxReads` = 20, and 101 blocks, the last node will have just 1 block.

I had this snippet (but I'm sure there's a better way):
```go
func Split[S ~[]E, E any](slice S, batchSize int) []S {
	size := len(slice)
	if size == 0 || batchSize < 1 {
		return nil
	}
	n := int(math.Ceil(float64(size) / float64(batchSize)))
	base := size / n
	remainder := size % n
	batches := make([]S, 0, n)
	var offset int
	for i := 0; i < n; i++ {
		end := offset + base
		if i < remainder {
			// Distribute remaining elements among the first few batches.
			end++
		}
		batches = append(batches, slice[offset:end])
		offset = end
	}
	return batches
}
```

[Example](https://go.dev/play/p/msFfGzOg5K6):
```
data   []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}
batch  3

Batch 1: [1 2 3]
Batch 2: [4 5 6]
Batch 3: [7 8]
Batch 4: [9 10]
```

Probably the same for merge nodes, as far as I understand.

kolesnikovae (approved):
Awesome work Aleks, thank you so much for solving this!

PR #3691, opened 2024-11-14
feat(v2): add stack trace selector support

aleks-p:
Adding it for the pprof endpoint only, I don't think there are any real usages of this in SelectSeries (I could be wrong).


kolesnikovae (approved):
LGTM

PR #3698, opened 2024-11-15
examples: add a java span profiling example without the Pyroscope sdk

aleks-p:
Adding a simpler example for Java span profiling, where we only load the OTEL agent and the Pyroscope OTEL extension. The example lives in `examples/tracing/java`.

It might be useful to move the other tracing examples (go, dotnet & python) to their own folders as well. This makes them easier to digest and run.


bryanhuhta (approved):
LGTM

marcsanmi (approved):

PR #3699, opened 2024-11-17
chore(v2): raft module refactoring

kolesnikovae:
The change moves raft component to a separate package `metastore/raftnode`, `RaftNodeService` is refactored as well. In addition, the state reader abstraction should prevent any confusion mentioned in [#3502](https://github.com/grafana/pyroscope/pull/3679#pullrequestreview-2434527731).

The change also fixes the way FSM is initialized: now that we explicitly separate read-only and write transaction, the state should be initialized explicitly as well.


pkg/experiment/metastore/fsm/boltdb.go
```diff
@@ -53,9 +53,15 @@ func (db *boltdb) open(readOnly bool) (err error) {
 	}
 
 	opts := *bbolt.DefaultOptions
+	// open is called with readOnly=true to verify the snapshot integrity.
```
aleks-p, line 56: I think we call open with readOnly=true and readOnly=false in different contexts, can we expand this comment to make that more clear?

aleks-p (approved):
Looks great!

pkg/experiment/metastore/fsm/boltdb.go
```diff
@@ -35,6 +35,15 @@ func newDB(logger log.Logger, metrics *metrics, dir string) *boltdb {
...
 
+// open creates a new or opens an existing boltdb database.
+//
+// The only case when we open the database in read-only mode is when we
+// restore it from a snapshot: before closing the database being in-use,
+// we open the snapshot in read-only mode to verify its integrity.
+//
+// Read-only mode guarantees the snapshot won't be corrupted by the current
+// process, and allows loading the database more quickly (by skipping the
+// free page list preload).
```
kolesnikovae, line 46:
```suggestion
// open creates a new or opens an existing boltdb database.
//
// The only case in which we open the database in read-only mode is when we
// restore it from a snapshot: before closing the database in use, we open
// the snapshot in read-only mode to verify its integrity.
//
// Read-only mode guarantees the snapshot won't be corrupted by the current
// process and allows loading the database more quickly (by skipping the
// free page list preload).
```

PR #3709, opened 2024-11-19
chore: move VCS service handler from querier to query frontend

aleks-p:
This will make the migration path to v2 easier, since in v2 the querier component is no longer required.

@bryanhuhta let me know if I am missing anything


bryanhuhta (approved):
LGTM

PR #3720, opened 2024-11-26
fix: handle non-pointer values in the flag parser

kolesnikovae:
Currently, the parser panics in certain cases. For example, [here](https://github.com/grafana/pyroscope/blob/00f81b9eca840bfa958c8c1c659bddee5c00bc0f/pkg/experiment/compactor/compaction_worker.go#L60), a struct value is used as a flag implementing the `encoding.Text(Un)marshaler` _interface_.

It's recommended to use `Value.UnsafePointer()` to get `uintptr` of a value via reflection.

```go
// Pointer returns v's value as a uintptr.
// It panics if v's Kind is not [Chan], [Func], [Map], [Pointer], [Slice], or [UnsafePointer].
//
// If v's Kind is [Func], the returned pointer is an underlying
// code pointer, but not necessarily enough to identify a
// single function uniquely. The only guarantee is that the
// result is zero if and only if v is a nil func Value.
//
// If v's Kind is [Slice], the returned pointer is to the first
// element of the slice. If the slice is nil the returned value
// is 0.  If the slice is empty but non-nil the return value is non-zero.
//
// It's preferred to use uintptr(Value.UnsafePointer()) to get the equivalent result.
func (v Value) Pointer() uintptr
```


aleks-p:
Based on the context, this seems related to docs. Do we need to re-run doc generation or did the panic have no impact on this?

tools/doc-generator/parse/parser.go
```diff
@@ -115,8 +115,12 @@ func Flags(cfg flagext.Registerer, logger log.Logger) map[uintptr]*flag.Flag {
...
 		}
 
-		ptr := reflect.ValueOf(f.Value).Pointer()
-		flags[ptr] = f
+		val := reflect.ValueOf(f.Value)
+		if val.CanAddr() {
+			flags[val.UnsafeAddr()] = f
+		} else if val.Kind() == reflect.Ptr {
+			flags[val.Pointer()] = f
+		}
```
marcsanmi, line 123: Should we log or handle the case where a value is neither addressable nor a pointe (even this unlikely to happen considering the current flags usage).

marcsanmi (approved):
LGTM

kolesnikovae:
> Based on the context, this seems related to docs. Do we need to re-run doc generation or did the panic have no impact on this?

Yes, the flag parser is used to generate docs. This issue fix has no impact as long as v2 flag is not set. I discovered the problem when was testing v2 configs (just wanted to make sure that the options have meaningful description). This is a bug-to-bug compatible change :)

tools/doc-generator/parse/parser.go
kolesnikovae: Added one more option and a panic if we can't get the address

Issue #3721, opened 2024-11-26
Align `/ingest` HTTP API validation with Push API


kolesnikovae:
The `/ingest` HTTP API has a long history and imposes stricter requirements on the `service_name` label (previously referred to as the "app name" in pre-v1 versions). Specifically, in the Push API, `service_name` supports UTF-8, whereas in the `/ingest` HTTP API, the allowed character set is more restricted (e.g., it does not permit / in the service name).

kolesnikovae:
Related: https://github.com/grafana/alloy/issues/2158

kolesnikovae:
Should be addressed now. Reopen if it surfaces again.

PR #3722, opened 2024-11-27
fix(flameql): allow slashes in tag name

marcsanmi:
This was originally reported in Alloy grafana/alloy#2158.

The change extends service name validation to allow slashes, enabling hierarchical service naming patterns. After this change is validated, I'll create the proper PR in Alloy since the validation isn't reused in Alloy but ported from Pyroscope.

Closes grafana/pyroscope#3721

### DIFF
--- a/pkg/og/flameql/flameql.go
+++ b/pkg/og/flameql/flameql.go
@@ -97,7 +97,7 @@ func ValidateAppName(n string) error {
 }
 
 func IsTagKeyRuneAllowed(r rune) bool {
-	return (r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') || (r >= '0' && r <= '9') || r == '_' || r == '.'
+	return (r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') || (r >= '0' && r <= '9') || r == '_' || r == '.' || r == '/'
 }
 
 func IsAppNameRuneAllowed(r rune) bool {
--- a/pkg/og/flameql/flameql.go
+++ b/pkg/og/flameql/flameql.go
@@ -97,11 +97,11 @@ func ValidateAppName(n string) error {
 }
 
 func IsTagKeyRuneAllowed(r rune) bool {
-	return (r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') || (r >= '0' && r <= '9') || r == '_' || r == '.' || r == '/'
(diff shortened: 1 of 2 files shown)


pkg/og/flameql/flameql.go
```diff
@@ -97,7 +97,7 @@ func ValidateAppName(n string) error {
 }
 
 func IsTagKeyRuneAllowed(r rune) bool {
-	return (r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') || (r >= '0' && r <= '9') || r == '_' || r == '.'
+	return (r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') || (r >= '0' && r <= '9') || r == '_' || r == '.' || r == '/'
```
aleks-p, line 100: Is `IsTagKeyRuneAllowed` used for both label names and values? If yes, we probably want to change this. We should be more restrictive with label names and less with label values. E.g., we want to allow `-` in label values (a very common case if for services to be named `namespace-001/application-name`). If I am not mistaken, we currently allow any UTF-8 character (for label values) in the /Push path.

marcsanmi: After checking the code path more carefully, I see that label values aren't actually validated at all. If I'm not wrong, the `IsTagKeyRuneAllowed` is only used for:

- Label keys (via ValidateTagKey)
- App names (via ValidateAppName which uses `IsTagKeyRuneAllowed` plus some extra characters)

For our specific [issue](https://github.com/grafana/alloy/issues/2158) with app names being rejected when they contain slashes, we just need to allow slashes in `IsAppNameRuneAllowed`:
```golang
func IsAppNameRuneAllowed(r rune) bool {
    return r == '-' || r == '.' || r == '/' || IsTagKeyRuneAllowed(r)
}
```
Your point about label keys validation right but I'd say it's a separate enhancement we could tackle later (?) Related to prometheus allowing UTF-8 characters. Does it make sense?

aleks-p: > Your point about label keys validation right but I'd say it's a separate enhancement we could tackle later (?) Related to prometheus allowing UTF-8 characters. Does it make sense?

To clarify, I didn't mean we should change label key validation right now, I was just saying it should be more restrictive than the one for label values in general. 

The change in 83b1701 makes sense to me!

aleks-p (approved):
LGTM

PR #3731, opened 2024-11-30
fix: `enable-ipv6` option is ignored

kolesnikovae:
Our configuration includes an option to enable IPv6 interface discovery when a component's advertised address is not explicitly defined. However, for unknown reasons, this option is not currently utilized, which can be quite confusing.

This PR addresses the issue by ensuring the relevant configuration parameters are properly applied. It is a no-op in most cases but may impact deployments intended to use IPv6.


docs/sources/configure-server/reference-configuration-parameters/index.md
```diff
@@ -543,6 +543,183 @@ pool_config:
...
   # CLI flag: -distributor.health-check-timeout
   [remote_timeout: <duration> | default = 5s]
+
+ring:
+  # The key-value store used to share the hash ring across multiple instances.
+  kvstore:
+    # Backend storage to use for the ring. Supported values are: consul, etcd,
+    # inmemory, memberlist, multi.
+    # CLI flag: -distributor.ring.store
+    [store: <string> | default = "memberlist"]
```
kolesnikovae, line 553: For some reason, the distributor's ring configuration was omitted from the documentation, even though it is still available in the CLI help message. I've added it to the config spec doc.

docs/sources/configure-server/reference-configuration-parameters/index.md
```diff
@@ -846,7 +1023,16 @@ The `query_frontend` block configures the query-frontend.
 # IP address to advertise to the querier (via scheduler) (default is
 # auto-detected from network interfaces).
 # CLI flag: -query-frontend.instance-addr
-[address: <string> | default = ""]
+[instance_addr: <string> | default = ""]
```
kolesnikovae, line 1026: I bet many people have been caught by this unfortunate inconsistency: we use `instance_addr` everywhere else. In the CLI, it's actually `instance-addr`, what makes things even more confusing.

To prevent any future confusion, I’m changing address to `instance_addr` while maintaining backward compatibility with the previous version.

korniltsev (approved):
lgtm

PR #3737, opened 2024-12-02
chore(v2): Add span tags into BlockReader.Invoke

simonswine:
Was fairly useful, giving v2 a spin locally. As it improves the detail of query information that can be seen in distributed tracing.

I might be better placed in the earlier Invoke calls, but I this one has the validation and is closet to the actual work being done, hence I put it there.


kolesnikovae (approved):
Thanks for the addition @simonswine! We definitely should review the tracing instrumentation in the new components and add any missing parts.

PR #3741, opened 2024-12-05
fix: otlp pprof fixes 

korniltsev:
- Do not append sample labels to series labels (imagine two samples with process=firefox and process=chrome, these attributes should not be appended to Push request series labels. Instead add them as sample labels.
- Fix function name for unsymbolized functions, previously it was incorrectly using only one function per mapping. Now we create multiple (a lot of them actually) functions with format `libfoo.so 0x...`
- Remove debugging pprof dump to fs


pkg/ingester/otlp/convert.go
```diff
@@ -48,49 +57,44 @@ func ConvertOtelToGoogle(src *otelProfile.Profile) *googleProfile.Profile {
...
+	}
 
-	functionOffset := uint64(len(dst.Function)) + 1
 	dst.Location = []*googleProfile.Location{}
-	locationMappingIndexAddressMap := make(map[uint64]uint64)
 	// Convert locations and mappings
 	for i, loc := range src.Location {
 		gl := convertLocationBack(loc)
 		gl.Id = uint64(i + 1)
 		if len(gl.Line) == 0 {
```
simonswine, line 77: So my understanding is that we only create those "placeholder" functions currently, as our query path relies on them to deduplicate. 

I think a comment would be nice explaining that.

Maybe something worth considering doing later in the write path. Because if I sent currently Google pprofs without symbols I have the exact same problem. Or am I missing something 😂

simonswine (approved):
LGTM

pkg/ingester/otlp/convert.go
```diff
@@ -30,6 +30,15 @@ func ConvertOtelToGoogle(src *otelProfile.Profile) *googleProfile.Profile {
...
 	}
+	stringmap := make(map[string]int)
+	addstr := func(s string) int64 {
+		if _, ok := stringmap[s]; !ok {
+			stringmap[s] = len(dst.StringTable)
+			dst.StringTable = append(dst.StringTable, s)
+		}
+		return int64(stringmap[s])
+	}
+	addstr("")
```
marcsanmi, line 41: Is this needed?

pkg/ingester/otlp/convert.go
marcsanmi: Hmm... I didn't test it, but it should be the same if we happen to receive pprof without symbols (lines) 😅

marcsanmi (approved):
LGTM 👍

pkg/ingester/otlp/convert.go
korniltsev: I think it is better to do it explicitly for https://github.com/google/pprof/blob/1e3fdc7de4679692c8c3bafdb87274f45090c73a/proto/profile.proto#L66

petethepig (approved):
LGTM, thanks for doing this!

pkg/ingester/otlp/convert.go
korniltsev: I think frontend or read path just skips these locations.

image

vs 

image

I think it is mostly needed to show the correct shape of the flamegraph - display that there were some frames we just could not name them

PR #3744, opened 2024-12-06
Feat/metadata string interning

kolesnikovae:


simonswine (approved):
LGTM

While I have not fully tested this change, I went through all the changes.

I think your assumptions about the immutability of the partition duration is a worth while one. Thanks for refactoring this and this seems to be a great improvement to the snapshotting.

PR #3745, opened 2024-12-06
NOMERGE:  allow overriding service_name label from sample label

korniltsev:
POC Hack: allow overriding service_name label from sample label
otel POC Hack: add pid(numeric) label


kolesnikovae:
Think this is a good idea, IMO. I've been considering it for a while, waiting for a real use case

korniltsev:
If we go with this, I suspect we'll need to do something differently for v2 , because pprof-split is happening inside a segment which is created per `service_name`

korniltsev:
closed in favor of https://github.com/grafana/pyroscope/pull/3792

PR #3753, opened 2024-12-10
fix(docs): Make sure helm Grafana example works as expected.

### LINKED ISSUE #3523: Revisit default alloy scrape configuration
#### Describe the bug

I'm trying to follow the getting started guide here: https://grafana.com/docs/pyroscope/latest/deploy-kubernetes/helm/

However I'm finding it quite challenging to actually get Alloy to scrape Grafana itself.

There's a few issues in the docs, but I've managed to work around this. The main one being the annotations being used are `pyroscope.grafana.com` but the default config is expecting `profiles.grafana.com` along with the profile type being specified. However, even after fixing the pod annotations, it's still not working.

I believe the issue is because of the following line that's added to the Alloy scrape config: https://github.com/grafana/pyroscope/blob/main/operations/pyroscope/helm/pyroscope/templates/configmap-alloy.yaml#L86-L90

It seems to be the case that the Alloy config is expecting a container port to be exposed that matches the port configured in the scrape annotations. The problem is, I don't see a way to configure additional ports ion the Grafana pod via the Grafana helm chart: https://github.com/grafana/helm-charts/blob/f2ee54ada5fc925f224dc38986b28c237adaeaa5/charts/grafana/templates/_pod.tpl#L1012-L1021

#### To Reproduce

1. Follow the steps in https://grafana.com/docs/pyroscope/latest/deploy-kubernetes/helm/
2. Notice Grafana is not being scraped.

#### Expected behavior

Grafana profiles can be scraped.

#### Environment

- Infrastructure:Kubernetes
- Deployment tool: helm

#### Additional details

I think it's a pretty common occurance the pods do not define the profiling ports in the `ports` section of their containers. It's often overlooked when thinking about the ports a container exposes, especially since profiling is optional. It probably doesn't make sense to rely on the container ports defined on the pod, and to simply trust the annotation is referencing a port that is actually listening. This is how it worked for Phlare, and it makes things a lot simpler.

simonswine:
The problem with the grafana helm is that we don't have the profiling port exposed as part of the containerPorts of the container. So I had to reuse/abuse the gossip-tcp port in order to get it to be scraped.

It also was using the wrong annotation of "pyroscope.grafana.com" instead of "profiles.grafana.com".

Fixes #3523

### DIFF
--- a/docs/sources/deploy-kubernetes/helm.md
+++ b/docs/sources/deploy-kubernetes/helm.md
@@ -128,9 +128,13 @@ Use a custom namespace so that you don't have to overwrite the default namespace
      --set env.GF_AUTH_ANONYMOUS_ORG_ROLE=Admin \
      --set env.GF_DIAGNOSTICS_PROFILING_ENABLED=true \
      --set env.GF_DIAGNOSTICS_PROFILING_ADDR=0.0.0.0 \
-     --set env.GF_DIAGNOSTICS_PROFILING_PORT=6060 \
-     --set-string 'podAnnotations.pyroscope\.grafana\.com/scrape=true' \
-     --set-string 'podAnnotations.pyroscope\.grafana\.com/port=6060'
+     --set env.GF_DIAGNOSTICS_PROFILING_PORT=9094 \
+     --set-string 'podAnnotations.profiles\.grafana\.com/cpu\.scrape=true' \
+     --set-string 'podAnnotations.profiles\.grafana\.com/cpu\.port=9094' \
+     --set-string 'podAnnotations.profiles\.grafana\.com/memory\.scrape=true' \
+     --set-string 'podAnnotations.profiles\.grafana\.com/memory\.port=9094' \
+     --set-string 'podAnnotations.profiles\.grafana\.com/goroutine\.scrape=true' \
+     --set-string 'podAnnotations.profiles\.grafana\.com/goroutine\.port=9094'
    ```
 
    For details, see [Deploy Grafana on Kubernetes](/docs/grafana/latest/setup-grafana/installation/kubernetes/).


docs/sources/deploy-kubernetes/helm.md
```diff
@@ -128,9 +128,13 @@ Use a custom namespace so that you don't have to overwrite the default namespace
...
-     --set env.GF_DIAGNOSTICS_PROFILING_PORT=6060 \
-     --set-string 'podAnnotations.pyroscope\.grafana\.com/scrape=true' \
-     --set-string 'podAnnotations.pyroscope\.grafana\.com/port=6060'
+     --set env.GF_DIAGNOSTICS_PROFILING_PORT=9094 \
+     --set-string 'podAnnotations.profiles\.grafana\.com/cpu\.scrape=true' \
+     --set-string 'podAnnotations.profiles\.grafana\.com/cpu\.port=9094' \
+     --set-string 'podAnnotations.profiles\.grafana\.com/memory\.scrape=true' \
+     --set-string 'podAnnotations.profiles\.grafana\.com/memory\.port=9094' \
+     --set-string 'podAnnotations.profiles\.grafana\.com/goroutine\.scrape=true' \
+     --set-string 'podAnnotations.profiles\.grafana\.com/goroutine\.port=9094'
```
marcsanmi, line 137: Nit: Instead of reusing the gossip-tcp port, do you think leveraging `extraExposePorts` from Grafana helm chart would work to properly expose the profiling port? I didn't try, but I believe it could be added via Helm command as:
```yaml
--set extraExposePorts[0].name=profiling,extraExposePorts[0].port=6060,extraExposePorts[0].targetPort=6060
```

marcsanmi (approved):
LGTM 👍

docs/sources/deploy-kubernetes/helm.md
simonswine: I am afraid that won't do it. extraExposePorts works on service level, we would need it on the container around here: https://github.com/grafana/helm-charts/blob/897c25093da0572716a7cd441e8d3c3dba064b98/charts/grafana/templates/_pod.tpl#L1081

I am planning of submitting something to that chart repo, but that will be a while until it comes through...

simonswine: There we go https://github.com/grafana/helm-charts/pull/3472

PR #3760, opened 2024-12-11
chore(examples): add cron job to update examples

korniltsev:
This will create PR's like this https://github.com/korniltsev/pyroscope/pull/7/files

### DIFF
--- a/.github/workflows/update-examples-cron.yml
+++ b/.github/workflows/update-examples-cron.yml
@@ -1,0 +1,27 @@
+name: Update Examples Cron
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  update-examples-cron:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - run: |
+          make tools/update_examples
+          if ! git diff --exit-code;
+          then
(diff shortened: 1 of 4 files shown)


.github/workflows/update-examples-cron.yml
```diff
@@ -0,0 +1,27 @@
...
+        with:
+          persist-credentials: false
+      - run: |
+          make tools/update_examples
+          if ! git diff --exit-code;
+          then
+            make tools/update_examples_pr
+          fi
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
simonswine, line 26: Unsure if this is the right token, this does something very similar: https://github.com/grafana/pyroscope/blob/9b5c6d7f8c7eadf9bcb596c4b0ffc3f5b5607ddb/.github/workflows/update-contributors.yml#L32

simonswine (approved):
LGTM

The hard part remains the testing if it still compile/runs/sends profiles. When I ran this locally I mostly ran the one tracing example that contained all of the SDK, but we are breaking that up 😂

aleks-p:
> The hard part remains the testing if it still compile/runs/sends profiles. When I ran this locally I mostly ran the one tracing example that contained all of the SDK, but we are breaking that up 😂

Technically we can still maintain a setup with all SDKs together. I don't think it is the best example to show users, but we could convert it to an integration test of some sort (outside of this PR).

korniltsev:
> compile/runs/sends profiles

It would be nice if we could automate this somehow

PR #3764, opened 2024-12-11
examples: add python span profiling

marcsanmi:
### Note for reviewers
I'm currently reusing the Golang load generator since it's already instrumented with OpenTelemetry for tracing and profiling. 

Should we copy the load generator code into this example instead of referencing it? (I could have done the same with the Flask code, but I'd prefer the example to be self-contained...). That's why I'm leaning towards adding the load generator in this example as well.

Note that `span_name` is only available for the load generator since `otel-profiling-python` doesn't add it as label unlike `otel-profiling-go`.

### DIFF
--- a/examples/tracing/python/README.md
+++ b/examples/tracing/python/README.md
@@ -6,6 +6,8 @@ The docker compose consists of:
 - Pyroscope for continuous profiling
 - Grafana for visualization
 - Load Generator for simulating traffic
+  
+For a detailed guide about Python span profiles configuration, refer to the docs [Pyroscope documentation](https://grafana.com/docs/pyroscope/latest/configure-client/trace-span-profiles/python-span-profiles/).
  
 The `rideshare` app generates traces and profiling data that should be available in Grafana.
 Pyroscope and Tempo datasources are provisioned automatically.
@@ -34,11 +36,6 @@ and have `pyroscope.profile.id` attribute set to the corresponding span ID.
 Please note that presence of the attribute does not necessarily
 indicate that the span has a profile: stack trace samples might not be collected, if the utilized CPU time is
 less than the sample interval (10ms).
-
-### Instrumentation
-
-- `rideshare` demo application instrumented with OpenTelemetry: [OTel integration](https://github.com/grafana/otel-profiling-python)
-- `pyroscope` itself is instrumented with opentracing-go SDK and [spanprofiler](https://github.com/grafana/dskit/tree/main/spanprofiler) for profiling integration.
 
 ### Grafana Tempo configuration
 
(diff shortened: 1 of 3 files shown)


simonswine:
> Should we copy the load generator code into this example instead of referencing it? (I could have done the same with the Flask code, but I'd prefer the example to be self-contained...). That's why I'm leaning towards adding the load generator in this example as well.

Maybe I remembered it wrong, but in one of the team meetings I think we spoke about loadgen go vs. python and that we should use the go one everywhere, but I might be misremembering that. I personally think it is ok to reduce it to go. The python example is mainly the rideshare.

examples/tracing/python/docker-compose.yaml
```diff
@@ -0,0 +1,85 @@
+services:
+  pyroscope:
+    image: grafana/pyroscope
+    ports:
+      - "4040:4040"
+
+  rideshare-python-us-east:
```
simonswine, line 7: The existing java one and my #3765 uses just the region name as container name. Not too sure what is better, but we should do the same everywhere:

```suggestion
  us-east:
```

simonswine (approved):
LGTM

examples/tracing/python/README.md
```diff
@@ -0,0 +1,57 @@
+# Span Profiles with Grafana Tempo and Pyroscope
```
simonswine, line 1: I think the instructions also said we should link back and forth between https://grafana.com/docs/pyroscope/latest/configure-client/trace-span-profiles/python-span-profiles/ and this README.md

examples/tracing/python/README.md
```diff
@@ -0,0 +1,57 @@
...
+By default, only the root span gets labeled (the first span created locally): such spans are marked with the link icon
+and have `pyroscope.profile.id` attribute set to the corresponding span ID.
+Please note that presence of the attribute does not necessarily
+indicate that the span has a profile: stack trace samples might not be collected, if the utilized CPU time is
+less than the sample interval (10ms).
+
+### Instrumentation
+
+- `rideshare` demo application instrumented with OpenTelemetry: [OTel integration](https://github.com/grafana/otel-profiling-python)
+- `pyroscope` itself is instrumented with opentracing-go SDK and [spanprofiler](https://github.com/grafana/dskit/tree/main/spanprofiler) for profiling integration.
```
simonswine, line 41: I know you just taken that bit from `java/`, but as far as I see we don't enable tracing for Pyroscope itself. Wonder if we should be. Wdyt?

marcsanmi: Yes, good catch. I think we should either remove it from the README.md or add it in all the tracing examples 🤔 Do you think adding it would provide significant value to these examples?

aleks-p: I think the Java one was likely copied from the one for Go, I don't think we need this at all (we don't necessarily care how Pyroscope is instrumented in these examples, the focus is on the rideshare application).

PR #3765, opened 2024-12-11
chore(examples): Update all golang examples and split out golang-push tracing

simonswine:
* Split out tracing, from the tempo example into dedicated golang push
* Add all examples to go mod tidy and update go libraries 

Things I checked are:

Explore (classic) -> Tempo
* Can I find trace in Explore (classic) -> Tempo?
* Do I see flamegraphs within the trace spans? (Picture 2)
* Does the span to profiles link work

Explore > Profiles
* Do we have 3 services (pyroscope, loadgen and rideshare) ?
* Is the rideshare service having region, span_name and vehice labels? (Picture 1)
* Is car significantly slower than bike & scooter  (Picture 1)


examples/tracing/golang-push/README.md
```diff
@@ -0,0 +1,14 @@
+## Golang Example
```
marcsanmi, line 1: I like this concise README.md. However, I followed the approach of writing a more detailed (likely unnecessary) one following java's approach.

I think we should make all READMEs in the `/tracing` directory consistent. I can update both my example and the Java one to follow this simpler approach, wdyt?

marcsanmi (approved):
LGMT 👍

examples/tracing/golang-push/README.md
aleks-p: Fully agree!

Though perhaps this is a bit too short and we want to explain what this example is for.

kolesnikovae (approved):
LGTM!

Also, we need to update the README in:
* https://github.com/grafana/otel-profiling-go
* https://github.com/grafana/dskit/tree/main/spanprofiler

as they refer to the old path

examples/tracing/golang-push/README.md
kolesnikovae: +1

simonswine: Yeah I was not planning to keep it like that, but I also did't necessary like the other one 😂  Fill follow up with a PR

PR #3773, opened 2024-12-12
test(examples): Introduce some basic testing for examples/

simonswine:
I am sure this is something that we all felt during this docs week: It is quite hard keeping the examples working.

I spent a bit of time automating building and running the examples:

My observation was often problems happen already at built or startup stage. So for now that will only cover problems during docker-compose build and will only detect failures that occur during the first 5 seconds after docker-compose up.

### DIFF
--- a/.github/workflows/test-examples.yml
+++ b/.github/workflows/test-examples.yml
@@ -1,0 +1,27 @@
+name: Test Examples
+on:
+  workflow_dispatch:
+  # TODO: consider cronjob to run this at night
+  push:  # THIS SHOULD BE REMOVED BEFORE MERGE
+    branches:
+      - 20241212_testing-examples
+
+concurrency:
+  # Cancel any running workflow for the same branch when new commits are pushed.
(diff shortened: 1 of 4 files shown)


.github/workflows/test-examples.yml
```diff
@@ -0,0 +1,23 @@
+name: Test Examples
+on:
+  workflow_dispatch:
```
marcsanmi, line 3: Would it make sense to set this as a cron job as well? This way we could still manually run it when needed, but also have it run periodically to ensure we don't forget to check the examples

marcsanmi (approved):
LGTM, this will be really useful 👍

.github/workflows/test-examples.yml
simonswine: Yes good point, but I am not too sure if we will notice that is breaking. [c0122ab](https://github.com/grafana/pyroscope/pull/3773/commits/c0122ab82f38f2eda0aedba5ef76ea246eaebb06)

PR #3777, opened 2024-12-13
docs: ebpf-otel fixes

korniltsev:
- use one dockerfile, build the profiler in the dockerfile, remove prerequisites to build the profiler and copy the profiler separately
- add screenshot
- fix profiler,pyroscope,collector versions. It does not work on the latests.

Notes: ebpf profiler does not properly work in minikube. I've tested in a real k8s cluster by pushing an image to the public docker repo. I am not sure we should advertise minikube in the example if it does not work. We can consider submiting a PR with pid namespace inspection from pyroscope ebpf profiler.


examples/grafana-alloy-auto-instrumentation/ebpf-otel/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM ubuntu:22.04 as builder
+RUN apt-get update && apt-get -y install wget gcc
+RUN wget https://go.dev/dl/go1.22.10.linux-amd64.tar.gz
+RUN tar -C /usr/local -xzf go1.22.10.linux-amd64.tar.gz
+# fixed versions for pyroscope, otel-collector, otel-profiler due to protocol changes
+RUN wget https://github.com/open-telemetry/opentelemetry-ebpf-profiler/archive/9eb454541a3cf1282a8755dbcec9a9471b243e83.tar.gz
```
marcsanmi, line 6: I like this. However, I think we should update to avoid working with pinned versions and images once the profiler is stable enough

examples/grafana-alloy-auto-instrumentation/ebpf-otel/README.md
```diff
@@ -67,10 +39,10 @@ minikube image load test-ebpf-profiler:latest
 2. Deploy to Kubernetes:
 ```bash
```
marcsanmi, line 40: What errors did you find on `minikube`? Didn't it run at all?
It worked to me when I tested it some weeks ago... 😅

marcsanmi (approved):
LGTM 👍

examples/grafana-alloy-auto-instrumentation/ebpf-otel/README.md
korniltsev: it shows only kernel stacktraces. The minikube is running in a nested pid namespace and the ebpf program is reporting processess with PIDs from host pid namesapce and userspace is unable to find such processes

PR #3769, opened 2024-12-12
chore(examples): update python examples

kolesnikovae:

PR #3778, opened 2024-12-13
chore(examples): rename flask-ride-sharing-app in tutorial

kolesnikovae:
In #3769 I renamed the `flask-ride-sharing-app` to `ride-sharing-app` to align with the shared README, where it is referred to in this way. However, in the KillerCoda tutorial, we're still using `flask-ride-sharing-app`. I'm updating the tutorial to reflect the new name and add a link to README.


simonswine (approved):
LGTM

Was not really aware of killercoda.com. I imagine we need to backport those PRs, in order to update the "latest" version of the docs.

PR #3780, opened 2024-12-16
examples: add ruby span profiling

marcsanmi:
Adds Ruby span profiling.

- Added OpenTelemetry SDK configuration with OTLP HTTP exporter to send traces to Tempo
- Created custom handler spans with meaningful names (BikeHandler, CarHandler, etc.)
- Implemented trace context propagation to link handler spans with parent load generator traces


simonswine (approved):
LGTM

Issue #3451, opened 2024-07-31
Add BucketLookupType for s3 bucket


devmentality:
#### Is your feature request related to a problem? Please describe.

Hi, Pyroscope Team!
I encountered a problem with s3 configuration. We have an on-premise Ceph installation as our company's s3 provider. It supports only virtual-host-style requests. But it seems that underlying minio library, which is used in Pyroscope, chooses to use path-style requests. 

Why minio-go uses path-style requests in our case? 
By default Pyroscope uses AutoLookup strategy, which resolves to VirtualHostLookup or PathLookup by minio-go library. As I can see from the [function](https://github.com/minio/minio-go/blob/master/pkg/s3utils/utils.go#L71), the library chooses virtual-host style only for AWS, Google and Alibaba provider. It means that the library's AutoLookup strategy will choose path-style requests for my on-premise s3.

#### Describe the solution you'd like

It would be great to have `s3.bucket_lookup_type` configuration parameter to force Pyroscope use virtual-host-style requests. 
Example of such a parameter in Mimir: https://github.com/grafana/mimir/pull/7684/files#diff-8a8879ab8601d8d13855963d239c72b9f082bd2be2889bbbce3282d05fcddf7d. 

#### Additional context

I see that Pyroscope already has `s3.force-path-style` flag, which is has the opposite meaning to what we need. 
I can help with this feature by creating a PR, but we need to discuss how new parameter will get along with `s3.force-path-style` first.

simonswine:
@devmentality yes I agree with your suggestions and would more than welcome a PR from you ❤️ 

Not fully sure if this has drawbacks that I missed but, I would make sure:

* Error when force-path is true and lookup-type is `virtual-host`.
* To display a warning if force-path is set to true, that this flag is deprecated and might be removed going forward. We then use lookup mode `path` internally.

lzyan:
We also need this feature，hope to be supported by offical

dozer47528:
Same problem

devmentality:
Hi @simonswine!
Implemented this feature in the mentioned above PR

simonswine:
Yes thanks for your PR, is released as v1.12.0

PR #3788, opened 2024-12-17
Added bucket-lookup-type field to s3 config

devmentality:
Hi Pyroscope Team!
This pull request targets the [issue](https://github.com/grafana/pyroscope/issues/3451). 
There are a few details I would like to point out:
- We have discussed previously that it would be great to show `a warning if force-path is set to true, that this flag is deprecated and might be removed going forward`. Unfortunately, I haven't found an elegant solution to do that and just put a deprecation message to flag's usage. There are [flagext](https://github.com/grafana/dskit/blob/main/flagext/deprecated.go) package, but it is used to replace deprecated flag completely. [Pflag](https://github.com/spf13/pflag?tab=readme-ov-file#deprecating-a-flag-or-its-shorthand) library has deprecation feature, but you use plain flag library here.
- It seems that s3 config is not validated with the application start. I may fix that in this PR, if you wish. But it may require thorough testing to ensure that validation won't break anything.

### DIFF
--- a/pkg/objstore/providers/s3/bucket_client.go
+++ b/pkg/objstore/providers/s3/bucket_client.go
@@ -7,6 +7,7 @@ package s3
 
 import (
 	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
 	"github.com/prometheus/common/model"
 	"github.com/thanos-io/objstore"
 	"github.com/thanos-io/objstore/providers/s3"
@@ -19,6 +20,8 @@ func NewBucketClient(cfg Config, name string, logger log.Logger) (objstore.Bucke
 		return nil, err
 	}
 
+	warnForDeprecatedConfigFields(cfg, logger)
+
 	return s3.NewBucketWithConfig(logger, s3Cfg, name)
 }
 
@@ -28,6 +31,8 @@ func NewBucketReaderClient(cfg Config, name string, logger log.Logger) (objstore
 	if err != nil {
 		return nil, err
 	}
+
+	warnForDeprecatedConfigFields(cfg, logger)
 
 	return s3.NewBucketWithConfig(logger, s3Cfg, name)
 }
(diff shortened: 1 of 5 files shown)


pkg/objstore/providers/s3/config.go
```diff
@@ -108,6 +121,14 @@ func (cfg *Config) Validate() error {
 		return errUnsupportedSignatureVersion
 	}
 
+	if cfg.ForcePathStyle && cfg.BucketLookupType == VirtualHostedStyleLookup {
```
simonswine, line 124: Nit: Might be slightly more failsafe, once there is another lookup type added (I hope not 😆 )

```suggestion
	if cfg.ForcePathStyle && cfg.BucketLookupType != AutoLookup && cfg.BucketLookupType != PathStyleLookup {
```

pkg/objstore/providers/s3/bucket_client.go
```diff
@@ -39,8 +39,10 @@ func newS3Config(cfg Config) (s3.Config, error) {
 	}
 
 	bucketLookupType := s3.AutoLookup
-	if cfg.ForcePathStyle {
+	if cfg.ForcePathStyle || cfg.BucketLookupType == PathStyleLookup {
```
simonswine, line 42: I wonder if we also should add a deprecation log  warning when ForcePath  argument is used, so users can notice it and migrate to `BucketLookupType`

Wdyt?

simonswine (commented):
Thank you for working on this ❤️

pkg/objstore/providers/s3/config.go
devmentality: > Nit: Might be slightly more failsafe, once there is another lookup type added (I hope not 😆 )

Fixed

pkg/objstore/providers/s3/bucket_client.go
devmentality: Yeah, that's a great idea. I have tried to add a warning on bucket client initialization. Maybe there is a better place for such logic?

simonswine: I think we could have also put it in validate, but that might have been shown even when no bucket is in use.

So that fine as it is. I noticed btw that we don't validate anything at present 😓  #3837

simonswine (approved):
LGTM

Thank you for your work on this ❤️

PR #3792, opened 2024-12-18
fix(otel): split otel profile to multiple pprof profile if samples have service.name attribute

korniltsev:


pkg/ingester/otlp/convert.go
```diff
@@ -6,197 +6,238 @@ import (
...
+	if i, ok := p.mappingMap[om]; ok {
+		return i
+	}
+
+	gm := &googleProfile.Mapping{
+		MemoryStart:     om.MemoryStart,
+		MemoryLimit:     om.MemoryLimit,
+		FileOffset:      om.FileOffset,
+		Filename:        p.addstr(p.src.StringTable[om.Filename]),
+		BuildId:         p.addstr(p.src.StringTable[om.BuildId]),
```
marcsanmi, line 233: This isn't going to work with the [new spec](https://github.com/open-telemetry/opentelemetry-proto/blob/2bd940b2b77c1ab57c27166af21384906da7bb2b/opentelemetry/proto/profiles/v1development/profiles.proto#L402).

However, not related to this PR since we're relying on `om.BuildId` before. I created an issue last week https://github.com/grafana/pyroscope-squad/issues/295

pkg/ingester/otlp/ingest_handler_test.go
```diff
@@ -314,6 +315,181 @@ func TestSampleAttributes(t *testing.T) {
...
+	}, {
+		Name:       otlpb.addstr("serviceB_func2"),
+		SystemName: otlpb.addstr("serviceB_func2"),
+		Filename:   otlpb.addstr("service_b.go"),
+	}}
+
+	otlpb.profile.LocationIndices = []int64{0, 1, 2, 3}
+
+	// Create two samples with different service.name attributes and different stacktraces
+	otlpb.profile.Sample = []*v1experimental.Sample{{
```
marcsanmi, line 393: Could we add a third sample without service name? We should expect a separate profile with `service_name=unknown`

marcsanmi (approved):
LGTM! 👍

PR #3799, opened 2024-12-20
Prototype(symbolization): Add symbolization in Pyroscope read path

marcsanmi:
## Context
This PR introduces a comprehensive implementation for DWARF symbolization of unsymbolized profiles in the Pyroscope read path. It enables automatic symbolization of profiles for non-customer code (primarily open-source libraries and binaries) where symbol information isn't available at collection time.

### Symbolization
- DWARF parsing: Optimized parsing of debug information with minimal memory overhead
- Comprehensive symbol resolution: Support for function names, file paths, and line numbers
- Inline function resolution: Proper handling of inlined functions for accurate stack traces
- Address-based lookup: Fast address-to-symbol mapping with optimized data structures

### Multi-level Caching
- In-memory symbol cache: LRU cache for frequently accessed symbols
- Object storage for debug files: Persistent storage of debug files with configurable obj storage solution
- Configurable TTL: Control over cache expiration for both memory and storage caches

### Integration Points
- Read path symbolization: Automatically symbolize profiles during query time
- Remote debug info fetching: Integration with debuginfod for symbol discovery from public servers

### Configuration Example
```yaml
# Symbolizer requires the v2 experiment flag
export PYROSCOPE_V2_EXPERIMENT=1
```

```yaml
symbolizer:
  debuginfod_url: "https://debuginfod.elfutils.org"
```

CLI flag
```
# Enable symbolizer globally for all tenants by default
--symbolizer.enabled=true
```


pkg/experiment/symbolization/debuginfod_client.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("unexpected HTTP status: %s", resp.Status)
+	}
+
+	// Save the debuginfo to a temporary file
+	tempDir := os.TempDir()
+	filePath := filepath.Join(tempDir, fmt.Sprintf("%s.elf", buildID))
```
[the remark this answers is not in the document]
petethepig: I'd also add that generally speaking we should also be avoiding file operations as much as possible. If it's possible to do this in memory I'd do this in memory. If the file is too large to fit in memory I'd see if we could stream it

pkg/phlaredb/symdb/resolver.go
```diff
@@ -295,3 +306,18 @@ func (r *Symbols) Tree(
...
+
+func (r *Symbols) SetSymbolizer(sym *symbolizer.Symbolizer) {
+	r.Symbolizer = sym
+}
+
+func (r *Symbols) needsDebuginfodSymbolization(loc *schemav1.InMemoryLocation, mapping *schemav1.InMemoryMapping) bool {
+	if r.Symbolizer == nil {
+		return false
+	}
+	if len(loc.Line) == 0 {
```
korniltsev, line 318: I guess it's time to move this logic (unsymbolized fallback) https://github.com/grafana/pyroscope/blob/466aaa8194e06aab452b00f56efd5ddd8bcff311/pkg/ingester/otlp/convert.go#L166 somewhere here to Symbols/resolver

kolesnikovae (commented):
Good work, Marc! I'm excited to see some experimental results 🚀

I think we can implement a slightly more optimized version for production use:

```mermaid
sequenceDiagram
    autonumber

    participant QF as Query Frontend
    participant M  as Metastore
    participant QB as Query Backend
    participant SYM as Symbolizer

    QF ->>+M: Query Metadata
    Note left of M: Build identifiers are returned<br> along with the metadata records
    M ->>-QF: 

    par
        QF ->>+SYM: Request for symbolication
        Note left of SYM: Prepare symbols for<br>the objects requested
    and
        QF ->>+QB: Data retrieval and aggregation
        Note left of QB: The main data path<br>Might be serverless
    end

    QB ->>-QF: Data in pprof format
    Note over QF: Because of the truncation,<br> only a limited set of locations<br>make it here (16K by default) 

    QF --)SYM: Location addresses
    
    SYM ->>-QF: Symbols
    
    QF ->>QF: Flame graph rendering
```

Even without a parallel pipeline and dedicated symbolication service, we could implement something like this:

```mermaid
sequenceDiagram
    autonumber

    participant QF as Query Frontend
    participant M  as Metastore
    participant QB as Query Backend
    participant SYM as Symbols

    QF ->>+M: Query Metadata
    Note left of M: No build identifiers are returned
    M ->>-QF: 

    QF ->>+QB: Data retrieval and aggregation
    Note left of QB: The main data path<br>Might be serverless

    QB ->>-QF: Data in pprof format
    Note over QF: Because of the truncation,<br> only a limited set of locations<br>make it here (16K by default)

    QF ->>+SYM: Fetch symbols
    SYM ->>-QF: Symbols
    Note over QF: In terms of the added latency,<br>this approach is not worse than<br>block level symbolication
    
    QF ->>QF: Flame graph rendering
```

I think we should avoid symbolization at the block level if the symbols are not already present in the block itself. Otherwise, this approach leads to excessive processing, increased latency, and higher resource usage. Please keep in mind, that a query may span many thousands of blocks.

I won't delve too deeply into how we fetch and process ELF/DWARF files, but I strongly doubt we can bypass the need for an intermediate representation optimized for our access patterns. Additionally, we need a solution to prevent concurrent access to the debuginfod service.

korniltsev:
I have not look into the code yet, but I've tried to run it locally and it looks like it's trying to load a lot of unnecesarry debug files.

I run ebpf profiler with no ontarget symbolization , also run a simple `python -m http.server` to mock debug infod responses.

I then query  only one executable `process_cpu:cpu:nanoseconds:cpu:nanoseconds{service_name="unknown", process_executable_path="/home/korniltsev/.cache/JetBrains/IntelliJIdea2024.2/tmp/GoLand/___go_build_go_opentelemetry_io_ebpf_profiler"}` 

I see 268 GET  requests, with 13 requests to `"GET /buildid/fbce2598b34f1cf8d0c899f34c2218864e1da6d1/debuginfo HTTP/1.1" 200 -` (which is the profiler binary I put into mock server for testing and a bunch of 404 which I assume are build ids for the filles in the other processes which the query does not target. 

Other then that it works \M/  Can't wait to run it in dev.

image

pkg/phlaredb/symdb/resolver.go
marcsanmi: This was changed in https://github.com/grafana/pyroscope/pull/3792.

pkg/experiment/query_backend/backend.go
```diff
@@ -57,13 +62,27 @@ func New(
...
 		backendClient: backendClient,
 		blockReader:   blockReader,
+		symbolizer:    sym,
 	}
+
+	// Pass symbolizer to BlockReader if it's the right type
+	if br, ok := blockReader.(*BlockReader); ok {
+		br.symbolizer = sym
+	}
+
```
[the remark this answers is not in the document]
marcsanmi: I believe this no longer applies since I'll be moving it into the Query Frontend.

liaol:
Hi @marcsanmi , when can this PR be merged? 
Thanks

marcsanmi:
Hi @liaol,
It's still going to take a little while :)

marcsanmi:
I've created this diagram to outline the current Pyroscope Symbolization:
```mermaid
flowchart TD
    A[SymbolizePprof] --> B{Group by Mapping}
    B --> C[Symbolize Request]
    
    subgraph "Symbol Cache Layer (LRU)"
        D{Check Symbol Cache}
        D -->|Cache Hit| E[Return Cached Symbols]
        D -->|Cache Miss| F{Check ObjstoreCache}
    end
    
    C --> D
    
    subgraph "Persistent Cache Layer"
        F -->|Cache Hit| G[Read from ObjstoreCache]
        F -->|Cache Miss| H[Fetch from Debuginfod]
    end
    
    subgraph "Debuginfod Client Layer (Ristretto)"
        H --> I{Check Debuginfod Cache}
        I -->|Cache Hit| J[Return Cached Debug File]
        I -->|Cache Miss| K[HTTP Request to Debuginfod]
        K --> L[Store in Debuginfod Cache]
    end
    
    L --> M[Parse ELF/DWARF]
    G --> M
    
    subgraph "DWARF Resolution Layer"
        M --> N[Resolve Addresses]
        N --> O{Check Address Map}
        O -->|Map Hit| P[Return from Map]
        O -->|Map Miss| Q[Parse DWARF Data]
        Q --> R[Build Lookup Tables]
        R --> S[Store in Address Map]
        S --> T[Return Symbols]
        P --> T
    end
    
    T --> U[Update Symbol Cache]
    U --> V[Return to Caller]
    E --> V
```

pkg/test/integration/testdata/otel-ebpf-profiler-offcpu-cpu.json
kolesnikovae, line 1: You can rename fixtures to `.json.bin` to make GitHub and IDEs to treat the files as binary

kolesnikovae:
I might be missing some details, but I have doubts about the cache hierarchy.

Now it looks like we have: symbols_cache -> object_store -> in_memory_object_store (ristretto) -> debuginfod.

As far as I understand, we're going to read from object_store even if there just a single unresolved address.

I expected to see: symbols_cache -> in_memory_object_store (ristretto) -> object_store -> debuginfod.

Could you please elaborate about the decision?

cmd/pyroscope/help-all.txt.tmpl
alsoba13, line 1: I wonder if we should discard this while symbolizer remains "experimental"

pkg/experiment/symbolizer/cache.go
```diff
@@ -0,0 +1,113 @@
...
+		go func() {
+			delCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			defer cancel()
+			if err = c.bucket.Delete(delCtx, buildID); err != nil {
+				c.metrics.cacheRequestErrorsTotal.WithLabelValues("delete", "delete_error").Inc()
+			}
+		}()
+		return nil, fmt.Errorf("cached object expired")
+	}
+
```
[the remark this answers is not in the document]
marcsanmi: Yes, this should be handled by a separate job that runs periodically, removing unused objects. Instead of relying on last modification, we should track and remove based on last access patterns (afaik, there's no built-in attribute like that); I'll remove this for now and add a TODO.

marcsanmi:
> I might be missing some details, but I have doubts about the cache hierarchy.

You're right @kolesnikovae. I've just realized the problem is that the ristretto cache is coupled inside the debuginfod client. I'll decoupled it and placed it at symbolizer level. Thus, we'll be able to have the following path:

symbols_cache -> in_memory_object_store (ristretto) -> object_store -> debuginfod

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,469 @@
...
+
+	client := NewDebuginfodClient(cfg.DebuginfodURL, metrics)
+
+	// Use configured cache size or default
+	symbolCacheSize := DefaultSymbolCacheSize
+	if cfg.SymbolCacheSize > 0 {
+		symbolCacheSize = cfg.SymbolCacheSize
+	}
+
+	return NewProfileSymbolizer(client, cache, metrics, symbolCacheSize), nil
```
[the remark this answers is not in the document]
marcsanmi: I made objstore cache is optional because the symbolizer might just work with just in-memory caching. Making it opt-in gives us flexibility - simple setups can avoid the operational overhead while larger ones can enable it when needed.. LMKWYT

cmd/pyroscope/help-all.txt.tmpl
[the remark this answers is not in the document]
marcsanmi: Yep good point, thanks! I've just placed them under v2 umbrella in here fd39d9e6fe82cda59c39ffa76aeccda10737c6fa

pkg/experiment/ingester/segment.go
```diff
@@ -499,11 +506,13 @@ func (s *segment) ingest(tenantID string, p *profilev1.Profile, id uuid.UUID, la
 		tenant:  tenantID,
 		service: model.Labels(labels).Get(model.LabelNameServiceName),
 	}
+	ds := s.datasetForIngest(k)
+	ds.needsSymbolization = !hasSymbols(p)
```
kolesnikovae, line 510: I think it should be:
```go
if !ds.needsSymbolization {
  // ...
}
```
Otherwise, if the last profile does not need symbols we reset the flag that could be set previously.

As optimization, we probably should detect it just once in the very end at flush time (instead of calling `hasSymbols` for each profile)

pkg/frontend/read_path/query_frontend/query_frontend.go
```diff
@@ -81,22 +95,87 @@ func (q *QueryFrontend) Query(
...
+					continue
+				}
+
+				// 3. Convert back to TREE if the *original* was TREE
+				//    (You can track the original query from req.Query[i], for example.)
+				if i < len(req.Query) && req.Query[i].QueryType == queryv1.QueryType_QUERY_TREE {
+					treeBytes, err := model.ConvertProfileToTree(&prof, req.Query[i].Tree.MaxNodes)
+					if err == nil {
+						// Store the tree result
+						r.Tree = &queryv1.TreeReport{Tree: treeBytes}
```
liaol, line 157: here must add  `r.ReportType = queryv1.ReportType_REPORT_TREE` , because `querySingle()` in `pkg/frontend/read_path/query_frontend/compat.go`  will check ` x.ReportType == t`

pkg/experiment/ingester/segment.go
```diff
@@ -334,21 +334,31 @@ func concatSegmentHead(f *headFlush, w *writerOffset, s *metadata.StringTable) (
...
+		// lb.WithLabelSet(metadata.LabelNameNeedsSymbolization, "true")
+		lb.WithLabelSet(model.LabelNameServiceName, f.head.key.service, metadata.LabelNameNeedsSymbolization, "true")
+		labelSet := lb.Build()
+		allLabels = append(allLabels, labelSet...)
 	}
 
 	// Other optional labels:
 	// lb.WithLabelSet("label_name", "label_value", ...)
-	ds.Labels = lb.Build()
+	ds.Labels = allLabels
```
kolesnikovae, line 354: Well, `Build` should be called once, after all the labels where added. If it does not work, it's a bug in the label builder that should be fixed

pkg/frontend/read_path/query_frontend/query_frontend.go
marcsanmi: Thanks for checking, I've added it in the latest commit 👍

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,618 @@
...
+					for k, f := range profile.Function {
+						if f.Name == nameIdx && f.Filename == filenameIdx {
+							funcIdx = k
+							break
+						}
+					}
+					if funcIdx == -1 {
+						funcIdx = len(profile.Function)
+						profile.Function = append(profile.Function, &googlev1.Function{
+							Name:      nameIdx,
```
liaol, line 230: missing `Id:        uint64(funcIdx)`

marcsanmi: I have this fixed locally. Today I'll push a new commit with it along with some other  small fixes/improvements. Thanks again for taking a look at this :)

pkg/experiment/ingester/segment.go
marcsanmi: It was a non-related issue to this. Changed it back :)

pkg/experiment/symbolizer/debuginfod_client.go
```diff
@@ -0,0 +1,258 @@
...
+
+	cache, err := ristretto.NewCache(&ristretto.Config{
+		NumCounters: 1e7,     // number of keys to track frequency of (10M)
+		MaxCost:     2 << 30, // maximum cost of cache (2GB)
+		BufferItems: 64,      // number of keys per Get buffer
+	})
+
+	if err != nil {
+		cache = nil
+	}
```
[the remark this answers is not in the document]
marcsanmi: It makes sense; I've just made it mandatory

pkg/experiment/symbolizer/debuginfod_client.go
```diff
@@ -0,0 +1,258 @@
...
+type debuginfodClient struct {
+	cfg     debuginfodClientConfig
+	metrics *Metrics
+
+	// In-memory cache of build IDs to file paths
+	cache *ristretto.Cache
+
+	// Track in-flight requests to prevent duplicate fetches
+	inFlightRequests      map[string]*inFlightRequest
+	inFlightRequestsMutex sync.Mutex
```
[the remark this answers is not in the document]
marcsanmi: Didn't know about it, great suggestion!

pkg/test/integration/testdata/otel-ebpf-profiler-offcpu.json.bin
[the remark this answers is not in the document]
marcsanmi: I've tried with 
```
testdata/* binary
```
```
testdata/* -text
```
And running `git add --renormalize .`, but nothing changed... I'll further investigate.

pkg/experiment/symbolizer/cache.go
```diff
@@ -0,0 +1,113 @@
...
+	Get(ctx context.Context, buildID string) (io.ReadCloser, error)
+	Put(ctx context.Context, buildID string, reader io.Reader) error
+}
+
+// ObjstoreCache implements DebugInfoCache using S3 storage
+type ObjstoreCache struct {
+	bucket  objstore.Bucket
+	maxAge  time.Duration
+	metrics *Metrics
+}
```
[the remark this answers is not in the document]
marcsanmi: Updated it to `DebugInfoStore`

pkg/experiment/symbolizer/reader.go
```diff
@@ -0,0 +1,68 @@
...
+			zr.Close()
+			return nil, fmt.Errorf("decompress zstd data: %w", err)
+		}
+		zr.Close()
+
+		return bytes.NewReader(decompressed.Bytes()), nil
+	}
+
+	// Not compressed or unknown format
+	// Read all data into memory since we need to return an io.ReaderAt
```
liaol, line 61: This may use lots of memory when there is a big bianry,  using os.File.ReadAt would be better.

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,642 @@
...
+	// Group locations by mapping ID
+	type locToSymbolize struct {
+		idx int
+		loc *googlev1.Location
+	}
+	locsByMapping := make(map[uint64][]locToSymbolize)
+
+	for i, loc := range profile.Location {
+		if loc.MappingId > 0 {
+			mapping := profile.Mapping[loc.MappingId-1]
```
korniltsev, line 133: IMO this is incorrect. you can not use ID for index into Mapping/Function/Location tables

kolesnikovae: If the pprof is generated by the backend, it should work. This optimization is assumed along the entire read path

korniltsev: TIL

pkg/test/integration/testdata/otel-ebpf-profiler-offcpu-cpu.json.bin
```diff
@@ -51420,65 +51198,38 @@
...
         {
           "address": "0x3246730",
-          "lines": [
-            "libxul.so 0x3246730[]@:0"
-          ],
           "mapping": "0x2995000-0x8ddb000@0x2994000 libxul.so(ed82595fe72c9215b49e7e2cb2f85ea9269c1a61)"
         },
         {
           "address": "0x3246290",
-          "lines": [
```
korniltsev, line 51430: I think we should keep these synthetic frames if we do not symbolize them or fail to symbolize

Consider this flamegraph
Screenshot 2025-04-17 at 13 37 39

It hides the whole userspace stack. it is at least make harder to debug things for us, and at most may change the shape and look of flamegraph drastically if these frames are skipped in the middle of symbolized frames

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,753 @@
...
+}
+
+// ProfileSymbolizer implements Symbolizer
+type ProfileSymbolizer struct {
+	logger  log.Logger
+	client  DebuginfodClient
+	store   DebugInfoStore
+	metrics *Metrics
+
+	symbolCache     *lru.Cache[string, []lidia.SourceInfoFrame]
```
korniltsev, line 51: can the cache key be a struct with multiple fields instead of string?

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,753 @@
...
+
+func (s *ProfileSymbolizer) createNotFoundSymbols(binaryName string, loc *Location, addr uint64) []lidia.SourceInfoFrame {
+	prefix := "unknown"
+	if binaryName != "" {
+		prefix = binaryName
+	}
+
+	return []lidia.SourceInfoFrame{{
+		FunctionName: fmt.Sprintf("%s!0x%x", prefix, loc.Address),
+		FilePath:     fmt.Sprintf("mapped_0x%x", addr),
```
korniltsev, line 736: I think we can leave FilePath empty. I'ts not visible anywhere and not useful. Mapping file path is enough.

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,753 @@
...
+	"debug/elf"
+	"errors"
+	"flag"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/dgraph-io/ristretto"
```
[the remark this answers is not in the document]
marcsanmi: I wanted to avoid generics 😁 I'm updating it

pkg/test/integration/testdata/otel-ebpf-profiler-offcpu-cpu.json.bin
marcsanmi: I think this is happening for somewhat strange behaviour. I'm getting the following mappings:
```
1: 0x0/0x0/0x0   
2: 0x28000/0x1b0000/0x28000 libc.so.6 42c84c92e6f98126b3e2230ebfdead22c235b667 
3: 0x0/0x2000/0x0 linux-vdso.1.so
```
Mapping 1 has ~880 locations, so it's likely the one with userspace data. However, without a valid build id, we skip it; it's not even reaching symbolization. It's strange as we had valid build ids for `pyroscope-ebpf/profiler` before 🤔 

This is for the `pyroscopee-ebpf-otel-dev-001/otel-profiler`

Screenshot 2025-04-22 at 12 24 33

pkg/phlare/phlare.go
```diff
@@ -200,6 +202,11 @@ func (c *Config) RegisterFlagsWithContext(f *flag.FlagSet) {
 	c.API.RegisterFlags(f)
 	c.EmbeddedGrafana.RegisterFlags(f)
 	c.TenantSettings.RegisterFlags(f)
+
+	c.v2Experiment = os.Getenv("PYROSCOPE_V2_EXPERIMENT") != ""
```
korniltsev, line 206: don't we already have `c.v2Experiment = os.Getenv("PYROSCOPE_V2_EXPERIMENT") != ""`  in `registerServerFlagsWithChangedDefaultValues` , do we need this duplication?

pkg/experiment/ingester/segment.go
```diff
@@ -337,6 +337,10 @@ func concatSegmentHead(f *headFlush, w *writerOffset, s *metadata.StringTable) (
 		lb.WithLabelSet(model.LabelNameServiceName, f.head.key.service, model.LabelNameProfileType, profileType)
 	}
 
+	if f.flushed.HasNativeProfiles {
+		lb.WithLabelSet(model.LabelNameServiceName, f.head.key.service, metadata.LabelNameHasNativeProfiles, "true")
```
korniltsev, line 341: why do we set service name here second time?

pkg/experiment/ingester/segment.go
```diff
@@ -513,9 +520,36 @@ func (s *segment) ingest(tenantID string, p *profilev1.Profile, id uuid.UUID, la
...
+	return nil, false
+}
+
+// isHexAddress checks if a string looks like a hexadecimal memory address
+func isHexAddress(s string) bool {
+	// Most hex addresses in profiles start with "0x"
+	if len(s) < 3 || !strings.HasPrefix(s, "0x") {
+		return false
+	}
+
```
korniltsev, line 539: are these functions used or needed? getFunctionById isHexAddress

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,844 @@
...
+				"msg", "failed to resolve address on Lidia table lookup",
+				"addr", fmt.Sprintf("0x%x", addr),
+				"binary", req.BinaryName,
+				"build_id", req.BuildID,
+				"error", err,
+			)
+			loc.Lines = s.createNotFoundSymbols(req.BinaryName, loc, addr)
+			s.metrics.debugSymbolResolution.WithLabelValues(StatusErrorServerError).Observe(time.Since(resolveStart).Seconds())
+			continue
+		}
```
kolesnikovae, line 386: * We don't want update to metrics on each resolved location: it's expensive and not useful: me are measuring memory latency here.
* We don't want to write a log message on each failed `Lookup`. it will kill the system.

pkg/experiment/symbolizer/addrmapper.go
```diff
@@ -0,0 +1,192 @@
...
+		ProgramHeaders: loadableSegments,
+	}, nil
+}
+
+// MapRuntimeAddress translates a runtime address to its corresponding address
+// in the debug information. This translation is necessary because:
+// - The program might be loaded at a different address than it was linked for
+// - Different segments might need different adjustments
+// - Various ELF types (EXEC, DYN, REL) handle addressing differently
+func MapRuntimeAddress(runtimeAddr uint64, ei *BinaryLayout, m Mapping) (uint64, error) {
```
kolesnikovae, line 56: @korniltsev could you please take a careful look? I assume we should have something very similar in the profiler-side symbolizer.

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,844 @@
...
+	for _, loc := range req.Locations {
+		resolveStart := time.Now()
+		addr, err := MapRuntimeAddress(loc.Address, ei, Mapping{
+			Start:  loc.Mapping.Start,
+			Limit:  loc.Mapping.Limit,
+			Offset: loc.Mapping.Offset,
+		})
+		if err != nil {
+			return fmt.Errorf("normalize address: %w", err)
+		}
```
kolesnikovae, line 371: I'd try to normalize all addresses before trying to resolve them.

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,844 @@
...
+
+	// Create symbolization request for this mapping group
+	req := s.createSymbolizationRequest(binaryName, buildID, mapping, locs)
+
+	if err := s.Symbolize(ctx, &req); err != nil {
+		return err
+	}
+
+	// Store symbolization results back into profile
+	s.updateProfileWithSymbols(profile, mapping, locs, req.Locations)
```
kolesnikovae, line 218: The code generated by LLM is extremely inefficient and might be invalid. Please consider an alternative:
1. We split profile by mappings. Each such profile only contains Locations, Mappings (just one), and StringTable.
2. Each such sub-profile is symbolized independently in-parallel (bound)
3. Symbolized profiles are merged.

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,844 @@
...
+	if s.checkLidiaTableCache(ctx, req) {
+		return nil
+	}
+
+	if s.checkObjectStoreCache(ctx, req) {
+		return nil
+	}
+
+	return s.fetchAndCacheFromDebuginfod(ctx, req, &status)
+}
```
kolesnikovae, line 355: A comment explaining the sequence and cache layers would be helpful.

kolesnikovae (changes requested):
Please cleanup the code after LLM

pkg/experiment/symbolizer/metrics.go
```diff
@@ -0,0 +1,140 @@
...
+
+	for _, collector := range collectors {
+		util.RegisterOrGet(m.registerer, collector)
+	}
+}
+
+func (m *Metrics) Unregister() {
+	if m.registerer == nil {
+		return
+	}
```
kolesnikovae, line 126: Do we use it?

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,844 @@
...
+func (s *ProfileSymbolizer) createSymbolCacheKey(buildID string, address uint64) SymbolCacheKey {
+	return SymbolCacheKey{
+		BuildID: buildID,
+		Address: address,
+	}
+}
+
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	// TODO: change me back to false!
+	f.BoolVar(&cfg.Enabled, "symbolizer.enabled", true, "Enable symbolization for unsymbolized profiles")
```
kolesnikovae, line 815: Should probably be disabled by default

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,844 @@
...
+		if err := s.storeLidiaTableInCache(data, req.BuildID, cacheStart); err != nil {
+			return err
+		}
+	}
+
+	// Store in ObjstoreCache
+	if s.store != nil {
+		cacheStart := time.Now()
+		if cacheErr := s.store.Put(ctx, req.BuildID, bytes.NewReader(data)); cacheErr != nil {
+			s.metrics.cacheOperations.WithLabelValues("objstore_cache", "put", StatusErrorUpload).Observe(time.Since(cacheStart).Seconds())
```
kolesnikovae, line 620: We already measure it inside the bucket implementation.

kolesnikovae (changes requested):

pkg/experiment/symbolizer/symbolizer.go
marcsanmi: I think we talked to enabled it by default, but I might be wrong 🤔

lidia/builder.go
```diff
@@ -135,3 +137,43 @@ func (rc *rangeCollector) VisitRange(r *Range) {
...
+func newBinaryLayoutBuilder() *binaryLayoutBuilder {
+	return &binaryLayoutBuilder{
+		buf: make([]byte, 0, 256),
+	}
+}
+
+func (blb *binaryLayoutBuilder) write(layout *BinaryLayoutInfo) error {
+	if layout == nil {
+		return fmt.Errorf("nil binary layout")
+	}
```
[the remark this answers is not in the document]
marcsanmi: Not as of now.

lidia/constants.go
```diff
@@ -11,11 +11,14 @@ const (
...
 	// Size of the file header in bytes
-	headerSize = 0x80
+	headerSize = 0x98
 
 	// Number of fields in a line table entry
 	lineTableFieldsCount = 2
 
+	// Size of each line table entry (2 uint16s or 2 uint32s)
+	lineTableFieldsSize = 4
+
```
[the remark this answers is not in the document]
marcsanmi: It shouldn't invalidate existing Lidia files. The file format is self-describing through its header, which stores the actual field sizes used in that particular file. When reading a file, the code uses these stored values rather than relying on the constant in the source code.
Therefore, this change would be backward compatible with existing objects.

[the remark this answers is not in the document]
marcsanmi: `lineTableFieldsSize` primarily affect sthe calculation of section offsets during file creation. When reading files, the code uses the field size information stored in the file's header; it shouldn't invalidate existing Lidia files.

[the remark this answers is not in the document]
marcsanmi: I'll remove the constant to properly calculate this dynamically as:
```go
entrySize := 4
if hdr.lineTablesHeader.fieldSize == 4 {
    entrySize = 8
}
hdr.binaryLayoutHeader.offset = hdr.lineTablesHeader.offset + uint64(len(rc.lb.entries)*entrySize)
```
But answer your question, no, this should only affect the offset calculation when writing a new file, ensuring that the binary layout section starts at the correct position. It has no impact on how we read existing files.

pkg/experiment/ingester/segment.go
marcsanmi: not anymore, good catch 👍

pkg/experiment/symbolizer/addrmapper.go
korniltsev: 1. REL is not something can be loaded , we should not handle these 
2. for EXEC base address is always 0
3. for DYN, the otle ebpf profiler always outputs addresses relative to the elf base, so if an elf has a function at 0xcafe and the library is loaded at 0x7f0000000000, the profiler outputs just 0xcafe. We may need this addrmapper if in the future some other profiler outputs non relative addresses - 0x7f000000cafe. I remember we found a bug somewhere here in addrmapper  when debug elf and source elf had different PT_LOAD addresses and this code was outputting incorrect addresses for a debug-only file. This was an alloy binary. And in the case of full binary it was correct and the base offset was calculated as zero because it was an otel profile. I think we either need to solve this issue and make it work for both debug-only elfs and regular elfs including debug sections and have tests for these cases. Or just remove this addrmapper alltogether and assume the base is always zero (until we find some other non otel profiler which outputs  outputs non relative addresses)

kolesnikovae: This is my understanding as well: we do not want to normalize addresses at lookup time. For DYN, it should be done either in profiler, or during the translation to lidia.

marcsanmi: > the otle ebpf profiler always outputs addresses relative to the elf base

I wasn't aware of it. Considering this, I'd say we can remove it. As Tolyan mentioned, this will also simplify implementation since lidia binary layout won't be needed anymore (at least for now).

lidia/constants.go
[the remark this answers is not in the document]
marcsanmi: N/A anymore since we removed `MapRuntimeAddress` so no binaryLayout is needed anymore.

pkg/experiment/symbolizer/debuginfod_client.go
```diff
@@ -0,0 +1,350 @@
...
+	// CacheConfig configures the in-memory cache
+	CacheConfig struct {
+		// MaxSizeBytes is the maximum size of the cache in bytes
+		// Default is 2GB
+		MaxSizeBytes int64
+
+		// NumCounters is the number of keys to track frequency of
+		// Default is 10M
+		NumCounters int64
+	}
```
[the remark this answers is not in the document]
marcsanmi: > NumCounters – do we want to make it configurable?
Maybe isn't needed from the beginning. We can always expose it later.

pkg/experiment/symbolizer/debuginfod_client.go
```diff
@@ -0,0 +1,350 @@
...
+func NewDebuginfodClient(logger log.Logger, baseURL string, metrics *Metrics) (*DebuginfodHTTPClient, error) {
+	return NewDebuginfodClientWithConfig(logger, DebuginfodClientConfig{
+		BaseURL:    baseURL,
+		MaxRetries: 3,
+		UserAgent:  "Pyroscope-Symbolizer/1.0",
+		BackoffConfig: backoff.Config{
+			MinBackoff: 1 * time.Second,
+			MaxBackoff: 10 * time.Second,
+			MaxRetries: 3,
+		},
```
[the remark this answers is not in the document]
marcsanmi: yep, now we use `backoff.Config` from dskit.

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,844 @@
...
+		loc.Lines = frames
+		s.metrics.debugSymbolResolution.WithLabelValues(StatusSuccess).Observe(time.Since(resolveStart).Seconds())
+	}
+
+	return nil
+}
+
+// checkSymbolCache checks if all addresses are in the symbol cache
+func (s *ProfileSymbolizer) checkSymbolCache(req *Request) bool {
+	if s.symbolCache == nil {
```
korniltsev, line 409: when is s.symbolCache nil?

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,844 @@
...
+	*status = StatusErrorDebuginfod
+	s.metrics.debuginfodRequestDuration.WithLabelValues(StatusErrorDebuginfod).Observe(time.Since(debuginfodStart).Seconds())
+	s.metrics.debugSymbolResolution.WithLabelValues(StatusErrorDebuginfod).Observe(0)
+	return fmt.Errorf("fetch debuginfo: %w", err)
+}
+
+// updateAllCaches updates all caches with the fetched data
+func (s *ProfileSymbolizer) updateAllCaches(ctx context.Context, req *Request, data []byte) error {
+	// Store in Ristretto cache
+	if s.lidiaTableCache != nil {
```
korniltsev, line 609: when is it nil?

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,844 @@
...
+	if !success {
+		level.Warn(s.logger).Log("msg", "failed to store debug info in cache", "buildID", buildID, "size", len(lidiaData))
+	}
+	s.metrics.cacheSizeBytes.WithLabelValues("debuginfo_cache").Set(float64(len(lidiaData)))
+
+	return nil
+}
+
+func (s *ProfileSymbolizer) processELFData(data []byte) (lidiaData []byte, ei *BinaryLayout, err error) {
+	// Create a reader from the data
```
korniltsev, line 745: Lets remove all the "obvious" comments all over the new package. They are not very useful.

> // Create a reader from data
> ... bytes.NewReader
> 
> // Create lidia file
> lidia.CreateLidiaFromElf
> 
> 		// Convert to our internal binary layout format
>		ei = convertBinaryLayout(lidiaLayout)
> 
> // Open the Lidia table to extract binary layout
> lidia.Open

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,844 @@
...
+	// Store in Ristretto cache
+	if s.lidiaTableCache != nil {
+		cacheStart := time.Now()
+		if err := s.storeLidiaTableInCache(data, req.BuildID, cacheStart); err != nil {
+			return err
+		}
+	}
+
+	// Store in ObjstoreCache
+	if s.store != nil {
```
korniltsev, line 617: when is s.store nil?

pkg/phlare/modules_experimental.go
```diff
@@ -246,6 +247,34 @@ func (f *Phlare) initRecordingRulesClient() (services.Service, error) {
 	return c.Service(), nil
 }
 
+func (f *Phlare) initSymbolizer() (services.Service, error) {
+	level.Info(f.logger).Log("msg", "Initializing Symbolizer")
+	if !f.Cfg.Symbolizer.Enabled {
+		level.Info(f.logger).Log("msg", "Symbolizer is disabled")
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
marcsanmi: I thought as the following:

The symbolizer should have two levels of configuration: the global flag and per-tenant settings. The global flag (symbolizer.enabled) controls whether the symbolization is available at all. If enabled globally, then we can opt-in via the tenant configuration. If disabled globally, no symbolization will happen regardless of tenant settings.

pkg/experiment/symbolizer/symbolizer.go
marcsanmi: We simplifed a lot.

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,844 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+
+	googlev1 "github.com/grafana/pyroscope/api/gen/proto/go/google/v1"
+	"github.com/grafana/pyroscope/lidia"
+	"github.com/grafana/pyroscope/pkg/objstore"
+)
+
+const (
+	DefaultInMemorySymbolCacheSize     = 100_000
+	DefaultInMemoryLidiaTableCacheSize = 2 << 30 // 2GB default
```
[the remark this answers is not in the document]
marcsanmi: N/A

pkg/frontend/read_path/query_frontend/query_frontend.go
```diff
@@ -144,3 +223,36 @@ func (q *QueryFrontend) QueryMetadata(
...
+			// Get the key and value from string table
+			keyStr := prof.StringTable[label.Key]
+			valStr := prof.StringTable[label.Str]
+			if keyStr == phlaremodel.LabelNameOTEL && valStr == "true" {
+				return true
+			}
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
korniltsev: no, we don't add this label to samples. only series.

pkg/validation/limits.go
```diff
@@ -118,6 +118,9 @@ type Limits struct {
 	// RecordingRules allow to specify static recording rules. This is not compatible with recording rules
 	// coming from a RecordingRulesClient, that will replace any static rules defined.
 	RecordingRules RecordingRules `yaml:"recording_rules" json:"recording_rules" category:"experimental" doc:"hidden"`
+
+	// SymbolizerEnabled enables the symbolizer in the query frontend.
+	SymbolizerEnabled bool `yaml:"symbolizer_enabled" json:"symbolizer_enabled" doc:"hidden"`
```
korniltsev, line 123: is it not possible to enable this flag wit cli? is this on purpose?

pkg/experiment/symbolizer/symbolizer.go
marcsanmi: N/A as per https://github.com/grafana/pyroscope/pull/3799#discussion_r2063051024

pkg/validation/limits.go
marcsanmi: Should we enable per tenant flag from cli? In symbolizer.go we have defined:
```go

    f.BoolVar(&cfg.Enabled, "symbolizer.enabled", false, "Enable symbolization for unsymbolized profiles")
    ...
}
```

pkg/test/integration/testdata/otel-ebpf-profiler-pyrosymbolized-docker.json.bin
```diff
@@ -87,16 +87,10 @@
...
           "address": "0x405275",
-          "lines": [
-            "cat 0x405275[]@:0"
-          ],
           "mapping": "0x400000-0x40c000@0x0 cat(2efc66e3f4cae30a989532b37b1d5dc81026be68)"
         },
         {
           "address": "0x402462",
-          "lines": [
-            "cat 0x402462[]@:0"
```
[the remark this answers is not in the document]
marcsanmi: This is because we [removed](https://github.com/grafana/pyroscope/pull/3799/files#diff-30b0385316ef719e3cf6b51bae2226ac4d6afd84b5861f2d3a67abbccbefb762L221) this code from convert that was populating with line when none available:
```go
if len(gl.Line) == 0 {
		gl.Line = append(gl.Line, &googleProfile.Line{
			FunctionId: p.addfunc(fmt.Sprintf("%s 0x%x", p.src.StringTable[om.FilenameStrindex], ol.Address)),
		})
	}
```

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,459 @@
...
+}
+
+func (s *Symbolizer) Symbolize(ctx context.Context, req *Request) error {
+	start := time.Now()
+	status := statusSuccess
+	defer func() {
+		s.metrics.profileSymbolization.WithLabelValues(status).Observe(time.Since(start).Seconds())
+	}()
+
+	if s.checkObjectStoreCache(ctx, req) {
```
korniltsev, line 268: Let's rewrite `checkObjectStoreCache` and `fetchAndCacheFromDebuginfod` a bit.
Currently they are doing multiple things and I'd like them to do less.

Lets change them to be`fetchLidiaFromObjsotre` and `fetchLidiaFromDebuginfod`.  
The first one just fetches lidia bytes from objstore and does no symbolization. The second one will fetch debug elf and convert it to lidia and return lidia bytes and does not perform symbolization.

very approximate pseudocode
```go
func fetchLidia() []byte, error {


    lidiaBytes, err := s.bucket.Get(req.buildid)
    if err == nil {
        return lidiaBytes, nill
    }
    if !isNotFoundErr(err) {
        return nil, err
    }
    lidiaBytes, err = fetchLidiaFromDebuginfod(req.buildid)
    if err != nil {
        return nil,  err    
    }
    err := s.bucket.Upload(req.buildid, lidiaBytes) 
    if err != nil {
        return nil, err
    }
    return lidiaBytes, err          
}

...
lidiaBytes, err := fetchLidia(req.builid)
table, err := lidia.Open(lidiaBytes)
if err != nil {
    return err
}

s.symbolizeWithTable(table, req)
```

pkg/experiment/symbolizer/types.go
```diff
@@ -0,0 +1,41 @@
...
+// Location represents a memory address to be symbolized
+type Location struct {
+	ID      string
+	Address uint64
+	Lines   []lidia.SourceInfoFrame
+	Mapping *pprof.Mapping
+}
+
+// Request represents a symbolization request for multiple addresses
+type Request struct {
```
korniltsev, line 29: I think this is not used outside the package, we can make it not exporter

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,459 @@
...
+				Line:       int64(line.LineNumber),
+			}
+		}
+	}
+
+	mapping.HasFunctions = true
+	mapping.HasFilenames = true
+}
+
+func (s *Symbolizer) Symbolize(ctx context.Context, req *Request) error {
```
korniltsev, line 261: I think this is not used outside the package, we can make it unexported

pkg/test/integration/testdata/otel-ebpf-profiler-offcpu-cpu.json.bin
korniltsev: I think we still something is broken here and we now do not output the stubs.
I also think is mostly important for profiler developers and me, but not the profiling users. 
I will resolve this, we can address this later

pkg/experiment/symbolizer/debuginfo_store.go
```diff
@@ -0,0 +1,72 @@
...
+	// TODO: Implement a separate cleanup job for managing debug info storage
+	return reader, nil
+}
+
+func (c *ObjstoreDebugInfoStore) Put(ctx context.Context, buildID string, reader io.Reader) error {
+	return c.bucket.Upload(ctx, buildID, reader)
+}
+
+// NullDebugInfoStore implements DebugInfoStore but performs no caching
+type NullDebugInfoStore struct{}
```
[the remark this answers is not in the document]
marcsanmi: yes, this was here before having the mocks. Updated.

pkg/experiment/symbolizer/types.go
marcsanmi: Yes, I also unexported other types not used outside the packages :)

pkg/frontend/read_path/query_frontend/query_frontend.go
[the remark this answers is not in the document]
marcsanmi: We decided to not filter by otel profiler for now. See https://github.com/grafana/pyroscope-squad/issues/436

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,448 @@
...
+
+	mapping.HasFunctions = true
+	mapping.HasFilenames = true
+}
+
+func (s *Symbolizer) symbolize(ctx context.Context, req *request) error {
+	start := time.Now()
+	status := statusSuccess
+	defer func() {
+		s.metrics.profileSymbolization.WithLabelValues(status).Observe(time.Since(start).Seconds())
```
korniltsev, line 268: this metric is touched twice - once for the whole pprf and once for the permapping request. I suggest we remove the latter one or create a separate metric for it.

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,448 @@
...
+func (s *Symbolizer) fetchFromDebuginfod(ctx context.Context, buildID string) (io.ReadCloser, error) {
+	debuginfodStart := time.Now()
+
+	debugReader, err := s.client.FetchDebuginfo(ctx, buildID)
+	if err != nil {
+		var bnfErr buildIDNotFoundError
+		statusCode, isHTTPError := isHTTPStatusError(err)
+
+		if errors.As(err, &bnfErr) || (isHTTPError && statusCode == http.StatusNotFound) {
+			s.metrics.debuginfodRequestDuration.WithLabelValues(statusErrorNotFound).Observe(time.Since(debuginfodStart).Seconds())
```
korniltsev, line 403: please take another look at this metric. Something is off here. it looks like error statuses are accounted multiple times, in the DebugInfodClient and in the symbolizer. Can we remove all the usages of debuginfod metrics from the symbolizer?

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,448 @@
...
+
+// fetchLidiaFromDebuginfod fetches debug info from debuginfod and converts to Lidia format
+func (s *Symbolizer) fetchLidiaFromDebuginfod(ctx context.Context, buildID string) ([]byte, error) {
+	debuginfodStart := time.Now()
+	status := statusSuccess
+
+	debugReader, err := s.fetchFromDebuginfod(ctx, buildID)
+	if err != nil {
+		var bnfErr buildIDNotFoundError
+		if errors.As(err, &bnfErr) {
```
korniltsev, line 373: what is the purpose of this type assertion? both branches are the same and return err. Is this needed? can we jsust return err?

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,448 @@
...
+		s.metrics.profileSymbolization.WithLabelValues(status).Observe(time.Since(start).Seconds())
+	}()
+
+	var table *lidia.Table
+	var err error
+
+	lidiaBytes, err := s.getLidiaBytes(ctx, req.buildID)
+	if err != nil {
+		var bnfErr buildIDNotFoundError
+		if errors.As(err, &bnfErr) {
```
korniltsev, line 277: I suggest we do not fail the query if there are any errors during symbolization. Let's treat all the errors the same - createNotFoundSymbolz

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,448 @@
...
+		}
+
+		status = statusErrorServerError
+		return err
+	}
+
+	lidiaReader := NewReaderAtCloser(lidiaBytes)
+	table, err = lidia.OpenReader(lidiaReader, lidia.WithCRC())
+	if err != nil {
+		s.metrics.debugSymbolResolution.WithLabelValues("lidia_error").Observe(0)
```
korniltsev, line 291: I suggest we do not fail the query if there are any errors during symbolization. Let's treat all the errors the same - createNotFoundSymbolz

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,448 @@
...
+		return nil
+	}
+
+	locationsByMapping, err := s.groupLocationsByMapping(profile, mappingsToSymbolize)
+	if err != nil {
+		return fmt.Errorf("grouping locations by mapping: %w", err)
+	}
+
+	for mappingID, locations := range locationsByMapping {
+		if err := s.symbolizeLocationsForMapping(ctx, profile, mappingID, locations); err != nil {
```
korniltsev, line 90: I suggest we do not fail the query if there are any errors during symbolization. Let's treat all the errors the same - createNotFoundSymbolz

pkg/experiment/symbolizer/types.go
```diff
@@ -0,0 +1,32 @@
...
+// LidiaTableCacheEntry represents a cached Lidia table with its binary layout information
+type LidiaTableCacheEntry struct {
+	Data []byte // Processed Lidia table data
+}
+
+// location represents a memory address to be symbolized
+type location struct {
+	address uint64
+	lines   []lidia.SourceInfoFrame
+	mapping *pprof.Mapping
```
korniltsev, line 24: this field is unused. Let's remove it

pkg/experiment/symbolizer/types.go
```diff
@@ -0,0 +1,32 @@
+package symbolizer
+
+import (
+	googlev1 "github.com/grafana/pyroscope/api/gen/proto/go/google/v1"
+	"github.com/grafana/pyroscope/lidia"
+
+	pprof "github.com/google/pprof/profile"
+)
+
+type locToSymbolize struct {
```
korniltsev, line 10: I think we can replace locToSymbolize with just `*googlev1.Location` because `loc.Id == idx +1`

pkg/experiment/symbolizer/reader.go
```diff
@@ -0,0 +1,130 @@
...
+
+	if len(header) >= 2 && header[0] == 0x1f && header[1] == 0x8b { // gzip
+		gr, err := gzip.NewReader(br)
+		if err != nil {
+			return nil, fmt.Errorf("create gzip reader: %w", err)
+		}
+
+		var decompressed bytes.Buffer
+		if _, err := decompressed.ReadFrom(gr); err != nil {
+			gr.Close()
```
korniltsev, line 102: can we use defer gr.Close() ?

pkg/experiment/symbolizer/reader.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func (m *memoryBuffer) Bytes() []byte {
+	return m.data
+}
+
+// detectCompression reads the beginning of the input to determine if it's compressed,
+// and if so, returns a ReaderAt that decompresses the data.
+func detectCompression(r io.Reader) (io.ReaderAt, error) {
+	br := bufio.NewReader(r)
```
korniltsev, line 87: lets accept []byte as an argument to this function and do not use bufio - we already have everything in memory, no need to buffer. Let's return []byte here instead of ReaderAt - it will make it a bit more visible that we are reading and decompressing everything fully in memory

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,448 @@
...
+	}
+}
+
+func (s *Symbolizer) getLidiaBytes(ctx context.Context, buildID string) ([]byte, error) {
+	lidiaBytes, err := s.fetchLidiaFromObjectStore(ctx, buildID)
+	if err == nil {
+		return lidiaBytes, nil
+	}
+
+	level.Error(s.logger).Log(
```
korniltsev, line 331: nit: this is not an error if it is "not found"

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,448 @@
...
+				BuildID: buildID,
+			},
+		}
+	}
+
+	return req
+}
+
+// updateProfileWithSymbols updates the profile with symbolization results
+func (s *Symbolizer) updateProfileWithSymbols(profile *googlev1.Profile, mapping *googlev1.Mapping, locs []locToSymbolize, symLocs []*location) {
```
korniltsev, line 201: Can we reorganize the code a bit, to call this function once per pprof instead of once per mapping? Othervise we create all the maps multiple times

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,448 @@
...
+			key := funcKey{nameIdx, filenameIdx}
+			funcID, ok := funcMap[key]
+			if !ok {
+				maxFuncID++
+				funcID = maxFuncID
+				profile.Function = append(profile.Function, &googlev1.Function{
+					Id:        funcID,
+					Name:      nameIdx,
+					Filename:  filenameIdx,
+					StartLine: int64(line.LineNumber),
```
korniltsev, line 248: lets remove this line. On the left we have the start of the function and on the right we have the middle of the function. Also the line number is not included in the key. Also we don't really have line numbers at the moment.

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,448 @@
...
+	stringMap := make(map[string]int64, len(profile.StringTable))
+	for i, str := range profile.StringTable {
+		stringMap[str] = int64(i)
+	}
+
+	type funcKey struct {
+		nameIdx, filenameIdx int64
+	}
+	funcMap := make(map[funcKey]uint64)
+	maxFuncID := uint64(0)
```
korniltsev, line 211: can we do `maxFuncId := len(profile.Function) + 1` here? I know I told you sometime ago it's not a valid approach in general, but it is a valid approach for pprof files returned from our pprof queries. so should be fine?

pkg/experiment/symbolizer/symbolizer.go
korniltsev: we also don' really have filenames yet. I suggest we remove filename here as well.

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,448 @@
...
+	if err := s.store.Put(ctx, buildID, bytes.NewReader(lidiaBytes)); err != nil {
+		level.Warn(s.logger).Log("msg", "Failed to store debug info in objstore", "buildID", buildID, "err", err)
+	}
+
+	return lidiaBytes, nil
+}
+
+// fetchLidiaFromObjectStore retrieves Lidia data from the object store
+func (s *Symbolizer) fetchLidiaFromObjectStore(ctx context.Context, buildID string) ([]byte, error) {
+	objstoreReader, err := s.store.Get(ctx, buildID)
```
korniltsev, line 351: can we avoid store.Get invocations if we already know debuginfod returned NotFound ? it would probably require to pull the cache from debuginfod into symbolizer

pkg/frontend/read_path/query_frontend/query_frontend.go
```diff
@@ -148,3 +187,75 @@ func (q *QueryFrontend) QueryMetadata(
...
 }
+
+// hasUnsymbolizedProfiles checks if a block has unsymbolized profiles
+func (q *QueryFrontend) hasUnsymbolizedProfiles(block *metastorev1.BlockMeta) bool {
+	matcher, err := labels.NewMatcher(labels.MatchEqual, metadata.LabelNameUnsymbolized, "true")
+	if err != nil {
+		return false
+	}
+
+	return len(slices.Collect(metadata.FindDatasets(block, matcher))) > 0
```
korniltsev, line 198: @aleks-p Can you help me verify my assumption here please.
1. Blocks contain multiple datasets and so a block may have both unsymbolized and symbolized datasets.
2. I believe the `MetadataQueryService.QueryMetadata` does not return all the datasets present in the block ,but only the ones that matched query.  I've made this conclusion from reading this code
https://github.com/grafana/pyroscope/blob/d564c76454a93f75d008c3a0b223102cdaf55c5c/pkg/experiment/metastore/index/query.go#L163

If my understanding is correct, than this  `hasUnsymbolizedProfiles` is fine. If the second assumption is not correct, then this check would degrade the symbolized query. 

I think everything is good here, just want another :eyes: to doublecheck.

pkg/experiment/symbolizer/symbolizer.go
marcsanmi: hmm you're right. I've checked it and seems we do guarantee this so we can remove the loop to find `maxFuncID`, good one 👍

pkg/experiment/symbolizer/metrics.go
```diff
@@ -0,0 +1,125 @@
...
+	// Debuginfod metrics
+	debuginfodRequestDuration *prometheus.HistogramVec
+	debuginfodFileSize        prometheus.Histogram
+
+	// Cache metrics
+	cacheOperations *prometheus.CounterVec
+	cacheSizeBytes  *prometheus.GaugeVec
+
+	// Profile symbolization metrics
+	profileSymbolization *prometheus.HistogramVec
```
korniltsev, line 44: please check the usage of this metric. I think it is used twice - once per mapping, second per all mappings. One of them should go.

pkg/frontend/read_path/query_frontend/query_frontend.go
```diff
@@ -148,3 +187,75 @@ func (q *QueryFrontend) QueryMetadata(
...
+		if err := pprof.Unmarshal(r.Pprof.Pprof, &prof); err != nil {
+			return fmt.Errorf("failed to unmarshal profile: %w", err)
+		}
+
+		if err := q.symbolizer.SymbolizePprof(ctx, &prof); err != nil {
+			return fmt.Errorf("failed to symbolize profile: %w", err)
+		}
+
+		// Convert back to tree if originally a tree
+		if i < len(originalQueries) && originalQueries[i].QueryType == queryv1.QueryType_QUERY_TREE {
```
korniltsev, line 248: lets add an early return (with an error?) for a case if len(originalQueries) != len(resp.Reports)

pkg/experiment/symbolizer/debuginfo_store.go
```diff
@@ -0,0 +1,48 @@
...
+	objstoreclient "github.com/grafana/pyroscope/pkg/objstore/client"
+)
+
+// DebugInfoStoreConfig holds configuration for the debug info cache
+type DebugInfoStoreConfig struct {
+	Storage objstoreclient.Config `yaml:"storage"`
+}
+
+// DebugInfoStore handles caching of debug info files
+type DebugInfoStore interface {
```
korniltsev: why do we need this interface? can we use `objstore.Bucket` interface directly?

pkg/experiment/symbolizer/debuginfo_store.go
```diff
@@ -0,0 +1,48 @@
...
+func NewObjstoreDebugInfoStore(bucket objstore.Bucket) *ObjstoreDebugInfoStore {
+	return &ObjstoreDebugInfoStore{
+		bucket: bucket,
+	}
+}
+
+func (c *ObjstoreDebugInfoStore) Get(ctx context.Context, buildID string) (io.ReadCloser, error) {
+	reader, err := c.bucket.Get(ctx, buildID)
+	if err != nil {
+		if c.bucket.IsObjNotFoundErr(err) {
```
korniltsev: why do we need special case handling here? why not always wrap? or always just return the err?

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,428 @@
...
+			loc.lines = s.createNotFoundSymbols(req.binaryName, loc)
+		}
+		return
+	}
+	defer table.Close()
+
+	s.symbolizeWithTable(ctx, table, req)
+}
+
+func (s *Symbolizer) symbolizeWithTable(_ context.Context, table *lidia.Table, req *request) {
```
korniltsev: lets remove context if it is not needed

korniltsev (approved):
lgtm overall

I left a bunch of nitpick comments.

I am not excited we have to implement `memoryBuffer`. At the same time I don't know any ready to use structure from any library. Maybe in the future we can rework lidia writing api to operate on bytes.Buffer instead of `WriteSeeker`

I think it should not impact the regular read path. I think we should run this in test/dev once again and merge.

I did not test/run this revision locally manually this time.

pkg/frontend/read_path/query_frontend/query_frontend.go
aleks-p: Sorry I missed this @korniltsev . Both assumptions are correct. Depending on the query you can get a mix of mismatching datasets, but since this is in the context of symbolization the datasets should be for the same service.

cmd/pyroscope/help-all.txt.tmpl
```diff
@@ -1079,6 +1079,8 @@ Usage of ./pyroscope:
     	True to enable zone-awareness and replicate blocks across different availability zones. This option needs be set both on the store-gateway and querier when running in microservices mode.
   -store-gateway.tenant-shard-size int
     	The tenant's shard size, used when store-gateway sharding is enabled. Value of 0 disables shuffle sharding for the tenant, that is all tenant blocks are sharded across all store-gateway replicas.
+  -symbolizer.tenant-enabled
+    	[experimental] When symbolizer.enabled is true, this controls whether symbolization is enabled for tenants by default. Can be overridden per-tenant. (default true)
```
kolesnikovae: If it depends on v2, it should not be shown here unless v2 is enabled

I think it should be just `-symbolizer.enabled` (false by default). Then, the tenant level override controls it at the tenant level (true). LMKWYT

Could you please elaborate on why we need this parameter?

pkg/frontend/read_path/query_frontend/query_frontend.go
kolesnikovae: I confirm this.

cmd/pyroscope/help-all.txt.tmpl
marcsanmi: Maybe I understood overrides wrong, but I added that flag to set it to default as true for all tenants. How, otherwise we can set it to true for all tenants by default? Should we hook into the `UnmarshalYAML`?

kolesnikovae: Yeah, it's all cool, we only need to make sure:
1. It's not visible in the help message unless v2 is enabled.
2. The parameter name looks good (`-symbolizer.enabled` – without `tenant-` – looks good IMO).
3. We're absolutely sure we do want to enable it by default – I guess makes sense to keep it disabled by default until we mark the feature stable.

pkg/phlare/phlare.go
```diff
@@ -200,6 +202,10 @@ func (c *Config) RegisterFlagsWithContext(f *flag.FlagSet) {
 	c.API.RegisterFlags(f)
 	c.EmbeddedGrafana.RegisterFlags(f)
 	c.TenantSettings.RegisterFlags(f)
+
+	if c.v2Experiment {
+		c.Symbolizer.RegisterFlags(f)
+	}
```
kolesnikovae: Let's move it to `registerServerFlagsWithChangedDefaultValues`, where we handle v2 flags:
https://github.com/grafana/pyroscope/blob/main/pkg/phlare/phlare.go#L226-L255

pkg/experiment/symbolizer/debuginfo_store.go
marcsanmi: N/A as per https://github.com/grafana/pyroscope/pull/3799#discussion_r2123449669

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -0,0 +1,428 @@
...
+		return fmt.Errorf("grouping locations by mapping: %w", err)
+	}
+
+	stringMap := make(map[string]int64, len(profile.StringTable))
+	for i, str := range profile.StringTable {
+		stringMap[str] = int64(i)
+	}
+
+	funcMap := make(map[funcKey]uint64)
+	maxFuncID := uint64(len(profile.Function) + 1)
```
marcsanmi: @korniltsev I got a panic index out of bounds, so not sure we can rely on function IDs to be sequential (maybe something else is wrong?). For now I'm reverting the change back to:
```go
funcMap := make(map[funcKey]uint64)
maxFuncID := uint64(0)

for _, fn := range profile.Function {
    if fn.Id > maxFuncID {
	maxFuncID = fn.Id
    }
    funcMap[funcKey{fn.Name, fn.Filename}] = fn.Id
}
```

korniltsev: can you reproduce? I suggest we understand / reproduce why it paniced before reverting. And maybe have a testcase.

kolesnikovae: `func.ID - 1` points to the function index in the `Function` slice. If it's not the case, it's a bug, and it should be fixed. Note that 0 ID is a sentinel and it cannot be valid.

marcsanmi: The symbolizer was creating non-sequential function IDs when adding new functions to profiles with existing functions due to 
```go
maxFuncID := uint64(len(profile.Function) + 1)
``` 
However, tree conversion assumes sequential IDs and tries to access [profile.Function[loc.Line[l].FunctionId-1]](https://github.com/grafana/pyroscope/pull/3799/files#diff-acf4ab855649a6af94cf68eb16403eeaa957c082a192dd804699e1dfb13cfb67R471), causing index out of bounds panics. 

Fixed by ensuring new function IDs are always sequential starting from the last existing ID by removing the +1 
```go
maxFuncID := uint64(len(profile.Function))
``` 

I've also added a test explicitly testing this behaviour. You can check it in this commit 1a526a1c3526a82e307712e3fb1ecc366c4321c4.

LMKWYT

korniltsev (approved):

pkg/phlare/modules_experimental.go
```diff
@@ -83,12 +85,18 @@ func (f *Phlare) initQueryFrontendV1() (services.Service, error) {
...
+
 	queryFrontend := queryfrontend.NewQueryFrontend(
 		log.With(f.logger, "component", "query-frontend"),
 		f.Overrides,
 		f.metastoreClient,
 		f.metastoreClient,
 		f.queryBackendClient,
+		symbolizerParam,
 	)
```
kolesnikovae: Basically, it looks like we can pass `f.symbolizer` without the check

pkg/validation/symbolizer.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+type Symbolizer struct {
+	// Enabled enables the symbolizer in the query frontend.
+	Enabled bool `yaml:"enabled" json:"enabled" category:"experimental" doc:"hidden"`
+}
+
+func (s *Symbolizer) RegisterFlags(f *flag.FlagSet) {
+	f.BoolVar(&s.Enabled, "symbolizer.enabled", false, "Enable symbolization for tenants by default.")
+}
```
kolesnikovae, line 14: The description appears invalid

Never mind, we will review them before v2 release

kolesnikovae (approved):
LGTM! Thank you so much @marcsanmi! 🎉

PR #3808, opened 2025-01-02
chore(otlp): add  integration tests with fixtures

korniltsev:


pkg/og/convert/pprof/strprofile/profile.go
```diff
@@ -0,0 +1,192 @@
...
+type Profile struct {
+	SampleTypes   []SampleType `json:"sample_types"`
+	Samples       []Sample     `json:"samples"`
+	TimeNanos     string       `json:"time_nanos,omitempty"`
+	DurationNanos string       `json:"duration_nanos,omitempty"`
+	Period        string       `json:"period,omitempty"`
+}
+
+// Stringify converts a profile to a human-readable JSON representation
+func Stringify(p *profilev1.Profile, opts Options) (string, error) {
```
simonswine, line 70: Not really a suggestion/nit, just wondering what you think about that other format.

Not too sure if you considered this: https://github.com/google/pprof/blob/40e02aabc2ad57b2edbe892564a0730aa40b1e50/profile/profile.go#L553-L555

That's my goto, as I fairly familar with the output from `pprof -raw`

I guess you would need to convert to protobuf and parse it to a googleprofile.Profile.

simonswine (approved):
LGTM

pkg/og/convert/pprof/strprofile/profile.go
korniltsev: I think google profile string is harder to read and write by hand or code.
I also would like to have an option to not output function/location/mapping ids.
Although I like how google string is more compact that what I submit in this PR, I will try update the PR to make it more compact

Another "problem" is the "Time" field which is generated by pyroscope at query time. Although it is pretty easy to remove it from the data.

PR #3818, opened 2025-01-06
fix(distributor): validate string table accesses

korniltsev:


simonswine (approved):
LGTM

At which point are we checking that `StringTable[0] == ""`? I can find it right now, but I pretty sure we do it...

korniltsev:
I've added a case for empty string just in case

PR #3819, opened 2025-01-06
fix(distributor): validate string table accesses

korniltsev:
backport https://github.com/grafana/pyroscope/pull/3818 to f/97

### DIFF
--- a/pkg/validation/validate.go
+++ b/pkg/validation/validate.go
@@ -274,6 +274,10 @@ func ValidateProfile(limits ProfileValidationLimits, tenantID string, prof *goog
 			return NewErrorf(MalformedProfile, "function id is 0")
 		}
 	}
+
+	if err := validateStringTableAccess(prof); err != nil {
+		return err
+	}
 	for _, valueType := range prof.SampleType {
 		stt := prof.StringTable[valueType.Type]
 		if strings.Contains(stt, "-") {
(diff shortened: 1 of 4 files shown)


pkg/validation/validate.go
```diff
@@ -274,13 +274,18 @@ func ValidateProfile(limits ProfileValidationLimits, tenantID string, prof *goog
 			return NewErrorf(MalformedProfile, "function id is 0")
 		}
 	}
+
+	if err := validateStringTableAccess(prof); err != nil {
+		return err
+	}
 	for _, valueType := range prof.SampleType {
 		stt := prof.StringTable[valueType.Type]
```
kolesnikovae, line 282: I think this is the only access we need to fix. We duplicate checks done later in pprof validation – it's fine if we merge this, but let's try to avoid duplication. I'm wondering if the sample type check should be performed here, and not in the pprof validation (along with many other checks)

kolesnikovae (approved):

simonswine:
We should also make sure we have this fix in f100, which should be cut today

pkg/validation/validate.go
korniltsev: I'd like to remove the dupliacate checks, but I am not sure where they are.  If I remove the new validations, all new tests fail. Either return 200 when they should not or panic (sample type / sample label string index oob)

kolesnikovae: Let's merge it as is and sort it out later

Issue #3817, opened 2025-01-06
Compaction throughput degraded after 1.8 -> 1.9.1 upgrade


kolesnikovae:
From https://grafana.slack.com/archives/C049PLMV8TB/p1735948533932149:

> After upgrading from Pyroscope 1.8.0 -> 1.9.1, we observed a massive decrease in throughput of compaction. (Behavior persists in 1.10 and 1.11.) After the upgrade, we've observed a compactor taking ~12 hours to compact/write a ~12 GB block. So the compactor throughput is <<1 MB/s. All the while the compactor process is using 100% of the CPU allocated to it. This represents a >>10x slowdown versus 1.8. (Before the upgrade, we were able to perform 3 rounds of compaction (level 4) from 1 compactor process. After, we had to 10x the compactor CPU footprint just to perform 1 round of compaction. Attempting an additional level of compaction beyond 1 hour ranges just wasn't viable since it took too much CPU resources.)

> Also, I downgraded an instance to 1.8.0 hoping to see the old/fast performance come back. But the compactor flamegraph looks very similar as 1.11.0. I'm really confused because our time series metrics from Pyroscope clearly show the compactor performance degrading after the 1.8.0 -> 1.9.1 upgrade. So I'm at a loss here.

kolesnikovae:
We investigated the issue and determined that the main cause is an increase in stack trace cardinality, driven by issues specific to the ingested profiling data.

A typical deployment should handle moderate stack trace cardinality effectively. However, it relies on certain properties, such as most stack traces sharing the majority of their frames, which must be preserved. If a single service ingests more than a few million (>10 million) _unique_ stack traces within the duration of a block (default: 1 hour), both query and compaction performance will degrade.

PR #3825, opened 2025-01-08
fix: revive stack trace range chunking

kolesnikovae:
Resolves https://github.com/grafana/pyroscope/issues/3817.

The PR reverts https://github.com/grafana/pyroscope/pull/3583 as it causes a performance regression in compactors (issue https://github.com/grafana/pyroscope/issues/3817) in certain cases. You can find more details in our slack channel ([thread](https://grafana.slack.com/archives/C049PLMV8TB/p1735948533932149)).

The change affects only ingesters and is fully backward-compatible. I'm hesitant to make this configurable because it's very easy to misinterpret and misconfigure the parameter.


korniltsev (approved):

kolesnikovae:
I'm cancelling it for now as we can confirm this is helpful. The issue mentioned in the PR description has a different nature and origin, and reviving stack trace chunking is not the best way to tackle it

PR #3827, opened 2025-01-08
perf: cache stack hash to node in stacktreeTree

airbnb-gps:
See the inline comment for more context.

Profiling against my workload showed the compactor spending significant CPU time in `stacktraceTree.insert()`.

Staring at the code, I realized it was performing a full stack walk to populate/resolve the leaf-most tree node representing the stack.

Since many stacks will be identical, we can benefit from caching the mapping of the unique stack back to the node index.

The performance speedups are greater if there are many repeated stacks and when stacks are deep. My data set consists of a lot of JVM profiles and Java is infamous for having very deep stacks. So this yields a considerable speedup for me.

### DIFF
--- a/pkg/phlaredb/symdb/stacktrace_tree.go
+++ b/pkg/phlaredb/symdb/stacktrace_tree.go
@@ -16,7 +16,8 @@ const (
 )
 
 type stacktraceTree struct {
-	nodes []node
+	nodes        []node
+	hashedStacks map[uint64]uint32
 }
 
 type node struct {
@@ -31,7 +32,10 @@ func newStacktraceTree(size int) *stacktraceTree {
 	if size < 1 {
 		size = 1
 	}
-	t := stacktraceTree{nodes: make([]node, 1, size)}
+	t := stacktraceTree{
+		nodes:        make([]node, 1, size),
+		hashedStacks: make(map[uint64]uint32, size),
+	}
 	t.nodes[0] = node{
 		p:  sentinel,
 		fc: sentinel,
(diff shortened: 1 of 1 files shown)


korniltsev:
can we add a benchmark to measure the difference?

korniltsev:
Btw, there should not be fully identical samples in the pprof converted from JFR due to this https://github.com/grafana/jfr-parser/blob/a8d22a1cd731f0ef8b48f84c5bab58532a9af541/pprof/pprof.go#L66

How do you ingest your data? do you ingest jfr to pyroscope?

airbnb-gps:
For single pprof profiles, yes, there should be 1 entry for each distinct stack (assuming the pprof generator deduplicated, which the JFR converting code does).

But for the compactor, it will see multiple occurrences of the same stack from different source blocks, no?

(I found this hot code when looking at compactor performance.)

pkg/phlaredb/symdb/stacktrace_tree.go
```diff
@@ -45,6 +49,18 @@ const sentinel = -1
...
+	// walking the tree. We simply cache a map of stack digest back to the
+	// node index. If there's a hit, we avoid a stack walk to resolve
+	// the leaf node. If not, we pay a penalty for computing the hash
+	// and performing a map lookup.
+	digest := hashLocations(refs)
+	existing, ok := t.hashedStacks[digest]
+	if ok {
+		return existing
+	}
+
```
kolesnikovae, line 63: This is a valid approach, and your reasoning is correct, if I understand it correctly.

However, we already have a caching layer on top of trees (so-called chunks), and we never access a tree directly, bypassing the cache in the write path (valid for both ingesters and compactors): https://github.com/grafana/pyroscope/blob/26950d5cce3d5f70a87a5b91ea8de7a4bc8ac830/pkg/phlaredb/symdb/partition_memory.go#L63-L113.

And it works in the same way: locations hash => stack trace ID (leaf node index) lookup:
 * There is a chance of hash collisions, and we accept this risk.
 * We eliminate an extra map lookup by relying on insertion idempotence: since the stack traces we want to insert might be added between mutex locks, we may write them again. Given that there are very few such writes, this approach is more efficient than checking each stack trace after acquiring the write lock.

Therefore, I'm wondering if we benefit from adding a cache at the tree level. I can say for sure that it will increase memory consumption significantly, while I can't see how it improves performance.

This is how it looks in a loaded cluster (half of 10G link of ingress, samples over 6 hours):

In ingesters:
image

In compactors:
image
_Note that there's a certain issue with the compaction process, which makes the cache less helpful. We're aware of it and we have a solution to the problem – something we're actively working on._

kolesnikovae:
Thank you for the contribution @airbnb-gps! I believe the proposed optimisation has already been implemented – please check my comment

airbnb-gps:
I agree with your assessment.

I still don't have a great way to reproduce my performance tests so my methodology was flawed, leading to faulty conclusions on my part. Sorry for the noise!

kolesnikovae:
No worries at all! Thank you for a good PR, and I hope you can contribute something else another time :)

PR #3837, opened 2025-01-14
Fix: Acutally validate config struct

simonswine:
We seem to have not linked up the validation for a few components.

### DIFF
--- a/pkg/objstore/client/config.go
+++ b/pkg/objstore/client/config.go
@@ -89,6 +89,10 @@ func (cfg *StorageBackendConfig) RegisterFlagsWithPrefix(prefix string, f *flag.
 }
 
 func (cfg *StorageBackendConfig) Validate() error {
+	if cfg.Backend == None {
+		return nil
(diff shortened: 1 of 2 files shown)


marcsanmi (approved):
LGTM 👍

kolesnikovae:
This might be a breaking change. Does it make sense to add a flag that enables/disables config validation?

julienvey:
I think we hit this breaking change.

We were using a storage_prefix value for s3 storage (the bucket is shared for multiple application / envs)

```
storage_prefix: pyroscope/my-env/
```

and this is now failing on the latest versions of Pyroscope

> failed creating pyroscope: storage prefix contains invalid characters, it may only contain digits and English alphabet letters

I see the code for regex validation has been there for more than 3 years

Is there any workaround for us to keep the current storage prefix ? Otherwise we would need to set up a new one without the `/`, and we would either loose all existing data, or would need to move it for all environments

PR #3847, opened 2025-01-16
fix(v2): compaction-worker temp directory cleanup

kolesnikovae:
Currently, compaction-worker does not cleanup the directory it uses for "staging" (kind of _temp_).

Typically, this is not a problem, because the intermediate data is deleted during the compaction process. However, the disk space (including ephemeral volumes) might be exhausted in case if the compactor fails to finish the compaction.

### DIFF
--- a/pkg/experiment/compactor/compaction_worker.go
+++ b/pkg/experiment/compactor/compaction_worker.go
@@ -58,12 +58,11 @@ type Config struct {
 
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	const prefix = "compaction-worker."
-	tempdir := filepath.Join(os.TempDir(), "pyroscope-compactor")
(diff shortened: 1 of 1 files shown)


pkg/experiment/compactor/compaction_worker.go
```diff
@@ -369,6 +373,11 @@ func (w *Worker) runCompaction(job *compactionJob) {
 			block.WithObjectMaxSizeLoadInMemory(w.config.SmallObjectSize),
 			block.WithObjectDownload(sourcedir),
 		))
+	defer func() {
+		if err = os.RemoveAll(tempdir); err != nil {
+			level.Warn(logger).Log("msg", "failed to remove compaction directory", "path", tempdir, "err", err)
+		}
+	}()
```
kolesnikovae, line 380: Although the content is removed, the job directory isn't

kolesnikovae, line 380: Although the content is removed, the job directory isn't

korniltsev (approved):

PR #3848, opened 2025-01-16
chore: Refactor labelset from og/segment and og/flameql

simonswine:
In order to be able to parse the labelset in the classic ingest endpoints in Grafana Alloy, I am refactoring this bit of the OG code into the api package so I can use this within Grafana Alloy.


korniltsev (approved):
lgtm

PR #3852, opened 2025-01-17
Metastore client fix + admin test page

aleks-p:
**Background**
We have observed metastore client latency spikes during rollouts / leadership changes.

The client relies on error responses containing the new leader, however there is a mismatch in what the client expects and what metastore nodes return.

**Problem**
The metastore client discovers servers using the grpc port (9095) and maintains a map with addresses such as 
```pyroscope-metastore-0.pyroscope-metastore-headless.profiles-dev-003:9095```

Metastore nodes return the leader using the Raft port (9099) and the `svc.cluster.local` domain:
```pyroscope-metastore-0.pyroscope-metastore-headless.profiles-dev-003.svc.cluster.local.:9099```

This causes a failed check here: https://github.com/grafana/pyroscope/blob/f4cec5d93dfbb8b49940e569520655ef1a47731b/pkg/experiment/metastore/client/methods.go#L83-L85
As a result we resort to random selections until we hit the right node.

**Fix**
The main options are to “hack” the client and massage the data or change the Raft server identity. This PR does the former, it is however messy and ideally we should align the two components. I will give the second option a try as well.

Either way we go the latency becomes stable, however we can't go lower than 51ms because of the fixed backoff here: https://github.com/grafana/pyroscope/blob/f4cec5d93dfbb8b49940e569520655ef1a47731b/pkg/experiment/metastore/client/methods.go#L24
For this, we can consider switching to a more aggressive and maybe exponential backoff.

**Bonus**
I've added a metastore client test page to make testing easier. 

Screenshot 2025-01-17 at 17 53 19

### DIFF
--- a/pkg/api/api_experimental.go
+++ b/pkg/api/api_experimental.go
@@ -30,7 +30,9 @@ func (a *API) RegisterQueryBackend(svc *querybackend.QueryBackend) {
 
 func (a *API) RegisterMetastoreAdmin(adm *metastoreadmin.Admin) {
 	a.RegisterRoute("/metastore-nodes", adm.NodeListHandler(), false, true, "GET", "POST")
+	a.RegisterRoute("/metastore-client-test", adm.ClientTestHandler(), false, true, "GET", "POST")
 	a.indexPage.AddLinks(defaultWeight, "Metastore", []IndexPageLink{
 		{Desc: "Nodes", Path: "/metastore-nodes"},
+		{Desc: "Client Test", Path: "/metastore-client-test"},
 	})
 }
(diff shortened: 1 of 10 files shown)


aleks-p:
I tried changing the metastore server identity to align identifiers between the metastore client and metastore error responses, however the server identity is currently tied to the Raft peer discovery so we can't do much without doing similar hacks in the metastore itself. Therefore, I suggest that we proceed with the workaround as is, for now.

I have also switched to an exponential backoff with a lower initial delay. This improves latencies during normal cluster transitions even further.

kolesnikovae (approved):

kolesnikovae:
Thanks for addressing the issue Aleks! We will see if the discovery mechanism can be improved in the future

Issue #3853, opened 2025-01-19
Ruby example span profiles not working


nateberkopec:
#### Describe the bug

AFAICT span profiles/Tempo connection isn't working in the Ruby language example.

#### To Reproduce

1. Start the Ruby rideshare example.
2. I should see some traces in Explore -> Tempo, e.g. with a blank query. I don't see anything.

#### Environment

Following instructions using `docker compose`.

marcsanmi:
Hi, @nateberkopec,  

Thanks for reporting this.

I believe this is because of the breaking change introduced in the latest tempo [release](https://github.com/grafana/tempo/releases/tag/v2.7.0) where the Open-Telemetry Collector receiver will connect to localhost instead of all interfaces 0.0.0.0.

For more details see https://github.com/grafana/tempo/issues/4465.

I'd say is just a matter of updating `tempo.yml`. I'll work on this.

marcsanmi:
Closing this since it was already fixed in https://github.com/grafana/pyroscope/pull/3855. Feel free to reopen if something arises.

nateberkopec:
After this change, I no longer see Tempo in the `Explore` tab or under `Data Sources`. 

I also see in the logs:

```
eu-north-1        | I, [2025-01-22T03:24:45.547374 #7]  INFO -- : [357620f4-89c5-4436-8a60-d69f631909e0] Processing by CarController#index as */*
ap-south-1        | E, [2025-01-22T03:24:46.682577 #7] ERROR -- : OpenTelemetry error: unexpected error in Jaeger::CollectorExporter#export - Failed to open TCP connection to localhost:14268 (Connection refused - connect(2) for "localhost" port 14268)
ap-south-1        | E, [2025-01-22T03:24:46.682603 #7] ERROR -- : OpenTelemetry error: Unable to export 9 spans
eu-north-1        | E, [2025-01-22T03:24:46.837973 #7] ERROR -- : OpenTelemetry error: unexpected error in Jaeger::CollectorExporter#export - Failed to open TCP connection to localhost:14268 (Connection refused - connect(2) for "localhost" port 14268)
eu-north-1        | E, [2025-01-22T03:24:46.938776 #7] ERROR -- : OpenTelemetry error: Unable to export 3 spans
```

nateberkopec:
FWIW I would also be happy to try this out on play.grafana.org, but the Tempo connection there seems to only be setup for the Go rideshare demo.

benedictfischer09:
[...]

PR #3855, opened 2025-01-20
fix: update tracing tempo configs to point to localhost

marcsanmi:
See https://github.com/grafana/pyroscope/issues/3853


simonswine (approved):
LGTM

Thank you!

PR #3856, opened 2025-01-20
Update golang version to 1.22.11

simonswine:


examples/language-sdk-instrumentation/golang-push/rideshare/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.23.4
```
simonswine, line 1: There are a few downgrades from 1.23. Would like to keep them managed by that script. Once we get a 1.24, we will just upgrade all of pyroscope.

korniltsev (approved):

PR #3860, opened 2025-01-21
chore(v2): improve error handling in compaction-worker

kolesnikovae:
Currently, a compaction job is abandoned by compaction workers if the blocks cannot be found in storage or if the metadata for the blocks cannot be retrieved. Such jobs remain incomplete and linger in the scheduler table indefinitely. This change enforces that compaction workers report an "empty" result for jobs that are known a priori to be uncompletable. The job entry is then removed from the compaction planner queue and the job scheduler, without replacing any blocks.

We need a mechanism that will evict such jobs, but this is out of scope of the PR.


aleks-p (approved):
LGTM

PR #3861, opened 2025-01-21
feat: record metrics from rules and export to remote

alsoba13:
In this PR, we introduce a first version of the metrics recorder and metrics exporter.

Every level 1 compaction job will record metrics from profiles in the form of time series. The recording will follow some recording rules given by config or an external service (for now, this is hardcoded to a single recording rule). The recorded metrics are exported to a remote after the compaction process.

Generated metrics are aggregations of total values of some kind of dimension (or profile type). The aggregation process is explained below:
* Given a recording rule with a profile type _T_, a filter _F_ (made of a set of key-value) and a set of labels _E_ to export.
* Every profile seen during the compaction that matches _T_ and _F_ will be considered for the aggregation.
* To aggregate, profiles are grouped by _E_, resulting in multiple time series.
* Every time serie will have a single sample with `time = blockTime` and value equal to the sum of all _totalValues_ that match (_T_, _F_, _E_).
* Hence, as we are adding up all totalValues that fulfill the conditions, we are conceptually aggregating by time (we discard the original profile timestamp and use the block time), resulting into a single sample per series per compaction job.

### Example:
Let's consider the following profiles present in some blocks being compacted

| profile | profile type | labels | totalValue | stacktraces (ignored) | timestamp (ignored) |
|-|-|-|-|-|-|
| 1 | memory alloc_space | `{service_name="worker", job="batch_compress", region="eu"}` | 10 | ... | ... |
| 2 | cpu samples | `{service_name="worker", job="batch_compress", region="eu"}` | 20 | ... | ... |
| 3 |  cpu samples | `{service_name="API", region="eu"}` | 1 | ... | ... |
| 4 |  cpu samples | `{service_name="worker", job="batch_compress", region="ap"}` | 30 | ... | ... |
| 5 |  cpu samples | `{service_name="worker", job="batch_compress", region="us"}` | 40 | ... | ... |
| 6 |  cpu samples | `{service_name="worker", job="batch_compress", region="eu"}` | 100 | ... | ... |

And the following recording rule:
Name = "cpu_usage_compress_workers"
T = cpu samples
F = `{service_name="worker", job="batch_compress"}`
E = `"region"`

This will result in the following exported series and samples.
`{__name__="cpu_usage_compress_workers", service_name="worker", job="batch_compress", region="eu"} = (t, 120)`
`{__name__="cpu_usage_compress_workers", service_name="worker", job="batch_compress", region="ap"} = (t, 30)`
`{__name__="cpu_usage_compress_workers", service_name="worker", job="batch_compress", region="us"} = (t, 40)`

Note that Profile 1 was discarded by profile type. Profiles 2 and 6 were aggregated, and Profile 3 was discarded by filter. For all of the 3 exported samples, t = blockTime.

Given the distributed architecture and concurrent nature of compactors, and the chosen timestamp for samples, time collisions may happen. For that reason, an extra `__pyroscope_instance__` label has been added, so that two compaction jobs may write to prometheus without causing overwrites. This intance id is computed from a worker id and a shard id.

Next steps:
* Get the export config programmatically so every metric is exported to the expected datasource (tenant-wise)
* Read rules from external service (tenant-settings?) and config.
* Error handling: lack of error handling is evident. There's a lot of room here for improvement but we should strive to not interfere with compaction and consider retries vs metrics loss.

Out of scope right now:
* functions/stacktraces processing


kolesnikovae (commented):
Good work, Alberto! 🚀 

I'd like to discuss the queries we aim to answer. Have you analyzed how the exported metrics will be used? Just some example use cases

pkg/experiment/metrics/recorder.go
```diff
@@ -0,0 +1,149 @@
...
+			sort.Sort(exportedLabels)
+			f := AggregatedFingerprint(exportedLabels.Hash())
+			aggregatedFp = &f
+
+			recording.fps[fp] = aggregatedFp
+			recording.data[*aggregatedFp] = newTimeSeries(exportedLabels, r.recordingTime)
+		}
+		recording.data[*aggregatedFp].Samples[0].Value += float64(totalValue)
+	}
+}
```
kolesnikovae, line 75: We can optimize it in the way that we only call it once per series and keep aggregators prepared.

This is a tight loop – if we can avoid allocations, loops, lookups, etc. we should go for it. We don't have to implement low level optimizations, but we should avoid wasteful operations

pkg/experiment/metrics/recorder.go
```diff
@@ -0,0 +1,149 @@
...
+		return nil, false
+	}
+	for _, matcher := range r.rule.matchers {
+		// assume labels.MatchEqual for every matcher:
+		if labelsMap[matcher.Name] != matcher.Value {
+			return nil, false
+		}
+	}
+	return nil, true
+}
```
kolesnikovae, line 136: This is a very reduced version of the syntax we use in queries. I think we should support the full one.

pkg/experiment/block/compaction.go
```diff
@@ -330,6 +369,13 @@ func (m *datasetCompaction) open(ctx context.Context, path string) (err error) {
 	return nil
 }
 
+func pyroscopeInstanceHash(shard uint32, id uuid.UUID) string {
+	buf := make([]byte, 0, 40)
+	buf = append(buf, byte(shard>>24), byte(shard>>16), byte(shard>>8), byte(shard))
+	buf = append(buf, id.String()...)
+	return fmt.Sprintf("%x", xxhash.Sum64(buf))
+}
```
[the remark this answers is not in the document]
alsoba13: Fixed to use `CreatedBy` instead of worker id

pkg/experiment/block/compaction.go
```diff
@@ -62,6 +66,7 @@ func Compact(
 	ctx context.Context,
 	blocks []*metastorev1.BlockMeta,
 	storage objstore.Bucket,
+	workerId uuid.UUID,
```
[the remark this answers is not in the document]
alsoba13: Great, I'll be rewriting this next. Thank you for the tips

pkg/experiment/metrics/recorder.go
alsoba13: Talked in private, I'll rewrite it to take advantage of the order property of the rows.

pkg/experiment/metrics/recorder.go
alsoba13: Fixed to use the complete matcher

pkg/experiment/block/compaction.go
```diff
@@ -179,18 +198,27 @@ func newBlockCompaction(
...
+	tmpdir string,
+	observer SampleObserver,
+) (m *metastorev1.BlockMeta, err error) {
 	w := NewBlockWriter(dst, b.path, tmpdir)
 	defer func() {
 		err = multierror.New(err, w.Close()).Err()
 	}()
 	// Datasets are compacted in a strict order.
 	for _, s := range b.datasets {
+		s.registerSampleObserver(observer)
```
alsoba13, line 213: At this point I think it's better to register observer in the dataset instead of passing it through this long call chain `compact > mergeAndClose > merge > writeRow`

pkg/experiment/metrics/exporter.go
```diff
@@ -0,0 +1,108 @@
...
+	return e.client.Store(context.Background(), snappy.Encode(nil, buf.Bytes()), 0)
+}
+
+func newClient(cfg Config) remote.WriteClient {
+	wURL, err := url.Parse(cfg.url)
+	if err != nil {
+		panic(err)
+	}
+
+	c, err := remote.NewWriteClient("exporter", &remote.ClientConfig{
```
simonswine, line 81: Feedback from the Mimir team, we should set a custom user agent to identify this is the "pyroscope-metrics-exporter". See https://raintank-corp.slack.com/archives/C03NCLB4GG7/p1738233634244319

pkg/experiment/block/compaction.go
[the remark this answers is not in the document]
alsoba13: This may be addressed at https://github.com/grafana/pyroscope-squad/issues/336

pkg/experiment/block/compaction.go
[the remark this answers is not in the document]
alsoba13: Rewritten

pkg/experiment/metrics/exporter.go
alsoba13: done

pkg/experiment/metrics/observer.go
```diff
@@ -0,0 +1,44 @@
...
+	return fmt.Sprintf("%x", xxhash.Sum64(buf))
+}
+
+func (o *MetricsExporterSampleObserver) Observe(row block.ProfileEntry) {
+	o.recorder.RecordRow(row.Fingerprint, row.Labels, row.Row.TotalValue())
+}
+
+func (o *MetricsExporterSampleObserver) Flush() error {
+	go func() {
+		NewExporter(o.tenant, o.recorder.Recordings).Send() // TODO log error
```
simonswine, line 41: Before merging this should log some errors

pkg/experiment/metrics/exporter.go
```diff
@@ -0,0 +1,109 @@
...
+				Password: cfg.password,
+			},
+		},
+		SigV4Config:      nil,
+		AzureADConfig:    nil,
+		Headers:          nil,
+		RetryOnRateLimit: false,
+	})
+	if err != nil {
+		panic(err)
```
simonswine, line 97: Should return error instead

pkg/experiment/metrics/exporter.go
```diff
@@ -0,0 +1,109 @@
...
+	return c
+}
+
+func configFromTenant(tenant string) Config {
+	// TODO
+	return Config{
+		url:      "omitted",
+		username: "omitted",
+		password: "omitted",
+	}
```
simonswine, line 108: When this is hardcoded, how can this be used?

 Eventually you could read from environment variables until this is figured out.

pkg/experiment/metrics/exporter.go
```diff
@@ -0,0 +1,109 @@
...
+	if err := buf.Marshal(p); err != nil {
+		return err
+	}
+	return e.client.Store(context.Background(), snappy.Encode(nil, buf.Bytes()), 0)
+}
+
+func newClient(cfg Config) remote.WriteClient {
+	wURL, err := url.Parse(cfg.url)
+	if err != nil {
+		panic(err)
```
simonswine, line 79: Should return error instead

pkg/experiment/metrics/exporter.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	return exporter
+}
+
+func (e *Exporter) Send() error {
+	if len(e.data) == 0 {
+		return nil
+	}
+	if e.client == nil {
+		e.client = newClient(e.config)
```
simonswine, line 47: I would create the client at NewExporter time as well and then handle problems as they arise

pkg/experiment/metrics/exporter.go
simonswine: I am not too sure where? Can you point me to the piece of code

pkg/experiment/metrics/rules.go
```diff
@@ -0,0 +1,52 @@
...
+
+type RecordingRule struct {
+	profileType string
+	metricName  string
+	matchers    []*labels.Matcher
+	keepLabels  []string
+}
+
+func recordingRulesFromTenant(tenant string) []*RecordingRule {
+	// TODO
```
simonswine, line 15: Even though you want to hardcode and test it, I would prefer those coming from e.g. an environment variable or a file (and parsed as yaml)so we can change them quicker than recompiling a new version.

pkg/experiment/metrics/exporter.go
alsoba13: I re-read it. I thought we had to change the NewWriteClient name field. I'll do it with user agent instead. Thanks for pointing it out

pkg/experiment/metrics/exporter.go
alsoba13: I thought that lazy client initialization would be convenient to save some resources for recorders hitting no data in a given block.

pkg/experiment/metrics/exporter.go
alsoba13: I will inject it by config/env vars in a following PR.

pkg/experiment/metrics/exporter.go
alsoba13: Good catch. This came from the old code before Observer pattern, where exporter was created at the beginning and I was trying to save resources by delaying client creation. I'll create the client at the NewExporter once we ensure e.data is not empty

pkg/experiment/block/compaction.go
```diff
@@ -179,18 +212,28 @@ func newBlockCompaction(
...
 		if err = s.compact(ctx, w); err != nil {
 			return nil, fmt.Errorf("compacting block: %w", err)
 		}
 		b.meta.Datasets = append(b.meta.Datasets, s.meta)
 	}
+	if err = observer.Flush(); err != nil {
+		return nil, fmt.Errorf("flushing sample observer: %w", err)
+	}
```
kolesnikovae, line 236: 1. I see no reasons why we want to fail a compaction job because of the metric export errors.
2. We may want to skip the entire dataset if it cannot possibly match any rule. 
3. I'm not sure why we're flushing after all the datasets have been processed. My idea was to flush observer when the dataset is closed in `compact`: as we release resources referenced by `ProfileEntry`, observer must not access them after the `Flush` call.

pkg/experiment/metrics/observer.go
```diff
@@ -0,0 +1,64 @@
...
+			level.Error(o.logger).Log("msg", "error creating metrics exporter", "err", err)
+			return
+		}
+
+		if err = exporter.Send(); err != nil {
+			level.Error(o.logger).Log("msg", "error sending recording metrics", "err", err)
+		}
+	}(o.tenant, rec)
+	return nil
+}
```
kolesnikovae, line 64: I think the `Exporter` should be really a dependency of the  `MetricsExporterSampleObserver`. Also, please, consider alternative names. For example, `metrics.SampleObserver` would say everything we need to know. Moreover, please consider incorporating `MetricsExporterSampleObserver` and `Recorder` – as far as I can see, there's no reason to separate their responsibilities.

I would suggest implementing a simple channel/queue that is handled in the background (e.g., per tenant, or tenant/rule). Keep in mind that we want to flush all samples on shutdown.

pkg/experiment/metrics/recorder.go
```diff
@@ -0,0 +1,147 @@
...
+
+type TimeSeries struct {
+	Labels  labels.Labels
+	Samples []Sample
+}
+
+type Sample struct {
+	Value     float64
+	Timestamp int64
+}
```
kolesnikovae, line 62: I'm not very sure if I understand why we have a concept of time series here. In the code I only see a single sample

pkg/experiment/metrics/recorder.go
```diff
@@ -0,0 +1,147 @@
...
+	aggregatedFp := AggregatedFingerprint(exportedLabels.Hash())
+	timeSeries, ok := r.data[aggregatedFp]
+	if !ok {
+		timeSeries = newTimeSeries(exportedLabels, recordingTime)
+		r.data[aggregatedFp] = timeSeries
+	}
+	r.state.timeSeries = timeSeries
+}
+
+type AggregatedFingerprint model.Fingerprint
```
kolesnikovae, line 52: It's somewhat confusing. This is still a fingerprint, the context we use it in does not change its type. Such an alias does not really add type safety.

pkg/experiment/metrics/recorder.go
```diff
@@ -0,0 +1,147 @@
...
+	state *recordingState
+}
+
+type recordingState struct {
+	fp         *model.Fingerprint
+	matches    bool
+	timeSeries *TimeSeries
+}
+
+func (r *Recording) InitState(fp model.Fingerprint, lbls phlaremodel.Labels, pyroscopeInstance string, recordingTime int64) {
```
kolesnikovae, line 30: For me it's not clear why this is separate from `MetricsExporterSampleObserver`, given that the latter is practically a wrapper for `Recorder` – you can simply reset its state on `MetricsExporterSampleObserver.Init`

pkg/experiment/metrics/exporter.go
```diff
@@ -0,0 +1,114 @@
...
+		client, err := newClient(cfg)
+		if err != nil {
+			return nil, err
+		}
+		exporter.client = client
+	}
+	return exporter, nil
+}
+
+func (e *Exporter) Send() error {
```
kolesnikovae, line 47: I'd expect to see something like:
```go
func (e *Exporter) Send([]prompb.TimeSeries) error
```

I'm wondering if all these conversions are required. Shouldn't `prompb` be suffice?

pkg/experiment/metrics/observer.go
```diff
@@ -0,0 +1,64 @@
...
+
+func NewMetricsExporterSampleObserver(meta *metastorev1.BlockMeta, logger log.Logger) *MetricsExporterSampleObserver {
+	recordingTime := int64(ulid.MustParse(meta.Id).Time())
+	pyroscopeInstance := pyroscopeInstanceHash(meta.Shard, meta.CreatedBy)
+	return &MetricsExporterSampleObserver{
+		recordingTime:     recordingTime,
+		pyroscopeInstance: pyroscopeInstance,
+		logger:            logger,
+	}
+}
```
kolesnikovae, line 31: The entire thing with `pyroscopeInstance` label can be offloaded from this component. From its perspective, it is just an extra label it will be adding.

pkg/experiment/metrics/observer.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (o *MetricsExporterSampleObserver) Init(tenant string) {
+	o.tenant = tenant
+	o.recorder = NewRecorder(recordingRulesFromTenant(tenant), o.recordingTime, o.pyroscopeInstance)
+}
+
+func (o *MetricsExporterSampleObserver) Observe(row block.ProfileEntry) {
+	o.recorder.RecordRow(row.Fingerprint, row.Labels, row.Row.TotalValue())
+}
```
kolesnikovae, line 47: Just a note: explicit `Init` is not required, tbh: `block.ProfileEntry` has a deference to the dataset the row comes from, including its metadata

kolesnikovae:
I think before merging the PR we should:
1. Define the model of our rules. See [the comment](https://github.com/grafana/pyroscope/pull/3874#pullrequestreview-2592379742).
2. Improve composition of components and clarify their lifecycle.
   * Usually, components that have side effects (e.g., external calls) are injected as dependencies and are not created in-place (like exporter).
   * We should not abandon goroutines (like in exporter)
3. Implement component that manages recording rules.
   * We probably don't want to fetch them for each compaction job.
   * We don't want to fail compaction if the rules cannot be retrieved.
5. Add basic unit tests.

pkg/experiment/metrics/recorder.go
alsoba13: It does provide type safety `Cannot use 'r. state. fp' (type *model. Fingerprint) as the type AggregatedFingerprint`, but yeah, nothing stops you from casting it to skip the type safety. For me this was more on the semantics and readability, but given that it's introducing confusion I agree on get rid of it.

pkg/experiment/block/compaction.go
alsoba13: 1. yes, I forgot to remove this, I'll just log an error.
2. It makes sense to distinguish rules by service_name, may explore it later.
3. My idea was to optimize communication here, record them all and send just once for each tenant. I don't mind changing it a bit and remote-writing multiple times.

pkg/experiment/metrics/recorder.go
alsoba13: You are right. It was here since I started by exporting every sample separately (before aggregation). I get rid of it at [5d277d](https://github.com/grafana/pyroscope/pull/3861/commits/5d277dd5606e47c2a7018b28e934f68495fe85b9)

pkg/experiment/metrics/observer.go
alsoba13: solved at [4775a2](https://github.com/grafana/pyroscope/pull/3861/commits/4775a2ed68e811fc35ae9592493b8ad0d27c882a)

pkg/experiment/metrics/exporter.go
alsoba13: solved at [4775a2](https://github.com/grafana/pyroscope/pull/3861/commits/4775a2ed68e811fc35ae9592493b8ad0d27c882a)

pkg/experiment/metrics/exporter.go
alsoba13: solved at [4775a2](https://github.com/grafana/pyroscope/pull/3861/commits/4775a2ed68e811fc35ae9592493b8ad0d27c882a)

pkg/experiment/metrics/rules.go
alsoba13: Talked in slack, I'll merge https://github.com/grafana/pyroscope/pull/3874 first.

pkg/experiment/block/compaction.go
alsoba13: 1. Yes, I forgot to remove this, I'll just log an error.
2. Will be hard to implement from block package with such a simple api (observe and flush), what do you really mean? Do you mean implementing it behind the scenes at the metrics.SampleObserver?
3. My idea was to minimize network here, record them all and send just once for each rule/tenant. I don't mind changing it a bit and remote-writing multiple times.

pkg/experiment/metrics/observer.go
alsoba13: I did all at [b0f8c17](https://github.com/grafana/pyroscope/pull/3861/commits/b0f8c179d447e7bc50190ba6fdbbc86dd9375d1f).

Please note:
* Exporter being a dependency of observer means creating it at observer creation time. Keep in mind that a single Observer will export to multiple tenant's datasources, so client can't be created there. As a result, `Exporter` becomes an empty struct (I just preserved so we can discuss further and for future extensions). Data is now injected at `Send` as you requested in another comment.

I'll create an issue for the channel/queue part, it makes all the sense.

pkg/experiment/metrics/exporter.go
alsoba13: changed it at [b0f8c17](https://github.com/grafana/pyroscope/pull/3861/commits/b0f8c179d447e7bc50190ba6fdbbc86dd9375d1f).I did all at [b0f8c17](https://github.com/grafana/pyroscope/pull/3861/commits/b0f8c179d447e7bc50190ba6fdbbc86dd9375d1f).

Note that tenant or datasource config needs to be provided as well

pkg/experiment/metrics/recorder.go
alsoba13: I agree on the part of Recorder being merged to Observer. Already changed that part at [b0f8c17](https://github.com/grafana/pyroscope/pull/3861/commits/b0f8c179d447e7bc50190ba6fdbbc86dd9375d1f).


But I'm not sure I understand your last sentence. Restarting state here is important because `Recording` object will do aggregations based on samples that match the filters. We don't `InitState` once as `SampleObserver.Init`. When we find a sample that hits the filters, we store an state so we can fast aggregate rows of the same series. `InitState` here will prepare the state to fast append rows.

pkg/experiment/metrics/observer.go
alsoba13: > from this component
Clarification request: You mean to remove the `pyroscopeInstance` addition from the _compaction-worker_? Do you mean to create `pyroscopeInstance` label on the write path for every profile and include it on the export step?

pkg/experiment/metrics/observer.go
alsoba13: Yes, I considered. But it meant high effort:
* First, `block.Dataset.Meta()` won't have access to tenant name but tenant string index. Not sure if I'm missing something but couldn't extract from it.
* On the other hand, in the current implementation, not providing a tenant would give `Observe` more responsibilities that I would find convenient, i.e: fetching recording rules for that specific tenant. Note that there's a single Observer for all the tenants. Am I missing something? Were you imagining pre-fetching all the rules for all the tenants present in the source blocks at SampleObserver creation time?

pkg/experiment/metrics/observer.go
alsoba13: Issue for the background export https://github.com/grafana/pyroscope-squad/issues/354

pkg/experiment/metrics/recorder.go
alsoba13: Removed at [41de226](https://github.com/grafana/pyroscope/pull/3861/commits/41de2262b18e8e13352f8295b2497077ab363eaa)

pkg/experiment/metrics/observer.go
```diff
@@ -0,0 +1,197 @@
...
+		logger:            logger,
+		exporter:          &Exporter{},
+	}
+}
+
+func pyroscopeInstanceHash(shard uint32, createdBy int32) string {
+	buf := make([]byte, 0, 8)
+	buf = append(buf, byte(shard>>24), byte(shard>>16), byte(shard>>8), byte(shard))
+	buf = append(buf, byte(createdBy>>24), byte(createdBy>>16), byte(createdBy>>8), byte(createdBy))
+	return fmt.Sprintf("%x", xxhash.Sum64(buf))
```
kolesnikovae, line 56: You probably want to use `binary.AppendUint32`
```go
buf := make([]byte, 0, 8)
buf = binary.BigEndian.AppendUint32(buf, shard)
buf = binary.BigEndian.AppendUint32(buf, createdBy)
```
Or `Put`:
```go
buf := make([]byte, 8)
binary.BigEndian.PutUint32(buf[0:4], shard)
binary.BigEndian.PutUint32(buf[4:8], createdBy)
```
The order (including endianness) and does not matter.

pkg/experiment/metrics/observer.go
```diff
@@ -0,0 +1,197 @@
...
+				timeSeries = append(timeSeries, ts)
+			}
+		}
+
+		if err := o.exporter.Send(tenant, timeSeries); err != nil {
+			level.Error(o.logger).Log("msg", "error sending recording metrics", "err", err)
+		}
+	}(o.tenant, recs)
+	return nil
+}
```
kolesnikovae, line 118: I believe we should control concurrency in the `Exporter` implementation. Otherwise, it's implied that the `SampleObserver` knows implementation details of the `Exporter`. Moreover, a method called `Flush` is typically meant to be blocking.

pkg/experiment/metrics/exporter.go
```diff
@@ -0,0 +1,73 @@
...
+	if err := buf.Marshal(p); err != nil {
+		return err
+	}
+	cfg := configFromTenant(tenant)
+	client, err := newClient(cfg)
+	if err != nil {
+		return err
+	}
+	return client.Store(context.Background(), snappy.Encode(nil, buf.Bytes()), 0)
+}
```
kolesnikovae, line 36: As we're aimed to promote it from the prototype state, I think we should consider refactoring this bit.

1. We should avoid creating a new client on every `Send` call.
   * Exporter should maintain a pool of tenant-specific clients (see my other comment about the remote write targets, we probably should consider make the clients named, if a single tenant might have multiple write targets).
   * Exporter should poll for changes (of get notifications from outside) and reconcile the pool with the configuration.
   * Clients should be initialized lazily, on access.

2. Context: it is fine to use the background context here, however, we should consider tracing of the `Send` method and the outgoing call.

3. We may want to add metrics to keep track of the latencies and errors.

4. Limiting the amount of data per a `client.Store` call. This probably should be implemented as part of the background queue.

5. Here and elsewhere: if we can avoid allocating an empty buffer, we shouldn't miss the opportunity. I believe it's entirely possible to reuse a buffer that has already been allocated.

pkg/experiment/metrics/observer.go
kolesnikovae: > Exporter being a dependency of observer means creating it at observer creation time.

Not quite. See https://github.com/grafana/pyroscope/pull/3861#discussion_r1951967971

pkg/experiment/metrics/exporter.go
kolesnikovae: As discussed, we assume that a single remote write target is used (for all tenants, for all rules). This allows to simplify the p.1: as we only have a single client with a static configuration. The routing of the requests can be implemented on the other side.

We, however, still need to poll tenant-settings service for recording rules.

kolesnikovae: > 2. Context: it is fine to use the background context here, however, we should consider tracing of the Send method and the outgoing call.

We can add it later, if needed. Metrics are more important.

> 3. We may want to add metrics to keep track of the latencies and errors.

The client we use probably already has it. We just need to make sure these are enabled/configured (ideally, we need `pyroscope_` prefix).

> 4. Limiting the amount of data per a client.Store call. This probably should be implemented as part of the background queue.

Not now. Let's use a simple wait group (a goroutine per call _in_ `Exporter`) and wait for completion on shutdown of the exporter.

pkg/experiment/metrics/observer.go
kolesnikovae: From the observer. See https://github.com/grafana/pyroscope/pull/3861#discussion_r1951967971

pkg/experiment/block/compaction.go
```diff
@@ -52,11 +52,24 @@ func WithCompactionDestination(storage objstore.Bucket) CompactionOption {
...
+	destination    objstore.Bucket
+	tempdir        string
+	sampleObserver SampleObserver
+}
+
+type SampleObserver interface {
+	// Observe is called before the compactor appends the entry
+	// to the output block. This method must not modify the entry.
+	Observe(ProfileEntry)
 }
```
alsoba13, line 73: Opted for a simplified version of this interface for this stage, no need for a flush right now if we are not handling rows in the scope of datasets.

pkg/experiment/block/compaction.go
```diff
@@ -187,18 +200,23 @@ func newBlockCompaction(
...
 	w, err := NewBlockWriter(tempdir)
 	if err != nil {
 		return nil, fmt.Errorf("creating block writer: %w", err)
 	}
 	defer func() {
 		_ = w.Close()
 	}()
-
 	// Datasets are compacted in a strict order.
 	for i, s := range b.datasets {
```
alsoba13, line 217: Flush and Init removed, simplifying compaction

pkg/experiment/metrics/exporter.go
```diff
@@ -0,0 +1,106 @@
...
+	e.wg.Wait()
+}
+
+func newClient(remoteUrl, user, password string) (remote.WriteClient, error) {
+	wURL, err := url.Parse(remoteUrl)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: enhance client with metrics
```
alsoba13, line 91: remote.WriteClient does not record transport metrics. To write a new client in a separate PR.

pkg/experiment/metrics/observer.go
alsoba13, line 1: Observer states work as follow:
* When a new tenant is observed, **observer state** is initialized with a new set of recordings (from the rules).
* When observing a new fingerprint, every **recording state** will be precomputed. This way, following rows with the same fingerprint will be quickly identify as matching and where should their value be aggregated.

New tenant rows will cause an export.Send, also closing the SampleObserver will do so.

pkg/experiment/metrics/observer.go
```diff
@@ -0,0 +1,186 @@
...
+	o.state = nil
+}
+
+func (o *SampleObserver) Close() {
+	if o.state != nil {
+		o.flush()
+	}
+}
+
+func (r *Recording) initState(fp model.Fingerprint, rowLabels phlaremodel.Labels, externalLabels labels.Labels) {
```
alsoba13, line 127: Recording state works as follow:
* New fingerprint appears. 
* Labels are mapped to quick lookup.
* We determine if row matches the filters.
* If matches, we compute the labels to export and compute new fingerprint.
  * Then fetch/create the series (sample) it belongs.

pkg/experiment/metrics/rules.go
```diff
@@ -0,0 +1,117 @@
...
+
+type StaticAnonymousRuler struct {
+	rules []*RecordingRule
+}
+
+type RecordingRule struct {
+	Matchers       []*labels.Matcher
+	KeepLabels     []string
+	ExternalLabels labels.Labels
+}
```
alsoba13, line 40: I opted to create a new data model here to:
* have matchers parsed before hand.
* generalize `__profile_type__` in matchers and _metric_name_ in external labels.

pkg/experiment/compactor/compaction_worker.go
```diff
@@ -176,6 +190,9 @@ func (w *Worker) running(ctx context.Context) error {
 	ticker.Stop()
 	close(stopPolling)
 	<-pollingDone
+	// Force exporter to send all staged samples (depends on the implementation)
+	// Must be a blocking call.
+	w.exporter.Flush()
```
simonswine: This should go moved into its separate module

pkg/experiment/metrics/observer.go
```diff
@@ -0,0 +1,186 @@
...
+		o.state.recordings[i] = &Recording{
+			rule:  rule,
+			data:  make(map[model.Fingerprint]*Sample),
+			state: &recordingState{},
+		}
+	}
+}
+
+func (o *SampleObserver) Observe(row block.ProfileEntry) {
+	tenant := row.Dataset.TenantID()
```
simonswine, line 73: If I understood https://github.com/grafana/pyroscope/pull/3861#discussion_r1951967971, correctly this should be thread safe, this code is not. As two parallel calls will conflict on the `o.state`.

Maybe it makes sense to keep multiple instances of the state open.

pkg/experiment/metrics/observer.go
simonswine: Also worth code comment instead

pkg/experiment/metrics/observer.go
alsoba13: Multiple compaction workers will use ruler and exporter, but every job will have their own instance of SampleObserver, and will sequentially observe all rows within the job

pkg/experiment/metrics/rules.go
alsoba13: Also, I accept naming suggestions. I kept `KeepLabels` although `GroupBy` may be more convenient

pkg/experiment/metrics/observer.go
simonswine: Sounds good! Worth a comment

pkg/experiment/metrics/observer.go
```diff
@@ -0,0 +1,215 @@
...
+type recordingState struct {
+	fp      model.Fingerprint
+	matches bool
+	series  *prompb.TimeSeries
+}
+
+type Sample struct {
+	Labels labels.Labels
+	Value  float64
+}
```
kolesnikovae, line 44: Can't find where the type is used

pkg/experiment/metrics/observer.go
```diff
@@ -0,0 +1,215 @@
...
+
+	series, ok := r.data[aggregatedFp]
+	if !ok {
+		series = newTimeSeries(exportedLabels, recordingTime)
+		r.data[aggregatedFp] = series
+	}
+	r.state.series = &series
+}
+
+func newTimeSeries(exportedLabels labels.Labels, recordingTime int64) prompb.TimeSeries {
```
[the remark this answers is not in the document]
alsoba13: good catch!

simonswine (approved):
Thanks for patiently addressing our feedback, this is looks ready to go.


LGTM!

PR #3863, opened 2025-01-22
fix(v2): goroutine leak in DLQ recovery

kolesnikovae:


pkg/experiment/metastore/dlq/recovery.go
```diff
@@ -91,62 +91,71 @@ func (r *Recovery) recoverLoop(ctx context.Context) {
...
 		return nil
 	}
-	shard, _ := strconv.ParseUint(sshard, 10, 64)
-	if ulid != meta.Id || meta.Shard != uint32(shard) {
-		level.Error(r.logger).Log("msg", "unexpected block meta", "path", metaPath, "meta", fmt.Sprintf("%+v", meta))
+
+	switch _, err = r.metastore.AddRecoveredBlock(ctx, &metastorev1.AddBlockRequest{Block: &meta}); {
+	case err == nil:
 		return nil
-	}
```
kolesnikovae, line 122: Removing it from here because this is not a concern of the recovery process – as long as the object can be unmarshaled and the service accepts it, it's a valid metadata entry

pkg/experiment/metastore/dlq/recovery.go
```diff
@@ -91,62 +91,71 @@ func (r *Recovery) recoverLoop(ctx context.Context) {
...
-	if _, err = r.metastore.AddRecoveredBlock(ctx, &metastorev1.AddBlockRequest{Block: meta}); err != nil {
-		if raftnode.IsRaftLeadershipError(err) {
-			return err
-		}
-		level.Error(r.logger).Log("msg", "failed to add block", "err", err, "path", metaPath)
+	case status.Code(err) == codes.InvalidArgument:
+		level.Error(r.logger).Log("msg", "invalid block metadata", "err", err, "path", path)
 		return nil
```
kolesnikovae, line 142: Basically, at this point we used to drop a metadata entry due to a transient/network error

pkg/experiment/metastore/dlq/recovery.go
```diff
@@ -91,62 +91,71 @@ func (r *Recovery) recoverLoop(ctx context.Context) {
...
-		return nil, err
-	}
-	recovered := new(metastorev1.BlockMeta)
-	err = recovered.UnmarshalVT(metaBytes)
+func (r *Recovery) readObject(ctx context.Context, path string) ([]byte, error) {
+	rc, err := r.bucket.Get(ctx, path)
 	if err != nil {
 		return nil, err
 	}
-	return recovered, nil
```
kolesnikovae, line 151: Here's where we leak the goroutine: the `bucket.Get` returns an `io.ReadClosed`, and it references the storage object client implementation until the `Close` call.

pkg/experiment/metastore/dlq/recovery.go
```diff
@@ -91,62 +91,71 @@ func (r *Recovery) recoverLoop(ctx context.Context) {
...
 
 func (r *Recovery) recoverTick(ctx context.Context) {
-	err := r.bucket.Iter(ctx, block.DirNameDLQ, func(metaPath string) error {
-		if ctx.Err() != nil {
-			return ctx.Err()
-		}
-		return r.recover(ctx, metaPath)
+	err := r.bucket.Iter(ctx, block.DirNameDLQ, func(path string) error {
+		return r.recover(ctx, path)
 	}, objstore.WithRecursiveIter)
```
kolesnikovae, line 96: This is not helpful to check the context in this way. If the iterator doesn't handle context errors (implementations we use do handle it well), it's a bug.

korniltsev (approved):

PR #3865, opened 2025-01-22
Add collection rules to tenant settings service

simonswine:
Replaces #3416


api/settings/v1/setting.proto
```diff
@@ -26,3 +26,72 @@ message Setting {
...
+
+message JavaSettings {
+  bool enabled = 1;
+}
+
+message EBPFSettings {
+  bool enabled = 1;
+}
+
+// CollectionRuleUpsertRequest is used to create / update a collection rule.
```
aleks-p, line 50: nit, the comment here doesn't match the message name

pkg/settings/collection/collection.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+func (cfg *Config) Validate() error {
+	if !cfg.Enabled {
+		return nil
+	}
+
+	if cfg.PyroscopeURL == "" {
+		return fmt.Errorf(
+			"tenant_settings.collection_rule.spyroscope_url is required when the tenant_settings.collection_rules.enabled is set",
```
aleks-p, line 61: nit, should this be using dashes to make it consistent with how the flags are defined above? (same for the other check further down)

pkg/settings/collection/store.go
```diff
@@ -0,0 +1,303 @@
...
+//go:embed config.alloy.gotempl
+var alloyTemplate string
+
+func (b *bucketStore) template(s *settingsv1.CollectionRuleStore) (string, error) {
+	vars := b.varsFromStore(s)
+
+	// generate the pyroscope config
+	var config strings.Builder
+
+	configTemplate, err := template.New("config.alloy.gotempl").Parse(alloyTemplate)
```
aleks-p, line 106: would we benefit from parsing this only once at startup?

pkg/settings/collection/store.go
```diff
@@ -0,0 +1,303 @@
...
+
+	// reset cache
+	b.cache = nil
+
+	// write to bucket
+	return b.bucket.Upload(ctx, b.path, bytes.NewReader(dataJson))
+}
+
+// updates or create a new rule
+func (b *bucketStore) upsertRule(ctx context.Context, rule *settingsv1.UpsertCollectionRuleRequest) error {
```
aleks-p, line 204: If two users are updating the same rule (doesn't have to be concurrently), will one user's action overwrite what the other did? If yes, could we use the generation field to prevent such edits (tell the second user that they are editing outdated data)? Does this apply to other operations as well?

aleks-p (approved):
LGTM, left a few minor remarks / questions.

pkg/settings/collection/store.go
simonswine: I think I would mainly worry about the Upsert operation. While the generation is stored on the whole set for a tenant, the last updated, is scoped to the tenant and rule name. Implemented in d557246926f5113c126a675d7ea3f646e9e60810

pkg/settings/collection/store.go
simonswine: Not really yet, I don't expect more than acouple of requests a day, I also don't think text/template is made for concurrently Executing, so would require a bit more thinking.

pkg/settings/collection/store.go
aleks-p: looks good to me

PR #3872, opened 2025-01-30
fix(v2): honour series limit in query router

kolesnikovae:
The series limit parameter is ignored when the query backend is enabled and the query is not split.


marcsanmi (approved):
LGTM 👍

PR #3874, opened 2025-01-30
feat: fetch recording rules from tenant-settings

alsoba13:
This is a draft. Plenty of assumptions have been done. WIP

### DIFF
--- a/pkg/experiment/metrics/recorder.go
+++ b/pkg/experiment/metrics/recorder.go
@@ -66,7 +66,6 @@ func NewRecorder(recordingRules []*RecordingRule, recordingTime int64, pyroscope
 	for i, rule := range recordingRules {
 		recordings[i] = &Recording{
 			rule: *rule,
-			// fps:  make(map[model.Fingerprint]*AggregatedFingerprint),
 			data: make(map[AggregatedFingerprint]*TimeSeries),
 			state: &recordingState{
 				fp: nil,
@@ -111,7 +110,7 @@ func generateExportedLabels(labelsMap map[string]string, rec *Recording, pyrosco
 			Value: rec.rule.metricName,
 		},
 		labels.Label{
-			Name:  "__pyroscope_instance__",
+			Name:  "pyroscope_instance",
 			Value: pyroscopeInstance,
 		},
 	}
--- a/pkg/experiment/metrics/recorder.go
+++ b/pkg/experiment/metrics/recorder.go
@@ -114,28 +114,7 @@ func generateExportedLabels(labelsMap map[string]string, rec *Recording, pyrosco
 			Value: pyroscopeInstance,
 		},
 	}
-	// Add filters as exported labels
-	/*
-			TODO: do we want this?
-			Adding filters as exported labels seems unoffensive / ambiguous when using like this:
-				matchers -> [vehicle="car"]   (the exported metric will include a single-value label "vehicle"
-		  But it's different for more complex matchers:
-			  matchers -> [pod=~"prod-us-.*"]
-					here, if we export pod, we are generating new series for each production US pod. While if we don't export it,
-		      a single serie is generated that aggregates all prod US pods.
 
-			Maybe we should not include this and let the user choose if they want to keep the pod label or not.
-	*/
-
-	for _, matcher := range rec.rule.matchers {
-		labelValue, ok := labelsMap[matcher.Name]
-		if ok {
-			exportedLabels = append(exportedLabels, labels.Label{
-				Name:  matcher.Name,
-				Value: labelValue,
-			})
-		}
-	}
 	// Keep the expected labels
 	for _, label := range rec.rule.keepLabels {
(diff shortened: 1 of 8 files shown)


pkg/experiment/metrics/recorder.go
```diff
@@ -115,11 +115,26 @@ func generateExportedLabels(labelsMap map[string]string, rec *Recording, pyrosco
...
+					here, if we export pod, we are generating new series for each production US pod. While if we don't export it,
+		      a single serie is generated that aggregates all prod US pods.
+
+			Maybe we should not include this and let the user choose if they want to keep the pod label or not.
+	*/
+
 	for _, matcher := range rec.rule.matchers {
-		exportedLabels = append(exportedLabels, labels.Label{
-			Name:  matcher.Name,
-			Value: matcher.Value,
```
alsoba13, line 121: this was in fact a bug. I was appending name value pair, when the label value may not be the label filter value (for exmaple `vehicle!="car"`)-

Anyway, considering removing this part as the code comment argues

alsoba13: I finally removed.

pkg/experiment/metrics/recorder.go
```diff
@@ -135,7 +129,7 @@ func generateExportedLabels(labelsMap map[string]string, rec *Recording, pyrosco
 }
 
 func (r *Recording) matches(labelsMap map[string]string) bool {
-	if r.rule.profileType != labelsMap["__profile_type__"] {
+	if r.rule.profileType != labelsMap["__type__"] {
```
alsoba13, line 132: TODO: solve ambiguities: `__type__` is not enough! `block/contentions` and `block/delays` can't be distinguish from `mutex/contentions` and `mutex/delays`

pkg/experiment/metrics/rules.go
```diff
@@ -1,7 +1,23 @@
...
 	"github.com/prometheus/prometheus/model/labels"
+
+	settingsv1 "github.com/grafana/pyroscope/api/gen/proto/go/settings/v1"
+	"github.com/grafana/pyroscope/api/gen/proto/go/settings/v1/settingsv1connect"
+	connectapi "github.com/grafana/pyroscope/pkg/api/connect"
+	"github.com/grafana/pyroscope/pkg/tenant"
+	"github.com/grafana/pyroscope/pkg/util"
 )
 
 type RecordingRule struct {
```
simonswine, line 23: It would be better if this is a protobuf defined type.

pkg/experiment/metrics/rules.go
```diff
@@ -48,5 +87,101 @@ func recordingRulesFromTenant(tenant string) []*RecordingRule {
...
+
+	return &RecordingRule{
+		profileType: parsed.ProfileType,
+		metricName:  "pyroscope_exported_metrics_" + parsed.Name, // TODO sanitize
+		matchers:    parseMatchers(parsed.Matcher, parsed.ServiceName),
+		keepLabels:  parsed.Labels, // [] != All
+	}
+}
+
+func parseMatchers(matchersString string, serviceName string) []*labels.Matcher {
```
simonswine, line 145: `github.com/prometheus/prometheus/promql/parser.ParseMetricSelector` should be used

simonswine (commented):
Thanks for your work on this. I left a couple of comments in line.

In general I think this is a bit too hacky for my liking and I would prefer for use to defined types properly in protobuf and have typed API calls to return the metric rules.

The settings API currently has mostly settings about how the UI should be presenting things (collapsing, maxNodes,...), I think this is something more complex and I think this should be a separate `MetricsFromProfileRulesService` as part of the settingsv1 API group.

I had done very similar things for CollectorRules: https://github.com/grafana/pyroscope/pull/3865

A well speced out API will make the UI integration easier and will also allow it to implement proper RBAC, API permissions going forward.

pkg/experiment/metrics/rules.go
```diff
@@ -48,5 +87,101 @@ func recordingRulesFromTenant(tenant string) []*RecordingRule {
...
+	// TODO: refactor this. Very rudimentary, only intended for experimental
+	httpClient := util.InstrumentedDefaultHTTPClient()
+	opts := connectapi.DefaultClientOptions()
+	opts = append(opts, connect.WithInterceptors(tenant.NewAuthInterceptor(true)))
+	return settingsv1connect.NewSettingsServiceClient(
+		httpClient,
+		"http://fire-tenant-settings-headless.fire-dev-001.svc.cluster.local:4100", // TODO: get it from config (set it in deployment tools) use "http://localhost:4040" for local
+		opts...,
+	)
+}
```
kolesnikovae, line 127: Initialization should be done in `modules` and the `SettingsClient` component should to be injected as a dependency

alsoba13:
closed in favour of #3945

alsoba13:
closed in favour of https://github.com/grafana/pyroscope/pull/3945

PR #3875, opened 2025-01-31
chore(otel): support ingesting offcpu

korniltsev:
For off cpu profiles, the ebpf profiler sends timestamps and times in each sample. 
Each sample has different  `Value` file length. it seems impossible to represent this in pprof, so I'm aggregating them to a single value. 
Furthermore use a separate `off_cpu` instead of `process_cpu` profile name.


pkg/test/integration/testdata/otel-ebpf-profiler-offcpu.json
kolesnikovae, line 1: Should we "hide" to json extension to trick github? `.fixture`, `.bin`, `.golden`

kolesnikovae (approved):

pkg/ingester/otlp/convert.go
```diff
@@ -66,19 +85,35 @@ func newProfileBuilder(src *otelProfile.Profile) *profileBuilder {
...
+			if len(res.dst.SampleType) == 1 {
+				res.name = &typesv1.LabelPair{
+					Name:  pyromodel.LabelNameProfileName,
+					Value: "process_cpu",
+				}
+			}
+			res.sampleProcessingTypes[i] = sampleConversionTypeSamplesToNanos
+		}
+		if profileType == "events:nanoseconds::" && len(res.dst.SampleType) == 1 {
+			res.sampleProcessingTypes[i] = sampleConversionTypeSumEvents
```
marcsanmi, line 106:
```suggestion
                // Identify off cpu profiles
		if profileType == "events:nanoseconds::" && len(res.dst.SampleType) == 1 {
			res.sampleProcessingTypes[i] = sampleConversionTypeSumEvents
```

pkg/ingester/otlp/convert.go
```diff
@@ -198,16 +249,30 @@ func (p *profileBuilder) convertFunctionBack(of *otelProfile.Function) uint64 {
...
-	} else if len(gs.Value) == 1 && p.cpuConversion {
-		gs.Value[0] *= p.src.Period
+		return nil, fmt.Errorf("sample value is required")
 	}
+
+	for i, typ := range p.sampleProcessingTypes {
+		switch typ {
+		case sampleConversionTypeSamplesToNanos:
+			gs.Value[i] *= p.src.Period
+		case sampleConversionTypeSumEvents:
```
marcsanmi, line 264: Do you think adding a comment for the future us saying something like the following would help (feel free to omit it):
```
// For off-CPU profiles, aggregate all sample values into a single sum
// since pprof cannot represent variable-length sample values
```

marcsanmi (approved):
LGTM 👍

PR #3876, opened 2025-01-31
chore: migrate main builds from gcr to dockerhub

korniltsev:
- push to `grafana/` in dockerhub instead of `gcr`
- do not push `latest` tag


marcsanmi (approved):
LGTM 👍

Issue #3256, opened 2024-04-26
Service analytics


kolesnikovae:
Currently, we expect users to know what they are looking for and provide quite limited abilities for exploration. Pyroscope should provide users with insight into their data: it should be possible to identify "interesting" services or service instances and code hotspots at a glance, without querying anything specific.

## Global View

At the highest level, we should present an overview of the user environment (the piece we're aware of): global statistics on the services sending data to Pyroscope.

We should identify a subset of "interesting" (or important) services that the user may want to take a closer look into. It does not mean we should only collect statistics for those services exclusively, but we should inform the user about them first; like suggestions.

This ought to be user-specific, although we could probably address this by utilizing multi-tenancy; the target audience is software engineers. Some of the criteria (in no particular order):
 - Owned by the user, user favorites.
 - Data statistics (ingested/stored/queried), projections/predictions.
 - Pattern changes (a surge, or drop in traffic or samples).
 - New services (+version?).
 - Dynamics, like new code hotspots (see below), or dimension statistics changes (see below).
 - AI features.

## Service catalog (registry)

List of all the services, including details such as:
 - Number of active instances.
 - Profile types.
 - Language/runtime (+version).
 - SDK/agent (+version).
 - Data statistics (ingested/stored/queried).

A user should be able to get answers to questions like:
 - How much data is ingested, stored, and queried, both total and per dimension (service, language, profile type, SDK/agent).
 - How many instances are sending data, both total and per dimension.
 - Average rate per service instance, average profile size.

These are frequently asked questions, and we can greatly help users if we provide them with all the necessary information.

## Dimension statistics

We should provide a detailed breakdown of dimensions (labels) for each service. For each of the service labels, we identify top-K values and for each of the selected label values, we collect statistics such as: share of samples, ingestion traffic, and data on disk. Label query matches could be tracked as well.

Later on, this information could also help us to add new features like:
 - Query costs estimation.
 - Adaptive profiling (generate optimisation suggestions based on the access patterns)

## Code hot spots

We should provide a list of functions (or call sites) that the user might be interested in. In the simplest form, this is a list of top-K functions (both flat and cumulative). More sophisticated analysis is possible, because the statistics are supposed to be collected service-wide in the background.

## Recent activity / query history

This is somewhat unrelated, but we also may want to keep track of the user actions globally (kind of audit log):
 * It would be very useful to have access to the own recent queries and recently visited pages. I personally lack this very much.
[...]

PR #3885, opened 2025-02-03
feat(v2): tenant-wide datasets

kolesnikovae:
The PR adds _dataset indices_: a TSDB index structurally identical to the one we use for storing series indices, but with postings pointing to datasets within the block. This index helps serve queries that do not specify which dataset is being queried (we currently create datasets based on the `service_name` label).

The index is also used for TSDB queries that do not rely on the **SeriesIndex**: specifically, `LabelNames`, `LabelValue`, and `SeriesBy` APIs. This is particularly useful for our old UX, where we need to generate suggestions for query input _ for example, listing all values for the `pod` label.

The dataset index is written at the tenant level. To enable this, I introduce the concept of tenant-wide datasets, or _anonymous datasets_: data regions attributed to the tenant itself. This allows storing data that is intended to be accessed globally for the tenant (i.e., when no specific query expression is provided). I plan to leverage this approach to implement #3256.

The dataset index is only generated during compaction. While we could create it in the segment writer, I have decided to postpone this for now. If any issues arise, they should be easy to address – I've modified the way datasets are flushed into segments, so we now get a stream of "flushes" and can produce the index in the same way it is done in the compactors.

In the PR, I also optimized the way blocks are compacted, eliminating most of the staging files: as a result the overall job duration (and resource usage) has decreased:
image
The reason is quite obvious:
image

I experimented with direct writes to object storage, but the client we use does not handle them well. As a result, I had to revert the optimization – something we should explore further in the future.


pkg/experiment/query_backend/query.go
```diff
@@ -70,45 +75,143 @@ func registerQueryType(
...
-		agg: agg,
-		ds:  ds,
+func (b *blockContext) needsDatasetLookup() bool {
+	md := b.obj.Metadata()
+	if len(md.Datasets) != 1 {
+		// The blocks metadata explicitly lists datasets to be queried.
+		return false
+	}
+	ds := md.Datasets[0]
+	if ds.Format != 1 {
```
kolesnikovae, line 128: Use constant.

pkg/experiment/block/writer.go
```diff
@@ -1,107 +1,73 @@
...
-func (b *Writer) ReadFromFiles(files ...string) (toc []uint64, err error) {
-	toc = make([]uint64, len(files))
-	for i := range files {
-		toc[i] = b.off
-		if err = b.ReadFromFile(files[i]); err != nil {
-			break
-		}
-	}
-	return toc, err
+func (w *Writer) ReadFrom(r io.Reader) (n int64, err error) {
```
aleks-p, line 45: nit, is this technically `WriteFromReader` or similar? (I know it was already named this way, just noticing it now)

pkg/experiment/ingester/segment.go
```diff
@@ -209,15 +215,19 @@ func (s *segment) flush(ctx context.Context) (err error) {
 		close(s.doneChan)
 		s.sw.metrics.flushSegmentDuration.WithLabelValues(s.sshard).Observe(time.Since(t1).Seconds())
 	}()
-	pprof.Do(ctx, pprof.Labels("segment_op", "flush_heads"), func(ctx context.Context) {
-		heads = s.flushHeads(ctx)
-	})
-	s.debuginfo.movedHeads = len(heads)
-	if len(heads) == 0 {
+
```
aleks-p, line 218: we removed the pprof label wrapper here, was that intentional?

pkg/experiment/query_backend/query.go
```diff
@@ -70,45 +75,143 @@ func registerQueryType(
...
+		return false
 	}
+	b.idx = ds
```
aleks-p, line 131: nit, this feels like a side effect of a at-first-look read-only method and it might be easy to miss - can it be moved outside or rename the method?

pkg/experiment/query_backend/query.go
```diff
@@ -70,45 +75,143 @@ func registerQueryType(
...
-		req: req,
-		agg: agg,
-		ds:  ds,
+func (b *blockContext) needsDatasetLookup() bool {
+	md := b.obj.Metadata()
+	if len(md.Datasets) != 1 {
+		// The blocks metadata explicitly lists datasets to be queried.
+		return false
+	}
+	ds := md.Datasets[0]
```
aleks-p, line 127: Is it possible to reach this with no datasets?

aleks-p (approved):
LGTM! I left a few questions / minor remarks that are not blockers to move this forward.

kolesnikovae:
Thank you for the review Aleks!

pkg/experiment/block/writer.go
aleks-p: I am saying it is strange for a method that does writing to be called `ReadFrom`, although I might be misunderstanding what it does. If this is a Go convention, then it is fine - though I would still find it strange :)

pkg/experiment/block/compaction.go
```diff
@@ -179,26 +187,67 @@ func newBlockCompaction(
...
 	}
-	if err = w.Flush(ctx); err != nil {
-		return nil, fmt.Errorf("flushing block writer: %w", err)
+	if err = b.writeDatasetIndex(w); err != nil {
+		return nil, fmt.Errorf("writing tenant index: %w", err)
 	}
-	b.meta.Size = w.Offset()
 	b.meta.StringTable = b.strings.Strings
+	b.meta.MetadataOffset = w.Offset()
+	if err = metadata.Encode(w, b.meta); err != nil {
```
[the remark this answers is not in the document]
aleks-p: Makes sense, especially for point 2. 

Regarding point 3, I would expect that there are benefits to serving the metadata from the metastore, compared to always fetching it from the object storage, but lets leave that discussion for later.

Thank you for clarifying!

pkg/experiment/query_backend/query.go
aleks-p: misread the `if`, all good :)

pkg/experiment/block/writer.go
kolesnikovae: Completely unrelated, but since we've already started this conversation:

`ReaderFrom` and `WriterTo` are primarily meant for `io.Copy` and very useful: they allow to delegate copying to the implementation in: 
```go
io.Copy(dst io.Writer, src io.Reader)
```
https://cs.opensource.google/go/go/+/refs/tags/go1.24.0:src/io/io.go;l=407-416

This might be confusing at times.

For example, `bytes.Buffer.ReaderFrom` will actually grow your buffer by at least `bytes.MinRead` extra space, if you call `ReadFrom`, or copy to the buffer, and the source does not implement `WriterTo` (which would be called before `dst.ReadFrom` in `io.Copy`).

```go
// MinRead is the minimum slice size passed to a [Buffer.Read] call by
// [Buffer.ReadFrom]. As long as the [Buffer] has at least MinRead bytes beyond
// what is required to hold the contents of r, [Buffer.ReadFrom] will not grow the
// underlying buffer.
const MinRead = 512
```
https://cs.opensource.google/go/go/+/refs/tags/go1.24.0:src/bytes/buffer.go;l=196-206

This is quite a pain – when you allocate a buffer that you know has sufficient capacity (e.g., 8MB), but it gets reallocated to 8MB + 512 bytes. This is especially frustrating when you're optimizing for performance and using a buffer pool (since most of them rely on size classes).

This is one of the reasons why we have our own "yet-another" buffer pool:
https://github.com/grafana/pyroscope/blob/main/pkg/util/bufferpool/pool.go#L34

This trick helped to achieve quite good inuse/alloc ratio:

Alloc space:
Screenshot 2025-02-14 at 13 11 05

Inuse space:
image

aleks-p: Nice, thanks for explaining! 

> This trick helped to achieve quite good inuse/alloc ratio:

This seems like a useful "metric" that could be extracted from profiling data, e.g., for automated analysis / recommendations.

PR #3886, opened 2025-02-03
docs: clarify OpenTelemetry profiling Cloud support

marcsanmi:
Updates the documentation to clarify that OpenTelemetry profiling is not currently supported in Grafana Cloud.


marcsanmi:
No need to add it since it'll be available in the next release on the Cloud during this week.

PR #3889, opened 2025-02-05
chore(ci): run tests on both amd64 and arm64

### LINKED ISSUE #3880: Multi-arch unit and integration tests
While we're releasing Pyroscope for several platforms (`linux/amd64`, `linux/arm64`, `linux/arm7`), we only run tests for `linux/amd64`.

kolesnikovae:
* Enables Go unit tests on arm64.
* arm64 e2e: enabled for the `regular path` only. The canvas library cypress depends on is not released for `linux/arm64`, so I had to add some dependencies.
  * The test will take longer and will likely determine the overall CI workflow duration. Given that the test is not very extensive, we may consider disabling it. However, even a smoke test is better than nothing – this is why I kept it.
* Linting is still only run for `linux/amd64`. As long as we don't have arch-dependent code, this is fine. Let me know if you want to enable linting for arm64 as well.

Resolves #3880

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -14,7 +14,14 @@ concurrency:
 
 jobs:
   regular-path:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - arch: amd64
+            runner: ubuntu-latest
+          - arch: arm64
+            runner: github-hosted-ubuntu-arm64
+    runs-on: ${{ matrix.runner }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
@@ -27,10 +34,11 @@ jobs:
         with:
           go-version: "1.22.11"
           cache: true
-      - run: yarn --frozen-lockfile
+      - name: Install node-canvas
+        run: sudo apt-get update && sudo apt-get install build-essential libcairo2-dev libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev
       - run: make build
       - name: Cypress run
-        uses: cypress-io/github-action@v5
+        uses: cypress-io/github-action@v6
         with:
           wait-on: http://localhost:4040/ready
           start: make run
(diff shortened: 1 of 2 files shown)


korniltsev:
I think we can remove `yarn --frozen-lockfile` and all the dependencies from workflows alltogether. The legacy frontend is build in docker

.github/workflows/e2e.yaml
```diff
@@ -27,10 +34,11 @@ jobs:
         with:
           go-version: "1.22.11"
           cache: true
-      - run: yarn --frozen-lockfile
+      - name: Install node-canvas
+        run: sudo apt-get update && sudo apt-get install build-essential libcairo2-dev libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev
```
kolesnikovae, line 38: This is for cypress, not our frontend. This is not required for amd64, but I added there as well just in case

korniltsev (approved):

.github/workflows/e2e.yaml
```diff
@@ -14,7 +14,14 @@ concurrency:
...
 jobs:
   regular-path:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - arch: amd64
+            runner: ubuntu-latest
+          - arch: arm64
+            runner: github-hosted-ubuntu-arm64
```
simonswine, line 23: According to this https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/ this image might be free/cheaper
```suggestion
          - arch: amd64
            runner: ubuntu-24.04
          - arch: arm64
            runner: ubuntu-24.04-arm
```

simonswine (approved):
Thank you for your work

.github/workflows/e2e.yaml
kolesnikovae: TIL Azure Cobalt 100

PR #3891, opened 2025-02-05
doc: dotnet tracing example

alsoba13:


examples/tracing/dotnet/musl.Dockerfile
```diff
@@ -0,0 +1,48 @@
+ARG SDK_VERSION=8.0
```
aleks-p, line 1: Is this Dockerfile used anywhere? If valid and working, we could point one of the services in the docker-compose to it, otherwise we can remove it.

alsoba13: Good catch. We could use it in our docker-compose, but there are already more than one example using it. I'll will remove it from this example

aleks-p (approved):
LGTM

PR #3892, opened 2025-02-05
feat(v2): evict failed jobs from compaction scheduler queue

kolesnikovae:
This change introduces a mechanism to evict faulty jobs from the compaction scheduler queue. While queue overflow is primarily a concern during development, it could still potentially occur in other scenarios.

```
To avoid infinite reassignment loops, the scheduler keeps track of reassignments (failures) for each job. If the number
of failures exceeds a set threshold, the job is not reassigned and remains at the bottom of the queue. Once the cause of
failure is resolved, the error limit can be temporarily increased to reprocess these jobs.

The scheduler queue has a size limit. Typically, the only scenario in which this limit is reached is when the compaction
process is not functioning correctly (e.g., due to a bug in the compaction procedure), preventing blocks from being
compacted and resulting in many jobs remaining in a failed state. Once the queue size limit is reached, failed jobs are
evicted, meaning the corresponding blocks will never be compacted. This may cause read amplification of the data queries
and bloat the metadata index. Therefore, the limit should be large enough. The recommended course of action is to roll
back or fix the bug and restart the compaction process, temporarily increasing the error limit if necessary.
```


pkg/experiment/metastore/compaction/scheduler/scheduler_test.go
```diff
@@ -37,6 +39,9 @@ func TestScheduler_UpdateSchedule(t *testing.T) {
 		CompletedJobs: []*raft_log.CompletedCompactionJob{{
 			State: &raft_log.CompactionJobState{Name: "3"},
 		}},
+		EvictedJobs: []*raft_log.EvictedCompactionJob{{
```
simonswine, line 42: You probably also want a `scheduler.queue.put` for this

pkg/experiment/metastore/compaction/scheduler/schedule.go
```diff
@@ -42,6 +42,7 @@ func (p *schedule) AssignJob() (*raft_log.AssignedCompactionJob, error) {
 		State: state,
 		Plan:  plan,
 	}
+	p.evicted++
```
simonswine, line 45: Not too sure if this should be better called assigned instead (I was a bit confused about it):

```suggestion
	p.assigned++
```

simonswine (approved):
Thanks for your work on this

LGTM

pkg/experiment/metastore/compaction/scheduler/schedule.go
kolesnikovae: Actually, I think this is a bug of sorts. For some reason, I counted assigned jobs toward job queue decrease (thus incrementing the evicted counter), which is not necessarily correct – the job might be abandoned and remain in the queue.

I made the limits more conservative and deleted the line.

pkg/experiment/metastore/compaction/scheduler/scheduler_test.go
kolesnikovae: Added some more assertions. This test is fairly formal, to be honest – I'd like to add an integration test when I get to it.

simonswine (approved):
Thanks for the fixes, glad I could be of help 🙂 

LGTM

PR #3894, opened 2025-02-05
examples: Add pyroscope relabel componet in rideshare-alloy example

marcsanmi:
This PR extends the rideshare-alloy example by adding label enrichment capabilities to demonstrate how to add meaningful metadata to profiles through the `pyroscope.relabel` component.

### DIFF
--- a/examples/language-sdk-instrumentation/golang-push/rideshare-alloy/README.md
+++ b/examples/language-sdk-instrumentation/golang-push/rideshare-alloy/README.md
@@ -45,7 +45,7 @@ pyroscope.write "backend" {
 # Pull latest images
 docker pull grafana/pyroscope:latest
 docker pull grafana/grafana:latest
-docker pull grafana/alloy:latest
+docker pull grafana/grafana/alloy-dev:latest
 
 # Run the example
 docker-compose up --build
--- a/examples/language-sdk-instrumentation/golang-push/rideshare-alloy/config.alloy
+++ b/examples/language-sdk-instrumentation/golang-push/rideshare-alloy/config.alloy
@@ -7,35 +7,41 @@ pyroscope.receive_http "default" {
 }
 
 pyroscope.relabel "filter_profiles" {
-    // Convert region to datacenter
+    // Group regions into geographical areas for better aggregation
     rule {
         action = "replace"
         source_labels = ["region"]
-        target_label = "datacenter"
+        target_label = "geo_area"
+        regex = "(us-.*)"
+        replacement = "americas"
+    }
+    rule {
+        action = "replace"
+        source_labels = ["region"]
+        target_label = "geo_area"
+        regex = "(eu-.*)"
+        replacement = "emea"
+    }
+    rule {
+        action = "replace"
+        source_labels = ["region"]
+        target_label = "geo_area"
+        regex = "(ap-.*)"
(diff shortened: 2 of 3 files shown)


examples/language-sdk-instrumentation/golang-push/rideshare-alloy/README.md
```diff
@@ -45,7 +45,7 @@ pyroscope.write "backend" {
 # Pull latest images
 docker pull grafana/pyroscope:latest
 docker pull grafana/grafana:latest
-docker pull grafana/alloy:latest
+docker pull grafana/grafana/alloy-dev:latest
```
kolesnikovae, line 48: We should be careful with use of dev images in examples, esp. `latest`. Users may not notice that and replicate it production.

```suggestion
docker pull grafana/alloy-dev:latest
```

kolesnikovae (approved):

examples/language-sdk-instrumentation/golang-push/rideshare-alloy/config.alloy
```diff
@@ -3,6 +3,40 @@ pyroscope.receive_http "default" {
...
+
+    // Add component label based on application name
+    rule {
+        action = "replace"
+        source_labels = ["__name__"]
+        target_label = "component"
+        regex = "ride-sharing-app.*"
+        replacement = "backend"
+    }
+    rule {
```
korniltsev, line 32: I think we should add example of how to drop targets, for example drop 10% of service-a and 20% of service-b

examples/language-sdk-instrumentation/golang-push/rideshare-alloy/config.alloy
```diff
@@ -3,6 +3,40 @@ pyroscope.receive_http "default" {
...
+        action = "replace"
+        target_label = "environment"
+        replacement = "demo"
+    }
+
+    // Add component label based on application name
+    rule {
+        action = "replace"
+        source_labels = ["__name__"]
```
korniltsev, line 27: `__name__` is a profile name like `process_cpu` or `memory` . I am not understanding why is it compared against `ride-sharing-app.*`

examples/language-sdk-instrumentation/golang-push/rideshare-alloy/README.md
marcsanmi: This was done as a temporary workaround since there's no other image available than the dev one, but I can pin it to a specific dev tag. 

The idea is to change it back to the previous state when the next release is available.

examples/language-sdk-instrumentation/golang-push/rideshare-alloy/config.alloy
korniltsev: for history: now I udnerstand why `__name__` is used. This is because `__name__` is the special lable in OG flameql.

korniltsev: I think we should somehow translate `__name__` to `service_name` transparentyl

marcsanmi: I'll change it to avoid confusion here 👍 On the other hand, I've just opened a slack thread to decided what's the best path to follow here.

PR #3901, opened 2025-02-11
feat: add call tree data structure

kolesnikovae:
This is a draft of the data structure that can help us handling call tree search and filtration efficiently. 

In the nutshell, it is very similar with the `stack trace` tree we use to store stack traces (and should be replaced, ideally), but it also provides some nice features like `transform`, `merge`, and `trim`.


kolesnikovae:
Closing it for now – I'm not going to work on this any soon. @alsoba13 you may find some of the ideas in the PR helpful for function filtering

Issue #3900, opened 2025-02-11
ring: Use net.JoinHostPort to support IPv6 addresses


Nihasa350:
#### Describe the bug
Pyroscope querier cannot construct an IPv6 address correctly and errors with "too many colons in address". Tracked down the issue here: https://github.com/grafana/pyroscope/blob/ae0067f11a7fee693a7b2e6620bc6e8ea34fa5be/pkg/scheduler/schedulerdiscovery/ring.go#L45

#### To Reproduce

1. Start Pyroscope in kubernetes with IPv6 (1.12.0)

#### Expected behavior

Querier should be able to process requests.

#### Environment

- Infrastructure: 
 Kubernetes - with IPv6
- Deployment tool:
 Helm

#### Additional Context

Here is PR from mimir that fixes the IPv6 address construction: https://github.com/grafana/mimir/commit/b3d4f613a25631762667a6c80e801e4dd0f61d0d

marcsanmi:
Hi @Nihasa350,

Thanks for reporting this, I'll take a look.

gjacquet:
I believe I was facing the same issue trying to run Pyroscope on IPv6 and I tried the fix using docker image `grafana/pyroscope:main-0d7771a` and I am still seeing a similar error:

```
ts=2025-02-13T19:55:35.294346853Z caller=pool.go:250 level=warn component=querier-worker msg="removing frontend failing healthcheck" addr=xx:xx:xx:xx:xx::xx:4040 reason="rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp: address xx:xx:xx:xx:xx::xx:4040: too many colons in address\""
```

I believe there could be a few more places where the endpoint address is not being built properly but I could not really figure out where yet.

gjacquet:
Actually, I think I found it: https://github.com/grafana/pyroscope/blob/main/pkg/frontend/frontend.go#L148

gjacquet:
There could be a few other spots that could have the issue based on this search: https://github.com/search?q=repo%3Agrafana%2Fpyroscope%20%22%25s%3A%25d%22&type=code
* https://github.com/grafana/pyroscope/blob/a4297db31e8ba38c9f74663af5306c5e542e093d/pkg/api/version/version.go#L190
* https://github.com/grafana/pyroscope/blob/a4297db31e8ba38c9f74663af5306c5e542e093d/pkg/frontend/frontend.go#L148

[...]

PR #3903, opened 2025-02-12
fix(ring): use net.JoinHostPort to support IPv6 addresses

marcsanmi:
Use Go's `net.JoinHostPort` instead of string formatting in ring lifecycler configuration to properly handle IPv6 addresses.

Fixes https://github.com/grafana/pyroscope/issues/3900

### DIFF
--- a/pkg/compactor/compactor_ring.go
+++ b/pkg/compactor/compactor_ring.go
@@ -7,7 +7,8 @@ package compactor
 
 import (
 	"flag"
-	"fmt"
+	"net"
+	"strconv"
 	"time"
 
 	"github.com/go-kit/log"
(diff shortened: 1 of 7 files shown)


kolesnikovae (approved):
Thanks for fixing this!

May I ask you to fix one more occurrence in: https://github.com/grafana/pyroscope/blob/main/pkg/experiment/metastore/discovery/kuberesolver.go#L106

It's fine if not (as this is `fix(ring)`), we can do this later

marcsanmi:
In fact, I missed that. Just updated, thanks!

simonswine (approved):
LGTM

Issue #3907, opened 2025-02-12
ClusterRole in helm chart has a namespace


charlesverdad:
#### Describe the bug

The latest helm chart version 1.12.0 fails our tests because the `ClusterRole` in the chart has a `namespace` parameter

```yaml
---
# Source: pyroscope/templates/rbac.yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: default-pyroscope-dev
  namespace: default 
  labels:
    helm.sh/chart: pyroscope-1.12.0
    app.kubernetes.io/name: pyroscope
    app.kubernetes.io/instance: pyroscope-dev
    app.kubernetes.io/version: "1.12.0"
    app.kubernetes.io/managed-by: Helm
```

#### Expected behavior

ClusterRole [should not have namespaces](https://stackoverflow.com/a/64872627), unless this is meant to be just a Role

>A Role always sets permissions within a particular [namespace](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces); when you create a Role, you have to specify the namespace it belongs in.

>ClusterRole, by contrast, is a non-namespaced resource. The resources have different names (Role and ClusterRole) because a Kubernetes object always has to be either namespaced or not namespaced; it can't be both.

>If you want to define a role within a namespace, use a Role; if you want to define a role cluster-wide, use a ClusterRole.

https://kubernetes.io/docs/reference/access-authn-authz/rbac/

simonswine:
Thanks for reporting fixed in #3925

PR #3908, opened 2025-02-12
Remove namespace from ClusterRole

charlesverdad:
## Context

Fixes the issue in https://github.com/grafana/pyroscope/issues/3907
Linters/tests catch this redundant namespace parameter on ClusterRole

## Intent

Remove the namespace from ClusterRole and ClusterRoleBinding because they are non-namespaced objects

> If you want to define a role within a namespace, use a Role; if you want to define a role cluster-wide, use a ClusterRole.

https://kubernetes.io/docs/reference/access-authn-authz/rbac/

## Note

Alternatively we could use Role instead of ClusterRole, but I'm not very familiar with why pyroscope needs this role. (It seems to allow pod discovery but I thought alloy should do that)

### DIFF
--- a/operations/pyroscope/helm/pyroscope/templates/rbac.yaml
+++ b/operations/pyroscope/helm/pyroscope/templates/rbac.yaml
@@ -4,7 +4,6 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: {{ .Release.Namespace }}-{{ include "pyroscope.fullname" . }}
-  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "pyroscope.labels" . | nindent 4 }}
 rules:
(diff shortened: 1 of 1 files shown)


kolesnikovae:
Thank you for your contribution, Charles!

May I ask you to run `make helm/check` to update the rendered version of manifests?

Your assertions are well-founded: not only the namespace issue but also the very purpose of the role is questionable at best. It used to be the case in early versions when the server was responsible for collecting profiles. Now that this is the responsibility of the Alloy deployment, we should probably consider removing the role and binding altogether. I think we can make this change safely, as the namespace is effectively ignored, making it a no-op from a functional perspective.

I, however, cannot find any reference for a strict requirement of prohibiting the specification of a namespace for cluster resources (though this may make sense). Could you clarify which linter you are using? `kube-linter` and `helm lint` do not complain about this issue.

simonswine (approved):
LGTM

Thanks for the fix

simonswine:
Just read your reply @kolesnikovae. Yes you are totally correct this is now provided by Alloy. When the alloy chart is enabled. So ideally we should remove it all together.

simonswine:
Replaced by #3925 thanks for reporting this @charlesverdad

PR #3919, opened 2025-02-14
fix: add IPv6 support in missing places

marcsanmi:
Add IPv6 support in 2 missing places.

Follow-up of https://github.com/grafana/pyroscope/pull/3903.


korniltsev (approved):
lgtm

we should add integration tests running microservices ipv6 at some point in the future

PR #3920, opened 2025-02-14
protobuf of recording rules

alsoba13:
From the prototype:
* profile type will be set in `matchers`
* service name will be set in `matchers`
* metric name can be set in `external_labels`
* keep_labels / labels_to_export is renamed to `group_by`

We will extend it in the future to give support to datasources

Edit: I finally added `metric_name`. We'll do frontend/backend representation splits in a later PR.


api/settings/v1/metrics.proto
```diff
@@ -0,0 +1,17 @@
...
+message RecordingRuleSetting {
+  repeated RecordingRule rules = 1;
+}
+
+message RecordingRule {
+  // Unique id of the recording rule within the tenant.
+  string id = 1;
+  repeated string matchers = 2;
+  repeated string group_by = 3;
+  repeated types.v1.LabelPair external_labels = 4;
```
simonswine, line 16: > * profile type will be set in `matchers`
> 
> * service name will be set in `matchers`
> 
> * metric name can be set in `external_labels`
> 
> * keep_labels / labels_to_export is renamed to `group_by`

It makes sense to put them as comments into the `proto` file, maybe with an example value each.

simonswine (approved):
Seems to fit in with the other calls

LGTM

aleks-p:
> metric name can be set in external_labels

I would imagine some things would be easier and cleaner if this was an explicit field (metric, metric_name, record, etc.), both for rule management and evaluation. Have we considered that already? Apologies if I've missed a discussion on this elsewhere.

simonswine:
> I would imagine some things would be easier and cleaner if this was an explicit field (metric, metric_name, record, etc.), both for rule management and evaluation. Have we considered that already? Apologies if I've missed a discussion on this elsewhere.

We found that unclean in the call (maybe just me), esp. as it should be allowed to reuse the same metric name multiple time. (so is not a ID/unique field).

See those two rules:

```
# rule1
name: `process_cpu_profiled`
matchers: [
"{environment=~"dev|staging:"}"
]
group_by: ["environment", "pod"]
external_labels: { cost_center: r-and-d}

# rule2
name: `process_cpu_profiled`
matchers: [
"{environment="prod"}"
]
group_by: ["environment", "pod"]
external_labels: { cost_center: r-and-d}
```

Also an external label of `__name__: my_overwritten_metric_name` will always overwrite it later.

api/settings/v1/metrics.proto
```diff
@@ -0,0 +1,17 @@
...
+}
```
bryanhuhta, line 17: In general I think this structure is good enough. I think this layout is designed with the backend in mind, which means to render even basic information of a `RecordingRule` in the UI, we need to do a fair amount of data massaging.

I quite like what @simonswine did to get around this with `CollectionRule`s. He specified an API model which is much more UI friendly

https://github.com/grafana/pyroscope/blob/80bab17eb5c63c21cfc77f1b3db62fd58f0c7994/api/settings/v1/setting.proto#L63-L71

and separately defined a storage model which is more convenient for the backend

https://github.com/grafana/pyroscope/blob/80bab17eb5c63c21cfc77f1b3db62fd58f0c7994/api/settings/v1/setting.proto#L87-L93

Lastly, the server will map between the two models

https://github.com/grafana/pyroscope/blob/80bab17eb5c63c21cfc77f1b3db62fd58f0c7994/pkg/settings/collection/store.go#L164-L199

I would advocate for this approach to be taken here. I generally would prefer these conversion/data massaging tasks to take place on the server side where we operate with

- Type language guarantees
- Much more compute power

This means we can get away with shipping less JavaScript to our clients and we eliminate an entire class of possible frontend bugs since we aren't massaging data for every single `RecordingRule`.

To further illustrate this point, if we use the backend data structure as the API model for the frontend, we need to do something like this:

```js
function serviceName(rule) {
  const maybeName = rule.matchers.split(',')
    .map((m) => m.split(/(?:!=|=~|!~|=)/))
    .filter((fields) => fields[0] == 'service_name');

  if (!maybeName || maybeName.length !== 2) {
    // Somehow handle the fact we couldn't parse out service name field.
  }

  const [, serviceName] = maybeName;
  return serviceName;
}
```

Contrast this with a well-defined protobuf structure which has `service_name` set as an explicit field:

```js
rule.serviceName
```

We can avoid a bunch of possible headache this way (especially once we migrate the UI to using generate protobuf clients) and our users get less JavaScript sent to their browser, which is always a win. This is not a hill I will die on. If we want to accept the drawbacks I outlined here, I'm not going block this PR, but I would encourage considering splitting the data model into backend and frontend representations.

bryanhuhta (commented):
I agree with @aleks-p that it would be nice to have the metric name set explicitly (even if it was set in both `external_labels` and `metric_name`). The name of the metric is important enough where I think it would be annoying/cumbersome to have to parse it out of external labels every time it should be used.

api/settings/v1/metrics.proto
simonswine: Yes definitely feel free to add the required fields/services/class to this to aid frontend editing. (Probably best in a subsequent PR)

bryanhuhta (approved):
I agree with [Christian](https://github.com/grafana/pyroscope/pull/3920#discussion_r1956498755) that my prior review can be addressed as a follow up PR change.

LGTM!

simonswine:
> I agree with @aleks-p that it would be nice to have the metric name set explicitly (even if it was set in both `external_labels` and `metric_name`).

Yes I am not blocking if we go for that. I leave that down to @alsoba13 to change and then merge.

api/settings/v1/metrics.proto
kolesnikovae: Please keep in mind that recording rules are not service specific. A good chunk of rules will span multiple/all services.

kolesnikovae:
**TL;DR**: Let's add a field to specify the metric name (`metric_name`). Can be this PR, can be a separate one.

1. We haven't considered further development. When we add filtering by functions (or stacks), we will likely want to export multiple metrics (`function_self`, `function_total`, `profile_total`, etc.) for a single rule:
   * This will make use of the explicit metric name somewhat confusing regardless of how we pass it, but I think that use of labels for that is not the best idea.
   * We could also consider naming the field `metric_name_base`, but I'm not sure if makes anything any better.

2. We should not allow metric name overrides via `exported_labels`:
  * _all_ users (frontend, backend, CLI, terraform provider – you never know) will have to validate/sanitize this.
  *  we need to handle this special case in the UI: we probably want to let users know that the label is invalid in the context, if they are trying to add it as an external label.

3. `rule.name`, however, should be also added and used for displaying purposes.

kolesnikovae (approved):

PR #3924, opened 2025-02-18
fix(v2): use better seed for shard placement

kolesnikovae:
The PR fixed a bad seed chosen for Fisher-Yates permutations forming the shard mapping in the new distribution algorithm.

I've noticed that in case of unavailability, the data distribution has pronounces skews and decided to test this case. I discovered that with low number of nodes and shards, the mapping we generate is very poor – the random choice is extremely unfortunate: 

```
instance-0 [  35%][   6%] ###################################
instance-1 [  32%][  -3%] ################################
instance-2 [  32%][  -3%] ################################
-
instance-0 [   0%][ -99%]
instance-1 [  68%][ 105%] ####################################################################
instance-2 [  32%][  -3%] ################################
-
instance-0 [  35%][   6%] ###################################
instance-1 [   0%][ -99%]
instance-2 [  64%][  93%] ################################################################
-
instance-0 [  67%][ 102%] ###################################################################
instance-1 [  32%][  -3%] ################################
instance-2 [   0%][ -99%]
```

After some brute force (for 3-10 instances, where distribution skews are especially impactful) I found a better seed that ensures significantly better behaviour in case of node failures/unavailability: 

```
instance-0 [  33%][   0%] #################################
instance-1 [  33%][   0%] #################################
instance-2 [  33%][   0%] #################################
-
instance-0 [   0%][ -99%]
instance-1 [  52%][  57%] ####################################################
instance-2 [  48%][  45%] ################################################
-
instance-0 [  50%][  51%] ##################################################
instance-1 [   0%][ -99%]
instance-2 [  49%][  48%] #################################################
-
instance-0 [  48%][  45%] ################################################
instance-1 [  51%][  54%] ###################################################
instance-2 [   0%][ -99%]
```

I'm not adding the scripts I used to find the seed, because I do not plan to maintain it: it's better to replace the randomized permutations with a knapsack solver or something similar that gives deterministic optimal results.

### DIFF
--- a/pkg/experiment/distributor/distributor.go
+++ b/pkg/experiment/distributor/distributor.go
@@ -326,14 +326,20 @@ func (p *perm) resize(n int) {
 	}
 }
 
-// The value is a random generated with a crypto/rand.Read,
-// and decoded as a little-endian uint64. No fancy math here.
-const randSeed = 4349576827832984783
-
 var steps [4 << 10]uint32
 
 func init() {
-	r := rand.New(rand.NewSource(randSeed))
+	// The seed impacts mapping of shards to nodes.
+	// TODO(kolesnikovae):
+	//  Stochastic approach does not work well
+	//  in all the cases; it should be replaced
+	//  with a deterministic one.
+	const randSeed = -3035313949336265834
+	setSeed(randSeed)
+}
+
+func setSeed(n int64) {
+	r := rand.New(rand.NewSource(n))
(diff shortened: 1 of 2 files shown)


simonswine (approved):
Nice work! LGTM

kolesnikovae:
Thank you Christian! This is a temporary workaround – I'm planing to revisit the approach we use to map shards to nodes

PR #3925, opened 2025-02-18
chore(helm): Remove remnants of built-in Pyroscope agent

simonswine:
Previously Pyroscope included an agent, before there was support in Grafana Alloy/Agent for profiling. We forgot to remove RBAC objects in the helm chart. This cleans this up.


simonswine:
This is on the back of #3908, thanks @charlesverdad for reporting.

PR #3930, opened 2025-02-19
fix: add service_name label to the canary exporter

kolesnikovae:
The change adds `service_name` label to the profiles ingested by canary exporter, and the label selector used to query it back. Without the `service_name` label selector, in the newest versions, the query hits tenant-wide dataset index, which is only created after L0 compaction and is usually delayed by ~15-30s. Therefore, canary-exporter queries return no data until then, and the probe fails due to empty response.


simonswine (approved):
LGTM

Currently canary-exporters are deployed manually, so you need to update a reference to the version

PR #3933, opened 2025-02-19
test: Rerun failed go tests

simonswine:
I am a bit conflicted, as we are seeing a fair bit of test failures.
This change will obvisouly help getting PRs merged quicker, but it might
also lead to a downward slope in terms of test quality.


kolesnikovae (approved):
LGTM 

Thanks for adding this! I agree with your reasoning

PR #3934, opened 2025-02-20
metrics for exporter client

alsoba13:
note base branch is #3861

### DIFF
--- a/pkg/experiment/metrics/exporter.go
+++ b/pkg/experiment/metrics/exporter.go
@@ -3,15 +3,19 @@ package metrics
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"net/url"
 	"os"
+	"strconv"
 	"sync"
 	"time"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/gogo/protobuf/proto"
+	"github.com/grafana/dskit/instrument"
 	"github.com/klauspost/compress/snappy"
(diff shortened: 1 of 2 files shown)


pkg/experiment/metrics/exporter.go
```diff
@@ -103,4 +110,51 @@ func newClient(remoteUrl, user, password string) (remote.WriteClient, error) {
 		},
 		RetryOnRateLimit: false,
 	})
+	if err != nil {
+		return nil, err
+	}
+	t := client.(*remote.Client).Client.Transport
+	client.(*remote.Client).Client.Transport = &RoundTripper{m, t}
```
alsoba13, line 117: Is this too hacky? remote.WriteClient serves us well at performing remote_write. It doesn't offer a lot of client configurations, but at least Client member is exported.

simonswine (approved):
Generally I would be okay with the apporach, just think we should add a `tenant` label.

alsoba13:
> Generally I would be okay with the apporach, just think we should add a `tenant` label.

I agree, but now we have a single exporter for every tenant. I propose to metric only remote url until clients are tenant/datasource-specific, as it'll be easier to implement

kolesnikovae:
> I agree, but now we have a single exporter for every tenant. I propose to metric only remote url until clients are tenant/datasource-specific, as it'll be easier to implement

We could probably inspect http headers (where the tenant ID should be present, I guess)

PR #3940, opened 2025-02-21
fix: remove ellipsis stub from Go functions

kolesnikovae:
See https://github.com/grafana/explore-profiles/issues/420 for the context


simonswine (approved):
I think we should do it!

LGTM

PR #3943, opened 2025-02-24
chore: add canary probe delay

kolesnikovae:
The PR adds a delay between canary probe ingestion and query requests. The delay is set to 2s by default. This delay might be needed if ingestion is asynchronous, as an ingest followed by an immediate query might not retrieve the data.

@simonswine I'm somewhat reluctant to set a default value here. On the other hand, I don't see much harm in doing so. What do you think?


simonswine (approved):
LGTM

I think this is a reasonable change. We should try to remember calling this change out in the release notes.

PR #3944, opened 2025-02-24
feat: tenant settings ruler

alsoba13:
Tenant settings ruler is a thread-safe ruler that obtains rules from tenant-settings service and caches the result for later calls.

It holds a cache **for each tenant**, that expires after the default time (60s).
Every first call to a tenant for `RecordingRules(tenant string)` will initialize the per-tenant cache. When calling again the method, the cached rules will be served unless they expired. In that case, cache will be refreshed.

I discarded doing generics here becase (1) I would like to validate the idea and (2) there's no need to generalize the cache.


pkg/experiment/metrics/ruler.go
```diff
@@ -51,3 +54,74 @@ func NewStaticRulerFromEnvVars(logger log.Logger) (Ruler, error) {
...
+	ttl      time.Time
+	initFunc func() []*model.RecordingRule
+	mu       sync.Mutex
+}
+
+func (c *tenantCache) get() []*model.RecordingRule {
+	if c.value != nil && time.Now().Before(c.ttl) {
+		// value exists and didn't expired
+		return c.value
+	}
```
simonswine, line 122: I think you want to hold the`RLock` for this

alsoba13: absolutely

PR #3945, opened 2025-02-24
feat: tenant settings ruler

alsoba13:
CachedRemoteRuler is a thread-safe ruler that obtains rules from tenant-settings service and caches the result for later calls.

It holds a cache for each tenant, that expires after the default time (60s).
Every first call to a tenant for `RecordingRules(tenant string)` will initialize the per-tenant cache. When calling again the method, the cached rules will be served unless they expired. In that case, cache will be refreshed.

I discarded doing generics here becase (1) I would like to validate the idea first and (2) there's no need to generalize the cache.

Some detected todos:
* Expiry time by config (default 60s)


pkg/experiment/metrics/ruler.go
```diff
@@ -51,3 +54,83 @@ func NewStaticRulerFromEnvVars(logger log.Logger) (Ruler, error) {
...
+	mu       sync.RWMutex
+}
+
+// get returns the stored value if present and not expired.
+// Otherwise, a single call to initFunc will be performed to retrieve the value and hold it for future calls within
+// the ttl.
+func (c *tenantCache) get() []*model.RecordingRule {
+	c.mu.RLock()
+	if c.value != nil && time.Now().Before(c.ttl) {
+		defer c.mu.Unlock()
```
simonswine, line 121:
```suggestion
		defer c.mu.RUnlock()
```

pkg/experiment/metrics/ruler.go
```diff
@@ -51,3 +54,92 @@ func NewStaticRulerFromEnvVars(logger log.Logger) (Ruler, error) {
 func (r StaticRuler) RecordingRules(tenant string) []*model.RecordingRule {
 	return r.rules[tenant]
 }
+
+// CachedRemoteRuler is a thread-safe ruler that retrieves rules from an external service.
```
alsoba13, line 58: I opted to rename this to a more generic name. The origin of the rules depends on the client, not of this cache.

pkg/settings/recording/config.go
```diff
@@ -2,15 +2,18 @@ package recording
...
 type Config struct {
-	Enabled bool `yaml:"enabled" category:"experimental"`
+	Enabled bool   `yaml:"enabled" category:"experimental"`
+	Address string `yaml:"address" category:"experimental"`
 }
 
 const (
 	flagPrefix  = "tenant-settings.recording-rules."
 	flagEnabled = flagPrefix + "enabled"
+	flagAddress = flagPrefix + "client.address"
```
simonswine, line 16: I would put those client address flags onto where the client gets consumed: "compaction-worker". (I can't see anything else consuming a ExporterRulesClient.)

If someone enables `-compaction-worker.metrics-exporter.enabled` then and only then should be validate the address flag.

pkg/settings/recording/config.go
```diff
@@ -20,12 +23,20 @@ func (cfg *Config) RegisterFlags(fs *flag.FlagSet) {
...
+	)
 }
 
 func (cfg *Config) Validate() error {
 	if !cfg.Enabled {
 		return nil
 	}
-
+	if cfg.Address == "" {
+		return fmt.Errorf("%s is required", flagAddress)
```
simonswine, line 39: I don't think this flag should be required to be set when the tenant settings are running, but they maybe should be required when the compaction workers is enabled with metric exporter.

simonswine (commented):
The caching logic looks good to me. I just would like to change the structure/flags/validation to be more local towards the compaction worker.

simonswine (approved):
LGTM

Feel free to merge this once you did the suggested change

pkg/experiment/metrics/config.go
```diff
@@ -0,0 +1,21 @@
+package metrics
+
+import (
+	"flag"
+)
+
+type Config struct {
+	Enabled       bool   `yaml:"enabled"`
+	ClientAddress string `yaml:"client_address"`
```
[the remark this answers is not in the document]
alsoba13: Good catch, sorry about that

simonswine (approved):
LGTM

PR #3946, opened 2025-02-24
feat: Recording rules settings

bryanhuhta:
Adds a tenant settings submodule that can manage recording rules.

The current API allows for:

- Listing recording rules `/settings.v1.RecordingRulesService/ListRecordingRules`
- Creating a recording rule `/settings.v1.RecordingRulesService/InsertRecordingRule`
- Deleting a recording rule `/settings.v1.RecordingRulesService/DeleteRecordingRule`

This is the minimum API required to power the UI. Updates and reading single records are not required by the UI, though I am happy to add them in this PR if we want to attempt a future-proof of this API.

The general architecture of this submodule looks like this:

diagram


pkg/settings/recording/store.go
```diff
@@ -0,0 +1,106 @@
...
+func (b *bucketStore) Insert(ctx context.Context, newRule *settingsv1.RecordingRule_Store) (*settingsv1.RecordingRule_Store, error) {
+	err := b.store.Update(ctx, func(ctx context.Context, c *store.Collection[*settingsv1.RecordingRule_Store]) error {
+		for _, rule := range c.Elements {
+			if rule.Id == newRule.Id {
+				return fmt.Errorf("recording rule %s already exists", newRule.Id)
+			}
+
+			if rule.MetricName == newRule.MetricName {
+				return fmt.Errorf("recording rule with name %s already exists", newRule.MetricName)
+			}
```
alsoba13, line 57: We had some examples where creating rules with same name could be useful. For example, merging metrics from different sources with different group_by values (`env="dev"` and `env="prod"`) to a single metric `metric_name{env="X"}`.

I think this is fine to discuss. In some cases, uses will unintentionally misconfigure the same metric name twice, while they can still try to achieve metrics from different sources. I think we should support this and maybe improve the UI later to mark duplicates on the list.

**Risk:** There are situations where exporting with the same metric name twice can cause collisions/overrides in prometheus if rules are not correctly configured. For example, using group by labels with the same labelset space, i.e. failing to include the discriminant label (`env` in the example above), or using labels with empty values. Also, in that scenario, collision _may_ happen depending on pyroscope deployments (profiles processed by the same compaction_worker), which will cause a lot of confusion to the user.

pkg/settings/recording/recording.go
```diff
@@ -0,0 +1,232 @@
...
+	if ok {
+		return tenantStore, nil
+	}
+
+	tenantStore = newBucketStore(r.logger, r.bucket, key)
+	r.stores[key] = tenantStore
+	return tenantStore, nil
+}
+
+func validateInsert(req *settingsv1.InsertRecordingRuleRequest) error {
```
alsoba13, line 145: Shouldn't be matchers parsed for a validation?

pkg/settings/recording/store.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (_ *storeHelper) ToStore(rule *settingsv1.RecordingRule_Store) (json.RawMessage, error) {
+	return protojson.Marshal(rule)
+}
+
+func (_ *storeHelper) ID(rule *settingsv1.RecordingRule_Store) string {
+	return rule.Id
+}
+
+func (_ storeHelper) TypePath() string {
```
alsoba13, line 104: I guess this comes from copying  `collection/store.go` but my IDE usually mark those as smells:
```
Struct storeHelper has methods on both value and pointer receivers. Such usage is not recommended by the Go Documentation. 
```

pkg/settings/recording/recording.go
alsoba13: Hm.. I guess we could also check that GroupBy and ExternalLabels are good candidates to validate (I don't validate those either in compaction_worker, I guess I should)

api/settings/v1/recording_rules.proto
```diff
@@ -0,0 +1,70 @@
...
+message DeleteRecordingRuleResponse {}
+
+message RecordingRule_API {
+  string id = 1;
+  string metric_name = 2;
+  string profile_type = 3;
+  repeated string matchers = 4;
+  repeated string group_by = 5;
+  repeated types.v1.LabelPair external_labels = 6;
+  string prometheus_data_source = 7;
```
simonswine, line 44: Same here, I am unsure what data_source will be set to?

Is this a Grafana data source uid? How would we use it?

Grafana Datasources usually would only have read permissions, so we probably can't use this as "where should the generated metrics be ingested to"

api/settings/v1/recording_rules.proto
```diff
@@ -0,0 +1,70 @@
+syntax = "proto3";
+
+package settings.v1;
+
+import "types/v1/types.proto";
+
+service RecordingRulesService {
```
simonswine, line 7: In order to be consistent with CollectionRulesService, I would consider:

- Standardize either on `Upsert` or `Insert`. I assume updating of rules should be possilbe?
- Also provide a `Get`

alsoba13:
FYI, I merged main to this branch, so I can make #3945 branch from here and extend it to be consumed by compaction workers

pkg/settings/recording/store.go
bryanhuhta: Makes sense. I can remove this restriction 👍

bryanhuhta: addressed in https://github.com/grafana/pyroscope/pull/3946/commits/dfea5069fc7ce4eedd57f0d961aba238a02a9ff4

pkg/settings/recording/recording.go
bryanhuhta: Added more validation along with tests in https://github.com/grafana/pyroscope/pull/3946/commits/ad0a5848b3c30ea0691bc0d254f6fcedfdae9010

pkg/settings/recording/store.go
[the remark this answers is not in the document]
bryanhuhta: Fixed both occurrences in https://github.com/grafana/pyroscope/pull/3946/commits/8d0335d4b7f99457c30ac3f9e62bf6c41b96e7cd

pkg/api/api.go
```diff
@@ -209,6 +209,7 @@ func (a *API) RegisterRuntimeConfig(runtimeConfigHandler http.HandlerFunc, userL
 func (a *API) RegisterTenantSettings(ts *settings.TenantSettings) {
 	settingsv1connect.RegisterSettingsServiceHandler(a.server.HTTP, ts, a.connectOptionsAuthRecovery()...)
 	settingsv1connect.RegisterCollectionRulesServiceHandler(a.server.HTTP, ts, a.connectOptionsAuthRecovery()...)
+	settingsv1connect.RegisterRecordingRulesServiceHandler(a.server.HTTP, ts, a.connectOptionsAuthLogRecovery()...)
```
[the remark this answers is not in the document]
bryanhuhta: This is a typo. I probably incorrectly accepted an autocomplete option. I wanted to use `connectOptionsAuthRecovery` like the others.

Fixed in https://github.com/grafana/pyroscope/pull/3946/commits/bf83df61c6d032f10132daa650e89c9edc00e1ca

api/settings/v1/recording_rules.proto
```diff
@@ -0,0 +1,70 @@
...
+message DeleteRecordingRuleRequest {
+  string id = 1;
+}
+
+message DeleteRecordingRuleResponse {}
+
+message RecordingRule_API {
+  string id = 1;
+  string metric_name = 2;
+  string profile_type = 3;
```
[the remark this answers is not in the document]
bryanhuhta: In the UI, we want some way to signal what kind profile type the metric is recording. I have no objection to changing this from profile_type to something else, but it would be nice to show in the UI that the metric is tracking a cpu profile, memory profile, lock profile, etc.

Screenshot 2025-02-25 at 5 28 23 PM

api/settings/v1/recording_rules.proto
bryanhuhta: In UI we want to provide an option to the user what data source they want the recording rule to forward the metric to.

Here's the form in the UI:

Screenshot 2025-02-25 at 5 32 14 PM

Right now it doesn't allow users to choose the destination of their recording rule, but I would imagine this is something we want to support in the future. Or should I get rid of this field for now and we can add something in later when we expand support for this feature?

api/settings/v1/recording_rules.proto
bryanhuhta: I think Upsert is good. I will change to that.

I added a `Get` API in https://github.com/grafana/pyroscope/pull/3946/commits/951fd890cd8fcc8301a15adc6460bda5c029452c.

alsoba13 (approved):

api/settings/v1/recording_rules.proto
bryanhuhta: Removed in https://github.com/grafana/pyroscope/pull/3946/commits/96541dad320965a96e15785f0f1021a300c8e488.

api/settings/v1/recording_rules.proto
bryanhuhta: Switched to upsert in https://github.com/grafana/pyroscope/pull/3946/commits/dfbb10beb7c0fc54d1535b347ea6ee8e8c2e5591.

api/settings/v1/recording_rules.proto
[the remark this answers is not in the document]
bryanhuhta: I added clarifying comments in https://github.com/grafana/pyroscope/pull/3946/commits/80329dcb4544720a2e15887112cecb6674cbf354. LMK if you want me to add more detail.

simonswine (approved):
LGTM

PR #3947, opened 2025-02-25
perf: add log buffer

kolesnikovae:
Use of unbuffered synchronous logger affects ingestion latency. In the worst case, if the output is redirected to a file, and the disk system does not cope with the log rate, it may cause 10+ sec. p99 spikes. The problem here is that distributors are stateless but may suffer from side-effects caused by neighbours, which makes it hard to identify the culprit.

image

image

image

### DIFF
--- a/pkg/phlare/phlare.go
+++ b/pkg/phlare/phlare.go
@@ -14,6 +14,7 @@ import (
 	"slices"
 	"sort"
 	"strings"
+	"time"
 
 	"connectrpc.com/connect"
 	"github.com/go-kit/log"
@@ -326,8 +327,8 @@ func (c *Config) Clone() flagext.Registerer {
 
 type Phlare struct {
 	Cfg    Config
-	logger log.Logger
 	reg    prometheus.Registerer
+	logger *logger
 	tracer io.Closer
 
 	ModuleManager *modules.Manager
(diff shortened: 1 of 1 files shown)


pkg/phlare/phlare.go
```diff
@@ -726,19 +727,35 @@ func (f *Phlare) stopped() {
 			level.Error(f.logger).Log("msg", "error closing tracing", "err", err)
 		}
 	}
+	if err := f.logger.buf.Flush(); err != nil {
+		fmt.Println("error flushing logs:", err)
+	}
 }
 
-func initLogger(logFormat string, logLevel dslog.Level) log.Logger {
-	writer := log.NewSyncWriter(os.Stderr)
```
kolesnikovae, line 732: There's no particular reason to explicitly synchronise stdout/err writes: writes less than 4K (pipe buffer size) are already synchronous, and large outputs such as goroutine dump are not directed to the writer.

kolesnikovae:
I decided to use dskit `BufferedLogger` that does block writes while flushing the buffer. This might be causing issues; I'll revisit this if we encounter any issues

korniltsev (approved):

kolesnikovae:
Unfortunately, it did not really help:

PR #3950, opened 2025-02-25
chore(goreleaser): Fix error handling during release

simonswine:
This weeks release, did fail but the CI pipeline was successful.  This change "should" prevent this from happening.

I will reproduce it and doublecheck the CI build fails


simonswine:
It took a bit more and a few tries:

This is the error case: https://github.com/grafana/pyroscope/actions/runs/13527124425/job/37800413661

and this is when the build is fine: https://github.com/grafana/pyroscope/actions/runs/13527418259/job/37801366464

kolesnikovae (approved):

PR #3951, opened 2025-02-25
fix: prevent duplicate service_name and app_name labels during ingestion

marcsanmi:
This PR fixes two issues related to duplicate labels during profile ingestion:

- Prevents duplicate `service_name` labels in the `createLabels` function in the pprof converter
- Prevents duplicate `app_name` labels in the pyroscope ingester adapter

This is particularly important because in Alloy, we have [logic](https://github.com/grafana/alloy/blob/main/internal/component/pyroscope/receive_http/receive_http.go#L302) that ensures profiles always have a `service_name` label, and adds an `app_name` label when `service_name` already exists.

Our current implementation could lead to duplicate labels when this Alloy logic is applied.

Resolves https://github.com/grafana/pyroscope-squad/issues/357


pkg/ingester/pyroscope/ingest_adapter_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/grafana/pyroscope/pkg/og/storage"
+	"github.com/grafana/pyroscope/pkg/og/storage/tree"
+
+	"connectrpc.com/connect"
+	"github.com/go-kit/log"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type MockLabelCapturingPushService struct {
```
kolesnikovae, line 21: Just a note, not a suggestion: you may want to consider using mockery (see `.mockery.yaml`) to generate mocks and assertions in a more convenient way. (I agree that hand-written mocks can be more convenient in certain cases – it's totally up to you)

kolesnikovae (approved):
LGTM. Thanks for fixing this!

I'm not 100% sure I understand why we add the `app_name` label; I assume it's for backward compatibility, but I don't fully understand why we would create it if it's absent. This is clearly out of scope of the PR, but it would be nice if could clarify this moment

korniltsev:
I think I added an `app_name`  to not drop the app name from flameql if the `service_name` is also present.
Bsaically if we have flameql `appname{service_name=svc}`
Honestly I now regret doing this I think this is redundant and this logic can be dropped

korniltsev:
I think in may be useful to add an integration test ingesting `foo{service.name=bar}` . I have suspicuous it may also introduce duplicate labels after sanitization in distributor. Feel free to do this in a followup or just create an issue to do it later for someone else.

pkg/distributor/distributor.go
```diff
@@ -900,6 +897,64 @@ func extractSampleSeries(
...
+		} else if label.Name == "service.name" {
+			hasDotServiceName = true
+			serviceDotNameIndices = append(serviceDotNameIndices, j)
+			serviceDotNameValue = label.Value
+		}
+	}
+
+	if hasDotServiceName {
+		if hasServiceName && serviceDotNameValue != "" {
+			labels[serviceNameIndex].Value = serviceDotNameValue
```
marcsanmi, line 920: Do we want this? I've decided `service.name` value to replace `service_name` if both exists.

pkg/ingester/pyroscope/ingest_adapter_test.go
marcsanmi: Done in 6365cb043a942347c1ddf75570cbae500cd9edba.

simonswine (commented):
Thanks for looking into this. I have a few more questions before we merge this

pkg/distributor/distributor.go
```diff
@@ -900,6 +897,64 @@ func extractSampleSeries(
 	return result, bytesRelabelDropped, profilesRelabelDropped
 }
 
+func processServiceNameLabels(labels []*typesv1.LabelPair) []*typesv1.LabelPair {
```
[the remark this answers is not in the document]
marcsanmi: @simonswine quick update on this,

I updated it to support "one level" of sanitization-induced duplicates by adding a "_dup" suffix to the sanitized name. For example, if both "service.name" and "service_name" exist, "service.name" will be sanitized to "service_name_dup"; more than the same 2 duplicates exist because of sanitization, we still fail.

I've also added the support to add service_name when only service.name is present.

LMKWYT.

Alloy's related [PR](https://github.com/grafana/alloy/pull/2918)

pkg/validation/validate.go
```diff
@@ -122,8 +136,8 @@ func ValidateLabels(limits LabelValidationLimits, tenantID string, ls []*typesv1
 	if !model.IsValidMetricName(model.LabelValue(metricNameValue)) {
 		return NewErrorf(InvalidLabels, InvalidLabelsErrorMsg, phlaremodel.LabelPairsString(ls), "invalid metric name")
 	}
-	serviceNameValue := phlaremodel.Labels(ls).Get(phlaremodel.LabelNameServiceName)
 	if !isValidServiceName(serviceNameValue) {
+		fmt.Println(">> OUT 1")
```
simonswine, line 140: You have forgotten this :)

pkg/validation/validate.go
```diff
@@ -135,11 +149,23 @@ func ValidateLabels(limits LabelValidationLimits, tenantID string, ls []*typesv1
 		if len(l.Value) > limits.MaxLabelValueLength(tenantID) {
 			return NewErrorf(LabelValueTooLong, LabelValueTooLongErrorMsg, phlaremodel.LabelPairsString(ls), l.Value)
 		}
-		var origName string
+		var origName, sanitized string
 		var ok bool
-		if origName, l.Name, ok = SanitizeLabelName(l.Name); !ok {
+		if origName, sanitized, ok = SanitizeLabelName(l.Name); !ok {
```
simonswine, line 154: I find this a lot more readable
```suggestion
origName, sanitized, ok := SanitizeLabelName(l.Name)
if !ok {
```

pkg/validation/validate.go
```diff
@@ -135,11 +149,23 @@ func ValidateLabels(limits LabelValidationLimits, tenantID string, ls []*typesv1
...
+		// Check if a dup would be created because of sanitization
+		if origName != sanitized {
+			if sanitizedNames[sanitized] {
+				// Add "_dup" suffix to the conflicting sanitized label name
+				sanitized = sanitized + "_dup"
+			} else {
+				sanitizedNames[sanitized] = true
+			}
+		}
+		l.Name = sanitized
```
simonswine, line 167: Changing the name requires another `sort.Sort(phlaremodel.Labels(ls))` to ensure the order: 

Let's say I have I have label names with `service_name_aaaa` and `service_name_eee` , `service_name_dup` should be in-between them.

pkg/validation/validate.go
```diff
@@ -112,6 +112,20 @@ func ValidateLabels(limits LabelValidationLimits, tenantID string, ls []*typesv1
 	if len(ls) == 0 {
 		return NewErrorf(MissingLabels, MissingLabelsErrorMsg)
 	}
+
+	sanitizedNames := make(map[string]bool)
```
simonswine, line 116: This is an allocation in the hot path for any profile if needing sanitation or not. Also in the worst case you are adding at most two elements, if I read the code correctly.

pkg/validation/validate.go
```diff
@@ -135,11 +149,23 @@ func ValidateLabels(limits LabelValidationLimits, tenantID string, ls []*typesv1
...
-		if origName, l.Name, ok = SanitizeLabelName(l.Name); !ok {
+		if origName, sanitized, ok = SanitizeLabelName(l.Name); !ok {
 			return NewErrorf(InvalidLabels, InvalidLabelsErrorMsg, phlaremodel.LabelPairsString(ls), "invalid label name '"+origName+"'")
 		}
+
+		// Check if a dup would be created because of sanitization
+		if origName != sanitized {
+			if sanitizedNames[sanitized] {
+				// Add "_dup" suffix to the conflicting sanitized label name
+				sanitized = sanitized + "_dup"
```
simonswine, line 162: Nit: Adding `_dup` to the label name, might bring you over `MaxLabelNameLength`, not necessarily a problem.

pkg/validation/validate.go
```diff
@@ -112,6 +112,20 @@ func ValidateLabels(limits LabelValidationLimits, tenantID string, ls []*typesv1
 	if len(ls) == 0 {
 		return NewErrorf(MissingLabels, MissingLabelsErrorMsg)
 	}
+
+	sanitizedNames := make(map[string]bool)
+	//Check if we only have service.name label before sanitization
+	serviceNameValue := phlaremodel.Labels(ls).Get(phlaremodel.LabelNameServiceName)
```
simonswine, line 118: While looking for when the labels are sorted, I noticed that we already add unspecified as a value, so this if can never be triggered:

https://github.com/grafana/pyroscope/blob/efeabde8456ee1e1d68db0ceb026305f753866b7/pkg/distributor/distributor.go#L293-L295

pkg/validation/validate.go
marcsanmi: Yes, I somewhat miss to reorder the labels after adding the sanitised name, good catch.

pkg/validation/validate.go
marcsanmi: Not really. It's a safe check... in case anytime this is changed or we ingest labels throughout another path. However, this is likely not going to happen. I'll remove it.

pkg/validation/validate.go
marcsanmi: N/A anymore

pkg/validation/validate.go
marcsanmi: N/A anymore

pkg/validation/validate.go
simonswine: I don't think this has any benefit how it is implemented right now: The distributor will always go through. `PushParsed`(which is the only caller of `ValidateLabels`).

I think we need to test it form that level, it is not super straight forward, but that's how I tested your change:

https://github.com/grafana/pyroscope/commit/2535c8d1afc6e0d384be454fcc2a3651d07789aa

You will find that without the returning of the labels it will panic. This is because you modify the label slice without returning the new slice.

simonswine (approved):
Thanks for putting all the effort in, I do think we came to best variant in the end 🙂 

LGTM

PR #3954, opened 2025-02-26
chore: Update thanos-io/objstore

simonswine:
I would like to upgrade the otel libraries, in order to do that grpc will need to be updated, which in turn requires the thanos-io/objstore to be updated. 😹 

I tried to refactor a bit and replaced the custom mock for a bucket with the generated ones.


kolesnikovae (approved):
LGTM

simonswine:
Checked with @korniltsev, I am okay to go ahead and merge this

PR #3914, opened 2025-02-13
Usage group support for ingestion limits

aleks-p:
Continuation of #3879, adding support for reacting to externally configured ingestion limits for usage groups. The config model is extended from before:

```
ingestion_limit:
  period_type: hour
  period_limit_mb: 2048
  limit_reached: true
  next_limit_reset: 1738274400
  sampling:
    num_requests: 1
    period: 2m
  # new field
  usage_groups:
    group-1:
      limit_reached: true
      period_limit_mb: 1024
    group-2:
      limit_reached: false
      period_limit_mb: 128
```

When the limit is reached, we add the usage group to the error response:

```
resource_exhausted: limit of 1.0 GiB/hour reached for usage group group-1, next reset at 2025-02-13T19:00:00Z
```

Usage groups use the global (tenant-level) period type and sampling config for simplicity, this might change in the future if needed.

PR #3956, opened 2025-02-27
Annotate throttled profiles

aleks-p:
With #3879 and #3914, distributors reject profiles when an externally controlled ingest limit is reached. 

This change annotates the few profiles that are let through in this state, so that users of the profiling data can be made aware that data is being rejected. The information is stored in an extra column in the parquet table, and is applied to v1 and v2 read and write paths.

A few things are still not finished, namely:
- [x] deduplication in the read path
- [x] increase test coverage
- [x] solidify naming and API contract


pkg/phlaredb/schemas/v1/profiles.go
```diff
@@ -311,6 +325,8 @@ type InMemoryProfile struct {
 	DefaultSampleType int64
 
 	Samples Samples
+
+	Annotations []string
```
simonswine, line 329: In theory we could achieve the same thing with just using the Comments field above in L322.

I guess comment strings would be in the string table, unsure if that is a good or a bad thing.

I think if we would go for a separate type/field, I think we should make it Key,Value, because that feels more extensible to me than a `[]string`,

korniltsev: if we go with this, maybe at some point we can stop exploiting build Ids for github integration

aleks-p: Apologies for the delay, I had to switch focus.

> In theory we could achieve the same thing with just using the Comments field above in L322.
> I guess comment strings would be in the string table, unsure if that is a good or a bad thing.

I considered using `Comments` at first, but because the strings end up in the string table there is no efficient way to read them back.

> I think if we would go for a separate type/field, I think we should make it Key,Value, because that feels more extensible to me than a []string,

I agree. In parquet the data is stored in a more structured way:

```
phlareparquet.NewGroupField(AnnotationsColumnName, parquet.List(
			phlareparquet.Group{
				phlareparquet.NewGroupField("Body", parquet.Encoded(parquet.String(), &parquet.DeltaByteArray)),
			})),
```

`[]string` is only used in the in-memory representation currently, but a key/value will likely make merging data somewhat easier. I will give that a try.

aleks-p:
I plan to add more tests but the critical pieces should be in place now.

api/types/v1/types.proto
```diff
@@ -30,6 +30,12 @@ message Point {
   double value = 1;
   // Milliseconds unix timestamp
   int64 timestamp = 2;
+  repeated ProfileAnnotation annotations = 3;
+}
+
+message ProfileAnnotation {
```
simonswine, line 36: I think it is also worth documenting, how annotations are different from labels.

- They can't be used at query time to select on.
- There can be multiple annotations with the same key (with labels we don't allow duplicates).

simonswine (approved):
Nice one! I couldn't think of an areas that you forgot about. I also haven't tried this in action, but so far so good. A couple of thoughts I had inline.

api/push/v1/push.proto
```diff
@@ -23,6 +23,9 @@ message RawProfileSeries {
 
   // samples are the set of profile bytes
   repeated RawSample samples = 2;
+
+  // annotations provide generic metadata for profiles
+  repeated types.v1.ProfileAnnotation annotations = 3;
```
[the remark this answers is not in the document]
aleks-p: That is a great point.

I didn't intend for annotations to be exposed in the API but we have to do it this way because we reuse the proto messages between distributors and ingesters. In v2 

I like the idea with prefixing "trustworthy" annotations, as well as (at least temporarily) rejecting external annotations.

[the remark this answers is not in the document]
simonswine: Leave this up to you, both of them are better than not checking at all.

I guess we could always loosen the criteria.

PR #3957, opened 2025-02-27
Update gotestsum

simonswine:
Wonder if this fixes that tests are not rerun because of:

> rerun aborted because previous run had a suspected panic and some test may not have run


kolesnikovae (approved):
That's an interesting finding!

Might be relevant: I saw a race somewhere deep in the dskit memberlist during one of the integration tests, but I didn't investigate it

PR #3961, opened 2025-02-28
chore: Update otel to latest versions

simonswine:
This carries a lot of other upgrade to dependcies along with it.

It tried to sync all the code generation versions in Makefile with the updated go modules in go.mod.

I also had to switch the minimum version in go.mod from 1.22 to 1.22.0 (because that's what otel does).


pkg/ingester/otlp/convert.go
```diff
@@ -202,8 +203,11 @@ func (p *profileBuilder) convertLocationBack(ol *otelProfile.Location) uint64 {
 	if i, ok := p.locationMap[ol]; ok {
 		return i
 	}
-	lmi := ol.MappingIndex_.(*otelProfile.Location_MappingIndex)
-	om := p.src.MappingTable[lmi.MappingIndex]
+	var lmi int32 = 0
+	if ol.MappingIndex != nil {
```
korniltsev, line 207: What happens if it is nil? I am not sure there is a correct mapping placeholder at index 0. Maybe we should fail the conversion in this case?

ebpf/go.mod
```diff
@@ -1,9 +1,11 @@
 module github.com/grafana/pyroscope/ebpf
 
-go 1.22
+go 1.22.0
+
+toolchain go1.22.9
```
korniltsev, line 5: To be honest I don't really know what this change does and why is this needed. Do we need tolchain in main go.mod in the root of the repo?

korniltsev (approved):
LGTM

pkg/ingester/otlp/convert.go
simonswine: Panic, I guess before it would also have paniked. 

Handled the error now in dfe456645

ebpf/go.mod
simonswine: I don't understand it either, it seems to be coming in everytime I touch the `go get`.

```
git ls-files '**/go.mod' '**/go.work'  | xargs sed -i '/toolchain go/d'
```

PR #3969, opened 2025-03-04
Rename yaml representation of `metrics.Config`

alsoba13:
yaml and flag representations were not matching


simonswine (approved):
Thanks for fixing!

LGTM

PR #3973, opened 2025-03-05
chore: Recording rules overrides

alsoba13:
First approach to recording rules from overrides. This replaces recording rules from envs.

Missing some test fixes and I probably will need to add some more tests.

With this change, the config for recording rules would work like this:
```
compaction_worker:
  metrics_exporter:
    enabled: true
    rules_source:
      client_address: <address_to_tenant_settings>  # in the presence of this url, a RecordingRulesClient will be used to fetch rules
```

If no `client_address` was provided, then a static ruler will grab recording rules from _limits_. Static recording rules are defaulted to empty set. You can configure them globally:
```
limits:
  recording_rules:
  - metric_name: 'xxx'
    matchers: ['{__profile_type__="<profile_type>",env="prod"}']
    group_by: ['pod']
    external_labels:
    - name: 'foo'
      value: 'bar'
```

or you can define a tenant override:
```
overrides:
  "13":
    recording_rules:
    - metric_name: 'xxx'
      ...
```

You can configure them globally with limits flags as well: `-compaction-worker.metrics-exporter.rules-source.static=json rules`.

### DIFF
--- a/pkg/model/recording_rule.go
+++ b/pkg/model/recording_rule.go
@@ -6,8 +6,6 @@ import (
 	prometheusmodel "github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/promql/parser"
-
-	settingsv1 "github.com/grafana/pyroscope/api/gen/proto/go/settings/v1"
 )
 
 type RecordingRule struct {
@@ -16,19 +14,19 @@ type RecordingRule struct {
 	ExternalLabels labels.Labels
 }
 
-func NewRecordingRule(rule *settingsv1.RecordingRule) (*RecordingRule, error) {
+func NewRecordingRule(metricName string, matchers []string, groupBy []string, externalLabels labels.Labels) (*RecordingRule, error) {
 	// validate metric name
-	if !prometheusmodel.IsValidMetricName(prometheusmodel.LabelValue(rule.MetricName)) {
-		return nil, fmt.Errorf("invalid metric name: %s", rule.MetricName)
+	if !prometheusmodel.IsValidMetricName(prometheusmodel.LabelValue(metricName)) {
+		return nil, fmt.Errorf("invalid metric name: %s", metricName)
 	}
 
 	// ensure __profile_type__ matcher is present
-	matchers, err := parseMatchers(rule.Matchers)
+	ms, err := parseMatchers(matchers)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse matchers: %w", err)
 	}
 	var profileTypePresent bool
-	for _, matcher := range matchers {
+	for _, matcher := range ms {
 		if matcher.Name == LabelNameProfileType {
 			profileTypePresent = true
 			break
@@ -39,13 +37,13 @@ func NewRecordingRule(rule *settingsv1.RecordingRule) (*RecordingRule, error) {
 	}
 
 	r := &RecordingRule{
-		Matchers:       matchers,
-		GroupBy:        rule.GroupBy,
-		ExternalLabels: make(labels.Labels, 0, len(rule.ExternalLabels)+1),
+		Matchers:       ms,
+		GroupBy:        groupBy,
+		ExternalLabels: make(labels.Labels, 0, len(externalLabels)+1),
 	}
 
 	// ensure __name__ is unique
-	for _, lbl := range rule.ExternalLabels {
+	for _, lbl := range externalLabels {
 		if lbl.Name == prometheusmodel.MetricNameLabel {
 			// skip __name__
 			continue
@@ -58,7 +56,7 @@ func NewRecordingRule(rule *settingsv1.RecordingRule) (*RecordingRule, error) {
 	// trust rule.MetricName
 	r.ExternalLabels = append(r.ExternalLabels, labels.Label{
 		Name:  prometheusmodel.MetricNameLabel,
-		Value: rule.MetricName,
+		Value: metricName,
 	})
 	return r, nil
 }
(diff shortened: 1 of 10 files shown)


pkg/validation/limits.go
```diff
@@ -185,6 +191,9 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.Var(&l.IngestionRelabelingDefaultRulesPosition, "distributor.ingestion-relabeling-default-rules-position", "Position of the default ingestion relabeling rules in relation to relabel rules from overrides. Valid values are 'first', 'last' or 'disabled'.")
 	_ = l.IngestionRelabelingRules.Set("[]")
 	f.Var(&l.IngestionRelabelingRules, "distributor.ingestion-relabeling-rules", "List of ingestion relabel configurations. The relabeling rules work the same way, as those of [Prometheus](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config). All rules are applied in the order they are specified. Note: In most situations, it is more effective to use relabeling directly in Grafana Alloy.")
+
+	_ = l.RecordingRules.Set("[]")
+	f.Var(&l.RecordingRules, "compaction-worker.metrics-exporter.rules-source.static", "List of static recording rules of the type settingsv1.RecordingRule. Will only be use in the absence of a recording rules client.")
```
alsoba13, line 196: Flags for limits may be a bit awkward sometimes. While you can set a flag for `-compaction-worker.metrics-exporter.rules-source.static="<json rules>"`, you can't do it through regular config file with `compaction_worker.metrics_exporter.rules_source.static`. I opted to use a name similar to our config though, for consistency (although through override file it looks different -> `limits.recording_rules` or `overrides.tenant.recording_rules`).

I accept naming suggestions if this is missleading.

pkg/validation/recording_rules.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+type RecordingRules []*RulesConfig
+
+type RulesConfig struct {
+	MetricName     string              `yaml:"metric_name" json:"metric_name"`
+	Matchers       []string            `yaml:"matchers" json:"matchers"`
+	ExternalLabels []map[string]string `yaml:"external_labels" json:"external_labels"`
+	GroupBy        []string            `yaml:"group_by" json:"group_by"`
+}
```
alsoba13, line 17: I accept suggestions here as well. For parsing config, I opted to do the _limits_ style (dto with `yaml` and `json` tags). But, if we do this way we can't use model.RecordingRule (no other model structs in that packages use `yaml), nor settingsv1.RecordingRule (as they don't include the yaml tag).

So I created this yet another rules representation, which can be fairly easy outdated and unaligned with other models.

pkg/validation/recording_rules.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		panic(fmt.Errorf("error unmarshal yaml: %w", err))
+	}
+
+	jsonBytes, err := json.Marshal(temp)
+	if err != nil {
+		panic(fmt.Errorf("error marshal json: %w", err))
+	}
+	return string(jsonBytes)
+}
```
alsoba13, line 45: Copied from another Limits field

pkg/validation/limits.go
simonswine: Not too sure I understand, I suggest you also add a couple of tests for this, because that would make this a lot clearer.

pkg/validation/recording_rules.go
alsoba13: I went with the first option: https://github.com/grafana/pyroscope/pull/3973/commits/9e38ac9febb2ef6de2d6ebf16b38758188ca89c3

pkg/model/recording_rule.go
```diff
@@ -16,19 +14,19 @@ type RecordingRule struct {
 	ExternalLabels labels.Labels
 }
 
-func NewRecordingRule(rule *settingsv1.RecordingRule) (*RecordingRule, error) {
+func NewRecordingRule(metricName string, matchers []string, groupBy []string, externalLabels labels.Labels) (*RecordingRule, error) {
```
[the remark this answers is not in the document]
alsoba13: This was reverted after the change on https://github.com/grafana/pyroscope/pull/3973/commits/9e38ac9febb2ef6de2d6ebf16b38758188ca89c3

pkg/validation/limits.go
alsoba13: What I meant is that limits flags are just flags, not config opts.

So using `-compaction-worker.metrics-exporter.rules-source.static` will work. But using this `pyroscope.yml` won't work:
```
compaction_worker:
  metrics_exporter:
    rules_source:
      static: ...
```

But same happens with other limits flags defined here. Using `-distributor.ingestion-relabeling-rules` will work. But not this:
```
...
distributor:
  ingestion_relabeling_rules:
...
```

`line 6: field ingestion_relabeling_rules not found in type distributor.Config`


Said that, I'm just asking for naming suggestions. I chose this very specific path `compaction-worker.metrics-exporter.rules-source.static` to be consistent with recording rules client source, but given that it can't be combined, `compaction-worker.recording-rules` could be enough.


I agree this lacks some testing. Working on them 😃

alsoba13:
fyi, in last commit I moved all the flag logic under v2 umbrella, with the other v2 flags, so it's hidden in --help and so
https://github.com/grafana/pyroscope/pull/3973/commits/445739765e5ee7e18255a53df7f67f149cc741fc

simonswine (approved):
LGTM

Thanks for working on that. Looking good!

Hidden is fine, let's try to remember to undo this once v2 becomes available for OSS

PR #3980, opened 2025-03-07
Add missing benchmarks

inkel:
This PR adds two benchmarks:

* `*PartitionWriter.WriteProfileSymbols` benchmark, as it is one of the functions that allocates the most space and I wanted to drill down why;
* `*stacktraceTree.insert` benchmark refactor for comparing the performance when using different tree sizes.


inkel:
Opened it against the wrong repository.

PR #3984, opened 2025-03-10
pyroscope-ebpf: cilium bump to 0.17.3

korniltsev:


korniltsev:
CI was red before the change. I've tested `ebpf/cmd/playground` locally it does compile.
Merging as is without fixing CI, the whole `ebpf` package is to be removed soon.

PR #3988, opened 2025-03-11
chore(v2): add segment upload timeout

kolesnikovae:
This change adds an explicit, configurable static timeout for segment uploads.

Currently, the timeout is inherited from the distributor and is too high (5s by default). This makes retries inefficient if the bucket client happens to connect to a slow server or encounters other high-latency issues.

Additionally, I have disabled the FailFast option for hedged requests to prevent a situation where a failed hedged request (which is not retried) cancels the upload.


korniltsev (approved):

simonswine (approved):
LGTM

PR #3990, opened 2025-03-11
fix(v2): scheduler metrics collector panics

kolesnikovae:
I discovered a panic in the compaction scheduler stats collector. The issue is that the scheduler queue may have gaps in levels – this is a completely valid case that wasn’t taken into account:

```
github.com/grafana/pyroscope/pkg/experiment/metastore/compaction/scheduler.(*statsCollector).collectStats(0xc00090f740, 0xc0009c3f50)
/home/runner/work/pyroscope/pyroscope/pkg/experiment/metastore/compaction/scheduler/metrics.go:85 +0x185
/home/runner/work/pyroscope/pyroscope/pkg/experiment/metastore/compaction/scheduler/metrics.go:118 +0x7b
```


simonswine (approved):
LGTM

PR #4000, opened 2025-03-12
docs(v2): add readme

kolesnikovae:
The PR adds README file to the `pkg/experiment` package and also provides minimal instruction on how to run the new version. The change does not affect public dicumentation.


simonswine (approved):
Thanks for putting those together. The deployment does work, I guess we have quite a bit of overlap with `Makefile` targets, let's merge this and the we can work out how we align both.

LGTM

PR #4004, opened 2025-03-13
chore(v2): simplify metadata labels

kolesnikovae:
The change makes working with metadata labels easier.


marcsanmi (approved):
Thanks for making it easier to use :)

PR #4006, opened 2025-03-14
feat(v2): configurable write-path compression

kolesnikovae:
The PR disables pprof gzip compression in the distributor-to-segment-writer path by default. This should be handled at the transport level.

This allows to save roughly 5-10% of CPU time in both distributors and segment-writers. However, this is only beneficial for all-in-one and single-AZ deployments (in cross-AZ deployments, ingest traffic should not leave the home zone).


simonswine (approved):
LGTM

PR #4005, opened 2025-03-13
chore: add GCS HTTP config

kolesnikovae:
The change will allow tuning the HTTP conn pool: the default parameters may cause excessive TLS handshakes:

Defaults:
image

1K idle connections, 10m idle timeout:
image

N.B.: it's worth exploring the GCS gRPC interface as well.

PR #4007, opened 2025-03-14
chore: mark GCS HTTP options advanced

kolesnikovae:
Follow-up for #4005


simonswine (approved):
Thanks for fixing this

LGTM

PR #4012, opened 2025-03-15
fix(v2): tombstone queue iterator ignores oldest tombstone

kolesnikovae:
The PR fixes a bug that prevented the oldest batch of tombstones from being scheduled for removal, causing it to remain in the queue indefinitely.


simonswine (approved):
LGTM

PR #4038, opened 2025-03-21
feat: Align missing service_name behaviour with otel spec

simonswine:
When looking at otel-ebpf-profiler I noticed we ignored their specification:

https://github.com/open-telemetry/opentelemetry-go/blob/ecfb73581f1b05af85fc393c3ce996a90cf2a5e2/semconv/v1.30.0/attribute_group.go#L10011-L10025


marcsanmi (approved):
LGTM, thanks for checking this 👍

Issue #3895, opened 2025-02-06
Flaky aggregation test


kolesnikovae:
Periodically, the `Test_Aggregation_Error` test fails in CI:

```
=== Failed
=== FAIL: pkg/distributor/aggregator Test_Aggregation_Error/First_aggregation_failed (0.01s)
    aggregator_test.go:156: 
        	Error Trace:	/home/runner/work/pyroscope/pyroscope/pkg/distributor/aggregator/aggregator_test.go:156
        	Error:      	Target error should be in err chain:
        	            	expected: "context canceled"
        	            	in chain: 
        	Test:       	Test_Aggregation_Error/First_aggregation_failed
    aggregator_test.go:160: 
        	Error Trace:	/home/runner/work/pyroscope/pyroscope/pkg/distributor/aggregator/aggregator_test.go:160
        	Error:      	An error is expected but got nil.
        	Test:       	Test_Aggregation_Error/First_aggregation_failed
        	Messages:   	context canceled
    --- FAIL: Test_Aggregation_Error/First_aggregation_failed (0.01s)
```

Presumably, it should be caused by the configured 10ms period. One way to solve it is to run test multiple times, or increase the 10ms period (for that specific test)

PR #4045, opened 2025-03-24
fix(distributor): Test flake in the aggregator

simonswine:
@kolesnikovae I have seen quite a few flakes from this.

When the second aggregations happens (at ts=2), it sometimes might be fast enough to have the aggegrationKey deleated already.

I amended the test, that it accepts either outcome of the ts=2 following aggreagtions. But it also makes sure the aggregation eventually recovers.


kolesnikovae:
Thanks for fixing this @simonswine!

It addresses #3895

kolesnikovae (approved):
LGTM

PR #4046, opened 2025-03-25
feat(v2): X-Scope-OrgId in exporter headers

alsoba13:
In this PR:
* env vars associated to metrics datasource auth are removed.
* It is replaced by the addition of a X-Scope-OrgId (tenant). OSS user will require to create their own proxy in order to authenticate to remote metrics tenants.
* Exporter url is no longer injected by env var, but regular config is used.
* new round tripper metrics tenantId as well


pkg/experiment/metrics/exporter.go
```diff
@@ -124,7 +109,7 @@ func newMetrics(reg prometheus.Registerer, remoteUrl string) *clientMetrics {
 			Name:      "client_request_duration_seconds",
 			Help:      "Time (in seconds) spent on remote_write",
 			Buckets:   instrument.DefBuckets,
-		}, []string{}),
+		}, []string{"route", "status_code", "tenant"}),
 	}
 	if reg != nil {
 		// TODO(alsoba13): include tenant label once we have a client per tenant/datasource in place
```
simonswine, line 115: I think you can remove that

simonswine (approved):
LGTM

PR #4051, opened 2025-03-26
fix: drop malformed locations

kolesnikovae:
It's currently possible, that we ingest and store a profile with locations that refer to non-existent functions. This is causing panics at compaction, because we assume that such locations are sanitized at ingest time. If a profile does not have symbols, its locations should not include lines and functions.


kolesnikovae:
I'm going to figure out why we allow such locations to get to the storage before merging.

korniltsev:
I may have a repro for this issue (or some other issue)

1. start a fresh pyroscope server with no previous data
2. Ingest the profile

```bash
cat 547df064-46df-465a-9ebc-6e7506bc9111.profile | curl -X POST --data-binary @- http://localhost:4040/ingest\?\name\=amixir\&format\=pprof 
```
[547df064-46df-465a-9ebc-6e7506bc9111.profile.zip](https://github.com/user-attachments/files/19465384/547df064-46df-465a-9ebc-6e7506bc9111.profile.zip)

3. Qurey

image

4. Restart the pyroscope server
5. Query again

image

The results are different.


The pprof comes from https://github.com/grafana/alloy/pull/2920 
```
* c604b2feb 2025-03-25 | skip interpreter trampoline, add interpreter lines (HEAD -> korniltsev/otel-profiler, origin/korniltsev/otel-profiler) [Tolya Korniltsev]
```

I have not checked the pprof file itself, but it is likely mailformed or incorrect. I will take a look at the pprof sometime later next week.

Eventhough pprof is mailformed it should not cause different query results or compaction issues, we still need to tackle this in the database.

kolesnikovae:
Thanks for the sample! This is very weird, indeed – I'll take a careful look

korniltsev (approved):

PR #4052, opened 2025-03-26
feat(v2): metastore snapshot compression

kolesnikovae:
The PR enables `zstd` compression for FSM snapshots. It is not configurable, as there is no reason to write uncompressed snapshots. The change is fully backward compatible but not forward-compatible.

The compression ratio is generally very good, with the compressed snapshot being around 30–40% of the original size (given the default compression level). It is much better that `snappy`, and slightly better than `lzma`, while is comparable in terms of CPU time. `gzip` gives decent compression level as well, but utilizes significantly more CPU time.

### DIFF
--- a/pkg/experiment/metastore/fsm/boltdb.go
+++ b/pkg/experiment/metastore/fsm/boltdb.go
@@ -19,7 +19,7 @@ const (
 	boltDBCompactedName   = "metastore_compacted.boltdb"
 	boltDBInitialMmapSize = 1 << 30
 
-	boltDBCompactionMaxTxnSize = 64 << 10
+	boltDBCompactionMaxTxnSize = 1 << 20
 )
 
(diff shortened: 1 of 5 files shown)


kolesnikovae:
Yeah, I've been thinking about it. Ideally, we store snapshots on a separate dedicated disk, but if this option is not available, rate limiting will be helpful

kolesnikovae:
Well, compression itself is sufficient to slow down the process suffice to address the problem: in the chart we can see that it takes 5-7 seconds to write a 0.5GB snapshot:

image

_The snapshot size metric was broken – fixed in the latest commit_

p99 latency spikes are barely noticeable and should not be a big problem in 99.9% of cases. Further improvements are possible (e.g., tuning the compression level, concurrency, etc.), but I think we're good for now.

Issue #4050, opened 2025-03-25
Pyroscope 1.13.0 can no longer access S3 Bucket


sberz:
#### Describe the bug

Since the upgrade to Pyroscope 1.13, the access to S3 fails. This causes the storage-gateway to no longer start.
We run Pyroscope on AWS EKS and use [IAM roles for service accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) for authentication.

Likely most relevant log entries (full logs below):
```
ts=2025-03-25T16:16:29.803823714Z caller=bucket_stores.go:171 level=info msg="synchronizing Pyroscope blocks for all users"
ts=2025-03-25T16:16:29.84179215Z caller=reporter.go:201 level=debug msg="failed to read cluster seed file" err="Access Denied"
ts=2025-03-25T16:16:34.637233727Z caller=bucket_stores.go:176 level=warn msg="failed to synchronize Pyroscope blocks" err="Access Denied"
```

#### To Reproduce

1. Start Pyroscope 1.13.0 using S3 block storage 
2. Pyroscope fails to start with a generic `Access Denied` error

#### Expected behavior

Pyroscope should be able to authenticate using IRSA and access S3 block storage.

#### Environment

- Infrastructure: Kubernetes (AWS EKS)
- Deployment tool: helm 

#### Additional Context

I bisected the regression using the container images published for every commit to the main branch. 
The commit that introduced the regression is: 62c24abd9819c2473c26c918d87f0b57a7eb80ef

<details>
<summary>Storage Configuration</summary>

I use basically the same storage configuration for the latest Loki and Mimir releases where it also works as expected 

The helm release uses the following storage configuration, which did not change and worked with 1.12.0:
```yaml
        storage:
          backend: s3
          s3:
            bucket_name: <bucket-name>
            endpoint: s3.eu-central-1.amazonaws.com
```
The following variables environment variables are automatically set when using IRSA authentication:
```yaml
    - name: AWS_STS_REGIONAL_ENDPOINTS                                                                                                                                                       
      value: regional                                                                                                                                                                        
    - name: AWS_DEFAULT_REGION                                                                                                                                                               
      value: eu-central-1                                                                                                                                                                    
    - name: AWS_REGION                                                                                                                                                                       
      value: eu-central-1                                                                                                                                                                    
    - name: AWS_ROLE_ARN                                                                                                                                                                     
[...]

PR #4056, opened 2025-03-27
fix: Update minio-go to fix bug in AWS STS auth

simonswine:
With the upgrade of objstore libraries from #3954 we introduced a bug #4050.

This fixes the underlying problem and adds a tests to validate that form of auth.


kolesnikovae (approved):
LGTM

Thanks for solving this!

PR #4057, opened 2025-03-27
test: Skip otel ebpf test for now

simonswine:
This appears to be using quite a lot of disk

We also might want to investigate, how to clear unused images, between the test runs. 🙂


simonswine:
let's do it better #4061

PR #4061, opened 2025-03-27
test: Increase runner size and reduce parallelity

simonswine:
This will fix the cronjob which tests our SDK examples on a nightly basis.

- We were running out of disk: bigger runner
- The bigger runner wasn't able to run the test in time before they get shutdown: decrease parallelism


simonswine:
https://github.com/grafana/pyroscope/actions/runs/14107384623

korniltsev (approved):

PR #4067, opened 2025-03-27
fix(v2): make metastore index locks less coarse

kolesnikovae:
Relaxing locking in the metastore index. Currently, the metastore index relies on a single global mutex. While this works well in most cases, it becomes problematic when loading a large number of partitions from disk, potentially blocking writes for several seconds.

The PR also introduces a new test suite for the metadata index.

More fine-grained locking practically eliminates impact of the metadata queries to the raft command handling. In the charts you can see that even heavy metadata queries have no impact on the raft log processing:

image

The metadata index query performance will be improved in the future when we add an inverted index (likely a bitmap index), which will help avoid expensive scans.


pkg/experiment/metastore/index/index.go
```diff
@@ -18,17 +18,19 @@ import (
...
 	"github.com/grafana/pyroscope/pkg/experiment/block/metadata"
 	"github.com/grafana/pyroscope/pkg/experiment/metastore/index/store"
-	"github.com/grafana/pyroscope/pkg/iter"
 )
 
 const (
 	partitionDuration = 6 * time.Hour
 	// Indicates that partitions within this window are "protected" from being unloaded.
-	partitionProtectionWindow = 30 * time.Minute
+	partitionProtectionWindow = 24 * time.Hour
```
kolesnikovae, line 26: Before we separate the index into mutable and immutable parts, it's better to minimize partition evictions

pkg/experiment/metastore/index/index.go
```diff
@@ -18,17 +18,19 @@ import (
...
 	partitionTenantCacheSize  = 32
 )
 
-var ErrBlockExists = fmt.Errorf("block already exists")
+var (
+	ErrBlockExists = fmt.Errorf("block already exists")
+	ErrReadAborted = fmt.Errorf("read aborted")
+)
```
kolesnikovae, line 33: I've identified a flaw in the current implementation that can lead to an inconsistent memory state in the index. It's more of a theoretical issue that can be triggered by misconfiguration.

If the modified partition is unloaded within the same transaction – alongside a read transaction that started after the write transaction but was blocked by the global index lock – the read transaction may observe a stale database state (due to isolation) after acquiring the lock, causing stale data to be loaded into memory.

To eliminate this inconsistency, I've added a restriction: a partition can only be unloaded by the write thread and only in the absence of read transactions. A read transaction is aborted if it attempts to load a partition that has been modified after the transaction started. Even under artificially crafted contention, this is a relatively rare case (1 in 10,000 runs).

pkg/experiment/metastore/index/index.go
```diff
@@ -66,12 +68,28 @@ type Store interface {
...
 	store  Store
 
-	mu               sync.Mutex
-	loadedPartitions map[tenantPartitionKey]*indexPartition
+	// The global lock protects index partitions.
+	// Index partition shards have their own locks.
+	global sync.Mutex
+	// TODO: cache at the shard level.
+	tenantPartitions map[tenantPartitionKey]*indexPartition
 	partitions       []*store.Partition
```
kolesnikovae, line 79: I'm going to redesign the index soon, so this PR does not change its structure; the only goal is to avoid long locks. Ideally, we should separate the mutable write buffer – fully in-memory – from the read buffer – an immutable on-disk/in-store part that is loaded on demand. Queries should ideally be delegated to query-backend workers.

pkg/experiment/metastore/index/query_test.go
```diff
@@ -265,3 +271,348 @@ func TestIndex_Query(t *testing.T) {
 		query(t, tx, idx)
 	})
 }
+
+func Test_QueryConcurrency(t *testing.T) {
+	for i := 0; i < 10 && !t.Failed(); i++ {
+		q := new(queryTestSuite)
+		q.run(t)
+	}
+}
```
kolesnikovae, line 280: The test is quite intensive (tens of thousands of queries), but it should not take longer than a second on average. Running it locally for extended periods has not revealed any races or inconsistencies; if the test causes issues in CI, we should decrease the number of iterations.

Ideally, we should create a library of expensive tests to run in the background, with a shorter version – a smoke test – for the CI context.

pkg/experiment/metastore/index/query.go
```diff
@@ -94,48 +94,38 @@ func (q *metadataQuery) overlapsUnixMilli(start, end int64) bool {
...
-func (mi *blockMetadataIterator) At() *metastorev1.BlockMeta { return mi.metas[mi.cur] }
-
-func (mi *blockMetadataIterator) Next() bool {
-	if n := mi.cur + 1; n < len(mi.metas) {
-		mi.cur = n
-		return true
-	}
-	mi.cur = 0
-	mi.metas = mi.metas[:0]
+func (mi *blockMetadataQuerier) queryBlocks() ([]*metastorev1.BlockMeta, error) {
```
aleks-p, line 112: nit, the receiver (`mi`) should probably be named `mq` now

pkg/experiment/metastore/index/query.go
```diff
@@ -304,33 +307,52 @@ func (si *shardIterator) Next() bool {
...
+		if err != nil {
+			si.err = err
+			return
+		}
+		if shard != nil {
+			si.shards = append(si.shards, shard)
+		}
+		// If there are goroutines blocked on the global lock,
+		// we want to get them change to acquire it.
+		runtime.Gosched()
```
aleks-p, line 352: This is interesting, I wasn't familiar with this runtime function. Have you measured any benefits from using it (in benchmarks / real world tests)?

aleks-p (approved):
LGTM

pkg/experiment/metastore/index/query.go
kolesnikovae: It was a desperate attempt to improve fairness – just a leftover – it's not helpful in this situation, because the mutex will enter the starvation mode, and goroutines will be queued, so it's almost guaranteed that they will be executed before this goroutine will acquire the lock again. In the test, however, Gosched call is helpful – it drastically reduces the test duration.

kolesnikovae:
Thanks for the review Aleks!

I think I've found a better way to tackle the problem and will file a separate PR. This one is helpful but doesn't fully address the issue. Besides, the in-memory cache makes it extremely difficult to avoid bad locks while still ensuring isolation.

pkg/experiment/metastore/index/query.go
```diff
@@ -214,47 +235,60 @@ func newMetadataLabelQuerier(tx *bbolt.Tx, q *metadataQuery) *metadataLabelQueri
...
+	// See comments for blockMetadataQuerier.collectBlockMetadata.
+	q.buf = q.buf[:0]
+	var sTable *metadata.StringTable
+	err := s.view(q.shards.tx, func(shard *indexShard) error {
+		matcher := metadata.NewLabelMatcher(shard.StringTable, q.query.matchers, q.query.labels...)
+		if !matcher.IsValid() {
+			return nil
 		}
-		for _, ds := range md.Datasets {
-			if _, ok := mi.query.tenantMap[shard.StringTable.Lookup(ds.Tenant)]; !ok {
```
simonswine, line 248: This was fixed in #4160

PR #4072, opened 2025-04-03
feat(v2): raft wait-committed

kolesnikovae:
This change allows us to decouple the metadata read and write paths, ensuring that adding metadata entries does not have to wait for FSM application. This is particularly useful in cases where FSM application might be slow due to queries (which is something we need to optimize for better performance).

The new feature depends on https://github.com/hashicorp/raft/pull/640. The current PR version relies on a fork, but I'm happy to move it to Grafana if the PR takes longer to merge – I think that we should wait for the PR to be discussed: I'm not in favor of maintaining our own fork.


kolesnikovae:
Closing it for now because I got no feedback from hashicorp/raft maintainers

PR #4011, opened 2025-03-14
feat: Add Prometheus to the rideshare demo w/utf_8

shelldandy:
Enhances: https://github.com/grafana/pyroscope/pull/3905 with a couple more otel utf8 fields

PR #4075, opened 2025-04-03
fix: rideshare tweaks

shelldandy:
Continues from: https://github.com/grafana/pyroscope/pull/4011

Needed to add the utf8 metrics to our prometheus config so they're properly scrapped

image


examples/language-sdk-instrumentation/golang-push/rideshare/Dockerfile.load-generator
```diff
@@ -2,5 +2,5 @@ FROM golang:1.23.7
 
 WORKDIR /go/src/app
 COPY . .
-RUN go build loadgen.go
+RUN go build -tags loadgen -o loadgen .
```
marcsanmi, line 5: I think the way you're building the load generator is making it not to be profiled. Why do you need it this way?

examples/language-sdk-instrumentation/golang-push/rideshare/docker-compose.yml
```diff
@@ -1,5 +1,6 @@
 services:
   us-east:
+    container_name: us-east
```
marcsanmi, line 3: I'd say this isn't needed 🤔

examples/language-sdk-instrumentation/golang-push/rideshare/rideshare/rideshare.go
```diff
@@ -357,20 +332,10 @@ func MeterProvider(c Config) (*sdkmetric.MeterProvider, error) {
 		), nil
 	}
 
-	ctx := context.Background()
```
marcsanmi, line 360: The changes in the MeterProvider made the `handler.duration` histogram to stop working. You need to keep it with something like this:
```golang
	ctx := context.Background()
	exp, err := otlpmetrichttp.New(ctx)
	if err != nil {
		return nil, err
	}

	// Create a new tracer provider with a batch span processor and the otlp exporter.
	mp := sdkmetric.NewMeterProvider(
		sdkmetric.WithResource(newResource(c)),
		sdkmetric.WithReader(sdkmetric.NewPeriodicReader(exp,
			sdkmetric.WithInterval(interval))),
	)
```

examples/language-sdk-instrumentation/golang-push/rideshare/docker-compose.yml
```diff
@@ -8,27 +9,26 @@ services:
     - PARAMETERS_POOL_SIZE=1000
     - PARAMETERS_POOL_BUFFER_SIZE_KB=1000
     - OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
-    - OTEL_EXPORTER_OTLP_METRICS_ENDPOINT=http://prometheus:9090/api/v1/otlp/v1/metrics
```
marcsanmi, line 11: You need to keep this environment variable for otlp metrics to work. Same for the other services.

examples/language-sdk-instrumentation/golang-push/rideshare/rideshare/rideshare.go
```diff
@@ -263,51 +262,17 @@ func staticList(input []string) func() string {
...
-			label:        "room",
-			getNextValue: staticList([]string{`"Friends Don't Lie"`}),
-		},
-	}
 
-	dimensions := []string{}
-	for _, dim := range fakeUtf8Metrics {
-		dimensions = append(dimensions, dim.label)
+	// Initialize metrics
+	dimensions := []string{
```
marcsanmi, line 267: I'm adding this comment here since I missed this in the previous review, but I don't this this metric is going to work:
```golang
	utf8Metric := promauto.NewCounterVec(prometheus.CounterOpts{
		Name: "a.utf8.metric 🤘",
		Help: "a utf8 metric with utf8 labels",
	}, dimensions)
```
We don't support it. However the one with ut8 labels does work.

Screenshot 2025-04-04 at 17 33 43

examples/language-sdk-instrumentation/golang-push/rideshare/loadgen.go
```diff
@@ -1,9 +1,11 @@
+//go:build loadgen
```
marcsanmi, line 1: Same as above, I don't think this is needed

marcsanmi (changes requested):

examples/language-sdk-instrumentation/golang-push/rideshare/docker-compose.yml
shelldandy: its needed so i can add them in the prometheus.yml file without having to use the funky autogenerated name docker gives when not adding it

examples/language-sdk-instrumentation/golang-push/rideshare/prometheus-provisioning/prometheus.yaml
```diff
@@ -21,3 +21,6 @@ scrape_configs:
     static_configs:
       - targets:
           - host.docker.internal:9099
+  - job_name: 'rideshare'
+    static_configs:
+      - targets: ['us-east:5000', 'eu-north:5000', 'ap-south:5000']
```
shelldandy, line 26: @marcsanmi this is why i need the container names , without this scrape i don't see the metrics i added show up in the explorer

marcsanmi (commented):
Hi @shelldandy,
I still see the `a.utf8.metric` being added. It's breaking the experience in some screens, such as in the metrics explorer, as I showed in the last comment.

When addressing that, could you take the chance to update the branch? Thanks for all the work! 🙂

shelldandy:
> Hi @shelldandy,
> 
> I still see the `a.utf8.metric` being added. It's breaking the experience in some screens, such as in the metrics explorer, as I showed in the last comment.
> 
> 
> 
> When addressing that, could you take the chance to update the branch? Thanks for all the work! 🙂 

The metric is still there but without the emoji like before, is that still a problem? I'm only using normal characters in there. 

I'll update the branch as well no problem

examples/language-sdk-instrumentation/golang-push/rideshare/rideshare/rideshare.go
```diff
@@ -358,12 +326,22 @@ func MeterProvider(c Config) (*sdkmetric.MeterProvider, error) {
 	}
 
 	ctx := context.Background()
-	exp, err := otlpmetrichttp.New(ctx)
+	opts := []otlpmetrichttp.Option{otlpmetrichttp.WithEndpoint(c.OTLPUrl)}
```
marcsanmi, line 329: I've found another issue. I think this is the culprit causing this,  since it overrides the `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` environment variable configuration:
```
2025/04/23 07:47:43 failed to upload metrics: Post "http:///api/v1/otlp/v1/metrics": http: no Host in request URL
```

After the changes, try to run the example to see if the otlp metrics added in this PR https://github.com/grafana/pyroscope/pull/3905 are still available.

marcsanmi (commented):
One last fix and we should be able to merge 🙂 

Please take the chance to update the branch from main, sorry about this.

examples/language-sdk-instrumentation/golang-push/rideshare/rideshare/rideshare.go
marcsanmi: Still seeing the error above when running the example with `docker-compose up --build`. Could you please another look?

marcsanmi: Still seeing the error above when running the example with `docker-compose up --build`. Could you please another look?

shelldandy: I think it should be working properly now, please take another look @marcsanmi

simonswine (approved):
LGTM. Gave this is a quick spin locally all appears to be working as expected.

Thank you for updating the OTEL metrics in our rideshare example.

PR #4076, opened 2025-04-04
fix(ingest pyspy): remove line numbers from synthetic function names

korniltsev:
I believe we should remove the whole function name rewriting altogether, and solve this issue in frontend, but that's for another change.


kolesnikovae (approved):
Agreed.

This change is somewhat invasive, as it may affect users who do want to see the line numbers — but I think it's worth it. Am I right in thinking we can still access this info via the Function Details view, even without GitHub integration?

korniltsev:
> Am I right in thinking we can still access this info via the Function Details view, even without GitHub integration?

I was hoping so :rofl:

Let me double check before merging

korniltsev:
image
Ugly but possible

PR #4077, opened 2025-04-04
fix(ingest pyspy): remove line numbers from synthetic function names …

korniltsev:
…(#4076)

cherry-pick 368696096


kolesnikovae (approved):
I usually push straight to weekly branches :) – assuming we already have it in the main

korniltsev:
I was going to rollout this change to one of our GL prod cluster, but then we discussed and decided not to rush and rollout this change as usual next week

PR #4079, opened 2025-04-07
feat(v2): non-blocking metadata queries

kolesnikovae:
The PR addresses an issue with blocking metadata queries.

Currently, we have a thick cache layer over the index stored in BoltDB, which is fully loaded into memory. The problem is that the cache is too coarse: an entire partition shard is used as a cache item. This requires broad locks and sophisticated cache management to ensure consistency with the on-disk state. Worse still, in certain cases, reads block writes for an extended period (seconds).

In the new version, we have two caches: one for partition shards (without blocks) and another for individual blocks. The first is primarily used as a skip index, while the latter is used to minimize resources spent on unmarshaling.

Queries never block the writer for longer than the time it takes to load the shard into memory (without blocks, just the shard index and the string table – a few dozen keys or hundreds of kilobytes), and they always read block keys from the DB in their own transaction, making locks unnecessary.

### DIFF
--- a/pkg/experiment/metastore/index/index_cache.go
+++ b/pkg/experiment/metastore/index/index_cache.go
@@ -28,7 +28,7 @@ type shardCacheKey struct {
 
 type blockCache struct {
 	mu    sync.RWMutex
-	reads *lru.TwoQueueCache[blockCacheKey, *metastorev1.BlockMeta]
+	read  *lru.TwoQueueCache[blockCacheKey, *metastorev1.BlockMeta]
 	write *lru.Cache[blockCacheKey, *metastorev1.BlockMeta]
 }
 
@@ -68,7 +68,7 @@ func newBlockCache(rcs, wcs int) *blockCache {
 	reads, _ := lru.New2Q[blockCacheKey, *metastorev1.BlockMeta](rcs)
 	write, _ := lru.New[blockCacheKey, *metastorev1.BlockMeta](wcs)
 	return &blockCache{
-		reads: reads,
+		read:  reads,
 		write: write,
 	}
 }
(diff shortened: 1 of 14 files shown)


pkg/experiment/metastore/index/index_cache.go
```diff
@@ -0,0 +1,135 @@
...
+type shardCacheKey struct {
+	partition store.PartitionKey
+	tenant    string
+	shard     uint32
+}
+
+type blockCache struct {
+	mu    sync.RWMutex
+	reads *lru.TwoQueueCache[blockCacheKey, *metastorev1.BlockMeta]
+	write *lru.Cache[blockCacheKey, *metastorev1.BlockMeta]
```
aleks-p, line 32: Nit, should we call the caches `blocksRead` and `blocksWrite` for more consistency?

pkg/experiment/metastore/store/store.go
```diff
@@ -14,29 +14,40 @@ type KV struct {
...
 }
 
 type CursorIterator struct {
 	cursor *bbolt.Cursor
 	seek   bool
-	prefix []byte
 	k, v   []byte
+
+	// Prefix that keys must start with.
+	Prefix []byte
```
aleks-p, line 27: is this used?

aleks-p (approved):
Looks great to me!

pkg/experiment/metastore/store/store.go
kolesnikovae: It's not used in the code at the moment, but I use it extensively for experimenting and testing. I’ve written it so many times that I decided to keep it.

pkg/experiment/metastore/index/index_cache.go
kolesnikovae: Thanks for pointing this out! There's a typo – it should be `writes`.

Could you please elaborate on consistency?

aleks-p: Just referring to the other cache (`shardsCache`) having a `shards` field, while this cache has fields named `reads` & `writes`, but this is a very minor inconsistency.

PR #4080, opened 2025-04-07
feat(v2): metastore snapshot write rate limit

kolesnikovae:
The PR adds a configuration option to limit the rate at which FSM snapshots are written on disk. This is useful in cases when the same disk is used for both WAL and snapshots.

### DIFF
--- a/pkg/util/ratelimit/ratelimit.go
+++ b/pkg/util/ratelimit/ratelimit.go
@@ -1,0 +1,42 @@
+package ratelimit
+
+import "time"
+
+// Limiter implements a simple token-bucket rate limiter.
+// Tokens are replenished over time.
+type Limiter struct {
+	rate    float64
+	tokens  float64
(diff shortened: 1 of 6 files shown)


pkg/util/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,42 @@
...
+		}
+		if l.tokens >= float64(n) {
+			l.tokens -= float64(n)
+			return
+		}
+		missing := float64(n) - l.tokens
+		delay := time.Duration(missing/l.rate*1e9) * time.Nanosecond
+		l.sleep(delay)
+	}
+}
```
kolesnikovae, line 42: Unfortunately, the standard `x/time/rate` is unreasonably expensive

simonswine (approved):
LGTM

Thanks you for looking at that ❤️

PR #4081, opened 2025-04-07
chore(v2): add segment-writer metastore client options

kolesnikovae:
The PR adds configuration options to control the metastore client in the segment-writer service.

### DIFF
--- a/pkg/experiment/ingester/segment.go
+++ b/pkg/experiment/ingester/segment.go
@@ -130,7 +130,7 @@ func newSegmentWriter(l log.Logger, metrics *segmentMetrics, hm *memdb.HeadMetri
 		shards:      make(map[shardKey]*shard),
 		metastore:   metastoreClient,
 	}
-	sw.retryLimiter = retry.NewRateLimiter(sw.config.Upload.HedgeRateMax, int(sw.config.Upload.HedgeRateBurst))
+	sw.retryLimiter = retry.NewRateLimiter(sw.config.UploadHedgeRateMax, int(sw.config.UploadHedgeRateBurst))
 	sw.ctx, sw.cancel = context.WithCancel(context.Background())
 	flushWorkers := runtime.GOMAXPROCS(-1)
 	if config.FlushConcurrency > 0 {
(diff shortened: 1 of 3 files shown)


pkg/experiment/ingester/segment.go
```diff
@@ -36,8 +36,6 @@ import (
 	"github.com/grafana/pyroscope/pkg/validation"
 )
 
-var ErrMetastoreDLQFailed = fmt.Errorf("failed to store block metadata in DLQ")
```
kolesnikovae, line 39: It is not used anywhere.

pkg/experiment/ingester/service.go
```diff
@@ -78,12 +88,14 @@ func (cfg *UploadConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet)
...
-func (cfg *Config) Validate() error {
-	// TODO(kolesnikovae): implement.
-	if err := cfg.LifecyclerConfig.Validate(); err != nil {
-		return err
-	}
-	return cfg.GRPCClientConfig.Validate()
+type MetadataConfig struct {
+	DLQEnabled bool          `yaml:"dlq_enabled,omitempty" category:"advanced"`
+	Timeout    time.Duration `yaml:"timeout,omitempty" category:"advanced"`
+}
```
kolesnikovae, line 94: I don't quite like the naming and having too "hierarchical" configs. If anyone has good suggestions, I'm happy to change this (or we can do it later, we provide no compatibility guarantees for this yet).

```
-segment-writer.upload.timeout=1s
-segment-writer.metadata.timeout=1s
-segment-writer.metadata.dlq-enabled=false // declarative (this PR)
-segment-writer.metadata.disable-dlq=true  // imperative (also works, esp. for CLI)
```

```yaml
segment_writer:
  upload:
    timeout: 1s
    # ...
  metadata:
    timeout: 1s
    dlq_enabled: false
```

I'd probably prefer a flat set of options (even if they have verbose, repetitive prefixes): 
```yaml
segment_writer:
  segment_max_duration: 500ms
  upload_timeout: 1s
  upload_retry_max_retries: 3
  # ...
  upload_hedge_rate_burst: 2
  metadata_update_timeout: 1s
  metadata_dlq_enabled: 1s
```

simonswine: I would agree with flatter is better. 

I think the deeper than 2 level nesting only makes sense when we configuring alternatives (like different implementation of the same thing).

YAML doesn't really need to reflect the struct nesting, if you use `yaml:",inline"`

pkg/experiment/ingester/service.go
```diff
@@ -47,27 +47,37 @@ type Config struct {
...
+	}
+	return cfg.GRPCClientConfig.Validate()
 }
 
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	const prefix = "segment-writer"
 	cfg.GRPCClientConfig.RegisterFlagsWithPrefix(prefix, f)
 	cfg.LifecyclerConfig.RegisterFlagsWithPrefix(prefix+".", f, util.Logger)
 	cfg.Upload.RegisterFlagsWithPrefix(prefix+".upload.", f)
+	cfg.Metadata.RegisterFlagsWithPrefix(prefix+".metadata.", f)
```
simonswine, line 66: Not fully sure what the "gold standard" is, but this will add two dots.

simonswine (approved):
LGTM

PR #4086, opened 2025-04-08
refactor: Mark storage.prefix as non experimental

simonswine:
This PR migrates from the experimental flag `storage.storage-prefix` to non experimental `storage.prefix` flag.

This is on the back of a user that was not able to find this flag and
reaching out in the community slack.


aleks-p (approved):
LGTM

PR #4092, opened 2025-04-09
fix(v2): query series labels from dataset index

kolesnikovae:
If a `SeriesLabels` query hits the dataset index (tenant-wide) we cannot use `SeriesIndex` there as it points the dataset and not specific series within it.

### DIFF
--- a/pkg/experiment/query_backend/query_profile_entry.go
+++ b/pkg/experiment/query_backend/query_profile_entry.go
@@ -28,7 +28,7 @@ type ProfileEntry struct {
 func (e ProfileEntry) RowNumber() int64 { return e.RowNum }
 
 func profileEntryIterator(q *queryContext, groupBy ...string) (iter.Iterator[ProfileEntry], error) {
-	series, err := getSeriesLabels(q.ds.Index(), q.req.matchers, groupBy...)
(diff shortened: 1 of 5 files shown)


pkg/experiment/query_backend/query_profile_entry.go
```diff
@@ -93,6 +93,35 @@ func getSeriesLabels(reader phlaredb.IndexReader, matchers []*labels.Matcher, by
...
+	l := make(phlaremodel.Labels, 0, 6)
+	for postings.Next() {
+		fp, err := reader.SeriesBy(postings.At(), &l, &chunks, by...)
+		if err != nil {
+			return nil, err
+		}
+		_, ok := series[fp]
+		if ok {
+			continue
+		}
```
kolesnikovae, line 115: While it works, it's inefficient: we remember by series fingerprint, not the hash of the subset of labels – I'll fix it before merging

bryanhuhta:
Just tested this fix in dev and it works! Example test run: https://dev.grafana-dev.net/a/k6-app/runs/232003?tab=profile

bryanhuhta (approved):
LGTM

PR #4093, opened 2025-04-09
chore: add Lidia binary format module

marcsanmi:
This PR introduces the `Lidia` module, a custom binary format for efficient symbolization  of Go profiling data. The module provides functionality to create, read and query symbol tables optimized for fast address-to-function lookups.


lidia/constants.go
```diff
@@ -0,0 +1,30 @@
+package lidia
+
+import "hash/crc32"
+
+// File format constants
+const (
+	// Magic number for lidia files ("lidia" in little-endian ASCII)
```
korniltsev, line 7: ldia

lidia is 5 bytes :)

lidia/Makefile
```diff
@@ -0,0 +1,28 @@
+GO ?= go
```
korniltsev, line 1: do we need a separate makefile? can we just test & lint from the main Makefile, same as we currenlty do for ebpf module?

lidia/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Available Options
+
+- `WithCRC()`: Enables CRC32C checksums for data integrity
+- `WithFiles()`: Includes source file information
+- `WithLines()`: Includes line number information
+
+## File Format
+
+Lidia uses a binary format with the following sections:
```
korniltsev, line 90: We may want to describe the layout of the file in exact details. In this PR or a followup.

korniltsev (approved):

lidia/lidia.go
```diff
@@ -0,0 +1,261 @@
...
+	if hdr.magic != magic {
+		res.Close()
+		return nil, fmt.Errorf("invalid magic number: expected 0x%x, got 0x%x", magic, hdr.magic)
+	}
+
+	if hdr.version != version {
+		res.Close()
+		return nil, fmt.Errorf("unsupported version: expected %d, got %d", version, hdr.version)
+	}
+
```
kolesnikovae, line 90: Actually, it would be nice if we allowed for forward compatibility – let's discuss it separately

lidia/lidia.go
```diff
@@ -0,0 +1,261 @@
...
+		res.Close()
+		return nil, fmt.Errorf("failed to read VA table: %w", err)
+	}
+
+	if res.opt.crc {
+		if err = res.CheckCRC(); err != nil {
+			res.Close()
+			return nil, fmt.Errorf("CRC check failed: %w", err)
+		}
+	}
```
kolesnikovae, line 123: Keep in mind that `hdr` might be completely invalid – we never validate its checksum (we don't even have it). It means that things like `make([]byte, int(hdr.rangeTableHeader.fieldSize)*fieldsCount)` may cause OOM, overflows, and all sorts of issues caused by bad data.

I'd like to see a fuzz test – let's not worry about this too much right now, but we should definitely add it.

lidia/README.md
kolesnikovae: I personally would love to read how the new format better than e.g., gSYM

kolesnikovae (approved):

lidia/constants.go
marcsanmi: wrong comment, good catch!

lidia/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
simonswine, line 1: Can you also mention it here: https://github.com/grafana/pyroscope/blob/main/LICENSING.md

lidia/README.md
```diff
@@ -0,0 +1,105 @@
+# Lidia
+
+Lidia is a custom binary format for efficient symbolization of Go profiling data.
```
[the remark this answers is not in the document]
simonswine: I also would be more specific "symbolization" is vague and it doesn't need to be "profiling data" : 

Lidia is a binary format for efficient look up of symbols of a binary by virtual address.

simonswine (approved):
LGTM! Nice work for adapting this here

lidia/lidia.go
```diff
@@ -0,0 +1,261 @@
...
+type GoRange struct {
+	VA        uint64
+	Length    uint32
+	Function  string
+	File      string
+	CallFile  string
+	CallLine  uint32
+	Depth     uint32
+	LineTable LineTable
+}
```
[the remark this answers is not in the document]
korniltsev: In the otel repo it used to differentiate with cgo Range from the rust lib.
Here it does not make much sense to call it GoRange now.

lidia/lidia.go
```diff
@@ -0,0 +1,261 @@
...
+				FilePath:     file,
+			}
+
+			// Add line number information if available
+			//if it.lineTable.count > 0 {
+			// Line number could be extracted here if implemented
+			//}
+
+			result = append(result, res)
+		}
```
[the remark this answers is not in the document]
marcsanmi: Great suggestion; I'll keep it in mind for future optimisations 🙂

lidia/README.md
marcsanmi: I've added the layout. Maybe @korniltsev you can (in a separate PR if you want) some insights about lidia vs gSYM?

korniltsev: I dont think or know if lidia is better or worse than gsym.

I've tried this implementation and found it has at leas one OOB read and very slow implementation https://github.com/ChimeHQ/gogsym

The difference between gsym and lidia would be
1.  that lidia has no variable length structures. When profiled the bad implementation was causing too much io syscalls, this could have probably
2. inline ranges are stored in the same ranges table and sorted by va-depth representing an inline tree. instead of optional variable length substruct in gsym. 

I would maybe do a benchmark: Implement mmap lidia and benchmark against blazesym.

Another thing I remember is that I've covnerted my local libc from dwarf to gsym with llvm and then for some reason it did not output any inline information. Again, may be a bug of the gogsym

kolesnikovae (approved):
Looks awesome! Thank you @marcsanmi @korniltsev 🚀

PR #4094, opened 2025-04-10
chore: Update prometheus to version v3

simonswine:
Note this is change of behaviour as it will relax the label validation and allow the newer metric selectors format, which in turn needs to be used to use the full character set.

There is also some bigger impact to the remote write/metrics export, as there is a new version of remote write in prometheus v3: https://prometheus.io/docs/specs/prw/remote_write_spec_2_0/

This will be the new label selector format:

```
{"metric.name", "my.label.name"="bar"}
```

In order to override other validation at a later point, we also need to set:

```
import "github.com/prometheus/common/model"

func init() {
    model.NameValidationScheme = model.UTF8Validation
}
```

I am **not** planning to do this as part of this PR.


marcsanmi (approved):
LGTM 👍 

We need to update prometheus testutils that has changed and some tests that now with a more relaxed label validation might expect or fail with different results.

pkg/experiment/ingester/client/client.go
```diff
@@ -127,16 +127,30 @@ func shouldBeHandledByCaller(err error) bool {
...
+        "retryableStatusCodes": [
+          "UNAVAILABLE"
+        ]
+      }
+    }
+  ],
+  "healthCheckConfig": {
+    "serviceName": "pyroscope.segment-writer"
+  }
 }`
```
simonswine, line 154: @kolesnikovae please take a look at this, I think after https://github.com/grpc/grpc-go/pull/7905 we cannot do this anymore using the service config.

kolesnikovae: Thanks for pointing this out!

We do want to disable gRPC retries as we manage it on our own on the caller side.

I think we should just remove the `methodConfig`: 
```go
const grpcServiceConfig = `{
    "healthCheckConfig": {
         "serviceName": "pyroscope.segment-writer"
    }
}`
```

This was an extra measure to not enable retries via service config mistakenly as it may come, e.g. from DNS, xDS, etc. We actually disable retries explicitly when create the connection, so this is not needed (exp. if the client just ignores it)

simonswine: Yes that works!

PR #4097, opened 2025-04-12
fix(v2): segment upload timeout

kolesnikovae:
Currently, the segment upload timeout is applied to each individual attempt. In practice, I saw that a configuration with 3 attempts and 1s timeout may take up to 5 seconds, which is not desirable.

This change modifies the behavior so that the timeout covers the entire upload routine – including backoff intervals, delays, and other overhead – making the process more predictable.

If a single attempt takes too long (e.g., due to a slow connection), it may be recovered using a hedged request. Regular retries are intended to recover from transient errors.


simonswine (approved):
LGTM

This makes the retries quite a bit more predictable.

PR #4102, opened 2025-04-14
fix(v2): object download error branch panic

kolesnikovae:
An error that occurs during object downloading in compaction workers may result in a panic. The reason is that one of the named return values is closed in a defer statement, but it may be nil if the error path is taken.

### DIFF
--- a/pkg/objstore/read_only_file.go
+++ b/pkg/objstore/read_only_file.go
@@ -33,39 +33,36 @@ func Download(ctx context.Context, name string, src BucketReader, dir string) (*
 	return f, nil
 }
 
-func download(ctx context.Context, name string, src BucketReader, dir string) (f *ReadOnlyFile, err error) {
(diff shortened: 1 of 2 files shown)


korniltsev (approved):

pkg/objstore/read_only_file.go
```diff
@@ -33,39 +33,36 @@ func Download(ctx context.Context, name string, src BucketReader, dir string) (*
 	return f, nil
 }
 
-func download(ctx context.Context, name string, src BucketReader, dir string) (f *ReadOnlyFile, err error) {
+func download(ctx context.Context, name string, src BucketReader, dir string) (*ReadOnlyFile, error) {
```
kolesnikovae, line 36: I think a linter that catches such cases might be helpful. Naked return is generally considered harmful, but it's unavoidable (or convenient) in some cases

korniltsev (approved):

marcsanmi (approved):
Thanks for the fix! 🚀

PR #4107, opened 2025-04-16
feat(v2): function totals recording rules

alsoba13:
Adding a stacktrace filter field in the recording rules.

This way, we can define filter for function names and extend it in the future.

Examples of use:
```
# total cpu per service
- metric_name: 'pyroscope_metric_cpu_total_nanoseconds'
  matchers: ['{__profile_type__="..."}']
  group_by: ['service_name']
- metric_name: 'pyroscope_metric_ingester_cpu_gc_total_nanoseconds'
  matchers: ['{__profile_type__="..."}']
  group_by: ['service_name']
  stacktrace_filter:
    type: 'function_name'
    expr: 'runtime.gcBgMarkWorker'
    metric_type: total
```

### DIFF
--- a/api/settings/v1/recording_rules.proto
+++ b/api/settings/v1/recording_rules.proto
@@ -74,11 +74,34 @@ message RecordingRule {
   repeated string matchers = 4;
   repeated string group_by = 5;
   repeated types.v1.LabelPair external_labels = 6;
+  StacktraceFilter stacktrace_filter = 7;
 
   // The observed generation of this recording rule. This value should be
   // provided when making updates to this record, to avoid conflicting
   // concurrent updates.
-  int64 generation = 7;
+  int64 generation = 8;
+}
+
+message StacktraceFilter {
+  StacktraceFilterType type = 1;
+  StacktraceFilterFunctionName function_name = 2;
+  // StacktraceFilterFunctionNameRegExp functionNameRegExp = 3;
+  // ...
+}
+
+enum StacktraceFilterType {
+  STACKTRACE_FILTER_FUNCTION_NAME = 0;
+  // STACKTRACE_FILTER_FUNCTION_NAME_REGEXP = 1;
+}
+
+enum MetricType {
+  TOTAL = 0;
+  // SELF = 1;
+}
+
+message StacktraceFilterFunctionName {
+  string function_name = 1;
+  MetricType metric_type = 2;
 }
 
 message RecordingRuleStore {


api/settings/v1/recording_rules.proto
```diff
@@ -74,11 +74,34 @@ message RecordingRule {
   repeated string matchers = 4;
   repeated string group_by = 5;
   repeated types.v1.LabelPair external_labels = 6;
+  StacktraceFilter stacktrace_filter = 7;
```
alsoba13, line 77: Optional. In absence of it _total_ dimension (current implementation) is used.

api/settings/v1/recording_rules.proto
```diff
@@ -74,11 +74,34 @@ message RecordingRule {
...
   // The observed generation of this recording rule. This value should be
   // provided when making updates to this record, to avoid conflicting
   // concurrent updates.
-  int64 generation = 7;
+  int64 generation = 8;
+}
+
+message StacktraceFilter {
+  StacktraceFilterType type = 1;
+  StacktraceFilterFunctionName function_name = 2;
```
simonswine, line 87: Rather than type I think it is nicer to use optional field per type and validate that only one is enabled at a time.

Alternatively there is also `oneOf` in protobuf: https://protobuf.dev/programming-guides/proto3/#using-oneof 
```suggestion
  optional StacktraceFilterFunctionName function_name = 1;
  // And in the future
  // optional StacktraceFilterFunctionNameRegExp functionNameRegExp = 2;
```

api/settings/v1/recording_rules.proto
```diff
@@ -74,11 +74,34 @@ message RecordingRule {
   repeated string matchers = 4;
   repeated string group_by = 5;
   repeated types.v1.LabelPair external_labels = 6;
+  StacktraceFilter stacktrace_filter = 7;
 
   // The observed generation of this recording rule. This value should be
   // provided when making updates to this record, to avoid conflicting
   // concurrent updates.
-  int64 generation = 7;
+  int64 generation = 8;
```
simonswine, line 82: Careful in protobuf those field IDs should never change or be reused after deprecation, as that would break existing API users.

api/settings/v1/recording_rules.proto
simonswine: Ideally that should take the optional argument

```suggestion
  optional StacktraceFilter stacktrace_filter = 7;
```

simonswine: I guess changes you make here also need to be done to `message RecordingRuleStore`?

simonswine (commented):
Thanks for working on this. A few suggestions, let me know once you want me to take another look

api/settings/v1/recording_rules.proto
alsoba13: Cool, I like this approach as well. I'll do it this way!

(I copied mine from `Query`)

api/settings/v1/recording_rules.proto
alsoba13: Yes, I need to `make generate` too. Will do it now

api/settings/v1/recording_rules.proto
alsoba13: I tried `oneOf`, but I didn't loved how it looked like. Opted for optionals

api/settings/v1/recording_rules.proto
```diff
@@ -79,6 +81,25 @@ message RecordingRule {
...
   // concurrent updates.
   int64 generation = 7;
+
+  optional StacktraceFilter stacktrace_filter = 8;
+}
+
+message StacktraceFilter {
+  optional StacktraceFilterFunctionName function_name = 1;
+
+  // StacktraceFilterFunctionNameRegExp functionNameRegExp = 2;
```
simonswine, line 91: I would remove those comments, they might just mislead users

api/settings/v1/recording_rules.proto
```diff
@@ -79,6 +81,25 @@ message RecordingRule {
...
+}
+
+enum MetricType {
+  TOTAL = 0;
+
+  // SELF = 1;
```
simonswine, line 97: I would remove those comments, they might just mislead users

api/settings/v1/recording_rules.proto
```diff
@@ -79,6 +81,25 @@ message RecordingRule {
   // provided when making updates to this record, to avoid conflicting
   // concurrent updates.
   int64 generation = 7;
+
+  optional StacktraceFilter stacktrace_filter = 8;
```
simonswine, line 85: Good opportunity for a comment:

```suggestion
  // The stacktrace filter allows filtering on particular function names in the stacktrace.
  // This allows recording rules to focus on specific functions and calculate their "total"
  // resource usage.
  optional StacktraceFilter stacktrace_filter = 8;
```

simonswine (approved):
LGTM

PR #4123, opened 2025-04-20
chore(v2): add option to specify raft fsm snapshot directory

kolesnikovae:
Previously, raft configuration accepted a single directory parameter for all raft-related files. In practice, it's often desirable to separate the WAL and snapshot storage. To facilitate this, two new configuration parameters have been introduced:
 * `snapshots-dir`: specifies the directory where Raft snapshots are stored. Overrides the default location.
 * `snapshots-import-dir`: an optional parameter that simplifies migration of existing snapshots.

### DIFF
--- a/pkg/experiment/metastore/raftnode/snapshot.go
+++ b/pkg/experiment/metastore/raftnode/snapshot.go
@@ -1,0 +1,109 @@
+package raftnode
+
+import (
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+	"path/filepath"
+
+	"github.com/go-kit/log/level"
+)
+
+// importSnapshots moves snapshots from SnapshotsImportDir to SnapshotsDir.
+//
+// When initializing the snapshot store, Raft creates a "snapshots" subdirectory.
+// To maintain consistency, this behavior is replicated here – so "snapshots" should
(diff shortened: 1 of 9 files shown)


pkg/experiment/metastore/raftnode/snapshot.go
```diff
@@ -0,0 +1,89 @@
...
+	for _, entry := range entries {
+		srcPath := filepath.Join(srcDir, entry.Name())
+		dstPath := filepath.Join(dstDir, entry.Name())
+		if _, err = os.Stat(dstPath); err == nil {
+			// While move is not guaranteed to be atomic, we do expect
+			// that if a file is found, it is a valid snapshot file.
+			level.Info(n.logger).Log("msg", "file already exists, skipping", "file", entry.Name())
+			continue
+		}
+		if err = os.Rename(srcPath, dstPath); err != nil {
```
korniltsev, line 83: The comment above says it will be moving across different filesystems.
If I remember correctly rename syscall does not work across different file systems.

kolesnikovae: That's a good point – I overlooked quite a few things there. Fixed!

korniltsev (approved):

PR #4124, opened 2025-04-20
fix(v2): allow empty ring in segment-writer client

kolesnikovae:
A follow-up to #4121: In all-in-one mode, the segment writer might not be available when the distributor attempts to start. Therefore, the "empty ring" situation is expected. Eventually, once all components are initialized, the client will establish connections.

A proper solution would be to introduce component dependencies for the all-in-one mode (e.g., the distributor should wait for ingesters, and so on) in addition to module dependencies.

### DIFF
--- a/pkg/experiment/ingester/client/client.go
+++ b/pkg/experiment/ingester/client/client.go
@@ -187,17 +187,21 @@ func (c *Client) starting(ctx context.Context) error {
 	// Warm up connections. The pool does not do this.
 	instances, err := c.ring.GetAllHealthy(ring.Reporting)
 	if err != nil {
-		return fmt.Errorf("get all healthy instances: %w", err)
+		// The ring might be empty initially if the segment-writer service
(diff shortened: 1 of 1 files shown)


korniltsev (approved):

pkg/experiment/ingester/client/client.go
```diff
@@ -186,18 +186,17 @@ func (c *Client) Service() services.Service { return c.service }
...
-		return fmt.Errorf("get all healthy instances: %w", err)
-	}
-	var wg sync.WaitGroup
-	for _, x := range instances.Instances {
-		wg.Add(1)
-		go func(x ring.InstanceDesc) {
-			defer wg.Done()
-			_, _ = c.pool.GetClientFor(x.Addr)
-		}(x)
+	if err == nil {
```
korniltsev, line 189: consider adding a comment why we ignore error in the else case

PR #4132, opened 2025-04-23
chore(helm): extra volume claim templates

kolesnikovae:
An option to include extra volume claim templates. Might be very helpful, e.g., to place different data (e.g. split by tenant) on different volumes.

The new `extraVolumeClaimTemplates` option is part of the `persistence.` member, in spite of other `extra*`. This is done so to avoid ambiguity.


simonswine (approved):
LGTM

PR #4136, opened 2025-04-24
feat: add enabling symbolizer override

marcsanmi:
This PR adds the capability to enable and disable symbolizer through internal system overrides. 
The whole flow can be summarised as:
- First, check if the symbolizer service is globally enabled (via v2 experiment flag)
- Then, for each request, check if symbolization is enabled for the tenant(s)


pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -109,12 +115,30 @@ func NewProfileSymbolizer(logger log.Logger, client DebuginfodClient, store Debu
...
+	tenantID := tenantIDs[0]
+
+	if !s.limits.SymbolizerEnabled(tenantID) {
+		level.Debug(s.logger).Log(
+			"msg", "symbolization disabled for tenant",
+			"tenant", tenantID,
+		)
+		return nil
+	}
+
```
kolesnikovae, line 141: I'm wondering if the check should be performed by the `SymbolizePprof` caller

kolesnikovae (approved):

pkg/experiment/symbolizer/symbolizer.go
marcsanmi: Initially I considered moving this logic to the query frontend level, where `SymbolizePprof` is called. However, this would require either adding symbolizer limits to frontend limits or passing additional limits to the query frontend. I opted to keep it in the symbolizer to maintain encapsulation, though moving the check outside remains a valid alternative if we want to make symbolization decisions more explicit. 
Do you think the code would be more readable?  I can do the change, since I wasn't sure about it neither.

kolesnikovae: I suggest that we perform the check on the caller side. It's more explicit, and it's how such checks are implemented in other places

marcsanmi: Updated in 821610f57e328cda7f02da29c39c8d9a2682402f

pkg/frontend/read_path/query_frontend/query_frontend.go
```diff
@@ -133,7 +136,16 @@ func (q *QueryFrontend) Query(
...
+	if len(tenants) == 0 {
+		return nil, fmt.Errorf("no tenant IDs found in context")
+	}
+	if len(tenants) > 1 {
+		return nil, fmt.Errorf("symbolization does not support multi-tenant requests")
+	}
+
+	symbolizationEnabled := q.symbolizerLimits.SymbolizerEnabled(tenants[0])
+
+	if symbolizationEnabled && hasNativeProfiles && q.symbolizer != nil {
```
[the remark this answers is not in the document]
marcsanmi: > Do we need the q.symbolizer != nil check? If the following is possible, we should find a way to avoid that:

No, we don't need it. The following isn't possible:
```
q.symbolizerLimits.SymbolizerEnabled && q.symbolizer == nil
```
I extended the PR description; we basically if the symbolizer si not enabled  we can't get in a `q.symbolizerLimits.SymbolizerEnable= true` situation.

pkg/frontend/read_path/query_frontend/query_frontend.go
```diff
@@ -95,23 +93,16 @@ func (q *QueryFrontend) Query(
...
+	shouldSymbolize := q.shouldSymbolize(tenants, blocks)
 
 	// Modify queries based on symbolization needs
 	modifiedQueries := make([]*queryv1.Query, len(req.Query))
 	for i, originalQuery := range req.Query {
 		modifiedQueries[i] = proto.Clone(originalQuery).(*queryv1.Query)
 
 		// If we need symbolization and this is a TREE query, convert it to PPROF
-		if hasNativeProfiles && originalQuery.QueryType == queryv1.QueryType_QUERY_TREE {
+		if shouldSymbolize && originalQuery.QueryType == queryv1.QueryType_QUERY_TREE {
```
kolesnikovae, line 105: Super minor note about `proto.Clone` a line above: please, use `type.CloneVT()`

pkg/frontend/frontend.go
```diff
@@ -112,6 +113,7 @@ type Limits interface {
 	MaxQueryLength(tenantID string) time.Duration
 	MaxQueryLookback(tenantID string) time.Duration
 	QueryAnalysisEnabled(string) bool
+	symbolizer.Limits
```
kolesnikovae, line 116:
```suggestion
	SymbolizerEnabled(string) bool
```

pkg/experiment/symbolizer/symbolizer.go
```diff
@@ -41,6 +41,10 @@ type Config struct {
 	PersistentDebugInfoStore    DebugInfoStoreConfig `yaml:"persistent_debuginfo_store"`
 }
 
+type Limits interface {
+	SymbolizerEnabled(string) bool
+}
+
```
kolesnikovae, line 47: I think we can optimize it: let's extend the `Limits` interface defined by frontend.

You may find it interesting:
 * https://go.dev/wiki/CodeReviewComments#interfaces
 * https://100go.co/#interface-on-the-producer-side-6

```suggestion
```

.mockery.yaml
```diff
@@ -64,3 +65,4 @@ packages:
     interfaces:
       DebuginfodClient:
       DebugInfoStore:
+      Limits:
```
kolesnikovae, line 68:
```suggestion
      Limits:
```

pkg/frontend/read_path/query_frontend/query_frontend.go
```diff
@@ -242,6 +208,73 @@ func (q *QueryFrontend) hasNativeProfiles(block *metastorev1.BlockMeta) bool {
...
+		return false
+	}
+
+	// Check if all tenants have symbolization enabled
+	// (once we support multi-tenant queries)
+	for _, t := range tenants {
+		if q.limits != nil && !q.limits.SymbolizerEnabled(t) {
+			return false
+		}
+	}
```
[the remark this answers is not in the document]
marcsanmi: As of now, it can't since it's always instantiated in the initOverrides. It's me being defensive here since I forgot to pass limits in a test when using `NewQueryFrontend`, but this should never happen with out current modules initiatilisation - going to remove it.

pkg/frontend/read_path/query_frontend/query_frontend.go
```diff
@@ -242,6 +208,73 @@ func (q *QueryFrontend) hasNativeProfiles(block *metastorev1.BlockMeta) bool {
 	return hasNativeProfiles
 }
 
+// shouldSymbolize determines if we should symbolize profiles based on tenant settings
+func (q *QueryFrontend) shouldSymbolize(tenants []string, blocks []*metastorev1.BlockMeta) bool {
+	if q.symbolizer == nil {
+		return false
+	}
```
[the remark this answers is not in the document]
marcsanmi: Same as https://github.com/grafana/pyroscope/pull/4136#discussion_r2060213326.

kolesnikovae (approved):
LGTM 🚀 

Thank you for your patience – and sorry for the many notes!

PR #4149, opened 2025-04-28
perf(v2): optimize compaction planning

kolesnikovae:
It's been observed that the compaction process might be very wasteful in terms of CPU resources. The reason is that the compaction planner inspects the block queues, cancelling the planned compaction job if the conditions are not met. The problem is that we do not skip the queues that we could have already checked.

image

### DIFF
--- a/pkg/experiment/metastore/compaction/compactor/compaction_queue.go
+++ b/pkg/experiment/metastore/compaction/compactor/compaction_queue.go
@@ -385,10 +385,6 @@ type blockIter struct {
 }
 
 func newBlockIter() *blockIter {
-	// Assuming that block IDs (16b ULID) are globally unique.
-	// We could achieve the same with more efficiency by marking visited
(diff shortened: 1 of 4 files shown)


pkg/experiment/metastore/compaction/compactor/compaction_queue.go
```diff
@@ -385,10 +385,6 @@ type blockIter struct {
 }
 
 func newBlockIter() *blockIter {
-	// Assuming that block IDs (16b ULID) are globally unique.
-	// We could achieve the same with more efficiency by marking visited
-	// batches. However, marking visited blocks seems to be more robust,
-	// and the size of the map is expected to be small.
```
kolesnikovae, line 391: It was sort of expected, but was pessimized prematurelly.

kolesnikovae:
After #4216 things look much better

PR #4151, opened 2025-04-28
fix: initialize phlaredb instance on query

kolesnikovae:
#4144 has introduced a bug where queries may ignore local data altogether as they wouldn't initialize the phlaredb instance and wouldn't sync metadata from the disk, if there were no pushes.


bryanhuhta (approved):
LGTM

simonswine (approved):
LGTM

PR #4152, opened 2025-04-28
fix: Panic in vcs getCommit when author empty

simonswine:
Something that we had encountered recently the ops cell.


bryanhuhta (approved):
LGTM

pkg/frontend/vcs/client/github_test.go
```diff
@@ -0,0 +1,122 @@
...
+			}
+
+			require.NoError(t, err)
+			assert.Equal(t, tt.expected, result)
+		})
+	}
+}
+
+// mockRepositoriesService implements the repositoryService interface
+type mockRepositoriesService struct {
```
marcsanmi, line 112: Can we use `.mockery.yaml` for the new interface mocks?

marcsanmi (approved):
LGTM 👍

PR #4160, opened 2025-04-29
fix(v2): add missing tenant filter for metadata label queries

aleks-p:
Fixes a rare case in v2 where datasets in L0 blocks are not correctly filtered by tenant.

This code path is only used in the `Series` endpoint, and only when requesting these 2 sets of labels AND no query selector is provided:

Set 1:
- `__profile_type__` 
- `service_name`

Set 2:
- `__name__` 
- `__profile_type__` 
- `__type__` 
- `pyroscope_app`
- `service_name`


bryanhuhta (approved):

korniltsev (approved):

simonswine (approved):
LGTM. Thanks for fixing this

PR #4162, opened 2025-04-30
reenable some more  workflows

korniltsev:


.github/workflows/update-examples-cron.yml
```diff
@@ -27,7 +26,7 @@ jobs:
 
       - name: Generate token
         id: generate_token
-        uses: tibdex/github-app-token@v2
+        uses: tibdex/github-app-token@3beb63f4bd073e61482598c45c71c1019b59b73a # v2.1.0
```
simonswine, line 29: Should migrate this to: https://github.com/actions/create-github-app-token

simonswine (approved):
Apart from my comment LGTM

.github/workflows/update-examples-cron.yml
korniltsev: Oh yes, I missed this one

PR #4163, opened 2025-04-30
fix: Do not initialise local querier, when there are no local blocks

simonswine:
This builds on top of #4151 and #4144, as previously initialised heads already created a folder on all ingesters. 

Hence the check is not detecting that there are actually no local blocks to be considered.


aleks-p (approved):
LGTM

PR #4173, opened 2025-05-02
feat: add __unsymbolized__ label on ingest path

marcsanmi:
This PR adds the __has_native_unsymbolized_profiles__ label on the ingest path.

Original [PR](https://github.com/grafana/pyroscope/pull/4147) that made us discover the panics.

### DIFF
--- a/pkg/experiment/ingester/memdb/head.go
+++ b/pkg/experiment/ingester/memdb/head.go
@@ -21,10 +21,11 @@ import (
 )
 
 type FlushedHead struct {
-	Index    []byte
-	Profiles []byte
-	Symbols  []byte
-	Meta     struct {
+	Index                   []byte
+	Profiles                []byte
+	Symbols                 []byte
+	HasUnsymbolizedProfiles bool
+	Meta                    struct {
 		ProfileTypeNames []string
 		MinTimeNanos     int64
 		MaxTimeNanos     int64
@@ -153,6 +154,8 @@ func (h *Head) flush(ctx context.Context) (*FlushedHead, error) {
 		return res, nil
 	}
 
+	res.HasUnsymbolizedProfiles = HasUnsymbolizedProfiles(h.symbols.Symbols())
+
 	symbolsBuffer := bytes.NewBuffer(nil)
 	if err := symdb.WritePartition(h.symbols, symbolsBuffer); err != nil {
 		return nil, err
@@ -172,4 +175,16 @@ func (h *Head) flush(ctx context.Context) (*FlushedHead, error) {
 		return nil, fmt.Errorf("failed to write profiles parquet: %w", err)
 	}
 	return res, nil
+}
+
+// TODO: move into the symbolizer package when available
+func HasUnsymbolizedProfiles(symbols *symdb.Symbols) bool {
+	locations := symbols.Locations
+	mappings := symbols.Mappings
+	for _, loc := range locations {
+		if !mappings[loc.MappingId].HasFunctions {
+			return true
+		}
+	}
+	return false
 }
--- a/pkg/experiment/ingester/memdb/head.go
+++ b/pkg/experiment/ingester/memdb/head.go
@@ -21,11 +21,11 @@ import (
 )
 
 type FlushedHead struct {
-	Index                   []byte
-	Profiles                []byte
-	Symbols                 []byte
-	HasUnsymbolizedProfiles bool
-	Meta                    struct {
+	Index        []byte
+	Profiles     []byte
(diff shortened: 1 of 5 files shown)


pkg/experiment/ingester/segment.go
```diff
@@ -216,7 +216,7 @@ func (sw *segmentsWriter) newSegment(sh *shard, sk shardKey, sl log.Logger) *seg
 func (s *segment) flush(ctx context.Context) (err error) {
 	span, ctx := opentracing.StartSpanFromContext(ctx, "segment.flush", opentracing.Tags{
 		"block_id": s.ulid.String(),
-		"datasets": len(s.heads),
+		"datasets": len(s.datasets),
```
korniltsev, line 219: Is this rename needed for the change?

korniltsev:
What was the issue causing panic? What is the difference with the previous attempt?

pkg/experiment/ingester/segment.go
marcsanmi: Not at all. It was a rename we did with Anton when review this, but to keep minimal changes we can leave it as `heads`. 

PS: I realised I didn't rename `headForIngest`.

marcsanmi:
> What was the issue causing panic?

That's the fun part. I've basically recreated the changes on top of yours and everything works now 😕 I'll take another look on Monday.

korniltsev:
Screenshot_20250505_141322
it's a spot the difference game :D
Screenshot_20250505_141322

korniltsev (changes requested):
Let's do renaming in a separate PR, I think the bug was hard to spot during review because of rename. Otherwise looks good to me - same as with previous PR.

korniltsev:
Bonus points if you add a unit test that would catch the bug (without integration tests)

pkg/experiment/ingester/memdb/head.go
```diff
@@ -21,10 +21,11 @@ import (
...
 
 type FlushedHead struct {
-	Index    []byte
-	Profiles []byte
-	Symbols  []byte
-	Meta     struct {
+	Index                   []byte
+	Profiles                []byte
+	Symbols                 []byte
+	HasUnsymbolizedProfiles bool
```
korniltsev, line 27: nit: Should we maybe rename this to `Unsymbolized` ? to make it look less like java and also to look more like the new label we add (I rememver it used to be `__has_unsymbolized_profiles__` but then we changed it)

pkg/experiment/ingester/segment_test.go
```diff
@@ -445,6 +447,46 @@ func TestDLQRecovery(t *testing.T) {
 	sw.queryInputs(clients, inputs)
 }
 
+func TestUnsymbolizedLabelIsSet(t *testing.T) {
```
korniltsev, line 450: It looks like this new test does not catch the panics we had.

I did the following test:
```diff
~/p/pyroscope (marcsanmi/add-unsymbolized-label) > git diff
diff --git a/pkg/experiment/ingester/segment.go b/pkg/experiment/ingester/segment.go
index ba864d9f3..38fe09bf3 100644
--- a/pkg/experiment/ingester/segment.go
+++ b/pkg/experiment/ingester/segment.go
@@ -544,7 +544,7 @@ func (v *sampleAppender) VisitSampleSeries(labels []*typesv1.LabelPair, samples
        }
        var n profilev1.Profile
        v.exporter.ExportSamples(&n, samples)
-       v.head.Ingest(&n, v.id, labels, v.annotations)
+       v.head.Ingest(v.profile, v.id, labels, v.annotations)
 }
```

I believe the bug is triggered by a pprof with sample labels.

marcsanmi: Updated the test to trigger the panic when changed like this in commit e7381020fe11e0c41084e2089a576c0b0779f135

korniltsev (approved):

pkg/experiment/ingester/memdb/head.go
```diff
@@ -173,3 +176,15 @@ func (h *Head) flush(ctx context.Context) (*FlushedHead, error) {
 	}
 	return res, nil
 }
+
+// TODO: move into the symbolizer package when available
+func HasUnsymbolizedProfiles(symbols *symdb.Symbols) bool {
+	locations := symbols.Locations
+	mappings := symbols.Mappings
+	for _, loc := range locations {
+		if !mappings[loc.MappingId].HasFunctions {
```
simonswine, line 185: Can we trust the mapping id here?

PR #4174, opened 2025-05-05
chore: run tests on both go1.23 and go1.24

korniltsev:


simonswine:
Any particular reason to do this now? I think we should make sure we are seeing less flaky tests, before increasing the amount of test runs. Wdyt?

korniltsev:
I've just noticed a failing test in go1.24 and want to investigate why is it failing and fix it (likely the test).

PR #4177, opened 2025-05-06
chore: Enable helm release workflow and tag v1.31.2

simonswine:
This enables the helm release workflow and also update the chart to point to the latest release (v.1.13.2), which we forgot at the time.


.github/workflows/helm-release.yml
```diff
@@ -0,0 +1,21 @@
...
+      contents: "write"
+      id-token: "write"
+      packages: "write"
+    uses: grafana/helm-charts/.github/workflows/update-helm-repo.yaml@64b1c8260d69a9b1d9a186df25c51358b92f7662
+    with:
+      charts_dir: operations/pyroscope/helm/
+      cr_configfile: operations/pyroscope/helm/cr.yaml
+      ct_configfile: operations/pyroscope/helm/ct.yaml
+    secrets:
+      vault_repo_secret_name: grafana-pyroscope-bot
```
korniltsev, line 21: nit: I think we're switching apps here, I am not 100% sure but I think we used to use `grafana-pyroscope-releases`. I also think it does not matter and probably even better to use less apps

korniltsev (approved):
lgtm

.github/workflows/helm-release.yml
simonswine: Yes that was intentional, as the vault prefix did not have that other app. Should have mentioned it though. Let's see if it works 🤞

PR #4183, opened 2025-05-07
chore: Add integration tests about tenant isolation

simonswine:
This adds multi tenant read path tests to the existing micro-services integration tests.

This tests currently only cover the default architecture. I plan to address v2 with a follow up PR.


aleks-p (approved):
Looks good to me, added a couple of minor remarks.

PR #4184, opened 2025-05-07
pyroscope(v2): query-backend tenant isolation check

alsoba13:
Validating the read block request so we ensure no foreign dataset is read

### DIFF
--- a/pkg/experiment/query_backend/block_reader.go
+++ b/pkg/experiment/query_backend/block_reader.go
@@ -120,6 +120,10 @@ func validateRequest(req *queryv1.InvokeRequest) (*request, error) {
 	if req.QueryPlan == nil || len(req.QueryPlan.Root.Blocks) == 0 {
 		return nil, fmt.Errorf("no blocks to query")
 	}
+	err := validateTenant(req)
+	if err != nil {
+		return nil, fmt.Errorf("querying other tenants' dataset: %w", err)
+	}
 	matchers, err := parser.ParseMetricSelector(req.LabelSelector)
 	if err != nil {
 		return nil, fmt.Errorf("label selection is invalid: %w", err)
@@ -131,4 +135,20 @@ func validateRequest(req *queryv1.InvokeRequest) (*request, error) {
 		endTime:   model.Time(req.EndTime).UnixNano(),
 	}
 	return &r, nil
+}
+
+func validateTenant(req *queryv1.InvokeRequest) error {
+	tenantMap := make(map[string]struct{})
+	for _, tenant := range req.Tenant {
+		tenantMap[tenant] = struct{}{}
+	}
+	for _, b := range req.QueryPlan.Root.Blocks {
+		for _, dataset := range b.Datasets {
+			_, ok := tenantMap[b.StringTable[dataset.Tenant]]
+			if !ok {
+				return fmt.Errorf("dataset %d doesn't belong to tenant %q", dataset.Name, req.Tenant)
+			}
+		}
+	}
+	return nil
 }
--- a/pkg/experiment/query_backend/block_reader.go
+++ b/pkg/experiment/query_backend/block_reader.go
@@ -88,7 +88,7 @@ func (b *BlockReader) Invoke(
 	for _, md := range req.QueryPlan.Root.Blocks {
 		md.Datasets, err = filterNotOwnedDatasets(md, tenantMap)
 		if err != nil {
-			b.metrics.tenantIsolationViolationAttempt.Inc()
+			b.metrics.datasetTenantIsolationFailure.Inc()
 			traceId, _ := tracing.ExtractTraceID(ctx)
 			level.Error(b.log).Log("msg", "trying to query datasets of other tenants", "valid-tenant", strings.Join(req.Tenant, ","), "block", md.Id, "err", err, "traceId", traceId)
 		}
--- a/pkg/experiment/query_backend/block_reader.go
+++ b/pkg/experiment/query_backend/block_reader.go
@@ -86,10 +86,13 @@ func (b *BlockReader) Invoke(
 	}
 
 	for _, md := range req.QueryPlan.Root.Blocks {
-		if err = validateTenant(md, tenantMap); err != nil {
+		md.Datasets, err = filterInvalidDatasets(md, tenantMap)
+		if err != nil {
 			b.metrics.mismatchingTenantDataset.WithLabelValues(req.Tenant[0]).Inc()
 			traceId, _ := tracing.ExtractTraceID(ctx)
 			level.Error(b.log).Log("msg", "trying to query datasets of other tenants", "valid-tenant", strings.Join(req.Tenant, ","), "block", md.Id, "err", err, "traceId", traceId)
+		}
+		if len(md.Datasets) == 0 {
 			continue
 		}
 		obj := block.NewObject(b.storage, md)
@@ -153,14 +156,17 @@ func validateRequest(req *queryv1.InvokeRequest) (*request, error) {
 	return &r, nil
 }
 
-func validateTenant(b *metastorev1.BlockMeta, tenantMap map[string]struct{}) error {
+func filterInvalidDatasets(b *metastorev1.BlockMeta, tenantMap map[string]struct{}) ([]*metastorev1.Dataset, error) {
 	errs := multierror.New()
+	datasets := make([]*metastorev1.Dataset, 0, len(b.Datasets))
 	for _, dataset := range b.Datasets {
 		datasetTenant := b.StringTable[dataset.Tenant]
 		_, ok := tenantMap[datasetTenant]
-		if !ok {
+		if ok {
+			datasets = append(datasets, dataset)
+		} else {
 			errs.Add(fmt.Errorf(`dataset "%s" belongs to tenant "%s"`, b.StringTable[dataset.Name], datasetTenant))
 		}
 	}
-	return errs.Err()
+	return datasets, errs.Err()
 }
--- a/pkg/experiment/query_backend/block_reader.go
+++ b/pkg/experiment/query_backend/block_reader.go
@@ -86,9 +86,9 @@ func (b *BlockReader) Invoke(
 	}
 
 	for _, md := range req.QueryPlan.Root.Blocks {
-		md.Datasets, err = filterInvalidDatasets(md, tenantMap)
+		md.Datasets, err = filterNotOwnedDatasets(md, tenantMap)
 		if err != nil {
-			b.metrics.mismatchingTenantDataset.WithLabelValues(req.Tenant[0]).Inc()
+			b.metrics.tenantIsolationViolationAttempt.Inc()
 			traceId, _ := tracing.ExtractTraceID(ctx)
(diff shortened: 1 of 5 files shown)


pkg/experiment/query_backend/block_reader.go
```diff
@@ -120,6 +120,10 @@ func validateRequest(req *queryv1.InvokeRequest) (*request, error) {
 	if req.QueryPlan == nil || len(req.QueryPlan.Root.Blocks) == 0 {
 		return nil, fmt.Errorf("no blocks to query")
 	}
+	err := validateTenant(req)
+	if err != nil {
+		return nil, fmt.Errorf("querying other tenants' dataset: %w", err)
```
aleks-p, line 125: I am not entirely sure, but this error will probably be returned back to the calling client.

I wonder if we should instead be more vague ("tenant validation failure for block dataset"). Alternatively, we can also consider not failing the request, but instead skipping offending datasets (and increase a metric, log an error).

Any thoughts @alsoba13 , @simonswine ?

pkg/experiment/query_backend/block_reader.go
```diff
@@ -132,3 +136,19 @@ func validateRequest(req *queryv1.InvokeRequest) (*request, error) {
...
+
+func validateTenant(req *queryv1.InvokeRequest) error {
+	tenantMap := make(map[string]struct{})
+	for _, tenant := range req.Tenant {
+		tenantMap[tenant] = struct{}{}
+	}
+	for _, b := range req.QueryPlan.Root.Blocks {
+		for _, dataset := range b.Datasets {
+			_, ok := tenantMap[b.StringTable[dataset.Tenant]]
+			if !ok {
```
aleks-p, line 148: We probably want to log the block id and some identifier for the dataset, so that an issue could be better investigated. Increasing a metric counter could also make sense.

pkg/experiment/query_backend/block_reader.go
simonswine: I agree, this error would look pretty concerning to a client. 

I think we should be not impacting the query experience without a good reason, so I think we should skip blocks that don't validate log a an error as described by @aleks-p below and introduce a metric_counter/alert.

alsoba13: On it

pkg/experiment/query_backend/block_reader.go
simonswine: Everything that @aleks-p said and also an identifier for the original query traceID would be handy.

pkg/experiment/query_backend/block_reader_test.go
```diff
@@ -250,3 +263,24 @@ func (s *testSuite) Test_QueryTimeSeries() {
...
+	resp, err := s.reader.Invoke(s.ctx, &queryv1.InvokeRequest{
+		EndTime:       time.Now().UnixMilli(),
+		LabelSelector: "{}",
+		QueryPlan:     s.plan,
+		Query: []*queryv1.Query{{
+			QueryType: queryv1.QueryType_QUERY_TREE,
+			Tree:      &queryv1.TreeQuery{MaxNodes: 16},
+		}},
+		Tenant: []string{queryTenant},
+	})
```
alsoba13, line 281: example log message:
```
    log.go:168: level=error msg="trying to query datasets of other tenants" valid-tenant=some-tenant block=01JKT2S01VXGG0YS6TG8QC2JWD err="dataset \"\" belongs to tenant \"anonymous\"" traceId=
    log.go:168: level=error msg="trying to query datasets of other tenants" valid-tenant=some-tenant block=01JKT2XEGT4VA4HHR0EC8668MM err="dataset \"\" belongs to tenant \"anonymous\"" traceId=
    log.go:168: level=error msg="trying to query datasets of other tenants" valid-tenant=some-tenant block=01JKT2XEGT9M0S3P8Y8Q3E7SQ2 err="dataset \"\" belongs to tenant \"anonymous\"" traceId=
    log.go:168: level=error msg="trying to query datasets of other tenants" valid-tenant=some-tenant block=01JKT3DJRETWN25NKEFZNEP1V0 err="dataset \"\" belongs to tenant \"anonymous\"" traceId=
    log.go:168: level=error msg="trying to query datasets of other tenants" valid-tenant=some-tenant block=01JKT426996H1XWCX1M1ZQFRJ6 err="dataset \"test-app\" belongs to tenant \"anonymous\"" traceId=
    log.go:168: level=error msg="trying to query datasets of other tenants" valid-tenant=some-tenant block=01JKT4333784C9PCQ6VM8VNAWC err="dataset \"test-app\" belongs to tenant \"anonymous\"" traceId=

```

pkg/experiment/query_backend/block_reader.go
```diff
@@ -71,7 +80,18 @@ func (b *BlockReader) Invoke(
...
+	tenantMap := make(map[string]struct{})
+	for _, tenant := range req.Tenant {
+		tenantMap[tenant] = struct{}{}
+	}
+
 	for _, md := range req.QueryPlan.Root.Blocks {
+		if err = validateTenant(md, tenantMap); err != nil {
+			b.metrics.mismatchingTenantDataset.WithLabelValues(req.Tenant[0]).Inc()
+			traceId, _ := tracing.ExtractTraceID(ctx)
+			level.Error(b.log).Log("msg", "trying to query datasets of other tenants", "valid-tenant", strings.Join(req.Tenant, ","), "block", md.Id, "err", err, "traceId", traceId)
```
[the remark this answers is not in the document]
alsoba13: Is `tracing.ExtractTraceID(ctx)` enough to get the original query's traceID @simonswine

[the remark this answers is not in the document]
simonswine: If we propagate everything correctly, yes. I think something we need to try out.

pkg/experiment/query_backend/block_reader.go
```diff
@@ -71,7 +80,18 @@ func (b *BlockReader) Invoke(
 	g, ctx := errgroup.WithContext(ctx)
 	agg := newAggregator(req)
 
+	tenantMap := make(map[string]struct{})
+	for _, tenant := range req.Tenant {
+		tenantMap[tenant] = struct{}{}
+	}
```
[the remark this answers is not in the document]
simonswine: In some of the other databases it is possible to query multiple tenants at the same time, the API is ready for it, but we don't really officially support it.

I don't think `len(req.Tenant) == 0` is a valid request, I probably would just panic it.

pkg/experiment/query_backend/block_reader.go
```diff
@@ -71,7 +80,18 @@ func (b *BlockReader) Invoke(
...
+
 	for _, md := range req.QueryPlan.Root.Blocks {
+		if err = validateTenant(md, tenantMap); err != nil {
+			b.metrics.mismatchingTenantDataset.WithLabelValues(req.Tenant[0]).Inc()
+			traceId, _ := tracing.ExtractTraceID(ctx)
+			level.Error(b.log).Log("msg", "trying to query datasets of other tenants", "valid-tenant", strings.Join(req.Tenant, ","), "block", md.Id, "err", err, "traceId", traceId)
+			continue
```
simonswine, line 93: I might be missing something, but instead of skipping the whole block, we could just filter the datasets out, that are not part of the tenantMap.

alsoba13: You are probably right, on it.

alsoba13: addressed at https://github.com/grafana/pyroscope/pull/4184/commits/3bb130f896b0bc0633ab68a8ed909a63f311aefe

pkg/experiment/query_backend/metrics.go
```diff
@@ -0,0 +1,22 @@
...
+}
+
+func newMetrics(reg prometheus.Registerer) *metrics {
+	m := &metrics{
+		mismatchingTenantDataset: prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Namespace: "pyroscope",
+				Subsystem: "query_backend",
+				Name:      "invalid_tenant_query_total",
+			}, []string{"tenant"}),
```
simonswine, line 16: I think we probably don't need the break down per tenant in the metric, it basically should not happen for any tenant, if it does we have to identify via logs anyhow what exactly happened.

pkg/experiment/query_backend/block_reader.go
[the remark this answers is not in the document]
simonswine: Can you add a check to `validateRequest` that req.Tenant has at least 1 element?

pkg/experiment/query_backend/block_reader.go
```diff
@@ -132,3 +155,18 @@ func validateRequest(req *queryv1.InvokeRequest) (*request, error) {
 	}
 	return &r, nil
 }
+
+func filterInvalidDatasets(b *metastorev1.BlockMeta, tenantMap map[string]struct{}) ([]*metastorev1.Dataset, error) {
```
simonswine, line 159: I don't think `Invalid` is good, maybe this works better (open to other suggestions)

```suggestion
func filterNotOwnedDatasets(b *metastorev1.BlockMeta, tenantMap map[string]struct{}) ([]*metastorev1.Dataset, error) {
```

```suggestion
func filterAlienDatasets(b *metastorev1.BlockMeta, tenantMap map[string]struct{}) ([]*metastorev1.Dataset, error) {
```

```suggestion
func filterOtherTenantsDatasets(b *metastorev1.BlockMeta, tenantMap map[string]struct{}) ([]*metastorev1.Dataset, error) {
```

simonswine, line 159: Also a comment would be nice, why we do that:

```suggestion
// While the metastore is expected to already filter datasets of other tenants, we do an additional check to avoid processing blocks or datasets belonging to the wrong tenant.
func filterInvalidDatasets(b *metastorev1.BlockMeta, tenantMap map[string]struct{}) ([]*metastorev1.Dataset, error) {
```

pkg/experiment/query_backend/block_reader.go
```diff
@@ -132,3 +155,18 @@ func validateRequest(req *queryv1.InvokeRequest) (*request, error) {
...
+	errs := multierror.New()
+	datasets := make([]*metastorev1.Dataset, 0, len(b.Datasets))
```
simonswine, line 161: We probably could get rid of that allocation, esp. since we would not expect this to ever be hit (the normal case is our dataset tenant filtering works).

simonswine (approved):
A few more comments from me, once addressed I would like this to be deployed to a dev cell before we merge.

Otherwise LGTM!

pkg/experiment/query_backend/metrics.go
```diff
@@ -0,0 +1,22 @@
...
+	tenantIsolationViolationAttempt prometheus.Counter
+}
+
+func newMetrics(reg prometheus.Registerer) *metrics {
+	m := &metrics{
+		tenantIsolationViolationAttempt: prometheus.NewCounter(
+			prometheus.CounterOpts{
+				Namespace: "pyroscope",
+				Subsystem: "query_backend",
+				Name:      "tenant_isolation_violation_query_attempt_total",
```
aleks-p, line 15: Nit, "attempt" sounds like someone is trying to extract data they don't own. I think the idea is more to report a failure of the query path itself. Perhaps we can go with something like `dataset_tenant_isolation_failure`?

aleks-p (approved):
LGTM

PR #4186, opened 2025-05-08
chpore: add v2 labels sanitization test

korniltsev:


kolesnikovae:
This problem involves three main aspects:
 1. Validation
 2. Relabeling
 3. Profile split

Those three aspects are tightly coupled:
 1. Relabeling must be performed in distributors because it may change the "service_name" label. It's used as the distribution key that must never change after we determined which instance, shard, and dataset to use.
 3. Relabeling rules must be applied before validation. Otherwise, e.g., labels with spaces will cause validation errors before we can handle it with relabeling. This is somewhat risky, but this is what we currently do; we may want to make validation more granular and validate limits, semantics, and syntax separately.
 2. Relabeling is currently done after profile split where we generate new "series labels" based on the pprof sample labels.
 4. Profile split must be performed in segment-writers. Otherwise, we duplicate symbolic information by distributing profiles over distinct blocks. In addition, it significantly complicates post-processing (e.g., profile export, or building a distribution/histogram/heatmap, etc. – some of the features we want to add in the future).
 5. Validation can be performed in both distributors and segment-writers. We _want_ to apply limits and validate user input in distributors due to many reasons.

Consider a scenario where the "service_name" label (in fact, any series label) is set based on a pprof sample label – it would force us to split the profile in distributors, because labels are part of the distribution key.

   We could probably tolerate changes in any label (it'd be _very_ hacky), but changing the "service_name" label is a no-go, because we use it to pick the dataset.

I considered the following options:
 1. Split profiles by "service_name" in distributors. It's not clear how to create series labels, especially after relabeling where we mix up series and pprof labels.
 2. Ignore relabeling rules that change the "service_name" label, but this is not a good idea, because we may actually need it.
 3. Ignore attempts to change the "service_name" label based on pprof labels. Too complex and confusing.
 4. Handle relabeling independently for series and pprof labels. Too confusing. Breaks some known use cases, e.g., when we want to relabel pprof labels based on series labels.

---

We can split profiles by "service_name" in distributors, and then split by all dimensions in segment writers. Series label set is the longest common prefix of pprof labels – this is what I'm currently working on.

aleks-p:
> Profile split must be performed in segment-writers. Otherwise, we duplicate symbolic information by distributing profiles over distinct blocks. In addition, it significantly complicates post-processing (e.g., profile export, or building a distribution/histogram/heatmap, etc. – some of the features we want to add in the future).

This to me seems to be the major contention point. We want to postpone splitting but that postpones the validation. Just curious, can we perform the split in distributors and still ensure profiles end up in the same block via other means (e.g., by sending them in a batch)?

kolesnikovae:
> This to me seems to be the major contention point. We want to postpone splitting but that postpones the validation. Just curious, can we perform the split in distributors and still ensure profiles end up in the same block via other means (e.g., by sending them in a batch)?

That also works, but it forces us to use the batch API, which was intentionally removed. I'm considering this as an alternative.

It has some complications. Before sending data to segment-writer we need to know what `service_name` and series labels are _after_ relabeling and validation. So we end up in the same situation as in:
> Split profiles by "service_name" in distributors. It's not clear how to create series labels, especially after relabeling where we mix up series and pprof labels.

But also adds some overhead and batch API.

korniltsev:
https://github.com/grafana/pyroscope/pull/4236

PR #4188, opened 2025-05-12
[profile-cli.md] Update link to recent profilecli release

rphibel:
The documentation for Linux propose to download version **1.1.5** of profilecli.
But version 1.1.5 does not support the `profile` subcommand. So I propose to update the link to a more recent version.


aleks-p:
looks like this needs an update from main before we can merge it

PR #4189, opened 2025-05-15
chore: update Java OTEL agent version in tracing example

aleks-p:
Update the OTEL Java agent to the latest version. This is a major version upgrade, but it only required a small change to keep tracing happy.


simonswine (approved):
LGTM

PR #4194, opened 2025-05-19
20250507 add v2 microservices integration test

simonswine:
- **chore: Add v2 support for microservices integration test**
- **Skip inconsistent query results with v2**
- **Add @aleks-p's suggestions**
- **Fix data races**
- **Reduce segment-writer heartbeats**
- **Move the lock up**
- **Check if all service names are available before starting the test**


pkg/experiment/metastore/compaction/scheduler/metrics.go
```diff
@@ -118,7 +115,10 @@ func (c *statsCollector) collectStats(fn func(level int, stats queueStats)) {
 }
 
 func (c *statsCollector) collectMetrics() []prometheus.Metric {
-	metrics := make([]prometheus.Metric, 0, 8*len(c.s.queue.levels))
+	c.s.mu.Lock()
+	defer c.s.mu.Unlock()
+
+	var metrics = make([]prometheus.Metric, 0, 8*len(c.s.queue.levels))
```
simonswine, line 121: The integration test revealed that lock problem here

alsoba13 (approved):

aleks-p (approved):
Looks good to me, thanks for adding this!

PR #4199, opened 2025-05-22
fix(v2): make the metastore index search more broad

aleks-p:
Currently when the metastore is asked for blocks within a time range, the range check is exclusive (for the most part). Shards and blocks that start / end exactly at the query boundaries are skipped. The impact is more noticeable in low-ingest-throughput environments.

This PR changes this behavior so that the metastore will return all shards and blocks at the query boundaries. The data within the blocks is still expected to be filtered by each read endpoint individually.

Additionally, `QuerierService.SelectSeries` in v1 adjusts the query start by the step. I've applied the same logic in v2 for consistency.

### DIFF
--- a/pkg/experiment/metastore/index/store/index_store.go
+++ b/pkg/experiment/metastore/index/store/index_store.go
@@ -293,7 +293,8 @@ func (s *Shard) Overlaps(start, end time.Time) bool {
 	if s.MinTime == 0 || s.MaxTime == 0 {
 		return true
 	}
-	return start.Before(time.UnixMilli(s.MaxTime)) && !end.Before(time.UnixMilli(s.MinTime))
+	// adjust by 1ms to make the check inclusive for start==s.MaxTime or end==s.MinTime
+	return start.Before(time.UnixMilli(s.MaxTime+1)) && !end.Before(time.UnixMilli(s.MinTime-1))
 }
 
 type stringIterator struct {
--- a/pkg/experiment/metastore/index/store/index_store.go
+++ b/pkg/experiment/metastore/index/store/index_store.go
@@ -293,8 +293,16 @@ func (s *Shard) Overlaps(start, end time.Time) bool {
 	if s.MinTime == 0 || s.MaxTime == 0 {
 		return true
 	}
-	// adjust by 1ms to make the check inclusive for start==s.MaxTime or end==s.MinTime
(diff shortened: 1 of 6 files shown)


pkg/experiment/metastore/index/store/index_store.go
```diff
@@ -293,7 +293,16 @@ func (s *Shard) Overlaps(start, end time.Time) bool {
...
+
+	if start.After(time.UnixMilli(s.MaxTime)) {
+		return false
+	}
+
+	if end.Before(time.UnixMilli(s.MinTime)) {
+		return false
+	}
+
+	return true
```
aleks-p, line 305: this allows for shards at the query boundaries to be visited

pkg/frontend/read_path/query_frontend/query_select_time_series.go
```diff
@@ -38,12 +39,15 @@ func (q *QueryFrontend) SelectSeries(
 		return connect.NewResponse(&querierv1.SelectSeriesResponse{}), nil
 	}
 
+	stepMs := time.Duration(c.Msg.Step * float64(time.Second)).Milliseconds()
+	start := c.Msg.Start - stepMs
```
aleks-p, line 43: this is done in querier.go for v1

https://github.com/grafana/pyroscope/blob/7514fb9c22c70112355bfe9c4443a19c237d914b/pkg/querier/querier.go#L982

alsoba13:
LGTM

alsoba13 (approved):

simonswine (approved):
LGTM

Thanks for fixing this.

pkg/experiment/metastore/index/store/index_store.go
kolesnikovae: Thanks for the fix Aleks! I believe the issue here was that the "range" was used instead of "query":
```go
// This is what it should have been:
// [queryStart:queryEnd)
rangeStart.Before(queryEnd) && !rangeEnd.Before(queryStart)
```
There was an issue in test fixtures, therefore it went unnoticed.

I think we should discuss internally what kind of query we want: `[start:end]` or `[start:end)`. The former results in some data beyond the query time range because profile timestamps point to the moment when the collection started. Not a big deal, but inaccurate.

NB:
 * In Parquet filtering, we currently have `[start:end]`.
 * In Partition handling we use `[start:end)`.

PR #4202, opened 2025-05-23
docs: Add instructions about profiling on macOS

simonswine:
Unfortunately SIP gets in the way of profiling system binaries on macOS using pyspy.

https://github.com/benfred/py-spy/issues/397#issuecomment-2627389627


tomglenn (approved):

tomglenn:
Thanks for this @simonswine

PR #4208, opened 2025-05-28
perf(v2): static parquet page buffer size

kolesnikovae:
Currently, parquet buffer sizes are determined based on the table size, with lower and upper bounds of 64KB and 1MB, respectively. This approach helps prevent high memory consumption in the segment-writer, compaction-worker, and query-backend services, which typically handle many small tables concurrently. While this strategy works well in the common case – keeping memory usage low – it can negatively impact query performance in certain scenarios.

In this PR, I'm increasing the buffers and making them static.


pkg/experiment/block/block.go
```diff
@@ -23,29 +23,12 @@ const (
...
-		return minSize
-	}
-	return int(min(e, maxSize))
-}
+const (
+	// Each 2MB translates to an I/O read op.
+	parquetReadBufferSize = 2 << 20
+	// Encoding and compression
+	parquetPageWriteBufferSize = 2 << 20
+)
```
kolesnikovae, line 31: It does make sense to increase the size further, e.g., 4MB. However, we need to estimate the actual impact on memory consumption.

pkg/experiment/block/block.go
```diff
@@ -23,29 +23,11 @@ const (
...
-	if e < minSize {
-		return minSize
-	}
-	return int(min(e, maxSize))
-}
+const (
+	// Each 2MB translates to an I/O read op.
+	parquetReadBufferSize      = 2 << 20
+	parquetPageWriteBufferSize = 1 << 20
+)
```
kolesnikovae, line 30: It does make sense to increase the `parquetReadBufferSize` size further, e.g., 4MB. However, we need to estimate the actual impact on memory consumption.

I'm less confident about `parquetPageWriteBufferSize`.

kolesnikovae:
I recently have discovered quite an interesting page access log:

image

25MB, ~40 page reads, 9/10 of fetched rows are discarded. For the given query (`{service_name="x", profile_type="y"}`), the access should be fully sequential, without gaps.

kolesnikovae:
While investigating the issue, I discovered that #4036 causes parquet reader to perform `SeekToRow` in an extremely inefficient way: on every call, it scans the entire column chunk up to the specified row from the very beginning (even if advancing by just one row); #4036 is fixed in #4209 – enabling the index significantly improves performance in pathological cases with high read amplification.

Regarding the previous comment on gaps in row ranges: this is expected due to the row ordering – first by timestamp, then by series ID. As a result, when the time range isn't fully covered, we must skip some rows for each series. Unfortunately, due to the way parquet-go is implemented, this almost always leads to a request to object storage. Changing the row order is a non-trivial task.

Another interesting finding: `parquet.List` of structs colocates fields on the same page. This means that we should never create an individual iterator for each of the columns in such cases ([example](https://github.com/grafana/pyroscope/blob/main/pkg/phlaredb/sample_merge.go#L120)): in fact, we fetch same pages repeatedly.

simonswine (approved):
LGTM! Thanks for tweaking the buffer sizes and the investigations/trialing around it.

I think we should make sure that we don't miss the ideas of that issue and create follow up tasks.

PR #4210, opened 2025-05-28
feat: dynamically named usage groups

aleks-p:
Adds a special meaning to usage groups with placeholders (`${labels.label_name}`) in their names. When encountered AND there is a match with the provided selector, we replace placeholders with the actual label values.

This allows for a simpler and more powerful configuration of usage groups where this:

```
    distributor_usage_groups:
      cart-service: '{service_name="cart-service"}'
      user-service: '{service_name="user-service"}'
      order-service: '{service_name="order-service"}'
```

can be replaced with:

```
    distributor_usage_groups:
      '${labels.service_name}': '{}'
```

Benchmark results:

```
go test -bench=BenchmarkUsageGroups -benchmem ./pkg/validation
goos: darwin
goarch: arm64
pkg: github.com/grafana/pyroscope/pkg/validation
cpu: Apple M2 Pro
BenchmarkUsageGroups_Regular-10            	12253587	        93.08 ns/op	      80 B/op	       1 allocs/op
BenchmarkUsageGroups_Dynamic-10            	 4551181	       265.0 ns/op	     200 B/op	       6 allocs/op
```

Notes:
- Both styles of usage groups can co-exist (though if there is overlap, attribution would count towards all groups)

### DIFF
--- a/pkg/validation/usage_groups.go
+++ b/pkg/validation/usage_groups.go
@@ -62,8 +62,8 @@ type templatePart struct {
 // usageGroupEntry represents a single usage group configuration
 type usageGroupEntry struct {
 	matchers []*labels.Matcher
-	// For static names, template is nil and staticName is used
-	staticName string
+	// For static names, template is nil and name is used
+	name string
 	// For dynamic names, template contains the parsed template parts
 	template []templatePart
 }
@@ -103,20 +103,18 @@ func (e *UsageGroupEvaluator) GetMatch(tenantID string, c *UsageGroupConfig, lbl
 					level.Warn(e.logger).Log(
 						"msg", "failed to expand usage group template, skipping usage group",
 						"err", err,
-						"labels", lbls,
-						"usage_group", entry.staticName)
+						"usage_group", entry.name)
 					continue
 				}
 				if dynamicName == "" {
 					level.Warn(e.logger).Log(
 						"msg", "usage group template expanded to empty string, skipping usage group",
-						"labels", lbls,
-						"usage_group", entry.staticName)
+						"usage_group", entry.name)
 					continue
 				}
 				match.names = append(match.names, dynamicName)
 			} else {
-				match.names = append(match.names, entry.staticName)
+				match.names = append(match.names, entry.name)
 			}
 		}
 	}
@@ -230,6 +228,7 @@ func parseUsageGroupEntries(m map[string]string) ([]usageGroupEntry, map[string]
 
 		entry := usageGroupEntry{
 			matchers: matchers,
+			name:     name,
 		}
 
 		if strings.Contains(name, dynamicLabelNamePrefix) {
@@ -238,8 +237,6 @@ func parseUsageGroupEntries(m map[string]string) ([]usageGroupEntry, map[string]
 				return nil, nil, fmt.Errorf("failed to parse template for usage group %q: %w", name, err)
 			}
 			entry.template = template
-		} else {
-			entry.staticName = name
 		}
 
 		entries = append(entries, entry)
--- a/pkg/validation/usage_groups.go
+++ b/pkg/validation/usage_groups.go
@@ -72,6 +72,11 @@ type UsageGroupConfig struct {
 	parsedEntries []usageGroupEntry
 }
 
+var (
+	dynamicLabelNamePrefix       = "${labels."
+	dynamicLabelNamePrefixLength = len(dynamicLabelNamePrefix)
+)
+
 func (c *UsageGroupConfig) GetUsageGroups(tenantID string, lbls phlaremodel.Labels) UsageGroupMatch {
 	match := UsageGroupMatch{
 		tenantID: tenantID,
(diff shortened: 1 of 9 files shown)


pkg/validation/usage_groups.go
```diff
@@ -61,8 +67,18 @@ func (c *UsageGroupConfig) GetUsageGroups(tenantID string, lbls phlaremodel.Labe
...
 	for name, matchers := range c.config {
-		if matchesAll(matchers, lbls) {
-			match.names = append(match.names, name)
+		matches, captureGroups := c.matchesAllWithCaptureGroups(matchers, lbls)
+		if matches {
+			if strings.Contains(name, "$") && len(captureGroups) > 0 {
+				dynamicName := name
+				for i, captureGroup := range captureGroups {
+					placeholder := fmt.Sprintf("$%d", i+1)
+					dynamicName = strings.ReplaceAll(dynamicName, placeholder, captureGroup)
```
simonswine, line 76: I guess that would stop working with `$1` and `$10`.

Also there is a bit of second level of replacement depending on the order: Imagine usage group: `$1-$2`

Theoretically the first capture group $1 could also contain a `haha-$2`, that then would be replaced later, but not if it would be `haha-$1`.

You probably need to tokenize it first. Then do the expanding in one go.

More nitpicking: Also probably if someone needs an explicit `$[0-9]` in the label, which has capture groups they might need an escape `$$12`.

Also as source for inspiration might be: https://cs.opensource.google/go/go/+/refs/tags/go1.24.3:src/regexp/regexp.go;l=926

simonswine (commented):
Not 100% sure myself yet if I like it. It obviously is pretty aligned with label selectors.

I also could think something more like `${label.service_name}` could be quite nice, but then that's something completely different to the label selector.

pkg/validation/usage_groups.go
```diff
@@ -61,8 +67,18 @@ func (c *UsageGroupConfig) GetUsageGroups(tenantID string, lbls phlaremodel.Labe
 	}
 
 	for name, matchers := range c.config {
-		if matchesAll(matchers, lbls) {
-			match.names = append(match.names, name)
+		matches, captureGroups := c.matchesAllWithCaptureGroups(matchers, lbls)
+		if matches {
+			if strings.Contains(name, "$") && len(captureGroups) > 0 {
+				dynamicName := name
```
simonswine, line 73: Wondering if new need to treat the empty string specially?

aleks-p:
Switched to named placeholders now. One added benefit is that we can now pre-parse the placeholders (once when the config is read) and can apply it more efficiently. The added overhead decreased quite a bit compared to using regex capture groups.

kolesnikovae:
Just in case: have we considered adding another internal/private label, like `__usage_group`, to be used in conjunction with relabeling rules?

This would:
 * Unify the different implementations
 * Allow users to explicitly specify usage groups
 * Simplify everything significantly

As I understand it, relabeling rules should be able to do everything that both the proposed and current implementations can do. In 99.(9)% of cases, it's just a simple mapping from `service_name`.

aleks-p:
> Just in case: have we considered adding another internal/private label, like `__usage_group`, to be used in conjunction with relabeling rules?
> 
> This would:
> 
> * Unify the different implementations
> * Allow users to explicitly specify usage groups
> * Simplify everything significantly
> 
> As I understand it, relabeling rules should be able to do everything that both the proposed and current implementations can do. In 99.(9)% of cases, it's just a simple mapping from `service_name`.

@kolesnikovae that is an interesting idea. It can work for simpler cases (like the `service_name` example) but usage groups can and often span multiple dimensions. E.g., a profile can contribute towards the usage statistics for a service, environment, team/department. A single label wouldn't work for this. We could bend the relabelling rules system to accommodate this, but ultimately it serves a different purpose.

pkg/validation/usage_groups_test.go
```diff
@@ -94,36 +88,54 @@ func TestUsageGroupConfig_GetUsageGroups(t *testing.T) {
...
+			},
+			Labels: phlaremodel.Labels{
+				{Name: "service_name", Value: "foo"},
+			},
+			Want: UsageGroupMatch{
+				tenantID: "tenant1",
+				names: []string{
+					"app/foo",
+				},
 			},
```
simonswine, line 128: I think we also should handle empty label value like missing label name (unlikely empty labels make makes it that far). Here roughly the test case:

```suggestion
			},
		},	
		{
			Name:     "dynamic_usage_group_names_missing_label",
			TenantID: "tenant1",
			Config: map[string]string{
				"app/${labels.service_name}/${labels.env}": `{service_name=~"(.*)"}`,
			},
			Labels: phlaremodel.Labels{
				{Name: "service_name", Value: "foo"},
			},
			Want: UsageGroupMatch{
				tenantID: "tenant1",
				names: []string{},
			},
		},	
		{
			Name:     "dynamic_usage_group_names_empty_label",
			TenantID: "tenant1",
			Config: map[string]string{
				"app/${labels.service_name}": `{service_name=~"(.*)"}`,
			},
			Labels: phlaremodel.Labels{
				{Name: "service_name", Value: ""},
			},
			Want: UsageGroupMatch{
				tenantID: "tenant1",
				names: []string{},
			},
		},				
```

Issue #4095, opened 2025-04-10
Support feature flags to propagate feature support from API backend to consumers


simonswine:
#### Is your feature request related to a problem? Please describe.

Currently it is hard for consumers of the Pyroscope APIs to figure out when a new feature has become available.

For example let's say we have a new feature, that adds utf8 support for label names.

Now in order for Drilldown Profiles and/or Alloy to know, that it can use utf-8 labels, it either needs to probe for the support or just trust that the other side has been updated in the same go.

#### Describe the solution you'd like

I would like Pyroscope to support feature flags, which are accessible for any authenticated Pyroscope API read/write client.

Right now I would like them to be fairly simple and straightforward and only depend on the Pyroscope backend version and it's configuration not necessarily be a thing that is dyanmic per tenant.

As it should be accessible with the same api context, I propose to introduce two calls to `querier.v1` and `push.v1`

```
rpc GetFeatureFlags({}) returns (types.v1.GetFeatureFlagsResponse) {}
````

with

```
message GetFeatureFlagsRespons {
  repeated string support_features = 1;
}

```

Features should be just strings like:

```
metricsFromProfiles
metricsFromProfilesFunctions
labelNamesUTF9
```

The Profiles Drilldown app should ideally only request the backendFeatureFlags once per datasource and then it knows how to treat the backend in the right way.

#### Describe alternatives you've considered

* We could continue to create a new API call, when something within fields change and probe for it
* For Drilldown we could configure the expected features in the plugin settings and/or data source config. E.g. Prometheus data source is doing that:

![Image](https://github.com/user-attachments/assets/f29355bd-0790-4aad-9026-7473ca9c6a3d)

#### Additional context

Talked about this with @ifrost in general and @alsoba13 and @bryanhuhta  with enhancing metricsFromProfiles specifically

**Comment by aleks-p** [comment]
I agree with the general idea!

Perhaps we want to make it slightly more extensible with something like this:

```
message GetFeatureFlagsRespons {
  repeated Feature features = 1;
}

message Feature {
  string name = 1;
  bool enabled = 2;
}
```

This could be helpful if a feature has some configuration attached to it.

ifrost:
Great idea 🚀  if we also added some meta data to the toggle (e.g. description, link to doc, etc.) we could nicely show in the UI what features are available and how to edit them.

PR #4211, opened 2025-05-29
feat: Add feature flags for read path

simonswine:
This implements #4095

Contrary to the initial plan, it provides a CapabilityService with FeatureFlag API targeting the read path only. It is exposed by QueryFrontends/Queriers.

The state of those feature flags is directly derived from the config and can't be changed during runtime. Those feature flags are exposed via metrics.


api/querier/v1/querier.proto
```diff
@@ -29,6 +29,9 @@ service QuerierService {
   // GetProfileStats returns profile stats for the current tenant.
   rpc GetProfileStats(types.v1.GetProfileStatsRequest) returns (types.v1.GetProfileStatsResponse) {}
   rpc AnalyzeQuery(AnalyzeQueryRequest) returns (AnalyzeQueryResponse) {}
+
+  // GetFeatureFlags returns the enabled backend feature flags for the current tenant
+  rpc GetFeatureFlags(types.v1.GetFeatureFlagsRequest) returns (types.v1.GetFeatureFlagsResponse) {}
```
kolesnikovae: I think we should have a dedicated service.

It's more convenient for both the client and the server. Otherwise, we would need to add the handler to every service, along with its implementation (so we also need to support this in the client). I've already counted 6 implementations, and we might also need handlers for VCS, ad-hoc, tenant settings, OTLP, and other services we may add later.

The API method feels rather out of place here – the querier service should be responsible for queries only.

As for scope (not sure if we really need it), it could just be a parameter in the request.

pkg/featureflags/featureflags.go
```diff
@@ -0,0 +1,181 @@
...
+
+func (f *FeatureFlags) HandlerForScope(scope Scope) Handler {
+	flags := make(map[string]*typesv1.FeatureParameters)
+	for name, flag := range f.flags {
+		if !slices.Contains(flag.apiScopes, scope) {
+			continue
+		}
+		if !flag.Parameters.Enabled {
+			continue
+		}
```
kolesnikovae: I'm not sure I understand the purpose of the `Enabled` flag. If we filter out disabled features, we probably don't need the `Enabled` attribute (at least at the API level)

pkg/featureflags/featureflags.go
```diff
@@ -0,0 +1,181 @@
...
+
+const (
+	ScopePusherV1  Scope = "pusherV1"  // this feature flag is relevant for the pusher.v1 API.
+	ScopeQuerierV1 Scope = "querierV1" // this feature flag is relevant for the querier.v1 API.
+)
+
+type featureDependency struct {
+	requires  []string // the feature cannot work without this other feature flag enabled.
+	conflicts []string // the feature cannot work with this other feature flag enabled.
+}
```
kolesnikovae: I'm wondering if this is necessary.

I think that feature flags (capabilities) should just reflect the _actual_ configuration in use. E.g., if the "ruler" is enabled in the config/overrides, it should be visible in "feature flags", regardless of the logic that allows enabling the ruler.

If a parameter (config/override) depends on others, it should not be possible to set it to "impossible" value – such a configuration would be invalid. For example, we should never enable "ruler" or "read path symbolication" if v2 is not enabled.

api/types/v1/types.proto
```diff
@@ -111,3 +111,20 @@ message GetProfileStatsResponse {
...
+
+// FeatureParameters contains the per feature flag parameters.
+message FeatureParameters {
+  // Wether the feature flag is enabled or disabled.
+  bool enabled = 1;
+  // Optional description of the feature flag.
+  optional string description = 2;
+  // Optional URL to the documentation page of the feature flag.
+  optional string documentation_url = 3;
+}
```
kolesnikovae: Given that the system's capabilities are well-defined, we could consider using a struct with optional fields instead of a map. The `map<string, FeatureParameters>` approach might be too generic, while the `FeatureParameters` struct itself might be too rigid. If we truly want to parameterize features, we won't be able to do it cleanly – introducing `any.Any` would likely be necessary.

kolesnikovae (commented):
I hope we'll get a chance to discuss the API – I'd seriously consider introducing a dedicated service rather than extending existing ones.

pkg/featureflags/featureflags.go
simonswine: I was mostly going down that direction, after thinking about how to handle per tenant overrides for feature flags and how some of them might depend on others and potentially are mutually exclusive. I don't think it makes sense to work on those now, so removed all of that.

api/types/v1/types.proto
simonswine: I decided against it, as I think it would require e.g. the UI always be update to date to show all the feature flags that are available, I would rather have that a little more flexible.

I also replaced the map with a classic repeated field.

Issue #3923, opened 2025-02-17
Flaky `Test_Concurrency` in the query backend client package


kolesnikovae:
As we use the real transport in tests, if fails from time to time:
 
```
=== FAIL: pkg/experiment/query_backend/client Test_Concurrency (0.01s)
    client_test.go:108: 
        	Error Trace:	/home/runner/work/pyroscope/pyroscope/pkg/experiment/query_backend/client/client_test.go:108
        	Error:      	Received unexpected error:
        	            	listen tcp 127.0.0.1:36467: bind: address already in use
        	Test:       	Test_Concurrency
```

https://github.com/grafana/pyroscope/actions/runs/13364982979/job/37321769113?pr=3885

The failure reasons vary (timouts, EOF, etc.) but the solution is the same – use in-memory transport for tests that involve networking.

callpraths:
Here's a few more tests with network dependency. Please consider having this issue tackle all of these / create a new issue for all of these. Whatever works best.

- [x] pkg/api/api_test.go:TestApiGzip - #4217
  - This doesn't seem to be testing anything within pyroscope. Probably best to delete it.
- [skip] pkg/pprof/merge_test.go:runEventsGatherer
  - This is actually a unix-specific file dependency (opening a file socket). Local filesystem access isn't as finicky as network access and I'm usually OK with unittests doing that if the alternative is too much mocking. Please let me know if we're OK with this or want to get rid of it.
- [x] pkg/scheduler/scheduler_test.go:* - #4233
- [x] pkg/util/gziphandler/gzip_test.go:TestGzipHandlerContentLength - #4234
- [x] pkg/experiment/metastore/client/client_test.go:* - #4235
  - via pkg/experiment/metastore/client/server_mock_test.go:createMockServers

kolesnikovae:
Thank you for you help with this @callpraths!

I do think `pkg/pprof/merge_test.go:runEventsGatherer` is OK. Moreover, I'm not sure if we still need it.

PR #4213, opened 2025-05-31
chore: use in-memory network connection to avoid query-backend concurrency test flake

callpraths:
Partial fix for #3923

### DIFF
--- a/pkg/experiment/query_backend/client/client.go
+++ b/pkg/experiment/query_backend/client/client.go
@@ -15,8 +15,8 @@ type Client struct {
 	grpcClient queryv1.QueryBackendServiceClient
 }
 
-func New(address string, grpcClientConfig grpcclient.Config) (*Client, error) {
-	conn, err := dial(address, grpcClientConfig)
+func New(address string, grpcClientConfig grpcclient.Config, dialOpts ...grpc.DialOption) (*Client, error) {
+	conn, err := dial(address, grpcClientConfig, dialOpts...)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


callpraths:
Hey folks.

Hoping to pick up a few simple bugfixes / cleanups of this variety to learn my way around the repo. There's a few more tests that can be updated in the same manner so I wouldn't close the issue just yet.

If this PR makes sense, I'll send a follow up migrating the remaining tests over in one go.

cheers,

pkg/experiment/query_backend/client/client.go
```diff
@@ -15,8 +15,8 @@ type Client struct {
 	grpcClient queryv1.QueryBackendServiceClient
 }
 
-func New(address string, grpcClientConfig grpcclient.Config) (*Client, error) {
-	conn, err := dial(address, grpcClientConfig)
+func New(address string, grpcClientConfig grpcclient.Config, dialOpts ...grpc.DialOption) (*Client, error) {
```
callpraths, line 18: Seems like the right thing to do to expose dial options.

e.g., `pkg/experiment/ingestor/client` does the same thing.

kolesnikovae: Yes, I believe this is the right thing to do.

Also, you may want to init `grpcclient.Config` with the default values (in the same way it's done in `ingester/client`):

```go
grpcClientConfig.RegisterFlags(flag.NewFlagSet("", flag.PanicOnError))
```

(this, however, could be a separate test case)

callpraths:
It makes sense to me to update the current test with the default flags, as it isn't intentionally exercising the set values (so those are magic values that look falsely significant). Updated.

Please help kick off CI.

kolesnikovae (approved):
LGTM

Thanks for taking care of this

PR #4216, opened 2025-06-01
fix(v2): compaction queue corruption

kolesnikovae:
This fixes an issue where compaction could stop completely.

Comments should provide exhaustive context:

1. We had a bug in time-based compaction. Specifically, we stopped tracking updates for a given compaction key. This is a rare but possible scenario that occurs when the oldest batch in a tenant-shard-level queue is removed from the priority queue used to track updates, while the queue itself is not removed. Batches added after this point won't be tracked. It was assumed that all batches would be processed by the time the oldest one is flushed, but this is not guaranteed in practice.

2. It can also happen that a compactor removes blocks from the staging batch, which was not expected. This is possible because the leader may decide to flush blocks that are still staged from the follower's perspective – for example, if time-based compaction has stopped on that follower.

3. When we remove blocks from a batch, we remove the batch from the queues once its size is reduced to zero.

4. There was a bug where attempting to remove a batch without a link to a previous batch (i.e., it appears to be the head) would replace the head with a link to the batch after the one being deleted.

Therefore, if we try to remove the staging batch from the queue, we end up replacing the head with nil.

---

Fortunately, this doesn't cause data-level inconsistency – it only renders the system inoperable. I've made some checks more offensive – it's better to fail fast than risk entering an infinite loop or blocking processing entirely.

### DIFF
--- a/pkg/experiment/metastore/compaction/compactor/compaction_queue.go
+++ b/pkg/experiment/metastore/compaction/compactor/compaction_queue.go
@@ -185,14 +185,8 @@ func (q *blockQueue) removeStaged(s *stagedBlocks) {
 		q.registerer.Unregister(s.collector)
 	}
 	delete(q.staged, s.key)
-	if s.heapIndex < 0 {
-		// We usually end up here since s has already been evicted
-		// from the priority queue via Pop due to its age.
-		return
-	}
-	if s.heapIndex >= q.updates.Len() {
-		// Should not be possible.
-		return
+	if s.heapIndex < 0 || s.heapIndex >= q.updates.Len() {
+		panic("bug: attempt to delete compaction queue with an invalid priority index")
 	}
 	heap.Remove(q.updates, s.heapIndex)
 }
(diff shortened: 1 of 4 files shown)


pkg/experiment/metastore/compaction/compactor/compaction_queue.go
```diff
@@ -248,21 +246,55 @@ func (s *stagedBlocks) delete(block string) blockEntry {
 	ref.batch.size--
 	s.stats.blocks.Add(-1)
 	if ref.batch.size == 0 {
-		s.queue.removeBatch(ref.batch)
+		// This is the last block in the batch, so we want to remove it
+		if ref.batch != s.batch {
```
kolesnikovae, line 249: This is the actual fix: we should never ever try to delete the staging batch from the queue – it hasn't been flushed and added to the queue yet.

pkg/experiment/metastore/compaction/compactor/plan_test.go
```diff
@@ -363,3 +365,51 @@ func TestPlan_time_split(t *testing.T) {
...
+			Level:  0,
+		}
+		c.enqueue(e)
+	}
+
+	p1 := &plan{compactor: c, blocks: newBlockIter()}
+	require.NotNil(t, p1.nextJob())
+	require.Nil(t, p1.nextJob())
+
+	// Add and blocks before they got to the compaction
```
simonswine:
```suggestion
	// Add extra blocks before they got to the compaction
```

PR #4217, opened 2025-06-01
chore: Delete redundant test with network dependency

callpraths:
Part of effort to drop network dependency from unittests as they cause test flakiness (See #3923).

This test file was originally added in https://github.com/grafana/pyroscope/commit/8c2988bf44ff91ab23406252125a9b4a5b1e4cc8 as a scaffolding test. The actual business logic was added to `pkg/util/gziphandler/gzip_go18_test.go` and this file was anticipated to hold package-boundary tests, that never came.

As it stands now, this test only exercises external gRPC packages and does not test anything in pyroscope.

But it potentially adds test flake because it opens network connections.


kolesnikovae (approved):
LGTM.

Good catch: `pkg/util/gziphandler` package tests cover all this. Thanks for fix!

callpraths:
Hmm, I wish CI would auto-trigger after a clean rebase, avoiding the maintainers having to push the button again...

aleks-p:
> Hmm, I wish CI would auto-trigger after a clean rebase, avoiding the maintainers having to push the button again...

Doesn't seem possible with the current setup

PR #4224, opened 2025-06-03
fix: Use tempdir for test instead ./testdata

simonswine:
This particular test creates the requires parquet files itself.


aleks-p (approved):
LGTM

PR #4226, opened 2025-06-04
Bump OTLP to `v1.7.0` and handle malformed OTLP profiles

### LINKED ISSUE #4220: Sanitize OTLP profiles
We should ensure that profiles which don't conform to the specification are handled gracefully, without causing data corruption or service failure.

Even though it's still an experimental feature in development, we should prioritize safety and a good user experience.

fandreuz:
Fixes #4220, and reintroduces the changes from #4212.

In addition to #4212, this has two more commits:
- 8814dcf521c625af529f5b37df78345eb3ac8814: graceful error handling for malformed OTLP data
- b8b03cdbe8cf16275881a3f65363c6fad07d19a7: new tests with malformed data

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -12,3 +12,50 @@
 
 # eBPF collector
 /ebpf/ @korniltsev
+
+# area/read-path
+/pkg/experiment/query_backend/ @marcsanmi @aleks-p
+/pkg/experiment/symbolizer/ @marcsanmi
+/pkg/storegateway/ @aleks-p
+/pkg/scheduler/
+/pkg/phlaredb/symdb/ @simonswine
+/pkg/phlaredb/tsdb/ @simonswine
+/pkg/querier/ @marcsanmi
+/pkg/frontend/ @marcsanmi
+/pkg/experiment/block/ @simonswine @marcsanmi @alsoba13
+/pkg/phlaredb/block @simonswine
+
+# area/write-path
+/pkg/distributor/ @aleks-p
+/pkg/experiment/distributor/
+/pkg/experiment/ingester/
+/pkg/og/convert/
+/pkg/pprof/
+/pkg/model/relabel/
+/pkg/model/
+/pkg/model/pprof_split/
+/pkg/validation/ @marcsanmi @aleks-p
+/pkg/ingester/otlp @marcsanmi
+/pkg/ingester/
+/pkg/phlaredb/shipper/
+
+# area/compactor
+/pkg/compactor/ @alsoba13
+/pkg/experiment/compactor/ @alsoba13
+/pkg/phlaredb/bucketindex/ @aleks-p
+/pkg/phlaredb/downsample/ @aleks-p
+
+# area/metastore
+/pkg/experiment/metastore/ @aleks-p
+
+# parquet
+/pkg/objstore/parquet/ @simonswine
+/pkg/phlaredb/query/ @simonswine
+/pkg/phlaredb/schemas/v1/ @simonswine @aleks-p
+/pkg/parquet/ @simonswine
+
+# api
+/api/ @simonswine @marcsanmi @aleks-p
+
+# objstore
+/pkg/objstore/ @simonswine
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -11,13 +11,11 @@ import (
 	"fmt"
 
 	"net/http"
-	"strings"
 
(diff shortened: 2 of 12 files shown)


kolesnikovae:
Thank you @fandreuz for adding this!

It looks very good overall, and I'm going to review it carefully tomorrow. @korniltsev, if you get a chance, it would be great if you could take a look as well

pkg/ingester/otlp/ingest_handler.go
```diff
@@ -80,12 +83,21 @@ func (h *ingestHandler) Export(ctx context.Context, er *pprofileotlp.ExportProfi
...
+	if dc == nil {
+		return &pprofileotlp.ExportProfilesServiceResponse{}, status.Errorf(codes.InvalidArgument, "missing profile metadata dictionary")
+	}
+
 	rps := er.ResourceProfiles
+	if rps == nil {
+		return &pprofileotlp.ExportProfilesServiceResponse{}, status.Errorf(codes.InvalidArgument, "missing resource profiles")
+	}
+
 	for i := 0; i < len(rps); i++ {
```
kolesnikovae, line 101: nit: I don't know what OTLP prescribes to do in this case, but returning an error seems reasonable. Also, it might be slightly better to check `len(rps) == 0` explicitly here, since `make([]T, 0, 0) != nil`: I'm just not 100% sure we can rely on the server proto codecs. Even though, if it's empty and non-nil, we just return an empty response without error, which is probably fine. Up to you.

pkg/ingester/otlp/ingest_handler.go
```diff
@@ -98,9 +110,10 @@ func (h *ingestHandler) Export(ctx context.Context, er *pprofileotlp.ExportProfi
 			for k := 0; k < len(sp.Profiles); k++ {
 				p := sp.Profiles[k]
 
-				pprofProfiles, err := ConvertOtelToGoogle(p)
+				pprofProfiles, err := ConvertOtelToGoogle(p, dc)
 				if err != nil {
-					return &pprofileotlp.ExportProfilesServiceResponse{}, fmt.Errorf("failed to convert otel profile: %w", err)
```
kolesnikovae, line 103: It would be really helpful if we logged the error here: that would allow us to see failure reasons on the server side

fandreuz: Sure: 6bb42104b646552d0669839b9404b80d2bd36a98

pkg/ingester/otlp/ingest_handler.go
fandreuz: I don't really have a strong preference here, both approaches make sense to me. I just feel that responding with an error might be a bit harsh, there's not really a problem to report. Empty requests are fine according to the spec. 

Let me know if you feel that any change is needed here though, happy to discuss it further.

kolesnikovae: Likewise, my understanding is that the intent here is to return an error on requests without RPs. I see the reasoning: if the profiler sends "empty" profiles, it might indicate a malfunction, and returning an error could help surface the issue earlier. From that standpoint, it seems reasonable.

That said, do we really need this? I'm not sure. If you don't have a strong opinion and the spec doesn't require it, let's keep things simple and remove the check altogether.

fandreuz: Sure: e5c0ea4a0e05bfb834e19aa56ae89da1551ecfb0

fandreuz:
Hi @kolesnikovae, I noticed Pyroscope enforces the presence of a mapping for each location. According to the spec, [that field is optional](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/profiles/v1development/profiles.proto#L437). Profiles not carrying it should probably be accepted without errors.

I'll fix the merge conflicts later today

simonswine (approved):
Thank you for your contribution! I have been running this against the latest otel-ebpf-profiler and works like a charm.


LGTM

simonswine:
> Hi @kolesnikovae, I noticed Pyroscope enforces the presence of a mapping for each location. According to the spec, [that field is optional](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/profiles/v1development/profiles.proto#L437). Profiles not carrying it should probably be accepted without errors.

@fandreuz This in fact will change with the next version, as very recently some PR marked this as no longer optional. But feel free to follow up with another PR

PR #4228, opened 2025-06-05
chore: Improve observability of the ingest handler

simonswine:


pkg/ingester/pyroscope/ingest_handler.go
```diff
@@ -123,13 +137,29 @@ func (h ingestHandler) ingestInputFromRequest(r *http.Request) (*ingestion.Inges
...
+		sp.SetTag("content_type", contentType)
+		sp.LogFields(
+			otlog.String("msg", "reading body from request"),
+		)
 	}
 
-	format := q.Get("format")
-	contentType := r.Header.Get("Content-Type")
+	buf := bytes.NewBuffer(make([]byte, 0, 64<<10))
+	if n, err := io.Copy(buf, r.Body); err != nil {
```
aleks-p, line 166: Is it possible for this to hang? If we get a clean error that the client cancelled the request, we are probably good. If we are just waiting we could end up with a gateway timeout instead. Not sure if it is worth to add a timeout here or if this is already handled elsewhere.

aleks-p (approved):
The added logging and spans should be useful, thanks for addressing this. I left one question / remark.

pkg/ingester/pyroscope/ingest_handler.go
```diff
@@ -43,14 +46,25 @@ func NewIngestHandler(l log.Logger, p ingestion.Ingester) http.Handler {
...
 		_ = h.log.Log("msg", "bad request", "err", err, "orgID", tenantID)
 		httputil.ErrorWithStatus(w, err, http.StatusBadRequest)
 		return
 	}
 
+	if err := readInputRawDataFromRequest(ctx, r, input); err != nil {
+		_ = h.log.Log("msg", "failed to read request body", "err", err, "orgID", tenantID)
+		httputil.ErrorWithStatus(w, err, http.StatusRequestTimeout)
+		return
+	}
```
kolesnikovae: I think it's fair to respond with 408 here, although it might not be correct in 100% of cases (server misbehaviour – such as 504 – should be reported to the client by the gateway/proxy)

kolesnikovae (approved):
Thanks for fixing this!

pkg/ingester/pyroscope/ingest_handler.go
simonswine: If it is possible to hang forever, I would think we would see errors elsewhere. I do think the [`ReadTimeout`](https://cs.opensource.google/go/go/+/refs/tags/go1.24.3:src/net/http/server.go;l=2976-2984) should also apply to those HTTPv2 multiplex requests, but I haven't tested this.

simonswine:
This is how the span looks like: 
image

PR #4231, opened 2025-06-06
feat: fine grained sampling

aleks-p:
Adds support for distributors to drop requests based on probabilistic sampling rules. 

The feature is tied to usage groups. Operators need to define usage groups first using the `distributor_usage_groups` tenant override and then add sampling rules via the `distributor_sampling` override. The sampling rules should target existing usage groups and provide a probability (0.0 means no profiles get through, 1.0 means all profiles are allowed).

The feature can make use of dynamically named usage groups (#4210). A valid example would be:

```
  anonymous: # tenant id
    distributor_usage_groups:
      service/${labels.service_name}: "{}"
    distributor_sampling:
      usage_groups:
        service/${labels.service_name}:
          probability: 0.5
        service/my-special-service:
          probability: 1.0
```

In this example, we drop 50% of all profiles but retain 100% of the profiles for a specific service.

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -8,6 +8,7 @@ import (
 	"flag"
 	"fmt"
 	"hash/fnv"
+	"math/rand"
 	"net/http"
 	"sort"
 	"sync"
@@ -39,6 +40,7 @@ import (
 	"github.com/grafana/pyroscope/pkg/distributor/aggregator"
 	"github.com/grafana/pyroscope/pkg/distributor/ingest_limits"
 	distributormodel "github.com/grafana/pyroscope/pkg/distributor/model"
+	"github.com/grafana/pyroscope/pkg/distributor/sampling"
 	writepath "github.com/grafana/pyroscope/pkg/distributor/write_path"
 	phlaremodel "github.com/grafana/pyroscope/pkg/model"
 	pprofsplit "github.com/grafana/pyroscope/pkg/model/pprof_split"
@@ -121,6 +123,7 @@ type Limits interface {
 	IngestionRateBytes(tenantID string) float64
 	IngestionBurstSizeBytes(tenantID string) int
 	IngestionLimit(tenantID string) *ingest_limits.Config
+	SamplingProbability(tenantID string) *sampling.Config
 	IngestionTenantShardSize(tenantID string) int
 	MaxLabelNameLength(tenantID string) int
 	MaxLabelValueLength(tenantID string) int
@@ -329,7 +332,20 @@ func (d *Distributor) PushParsed(ctx context.Context, req *distributormodel.Push
 
 		groups := d.usageGroupEvaluator.GetMatch(tenantID, usageGroups, series.Labels)
 		if err := d.checkUsageGroupsIngestLimit(tenantID, groups.Names(), req); err != nil {
+			groups.CountDiscardedBytes(string(validation.IngestLimitReached), req.TotalBytesUncompressed)
 			return nil, err
+		}
+
+		var shouldSample bool
+		if err, shouldSample = d.checkSampling(tenantID, groups.Names(), req); err != nil {
+			return nil, err
+		}
+		if !shouldSample {
+			level.Debug(d.logger).Log("msg", "skipping push request due to sampling", "tenant", tenantID)
+			validation.DiscardedProfiles.WithLabelValues(string(validation.SkippedBySamplingRules), tenantID).Add(float64(req.TotalProfiles))
+			validation.DiscardedBytes.WithLabelValues(string(validation.SkippedBySamplingRules), tenantID).Add(float64(req.TotalBytesUncompressed))
+			groups.CountDiscardedBytes(string(validation.SkippedBySamplingRules), req.TotalBytesUncompressed)
+			return connect.NewResponse(&pushv1.PushResponse{}), nil
 		}
 
 		profLanguage := d.GetProfileLanguage(series)
@@ -806,6 +822,36 @@ func (d *Distributor) checkUsageGroupsIngestLimit(tenantID string, groupsInReque
 	}
 
 	return nil
+}
+
+func (d *Distributor) checkSampling(tenantID string, groupsInRequest []string, req *distributormodel.PushRequest) (error, bool) {
+	l := d.limits.SamplingProbability(tenantID)
+	if l == nil {
+		return nil, true
+	}
+
+	// Determine the minimum probability among all matching usage groups.
+	minProb := 1.0
+	matched := false
+	for _, group := range groupsInRequest {
+		if probCfg, ok := l.UsageGroups[group]; ok {
+			matched = true
+			if probCfg.Probability < minProb {
+				minProb = probCfg.Probability
+			}
+		}
+	}
+
+	// If no sampling rules matched, accept the request.
+	if !matched {
+		return nil, true
+	}
+
+	// Sample once using the minimum probability.
+	if rand.Float64() <= minProb {
+		return nil, true
+	}
+	return nil, false
 }
 
 type profileTracker struct {
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -336,11 +336,7 @@ func (d *Distributor) PushParsed(ctx context.Context, req *distributormodel.Push
 			return nil, err
 		}
 
-		var shouldSample bool
-		if err, shouldSample = d.checkSampling(tenantID, groups.Names(), req); err != nil {
-			return nil, err
-		}
-		if !shouldSample {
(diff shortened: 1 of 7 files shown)


pkg/distributor/distributor.go
```diff
@@ -808,6 +820,36 @@ func (d *Distributor) checkUsageGroupsIngestLimit(tenantID string, groupsInReque
...
 }
 
+func (d *Distributor) shouldSample(tenantID string, groupsInRequest []string) bool {
+	l := d.limits.SamplingProbability(tenantID)
+	if l == nil {
+		return true
+	}
+
+	// Determine the minimum probability among all matching usage groups.
+	minProb := 1.0
```
simonswine, line 835: Wonder if it wouldn't be better with the `maximum` of those, you could have a very coarse background profiling `0.01` rule for all services and the profiles you are interested in (with service_name/pods selector) would be overriding them with out to 1.0.

simonswine (approved):
LGTM

pkg/distributor/distributor.go
aleks-p: This is up in the air.

The current system isn't able to apply a rule for all services. If we use `${labels.service_name}` for sampling (or ingest limits), the groups won't match as we'd compare the "resolved" name and the template. If we need this to work though, we can solve it easily by returning the source name for each "resolved" group.

Ultimately we might want a priority system because someone might want to have 0.5 on all services but less on an expensive one.

pkg/distributor/distributor.go
```diff
@@ -328,10 +334,22 @@ func (d *Distributor) PushParsed(ctx context.Context, req *distributormodel.Push
 		profName := phlaremodel.Labels(series.Labels).Get(ProfileName)
 
 		groups := d.usageGroupEvaluator.GetMatch(tenantID, usageGroups, series.Labels)
-		if err := d.checkUsageGroupsIngestLimit(tenantID, groups.Names(), req); err != nil {
+		if err := d.checkUsageGroupsIngestLimit(req, groups.Names()); err != nil {
+			level.Debug(d.logger).Log("msg", "rejecting push request due to usage group ingest limit", "tenant", tenantID)
+			validation.DiscardedProfiles.WithLabelValues(string(validation.IngestLimitReached), tenantID).Add(float64(req.TotalProfiles))
+			validation.DiscardedBytes.WithLabelValues(string(validation.IngestLimitReached), tenantID).Add(float64(req.TotalBytesUncompressed))
+			groups.CountDiscardedBytes(string(validation.IngestLimitReached), req.TotalBytesUncompressed)
 			return nil, err
```
kolesnikovae, line 342: Just a note – I'm not proposing to do anything in this PR.

This isn't specific to this particular spot, but rather a general issue:

```go
	groups.CountDiscardedBytes(string(validation.IngestLimitReached), req.TotalBytesUncompressed)
	return nil, err
```

We're reporting the entire request as discarded for the groups matching the series labels and rejecting the request. This isn't a major issue in the common case, since we assume there's only a single series (which is usually true), or that all series belong to the same set of groups. This is somewhat speculative, but it doesn't cause problems in 99.9% of cases.

However, this breaks with OTLP. In that case, we have a large system-wide profile that gets split, meaning the series might be completely unrelated. We should probably handle this on the caller side – in the OTLP handler – by pushing profiles for each service independently (i.e., building multiple push requests).

kolesnikovae (approved):

pkg/validation/usage_groups.go
```diff
@@ -89,29 +89,35 @@ func NewUsageGroupEvaluator(logger log.Logger) *UsageGroupEvaluator {
...
 						"usage_group", entry.name)
 					continue
 				}
-				if dynamicName == "" {
+				if resolvedName == "" {
 					level.Warn(e.logger).Log(
 						"msg", "usage group template expanded to empty string, skipping usage group",
 						"usage_group", entry.name)
 					continue
 				}
```
kolesnikovae, line 111: On a side note, these log messages are fairly noisy:

image

Do we absolutely need them?

kolesnikovae:
I understand that we're still exploring this area, and so far, it looks very promising! That said, after looking at the example, I have a few questions:

---

If I want to collect 100% of profiles of a specific pod when investigating, do I need to create a usage group?

```
  anonymous: # tenant id
    distributor_usage_groups:
      service/${labels.service_name}: "{}"
    distributor_sampling:
      usage_groups:
        service/${labels.service_name}:
          probability: 0.5
        service/my-special-service:
          probability: 1.0
```

From this example I understand that I don't have to create the `service/my-special-service` usage group (as we still would work in the context of the `service/${labels.service_name}` group).

But we still need a group to "catch" the pod – probably, something like this:

```
  anonymous: # tenant id
    distributor_usage_groups:
      service/${labels.service_name}: "{}"
      pod/my-special-pod: "{rule definition goes here}"
    distributor_sampling:
      usage_groups:
        service/${labels.service_name}:
          probability: 0.1
        service/my-special-service:
          probability: 0.5
        pod/my-special-pod:
          probability: 1.0
```

I think that it might be one of the most popular requests (based on the experience with adaptive signals).

---

For me it's not clear what probability should be selected (neither the `min` nor the `max` feels quite right, depending on the context). Without explicit priorities, it might be difficult to reason about a config like this:

```
  anonymous: # tenant id
    distributor_usage_groups:
      namespace/${labels.namespace}: "{}"
      service/${labels.service_name}: "{}"
      pod/my-special-pod: "{rule definition goes here}"
    distributor_sampling:
      usage_groups:
        namespace/${labels.namespace}:
          probability: 0.5
        service/my-special-service: # An override for a specific service – better match.
          probability: 0.1
        pod/my-special-pod: # An override for a specific pod of the my-special-service - even better match, but we can't know it.
          probability: 1.0
```

pkg/distributor/distributor.go
aleks-p: > We're reporting the entire request as discarded for the groups matching the series labels and rejecting the request. This isn't a major issue in the common case, since we assume there's only a single series (which is usually true), or that all series belong to the same set of groups. This is somewhat speculative, but it doesn't cause problems in 99.9% of cases.

We actually don't assume anything, we've added a validation rule saying "if any series in the request is part of a limited dataset, we reject the entire request". We return an error back and count the entire request as discarded bytes. Alternative solutions at this level will be hard to reason about. What do we respond with if we accept some of the series but reject others, with the current Push API?

> However, this breaks with OTLP. In that case, we have a large system-wide profile that gets split, meaning the series might be completely unrelated. We should probably handle this on the caller side – in the OTLP handler – by pushing profiles for each service independently (i.e., building multiple push requests).

Indeed. If we need to use some of these features on system-wide profiles they would have to be configured with system-wide labels, or the profiles would need to be split before reaching this point.

pkg/validation/usage_groups.go
aleks-p: This is a sign of misconfiguration or wrong assumptions about the data, which I think can be valuable information for an operator. I can convert it to a metric though, to reduce the noise.

Ideally we should have a way to control log levels for specific subsystems/packages/callers at runtime, to investigate issues when needed.

aleks-p:
@kolesnikovae 

Yes, a configuration like that will be non-deterministic with the current setup. We can solve it in a future iteration in a few different ways though, sorting groups by the number of matchers, an explicit priority field, order-based priority, etc.. I will create an issue for this.

pkg/distributor/distributor.go
kolesnikovae: > What do we respond with if we accept some of the series but reject others, with the current Push API? 

Handle as many profiles from the request as we can and report discarded data via logs, metrics, and response: we should extent Push API to include some diagnostics in the response. 200/OK if all processed, 400 otherwise.

OTLP [requires](https://github.com/open-telemetry/opentelemetry-proto/blob/68f9c6329ca91f96333ee5b240e305945b122b70/opentelemetry/proto/collector/profiles/v1development/profiles_service.proto#L45-L62) this.

@korniltsev as you're working on support for OTLP more than anyone else, I'd like you to be aware of this.

> We actually don't assume anything, we've added a validation rule saying "if any series in the request is part of a limited dataset, we reject the entire request". We return an error back and count the entire request as discarded bytes.

If we don't make such assumption, we presume that series in the request do not belong to the same group set (they might be completely independent). It means that we silently drop requests of groups that haven't exhausted their limits, not even reporting about discarded data attributed to them (but reporting more than needed against the matched groups) – I'm not sure the rule is fair; but I suspect it might be a source of confusion

pkg/validation/usage_groups.go
aleks-p: solved in the dev environment by configuring the label selector properly, `namespace/${labels.namespace}: {namespace!=""}`

PR #4232, opened 2025-06-06
feat(v2): recording rules of function names

alsoba13:
This PR approaches recording rules of functions.

There's one part I'd like to change (marked as _TODO_), but as it mostly involves code style, and is not compromising performance yet, I prefer to introduce this PR with a clearer presented core logic.

You may want to read this before jumping to reviewing the code, as this may help:

### Observer state

Data is organized and accessed sequentially this way:
<table><thead>
  <tr>
    <th>Tenant</th>
    <th>Dataset</th>
    <th>Series</th>
    <th>Rows/profiles</th>
  </tr></thead>
<tbody>
  <tr>
    <td rowspan="7">tenant1</td>
    <td rowspan="4">service_name1</td>
    <td rowspan="3">fingerprint1</td>
    <td>row1</td>
  </tr>
  <tr>
    <td>row2</td>
  </tr>
  <tr>
    <td>row3</td>
  </tr>
  <tr>
    <td>fingerprint2</td>
    <td>row4</td>
  </tr>
  <tr>
    <td rowspan="3">service_name2</td>
    <td>fingerprint3</td>
    <td>row5</td>
  </tr>
  <tr>
    <td rowspan="2">fingerprint4</td>
    <td>row6</td>
  </tr>
  <tr>
    <td>row7</td>
  </tr>
  <tr>
    <td>tenant2<br>...</td>
    <td>service_name3<br>...</td>
    <td>fingerprint5<br>...</td>
    <td>row8<br>...</td>
  </tr>
</tbody>
</table>

Observer only see single rows (last column), and need to deduce if a context switch has occurred.
There are 3 types of context/states/scopes:
* Tenant scope:
  * A tenant switch will cause a flush of the recorded metrics, and fetch and init new recording rules for the new tenant.
* Dataset scope:
  * Symbols information is scoped to every dataset. During the compaction process, symbols within a dataset are gathered from different blocks and rewritten to a new block. _Observe_ needs to detect dataset changes so symbolical information may be reset.
  * Tenant's rules are filtered and narrowed to those that target the specific dataset.
  * Dataset state holds lookup tables, that relate symbolical information to rules targeting those symbols.
* Series scope:
  * Every batch of rows with the same fingerprint may match some rules. On a series switch, every recording rule is evaluated to know which ones are relevant to the series.

### Observing symbols
We will try to avoid processing symbols that don't matter to any rule. With that purpose, row is *Observe*d before symbols. Observe will compute the state of the series so we know if the series matches some rule with a FunctionName.

Observer will completely ignore symbols when the current Series state has no FunctionName rules. In a later series, if there's one rule that matches and has a FunctionName, symbols will be observed and pointers symbol-to-rule will be computed. As we only want to process symbols _once_ for every new symbol rewritten, we will compute lookup tables of **all rules that contain FunctionNames**, not only the ones matching the current Series. The reason for this is that in a later series, a new rule with FunctionName may match and maybe all the symbols were already observed, and hence missed.

As a summary, requirements/properties we find:
* We want to Observe row before observing symbols.
* Symbols are only observed if we need to.
* The observed row has old symbols, so we need to observe again to get the rewritten ones. For this reason, I included a Flush function, that re-observes the symbol. I'd like to get rid of this, or in case we keep this 2-step observer, rename it (maybe we can have a Observe(row, state) function instead)


pkg/experiment/metrics/observer.go
```diff
@@ -56,60 +70,171 @@ func NewSampleObserver(recordingTime int64, exporter Exporter, ruler Ruler, labe
...
+				if !exists {
+					targetLocation = make(map[*recording]struct{})
+					o.state.targetLocations[uint32(o.state.seenLocations)] = targetLocation
+				}
+				for _, rec := range recs {
+					targetLocation[rec] = struct{}{}
+				}
+			}
+		}
+	}
```
alsoba13, line 200: One or more rules hit this series, and have symbols to look for.
Given this situation, we will look inside the rewritten locations slice for functions that matter to us. We target every function that matters in the dataset's rules, not only the ones matching this rule, as we won't have new opportunities to observe this slice again.

pkg/experiment/metrics/observer.go
```diff
@@ -56,60 +70,171 @@ func NewSampleObserver(recordingTime int64, exporter Exporter, ruler Ruler, labe
...
+					o.state.targetStacktraces[stacktraceIds[i]] = targetStacktrace
+				}
+				for rec := range recs {
+					targetStacktrace[rec] = struct{}{}
+				}
+			}
 		}
-		if rec.state.matches {
+	}
+}
```
alsoba13: At the end of this process we'll have a map stacktraceId -> matching rule. So later we can get stacktraces from the row and quicky lookup for matching rules

pkg/experiment/metrics/observer.go
```diff
@@ -56,60 +70,171 @@ func NewSampleObserver(recordingTime int64, exporter Exporter, ruler Ruler, labe
...
+
+func (o *SampleObserver) Flush(row block.ProfileEntry) {
+	for _, rec := range o.state.targetRecordings {
+		if rec.state.matches && rec.rule.FunctionName == "" {
 			rec.state.sample.Value += float64(row.Row.TotalValue())
 		}
 	}
```
alsoba13, line 227: Totals are computed as usual: for every rule that matches the series, we add the TotalValue

pkg/experiment/metrics/observer.go
alsoba13: Lookup tables values are sets of recording rules (avoid duplications on recursive functions is best to compute totals)

pkg/experiment/metrics/observer.go
```diff
@@ -56,60 +70,171 @@ func NewSampleObserver(recordingTime int64, exporter Exporter, ruler Ruler, labe
...
+		row.Row.ForStacktraceIdsAndValues(func(ids []parquet.Value, values []parquet.Value) {
+			for i, id := range ids {
+				for rec := range o.state.targetStacktraces[id.Uint32()] {
+					if rec.state.matches {
+						rec.state.sample.Value += float64(values[i].Int64())
+					}
+				}
+			}
+		})
+	}
```
alsoba13, line 239: On the other hand, functions are computed from the lookup tables only if the series hit some rule.

pkg/experiment/metrics/observer.go
```diff
@@ -119,24 +244,17 @@ func (o *SampleObserver) flush() {
...
 // initState compute labelsMap for quick lookups. Then check whether row matches the filters
-// if filter match, then labels to export are computed, and fetch/create the series where the value needs to be
+// if filters match, then labels to export are computed, and fetch/create the series where the value needs to be
 // aggregated. This state is hold for the following rows with the same fingerprint, so we can observe those faster
-func (r *recording) initState(fp model.Fingerprint, rowLabels phlaremodel.Labels, externalLabels labels.Labels, recordingTime int64) {
-	r.state.fp = fp
-	labelsMap := map[string]string{}
-	for _, label := range rowLabels {
-		labelsMap[label.Name] = label.Value
-	}
```
alsoba13, line 139: This was computed for every rule, but it belongs to the series, so we can compute it once in the caller

pkg/experiment/block/compaction.go
```diff
@@ -67,9 +67,13 @@ type compactionConfig struct {
 }
 
 type SampleObserver interface {
+	symdb.SymbolsObserver
+
 	// Observe is called before the compactor appends the entry
 	// to the output block. This method must not modify the entry.
 	Observe(ProfileEntry)
+	// TODO: Find an alternative to this
+	Flush(ProfileEntry)
```
alsoba13: Happy to find a way to avoid this. We want to observe row first, then symbols, then observe row again when symbols are rewritten. I tried some strategies but failed to make this interface simpler.

api/settings/v1/recording_rules.proto
```diff
@@ -79,6 +81,25 @@ message RecordingRule {
...
   // concurrent updates.
   int64 generation = 7;
+
+  optional StacktraceFilter stacktrace_filter = 8;
+}
+
+message StacktraceFilter {
+  optional StacktraceFilterFunctionName function_name = 1;
+
+  // StacktraceFilterFunctionNameRegExp functionNameRegExp = 2;
```
simonswine: Would remove those comments from here, to avoid confusing users.

api/settings/v1/recording_rules.proto
```diff
@@ -79,6 +81,25 @@ message RecordingRule {
...
+}
+
+enum MetricType {
+  TOTAL = 0;
+
+  // SELF = 1;
```
simonswine: Would remove those comments from here, to avoid confusing users.

pkg/experiment/block/compaction.go
simonswine: I think the most confusing I find the naming here:
- `Flush`, is used already in the exporter and also there is `flush()`.
- `Observe` is not really observing yet until `ObserveSymbols` or `Flush`.

Maybe naming it `Observe -> Prepare` and `Flush -> ObserveValues`.

In terms of interface, I am not too sure if I would prefer a callback here, but it could be a possibility. It feels weird though with `ObserveSymbols`:

```
	if m.observer != nil {
		observeValues := m.observer.Prepare(r)
		defer observeValues()
	}
```

pkg/experiment/metrics/observer.go
simonswine: This should be comment

pkg/experiment/metrics/observer.go
simonswine: This should be comment

pkg/experiment/metrics/observer.go
simonswine: This should be comment

pkg/phlare/phlare.go
```diff
@@ -593,6 +593,7 @@ func (f *Phlare) Run() error {
 					"service_git_ref":    serviceGitRef(),
 					"service_repository": "https://github.com/grafana/pyroscope",
 				},
+				//TenantID: "pyroscope-tenant",
```
simonswine: Clean me up

pkg/phlaredb/symdb/rewriter.go
```diff
@@ -15,12 +15,20 @@ type Rewriter struct {
 	symdb      *SymDB
 	source     SymbolsReader
 	partitions *lru.Cache[uint64, *partitionRewriter]
+	observer   SymbolsObserver
 }
 
-func NewRewriter(w *SymDB, r SymbolsReader) *Rewriter {
+type SymbolsObserver interface {
+	// ObserveSymbols is called once new symbols has been rewritten. This method must not modify the symbols.
```
simonswine: Worth noting that `Prepare()` should be called before

kolesnikovae (approved):
LGTM

Thanks Alberto for adding this! It looks very good. I think we can iterate over the API and performance improvements in the future

pkg/experiment/block/compaction.go
kolesnikovae: I like the idea of deferred observation. Something like:

```go
if observe := evaluate(entry); observe != nil {
    defer observe()
}
```

I'm curious how it performs on widely spread functions, e.g., Go `runtime.mallocgc`. What I had in mind is that we could resolve symbols once per aggregate (we'd need to aggregate all sample values first) or at dataset boundaries. That would also allow for slightly clearer contracts. I'm not proposing we implement this unless we encounter performance issues – just something that might be helpful in the future.

alsoba13:
Tests to cover observeSymbols will be added in a following PR

PR #4233, opened 2025-06-06
chore: Use in-memory connection for scheduler tests

callpraths:
... to avoid test flake from network dependency.

Partial fix for https://github.com/grafana/pyroscope/issues/3923
See https://github.com/grafana/pyroscope/pull/4213 for prior work.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -97,6 +97,10 @@ type Config struct {
 	QuerierForgetDelay      time.Duration             `yaml:"querier_forget_delay" category:"experimental"`
 	GRPCClientConfig        grpcclient.Config         `yaml:"grpc_client_config" doc:"description=This configures the gRPC client used to report errors back to the query-frontend."`
 	ServiceDiscovery        schedulerdiscovery.Config `yaml:",inline"`
+
+	// Dial options used to initiate outgoing gRPC connections.
+	// Intended to be used by tests to use in-memory network connections.
+	DialOpts []grpc.DialOption `yaml:"-"`
 }
 
 func (cfg *Config) RegisterFlags(f *flag.FlagSet, logger log.Logger) {
@@ -536,6 +540,7 @@ func (s *Scheduler) forwardErrorToFrontend(ctx context.Context, req *schedulerRe
 		return
 	}
 
+	opts = append(opts, s.cfg.DialOpts...)
 	conn, err := grpc.DialContext(ctx, req.frontendAddress, opts...)
 	if err != nil {
 		level.Warn(s.log).Log("msg", "failed to create gRPC connection to frontend to report error", "frontend", req.frontendAddress, "err", err, "requestErr", requestErr)
--- a/pkg/scheduler/scheduler_test.go
+++ b/pkg/scheduler/scheduler_test.go
@@ -12,7 +12,6 @@ import (
 	"net"
 	"net/http"
 	"net/http/httptest"
-	"net/url"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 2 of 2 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -76,6 +76,10 @@ type Scheduler struct {
 
 	schedulerpb.UnimplementedSchedulerForFrontendServer
 	schedulerpb.UnimplementedSchedulerForQuerierServer
+
+	// Dial options used to initiate outgoing gRPC connections.
+	// Intended to be used by tests to use in-memory network connections.
+	DialOpts []grpc.DialOption
```
callpraths: Unlike in https://github.com/grafana/pyroscope/pull/4213, here I chose to expose a public field instead of accepting one or more `DialOption...` in the constructor.

The scheduler is not a client, so it didn't make sense to have dial options as varargs. I also didn't want to force all production callsites having to pass in a nullish array for the options.

So I went with a public field that tests can set.

Please let me know if you'd like me to take a different approach.

kolesnikovae: Could we extend the `Config` type instead (omitting yaml and flag registration)?

callpraths: Yep, this makes sense to me.

I had shied away from this earlier because `setupScheduler` already accepted `handlerOpts` as varargs, so I couldn't add `dialOpts` there nicely.
But I think the `nil` argument being passed in to fill the positional argument before `handlerOpts` was smelly anyway.

I have now wrapped the arguments in a struct. PTAL.

kolesnikovae (approved):
LGTM

pkg/scheduler/scheduler.go
kolesnikovae: Sorry, our tool does not support this.

Now that I look at it: we probably should return the listener from from `setupScheduler`, then the caller will be able to create any clients it needs (instead of injecting options via server, which is rather complicated). WDYT?

pkg/scheduler/scheduler.go
```diff
@@ -97,6 +97,10 @@ type Config struct {
 	QuerierForgetDelay      time.Duration             `yaml:"querier_forget_delay" category:"experimental"`
 	GRPCClientConfig        grpcclient.Config         `yaml:"grpc_client_config" doc:"description=This configures the gRPC client used to report errors back to the query-frontend."`
 	ServiceDiscovery        schedulerdiscovery.Config `yaml:",inline"`
+
+	// Dial options used to initiate outgoing gRPC connections.
+	// Intended to be used by tests to use in-memory network connections.
+	DialOpts []grpc.DialOption
```
simonswine: This will prevent it from getting picked into the YAML representation
```suggestion
	DialOpts []grpc.DialOption `yaml:"-"`
```

simonswine (approved):
LGTM

pkg/scheduler/scheduler.go
callpraths: I didn't follow the last comment, sorry.

The reason I had to inject `DialOpts` into the `Scheduler` is because it creates a gRPC client connection on demand (currently in order to report an error back to the frontend).

Did you mean create the listener inside `setupScheduler` and still pass in the dial options to Scheduler as I do? The only difference there would be that the test doesn't create the listener separately, instead it receives a listener from `setupScheduler`. I prefer the current approach, where the test creates the in-memory listener, for two reasons:

- Most tests don't need this setup, so they skip it.
- Tests are more explicit in their setup when they do need an extra network mocking step.

Please let me know if I'm misunderstanding your comment.

pkg/scheduler/scheduler.go
callpraths: Thank you. Applied in latest patchset.

PR #4236, opened 2025-06-08
fix(v2): label sanitization

kolesnikovae:


pkg/pprof/pprof.go
```diff
@@ -1079,12 +1077,11 @@ var languageMatchers = map[string][]string{
 	"rust":   {"main.rs", "core.rs"},
 }
 
-func GetLanguage(profile *Profile, logger log.Logger) string {
+func GetLanguage(profile *Profile) string {
 	for _, symbol := range profile.StringTable {
 		for lang, matcherPatterns := range languageMatchers {
 			for _, pattern := range matcherPatterns {
 				if strings.HasPrefix(symbol, pattern) || strings.HasSuffix(symbol, pattern) {
-					level.Debug(logger).Log("msg", "found profile language", "lang", lang, "symbol", symbol)
```
kolesnikovae, line 1087: I hope this doesn't cause any objections

pkg/distributor/distributor.go
```diff
@@ -377,6 +382,15 @@ func (d *Distributor) PushParsed(ctx context.Context, req *distributormodel.Push
...
 	}
 
+	removeEmptySeries(req)
+	if len(req.Series) == 0 {
+		// TODO(kolesnikovae):
+		//   Normalization may cause all profiles and series to be empty.
+		//   We should report it as an error and account for discarded data.
+		//   The check should be done after ValidateProfile and normalization.
+		return connect.NewResponse(&pushv1.PushResponse{}), nil
+	}
```
kolesnikovae, line 392: A few lines above we do this:
```go
	if req.TotalProfiles == 0 {
		return nil, connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("no profiles received"))
	}
``` 
The intent is not quite clear: the counter never decreases, we should do it earlier.

In the meantime, normalization can easily discard all samples e.g., if they contain invalid references, or if they have no samples (or they are invalid i.e., negative because of the delta calculation), or if there's a bug. We should account for this.

I think it's reasonable to return an error in such cases: it's likely the profile is malformed.

pkg/distributor/distributor_test.go
```diff
@@ -1294,11 +1312,690 @@ func Test_SampleLabels(t *testing.T) {
...
+				for j, actualProfile := range actualSeries.Samples {
+					expectedProfile := expectedSeries.Samples[j]
+					assert.Equal(t, expectedProfile.Profile.Sample, actualProfile.Profile.Sample)
+				}
+			}
+		})
+	}
+}
+
+func Test_SampleLabels_SegmentWriter(t *testing.T) {
```
kolesnikovae, line 1341: A slightly adjusted copy of `Test_SampleLabels_Ingester`

pkg/distributor/distributor.go
```diff
@@ -885,54 +960,84 @@ func injectMappingVersions(series []*distributormodel.ProfileSeries) error {
...
+			req.DiscardedProfilesRelabeling += int64(visitor.discardedProfiles)
+			req.DiscardedBytesRelabeling += int64(visitor.discardedBytes)
+			if visitor.discardedBytes > 0 {
+				usageGroups.CountDiscardedBytes(string(validation.DroppedByRelabelRules), int64(visitor.discardedBytes))
 			}
-			result = append(result, v.series...)
-			bytesRelabelDropped += float64(v.discardedBytes)
-			profilesRelabelDropped += float64(v.discardedProfiles)
-			d.usageGroupEvaluator.GetMatch(req.TenantID, usage, series.Labels).
-				CountDiscardedBytes(string(validation.DroppedByRelabelRules), int64(v.discardedBytes))
```
kolesnikovae, line 913: Note that we use the _source_ series labels (before relabeling) to find usage groups, which may not what we want. We could move this to the series visitor where we visit all series; I leave it unchanged in the PR. (In the current version, we do it even before validation, so `usageGroupEvaluator` may see some trash.)

I wonder if we actually want to report discarded data per each group here. I can't find a case when we want to drop _samples_ via relabeling rules (dropping labels is fine) – that would make the profile invalid in almost all cases (all the supported cases).

I see multiple issues with how we report discarded data:
 * Both label and profile validation errors result in an early stop: we won't report discarded data for all the groups/profiles in the request once we encounter an invalid element. Moreover, we're reporting total profiles and bytes there, and not what actually belongs to the group.
 * We do not report data discarded due to ingest limits per group; we're reporting the request totals (see `checkUsageGroupsIngestLimit` – we may want to fix this).
 * We do not report data discarded due to rate limits per group.
 * We do not account for data discarded due to normalization at all.
 * In some places we report discarded without reporting received - this is probably fine, but hard to understand. Also, internally, we only care about `rate_limited` and `series_limit` reasons (not clear why we ignore invalid profiles, for example).
 * We do not account for requests we couldn't handle down the stream.

I believe we should carefully reassess the rules and their implementation. The API that allows specifying multiple series of profiles seems particularly prone to errors in this area: in some cases, it looks like we assume there's always just one series and one profile, which is not necessarily the case.

@aleks-p as you're working on this currently, I think it might be of interest to you.

pkg/model/labels.go
```diff
@@ -219,6 +219,52 @@ func (ls Labels) Delete(name string) Labels {
...
+			i++
+			j++
+		} else if cmp < 0 {
+			i++
+		} else {
+			j++
+		}
+	}
+	return ls[:k]
+}
```
kolesnikovae, line 266: This is one way to distinguish "sample labels" from "series labels": given a set of label sets, their intersection represents the "series labels".

This holds true in 99.9% of cases. If labels are set dynamically during program execution, some stack traces or samples should fall outside the scope where those labels can be applied, determining the intersection. In other cases – such as when we add static labels to all samples via the profiler – we treat the labels shared by all samples as "series labels".

This approach allows us to reconstruct the original pprof labels after a pprof split.

pkg/test/integration/ingest_pprof_test.go
```diff
@@ -319,6 +319,67 @@ func TestIngestPPROFFixPythonLinenumbers(t *testing.T) {
 	})
 }
 
+func TestIngestPPROFSanitizeOtelLabels(t *testing.T) {
```
kolesnikovae, line 322: Thank you @korniltsev for adding this!

pkg/test/integration/ingest_otlp_test.go
```diff
@@ -143,8 +146,13 @@ func TestIngestOTLP(t *testing.T) {
 
 					expectedBytes, err := os.ReadFile(metric.expectedJsonPath)
 					require.NoError(t, err)
+					var expected strprofile.CompactProfile
+					assert.NoError(t, json.Unmarshal(expectedBytes, &expected))
+					strprofile.SortProfileSamples(expected)
```
korniltsev, line 151: This sorting likely fixes the issue I saw with go 1.24 
https://github.com/grafana/pyroscope/actions/runs/14830299218/job/42561162180
Thanks :heart:

pkg/distributor/distributor.go
simonswine: > rate_limited and series_limit reasons (not clear why).

The idea behind that has been copied from Loki: Those two reasons are used for cost control for customers (at least therefore), we don't account them, while other validation reasons, we like to charge the user to incentivise them to fix the validation problem. This is maybe question when we (Grafana Labs), cause the problem and they rely on the fix. (e.g. alloy adding too many labels and causing them to send invalid labels.


> I can't find a case when we want to drop samples via relabeling rules (dropping labels is fine)

Yes agreed, that should not be a valid use case.

simonswine (approved):
LGTM. Thank you for cleaning this up. I have tested this locally with additional labels. 

The only problems I found are actually out side of this PR, will provide a separate fix for this.

In general this shows how problematic it is to throw together labels of different dimension, I do think we will iterate on this over time to maybe be able to distinguish between series and sample labels, eventually even more dimensions. This might be quite aligned to adoption OTLP attributes in-depth.

korniltsev (approved):
lgtm

pkg/distributor/distributor.go
aleks-p: Agreed, we should document this in an issue. Some of these are likely conscious decisions, some of them are bugs and things that we've overlooked.

pkg/distributor/distributor.go
kolesnikovae: I investigated it a little further: many Go mutex and block profiles are empty. Created a separate issue for that: https://github.com/grafana/pyroscope-squad/issues/468

Issue #4227, opened 2025-06-05
goroutine leak in  openIndexV12 and openTSDBIndex


liaol:
I identified a goroutine leak with  `github.com/minio/minio-go/v7.(*Client).GetObject.func1` when use S3 objstore 
The root cause is that the Close method is not called on the return value of bucket.Get in these two methods.

https://github.com/grafana/pyroscope/blob/81af7c65a7630ab92265e4dcc832de3418397f34/pkg/phlaredb/block_querier.go#L2045
https://github.com/grafana/pyroscope/blob/81af7c65a7630ab92265e4dcc832de3418397f34/pkg/phlaredb/symdb/block_reader.go#L259

If confirmed, I can submit a CR.

korniltsev:
Do you think it's possible to catch with https://github.com/uber-go/goleak ?

I have not looked into the code, but I believe any fixes are welcome!

liaol:
@korniltsev 

I created a simple unit test in `pkg/phlaredb/symdb`, `goleak` find a goroutine leak in `github.com/minio/minio-go/v7.(*Client).GetObject.func1`

Please replace your own s3 config and symdb file when run the test.


```
package symdb

import (
	"context"
	"testing"
	"time"

	"github.com/grafana/pyroscope/pkg/objstore/client"
	"github.com/grafana/pyroscope/pkg/objstore/providers/s3"
	"github.com/grafana/pyroscope/pkg/phlaredb/block"
	"go.uber.org/goleak"
)

func Test_Reader_Open_v3_leak(t *testing.T) {
	defer goleak.VerifyNone(t)
	cfg := client.Config{
		StoragePrefix: "your prefix",
		StorageBackendConfig: client.StorageBackendConfig{
			Backend: "s3",
			S3: s3.Config{
				SignatureVersion: s3.SignatureVersionV4,
				BucketName:       "your bucket",
				Region:           "your region",
				Endpoint:         "your endpoint",
				BucketLookupType: s3.AutoLookup,
			}}}
	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
	defer cancel()
	b, _ := client.NewBucket(ctx, cfg, "anonymous")
	testBlockMeta1 := &block.Meta{
		Files: []block.File{
			{RelPath: "anonymous/phlaredb/01JX2W31CCR1VE3C21ABAF60QF/symbols/index.symdb"},
		},
	}
	Open(context.Background(), b, testBlockMeta1)
	return
}

```

```
[...]

PR #4237, opened 2025-06-09
fix go routine leak

liaol:
@korniltsev 

fix https://github.com/grafana/pyroscope/issues/4227

### DIFF
--- a/pkg/phlaredb/block_querier.go
+++ b/pkg/phlaredb/block_querier.go
@@ -2046,7 +2046,9 @@ func (q *singleBlockQuerier) openTSDBIndex(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("opening index.tsdb file: %w", err)
 	}
-
+	defer func() {
+		_ = f.Close()
+	}()
 	var buf []byte
 	var tsdbIndexFile block.File
 	for _, mf := range q.meta.Files {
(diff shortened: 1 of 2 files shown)


kolesnikovae:
Thanks for fixing this! I found a few more candidates:
 * https://github.com/grafana/pyroscope/blob/main/pkg/adhocprofiles/adhocprofiles.go#L157
 * https://github.com/grafana/pyroscope/blob/main/pkg/settings/store/store.go#L204
 * https://github.com/grafana/pyroscope/blob/main/pkg/settings/bucket.go#L157
 * https://github.com/grafana/pyroscope/blob/main/pkg/storegateway/block.go#L43 <- this should be fine; pretty confusing, though – I'd say we should handle it explicitly

If you want, please feel free to include those in the PR

kolesnikovae (approved):

liaol:
> Thanks for fixing this! I found a few more candidates:
> 
> * https://github.com/grafana/pyroscope/blob/main/pkg/adhocprofiles/adhocprofiles.go#L157
> * https://github.com/grafana/pyroscope/blob/main/pkg/settings/store/store.go#L204
> * https://github.com/grafana/pyroscope/blob/main/pkg/settings/bucket.go#L157
> * https://github.com/grafana/pyroscope/blob/main/pkg/storegateway/block.go#L43 <- this should be fine; pretty confusing, though – I'd say we should handle it explicitly
> 
> If you want, please feel free to include those in the PR

@kolesnikovae  OK, I will send a new PR to fix those when available

PR #4238, opened 2025-06-09
Fix: Handle duplicate validation correctly when sanitizing

simonswine:
When replacing dots with underscores, we don't take care of the order of the labels slice, hence our duplicate check can be ineffective.

This change will fix this bug and also will handle the case when both the sanitized and unsanitized label values actually match and reduce them to the sanitized variant.

Discovered when reviewing #4236

### DIFF
--- a/pkg/validation/validate.go
+++ b/pkg/validation/validate.go
@@ -63,22 +63,23 @@ const (
 	// Those profiles were dropped because of relabeling rules
 	DroppedByRelabelRules Reason = "dropped_by_relabel_rules"
 
-	SeriesLimitErrorMsg                 = "Maximum active series limit exceeded (%d/%d), reduce the number of active streams (reduce labels or reduce label values), or contact your administrator to see if the limit can be increased"
-	MissingLabelsErrorMsg               = "error at least one label pair is required per profile"
-	InvalidLabelsErrorMsg               = "invalid labels '%s' with error: %s"
-	MaxLabelNamesPerSeriesErrorMsg      = "profile series '%s' has %d label names; limit %d"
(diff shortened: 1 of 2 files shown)


marcsanmi (approved):
Good catch, LGTM :thumbsup:

pkg/validation/validate.go
```diff
@@ -128,32 +129,91 @@ func ValidateLabels(limits LabelValidationLimits, tenantID string, ls []*typesv1
...
+			if err != nil {
+				return err
+			}
+			lastLabelName = ""
+			if idx > 0 && idx <= len(ls) {
+				lastLabelName = ls[idx-1].Name
+			}
+			continue
+		} else if !ok {
 			return NewErrorf(InvalidLabels, InvalidLabelsErrorMsg, phlaremodel.LabelPairsString(ls), "invalid label name '"+origName+"'")
```
kolesnikovae, line 157: We could also try finding duplicates outside the main loop:
1. Check limits
2. Check/sanitize lexis: character set, encoding, etc.
3. Check/sanitize semantics: what labels are specified (sorting is part of semantic sanitization)

For exmaple, we consider `service.name` invalid, as we check it before sanitization – this isn't a real-life problem, because we add a stub for that very specific label.

This is one of the issues I highlighted recently: we perform validation too late (but we have to): we may want to make it more granular

kolesnikovae (approved):

pkg/validation/validate.go
simonswine: Yes I do agree, I also think the we need to be rethinking the labels, at latest once we fully adopt the OTLP format. Created #4262 with this in mind.

PR #4239, opened 2025-06-09
fix: goroutine leak

liaol:
same as https://github.com/grafana/pyroscope/pull/4237

I did not change https://github.com/grafana/pyroscope/blob/main/pkg/storegateway/block.go#L43  since it works well 

@kolesnikovae please review this PR


kolesnikovae (approved):
Thanks for fixing this @liaol 🎉

PR #4244, opened 2025-06-11
chore: Refactor API handler registration

simonswine:
This change refactors how the API handlers are registered. I am doing
this as I am plannig to add an artificial delay middleware to both push
and /ingest endpoints.

Comparing the registration logs ideally there should have been no
mistake:

```
diff before-log.txt after-log.txt
19d18
< level=debug msg="api: registering route" methods=GET,POST,PUT,DELETE,HEAD,OPTIONS prefix=/api auth=false gzip=true
22a22
> level=debug msg="api: registering route" methods=all prefix=/api auth=false gzip=true
```

### DIFF
--- a/pkg/api/register_options.go
+++ b/pkg/api/register_options.go
@@ -1,0 +1,125 @@
+package api
+
+import (
+	"strings"
+
+	"github.com/grafana/dskit/middleware"
+
+	"github.com/grafana/pyroscope/pkg/util/gziphandler"
+)
+
+type registerMiddleware struct {
+	middleware.Interface
+	name string
+}
+
+type registerParams struct {
+	methods     []string
+	middlewares []registerMiddleware
+	isPrefix    bool
+}
+
+func (r *registerParams) logFields(path string) []interface{} {
+	gzip := false
+	auth := false
+	for _, m := range r.middlewares {
+		if m.name == "gzip" {
+			gzip = true
+		}
+		if m.name == "auth" {
+			auth = true
+		}
+	}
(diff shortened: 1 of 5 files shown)


pkg/api/register_options.go
```diff
@@ -0,0 +1,125 @@
...
+		WithGzipMiddleware(),
+		WithMethod("GET"),
+		WithPrefix(),
+	}
+}
+
+func (a *API) registerOptionsRingPage() []RegisterOption {
+	return append(
+		a.registerOptionsPublicAccess(),
+		WithMethod("POST"),
```
kolesnikovae, line 124: It looks like we also want to include `GET` – here's the handler:
```go
	if req.Method == http.MethodPost {
		ingesterID := req.FormValue("forget")
		// ... 
		return
	}

	ringDesc, err := h.r.getRing(req.Context())
	if err != nil {
		http.Error(w, err.Error(), http.StatusInternalServerError)
		return
	}

	// ...
```

kolesnikovae (approved):
LGTM

Thanks for refactoring this! It looks much cleaner now

Please check my comment about the ring handler

pkg/api/register_options.go
simonswine: It looks confusing but get is coming via the `a.registerOptionsPublicAccess()`, but I will make it explicit

kolesnikovae: Oh, right – I've missed that. Yeah, listing options explicitly may make things even better

PR #4248, opened 2025-06-12
chore: optimize metrics for discarded data in distributor

kolesnikovae:
Something I noticed while working on #4236: we unconditionally update the `DiscardedBytes` and `DiscardedProfiles` metrics after relabeling:
```go
validation.DiscardedBytes.WithLabelValues(string(validation.DroppedByRelabelRules), req.TenantID).Add(float64(req.DiscardedBytesRelabeling))
validation.DiscardedProfiles.WithLabelValues(string(validation.DroppedByRelabelRules), req.TenantID).Add(float64(req.DiscardedProfilesRelabeling))
```

It's not a major issue, but in large-scale deployments with hundreds of distributors and tenants, we generate many time series unnecessarily (it's very unlikely that we ever need to report anything in this particular place). Additionally, having to filter with `> 0` in each query is annoying


pkg/distributor/distributor.go
```diff
@@ -1075,10 +1080,7 @@ type sampleSeriesVisitor struct {
 }
 
 func (v *sampleSeriesVisitor) ValidateLabels(labels phlaremodel.Labels) error {
-	if err := validation.ValidateLabels(v.limits, v.tenantID, labels); err != nil {
-		return connect.NewError(connect.CodeInvalidArgument, err)
-	}
-	return nil
+	return validation.ValidateLabels(v.limits, v.tenantID, labels)
```
kolesnikovae, line 1083: It feels like it's the caller's responsibility to wrap the error

PR #4251, opened 2025-06-13
feat: Implement artificial ingestion delay.

simonswine:
This implements the ability to add artificial delay to the ingestion handlers.

This is useful during the v2 migration process in order to spot how client will behave 

Inspired by Mimir implementation:

- https://github.com/grafana/mimir/pull/10107
- https://github.com/grafana/mimir/pull/11265

This relies on #4249 

Once this is enabled, we also need to update dashboards to show the metrics without artificial delay

### DIFF
--- a/pkg/util/delayhandler/http_test.go
+++ b/pkg/util/delayhandler/http_test.go
@@ -2,7 +2,6 @@ package delayhandler
 
 import (
 	"context"
-	"errors"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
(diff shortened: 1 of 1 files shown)


pkg/util/delayhandler/http.go
```diff
@@ -0,0 +1,104 @@
...
+			ctx := r.Context()
+
+			delay := getDelay(ctx, limits)
+			var delayRw *delayedResponseWriter
+			if delay > 0 {
+				w, delayRw = wrapResponseWriter(w, start.Add(delay))
+			}
+
+			// now run the chain after me
+			h.ServeHTTP(w, r.WithContext(ctx))
```
kolesnikovae: I'm not sure why we replace the request context. If we haven't modified it, it should not be needed

kolesnikovae (approved):
LGTM

Thanks for adding this!

pkg/util/delayhandler/http.go
simonswine: Good spotted, I had a trace span created in some previous version

PR #4252, opened 2025-06-13
fix(v2): make segment upload hedging limiter respect context

kolesnikovae:
Previously, if the segment writer failed to upload to object storage, it could hit the hedge request rate limit and block clients ignoring the upload timeout context. This could lead to unbounded memory usage and potential OOM if object storage became unavailable. This change ensures the limiter respects the upload context.


kolesnikovae:
cc @alsoba13

alsoba13:
Sorry for the delay.

LGTM

PR #4253, opened 2025-06-14
feat: phlaredb parallel flushing

liaol:
In our high-traffic Pyroscope ingester, flushing a single head takes 20–30 seconds, and sequentially flushing all heads can take minutes. During this period, new heads arrive, causing memory usage to  increase, which keeps the GC busy and may even OOM.

Therefore, implementing parallel flushing is essential.

Changes:
1. Enable parallel flushing.
2. Fix the unused multierror.


liaol:
closed with wrong email

Issue #4133, opened 2025-04-23
How to improve the distributor's performance


liaol:
I am using Pyroscope with the [opentelemetry-ebpf-profiler](https://github.com/open-telemetry/opentelemetry-ebpf-profiler) to profile an entire host.

Here is my load test setup:
- **Distributor**: 16 cores, 32 GB RAM
- **Ingesters**: 8cores, 16GB RAM, SSD disk，3 instances
- **Method**: Uses the `/opentelemetry.proto.collector.profiles.v1development.ProfilesService/Export` gRPC endpoint
- **Workload**: Sends 8,000 samples (4 MB after gzip compression) every 5 seconds
- **Client Configuration**: Write timeout set to 5 seconds

The distributor's performance is suboptimal, handling only 7 requests per second, indicating a bottleneck. The ingesters, however, are performing well.

How can I optimize the distributor's performance to handle higher throughput? Thanks~

Here is my config

```
storage:
    backend: "s3"

distributor:
    ring:
        kvstore:
            store: memberlist
            prefix: collectors/

ingester:
    lifecycler:
        ring:
            kvstore:
                store: memberlist
                prefix: collectors/

limits:
    max_profile_size_bytes: 33554432
    ingestion_rate_mb: 1024000
    max_global_series_per_tenant: 0
    max_profile_stacktrace_depth: 0
    compactor_blocks_retention_period: 48h

```

korniltsev:
Do you have a pprof from the suffering distributor?

liaol:
> Do you have a pprof from the suffering distributor?

@korniltsev 

cpu usage: 10 of 16cores   [cpu profile](https://flamegraph.com/share/ea53ff0f-202d-11f0-8b23-2a8f4480f552)  

memory usage: 18GB rss before GC [heap profile ](https://flamegraph.com/share/f8b24c07-202d-11f0-8b23-2a8f4480f552)

liaol:
The distributor processes a request with 8,000 samples in 5 seconds, with the following breakdown:
1. **Read request & unmarshal**: 1 second
2. **Convert Otel to Google format**: 1 second
3. **Push parsed data**: 3 seconds
   - **Normalize**: 1 second
   - **Send to ingester**: 2 seconds

liaol:
I modified the setup to send pprof data directly to the /ingester endpoint, bypassing the need for conversion, and it works seamlessly.

Issue #2563, opened 2023-10-20
Flush queue


cyriltovena:
For large multitenant cluster we're at risk many blocks will be flushed at once, to avoid this right now we have a jitter per tenant. But ultimately we should have a global flush queue that limit the max concurrent flushes across all tenants.

PR #4254, opened 2025-06-14
feat: phlaredb parallel flushing

liaol:
In our high-traffic Pyroscope ingester, flushing a single head takes 20–30 seconds, and sequentially flushing all heads can take minutes. During this period, new heads arrive, causing memory usage to increase, which keeps the GC busy and may even OOM.

Therefore, implementing parallel flushing is essential.

Changes:

Enable parallel flushing.
Fix the unused multierror.

### DIFF
--- a/pkg/phlaredb/phlaredb.go
+++ b/pkg/phlaredb/phlaredb.go
@@ -233,16 +233,25 @@ func (f *PhlareDB) Flush(ctx context.Context, force bool, reason string) (err er
 	// lock is release flushing heads are available for queries in the flushing array.
 	// New heads can be created and written to while the flushing heads are being flushed.
 	errs := multierror.New()
-
+	successful := make([]*Head, 0, len(f.flushing))
+	var mu sync.Mutex
+	var wg sync.WaitGroup
 	// flush all heads and keep only successful ones
-	successful := lo.Filter(f.flushing, func(h *Head, index int) bool {
-		f.metrics.flushedBlocksReasons.WithLabelValues(reason).Inc()
-		if err := h.Flush(ctx); err != nil {
-			errs.Add(err)
-			return false
-		}
-		return true
-	})
+	for _, h := range f.flushing {
+		wg.Add(1)
+		go func(head *Head) {
+			defer wg.Done()
+			f.metrics.flushedBlocksReasons.WithLabelValues(reason).Inc()
+			if err := head.Flush(ctx); err != nil {
+				errs.Add(err)
+				return
+			}
+			mu.Lock()
+			successful = append(successful, head)
+			mu.Unlock()
+		}(h)
+	}
+	wg.Wait()
 
 	// At this point we ensure that the data has been flushed on disk.
 	// Now we need to make it "visible" to queries, and close the old
@@ -265,7 +274,7 @@ func (f *PhlareDB) Flush(ctx context.Context, force bool, reason string) (err er
 	}
 	f.flushing = nil
 	f.headLock.Unlock()
-	return err
+	return errs.Err()
 }
 
 func (f *PhlareDB) maxBlockDuration() time.Duration {


pkg/phlaredb/phlaredb.go
```diff
@@ -233,16 +233,25 @@ func (f *PhlareDB) Flush(ctx context.Context, force bool, reason string) (err er
...
-		}
-		return true
-	})
+	for _, h := range f.flushing {
+		wg.Add(1)
+		go func(head *Head) {
+			defer wg.Done()
+			f.metrics.flushedBlocksReasons.WithLabelValues(reason).Inc()
+			if err := head.Flush(ctx); err != nil {
+				errs.Add(err)
```
kolesnikovae, line 246: `errs` is not thread-safe, this is a potential race (quite harmless, but still)

pkg/phlaredb/phlaredb.go
```diff
@@ -233,16 +233,25 @@ func (f *PhlareDB) Flush(ctx context.Context, force bool, reason string) (err er
 	// lock is release flushing heads are available for queries in the flushing array.
 	// New heads can be created and written to while the flushing heads are being flushed.
 	errs := multierror.New()
-
+	successful := make([]*Head, 0, len(f.flushing))
+	var mu sync.Mutex
+	var wg sync.WaitGroup
```
kolesnikovae, line 238: If we _really_ want to flush heads concurrently, let's put a bound on it. Ideally, there should be a queue and a worker pool of configurable size. Please see #2563.

As a quick solution, we could use `errgroup` for that (yeah, it's not exactly what it's meant for):

```go
	g, ctx := errgroup.WithContext(context.Background())
	if limit == 0 {
		limit = 1
	}
	g.SetLimit(limit)
	for head := range heads {
		g.Go(func() error {
			// As there's no case where we want to cancel
			// all flushes, we always return.
			if err := head.Flush(ctx); err != nil {
				// ...
				return nil
			}
			// ...
			return nil
		})
	}
```

pkg/phlaredb/phlaredb.go
```diff
@@ -265,7 +274,7 @@ func (f *PhlareDB) Flush(ctx context.Context, force bool, reason string) (err er
 	}
 	f.flushing = nil
 	f.headLock.Unlock()
-	return err
+	return errs.Err()
```
kolesnikovae, line 277: Oh, thanks for fixing this! We never rely on it in the reality – it would be nice to add at least logging on the caller side

kolesnikovae (commented):
Thank you @liaol for the contribution!

> In our high-traffic Pyroscope ingester, flushing a single head takes 20–30 seconds, and sequentially flushing all heads can take minutes. During this period, new heads arrive, causing memory usage to increase, which keeps the GC busy and may even OOM.

I'd suggest scaling ingesters (and the rest of the system) horizontally – Pyroscope is designed for this. If you provide me with more details, I'd be happy to help with capacity planning.

> Therefore, implementing parallel flushing is essential.

I'm afraid that flushing heads concurrently will very likely make things worse.

A head flush is a very memory- and disk-intensive operation, and a 20-40% spike in memory usage is expected – we're trading disk IO and CPU for memory. If we flush heads concurrently, memory usage will increase: in the synchronous version, we need X memory to flush things; in the concurrent – X*n.

New and old heads co-exist besides of the flush process. Even if we decide to flush all heads at once, new ones are very likely have already been created – those are allocated on demand, based on the data timestamp (blocks are aligned by max_block_duration). In addition, the age-based flush has quite a significant jitter (`util.DurationWithJitter(10*time.Minute, 0.5)`) to account for out-of-order requests and to prevent flushing all heads simultaneously, since heads are flushed concurrently _across tenants_ (`PhlareDB` instances). In other words. What I think we must do is to ensure that flushed heads are not referenced in memory longer than needed. However, it's not that easy and will likely change the entire flow (e.g., employ queueing).

Lastly, at any given time, there should be very few heads being flushed (within a single `PhlareDB` instance/tenant) – typically around three: the current block time window, the next one, and the previous one. However, this depends on the configuration, mainly the ingestion window duration and the maximum block duration. If there are many heads to be flushed for a tenant, it's likely a bug or misconfiguration.

---

That said, I'm not against the change, but I want to give you a heads-up.

If we add this, let's make sure the change is backward compatible: existing deployments should continue to flush heads synchronously.

liaol:
Thank you for the detailed explanation. @kolesnikovae 

We have already scaled ingesters horizontally.

> Lastly, at any given time, there should be very few heads being flushed (within a single PhlareDB instance/tenant) – typically around three: the current block time window, the next one, and the previous one. However, this depends on the configuration, mainly the ingestion window duration and the maximum block duration. If there are many heads to be flushed for a tenant, it's likely a bug or misconfiguration.

In our scenario, the number of flushing heads is 7～8.
```
ts=2025-06-14T22:25:15.68113863+08:00 caller=phlaredb.go:230 level=debug tenant=anonymous msg="flushing heads to disk" reason=max-block-bytes max_size="3.2 GB" current_size="3.3 GB" num_heads=7
ts=2025-06-14T22:31:05.68110972+08:00 caller=phlaredb.go:230 level=debug tenant=anonymous msg="flushing heads to disk" reason=max-block-bytes max_size="3.2 GB" current_size="3.3 GB" num_heads=8
```

The configuration is as follows:
```
pyroscopedb:
  max_block_duration: 1m

hard code defaultParquetConfig.MaxBlockBytes = 3.2GB
```

I think the issue stems from a mismatch between `staleHeadTicker := time.NewTimer(util.DurationWithJitter(10*time.Minute, 0.5))` and `max_block_duration = 1m`, which results in more than three flushing heads(1 minute 1 head).

I plan to test adjusting max_block_duration to 10 minutes to address this.

kolesnikovae:
Hello @liaol,

Indeed, the configuration you shared is quite different from what we had in mind when designing Pyroscope :) Our current solution is built around a few key principles – primarily, keeping larger blocks in memory for longer periods of time, which helps reduce both write and read amplification.

By default, `max_block_duration` is set to 1h, and many components rely on that assumption. From our experience, things tend to stop working as intended when it's reduced to around 15m, and even a heavily tuned configuration may struggle to compensate for that.

Could you share more details about your deployment? Understanding the challenges you're facing at your scale would help us improve Pyroscope.

---

```
ts=2025-06-14T22:25:15.68113863+08:00 caller=phlaredb.go:230 level=debug tenant=anonymous msg="flushing heads to disk" reason=max-block-bytes max_size="3.2 GB" current_size="3.3 GB" num_heads=7
ts=2025-06-14T22:31:05.68110972+08:00 caller=phlaredb.go:230 level=debug tenant=anonymous msg="flushing heads to disk" reason=max-block-bytes max_size="3.2 GB" current_size="3.3 GB" num_heads=8
```

You may benefit from reverting `max_block_duration` to something more substantial – e.g., `1h`. With that, blocks will still flush based on size, but you'll avoid the overhead of managing too many concurrent heads. In practice, this depends on your ingestion window duration and out-of-order ingestion. It's _very_ expensive when data arrives after the corresponding block (based on the profile's timestamp) has already been flushed.

Also, ingesters scale very well – many our clusters have well over 100 nodes – and this is the preferred way to operate them in large-scale deployments.

---

We're also actively working on a new storage backend: a disk-less mode that writes data directly to object storage. This shifts the responsibility for durability to the object store provider and eliminates the need for expensive local disks (assuming a suitable replication factor). I assume your `max_block_duration = 1m` setting may be trying to achieve something similar.

You can find more here: https://github.com/grafana/pyroscope/tree/main/pkg/experiment. We've already tested it under heavy load (0.5-1GB/s, 10–20k RPS) and are happy with both performance and cost. We're planning to release it later this year.

Please let me know if you're interested – I'd be happy to answer any questions or help you evaluate it in your infrastructure.

liaol:
Thanks, @kolesnikovae 

Setting `max_block_duration` to 1h works like a charm. It also reduces S3 storage usage.

> I assume your max_block_duration = 1m setting may be trying to achieve something similar.

No, it's just a mistake.

Here are more details about our scenario:

* We use [opentelemetry-ebpf-profiler](https://github.com/open-telemetry/opentelemetry-ebpf-profiler) to collect profiles from all processes on our 256-core machines. The ebpf agent sends more than 10000 samples (approximately 4 MB after gzip compression) to Pyroscope every 5 seconds. Profiles are aggregated by service.name and routed to ingesters using consistent hashing.
* Due to performance limitations with the distributor, we send profiles directly to the ingester’s ingester.v1.IngesterService/Push endpoint. See https://github.com/grafana/pyroscope/issues/4133 for context.
* In load tests, a single 16-core, 64GB ingester pod can handle profiles from approximately 500 machines, equating to 100 QPS.

Additionally, I’ve added a compression option for Parquet in Pyroscope, which reduces disk or S3 storage usage by 50%. I’d be happy to submit a PR if contributions are welcome.

Thanks again!

kolesnikovae:
Hi @liaol!

> We use [opentelemetry-ebpf-profiler](https://github.com/open-telemetry/opentelemetry-ebpf-profiler) to collect profiles from all processes on our 256-core machines. The ebpf agent sends more than 10000 samples (approximately 4 MB after gzip compression) to Pyroscope every 5 seconds. Profiles are aggregated by service.name and routed to ingesters using consistent hashing.

A word of caution: sending profiles too frequently creates unnecessary load. It's better to send 15-30s profiles: many stack traces will get aggregated and the overall overhead will be significantly lower

> Due to performance limitations with the distributor, we send profiles directly to the ingester’s ingester.v1.IngesterService/Push endpoint. See https://github.com/grafana/pyroscope/issues/4133 for context.

Please note that without normalization and sanitization performed in distributors it's very likely that the stored profiles are malformed. It will cause compaction (if you use it) and query failures.

I'd recommend deploying more smaller instances of distributor: e.g., 32 2 CPU, 2 GB RAM. This is how we run Pyroscope ourselves.

>  Additionally, I’ve added a compression option for Parquet in Pyroscope, which reduces disk or S3 storage usage by 50%. I’d be happy to submit a PR if contributions are welcome.

This is super cool! Yes, PRs are always welcome! We haven't enabled compression because of the CPU costs (both in read and write paths). I'm wondering what algo you're using and how bad the CPU cost is. I think something like snappy, lz4, or zstd might be affordable.

Issue #3587, opened 2024-09-24
Time-series aggregation step in diff / comparison queries


kolesnikovae:
The time series always uses sum aggregation, and the aggregation window depends on the time range.

For example, when you query the last hour, each point represents the sum of values within a 5-second window. However, when you query the last 12 hours, the window increases to 30 seconds, so each point covers a larger time span and aggregates more data. This results in higher sums for each point, which can be misleading if you're expecting consistent, time-normalized values across different time ranges.

That said, I find this behavior quite confusing. I believe users expect to see series where the values are normalized across different time windows. This inconsistency feels like a bug, and I think we should fix it to provide more intuitive results.

In the screenshot below you can see that in the diff view, the difference is even more pronounced and confusing:

![Image](https://github.com/user-attachments/assets/45716c9b-b083-43c5-aa45-df6900a3a68c)

simonswine:
I do think this has to do how we treat the timeline and how we've chosen the wrong unit for cumulative profile types. So we need to address this slightly different per type.

I guess we have roughly two types of profiles:

- Cumulative:
  -  cpu / memory_allocation / block / mutex
  - Collected over a duration (represented in this [pprof field](https://github.com/google/pprof/blob/a4b03ec1a45e1b7f16c441907485dd0efb33cccb/proto/profile.proto#L80-L81).
- Instant
  - goroutines / memory_in_use
  - Collect as point in time snapshot.

So for cumulative profiles the unit we display is seconds or bytes, so we take the total sum of the profile and display it but this is not correct, it should be a rate instead:

Let's say we have serviceA 15s, that has a scrape interval of 15s with cpu usage of 15s and serviceB that has a scrape interval of 30s with cpu usage of 30s, both there fore using the exact same cpu usage. But we would display it in the way that serviceB uses double the CPU. So we need to normalize by dividing total value / duration. In this case and the unit becomes seconds per second for CPU/wait or bytes/s for allocaiton. This is a lot clearer and will stay consistent over step changes.

If we have 3 pods for a service, once we build the normalized form, we can sum the up all the values in the interval and they return the correct aggregate.

For instance profile types, we currently show the correct unit (e.g. count of goroutines, bytes of inuse memory), but we need to treat the aggregation differently:

When we have 3 pods in a service, we first need to build the avg over time per series, (so one each per pod) and then we can sum those timelines up.

I do think we should work on this soon, @marcsanmi has shown some interest last week, let's keep this conversation going.


[...]

Issue #4192, opened 2025-05-19
V2: Support DurationNanos


simonswine:
Currently DurationNanos is not set during merge

https://github.com/simonswine/pyroscope/blob/9fdea3b017b7335acf39b36df81e3b6ac8f3de1b/pkg/pprof/merge.go#L160-L162

alsoba13:
Just using the grafana/pyroscope link in case your fork is deleted: 
https://github.com/grafana/pyroscope/blob/9fdea3b017b7335acf39b36df81e3b6ac8f3de1b/pkg/pprof/merge.go#L160-L162

aleks-p:
I think you just found where the issue is :)

simonswine:
I think there is a bigger question about if duration should be determined like in v1 where it is basically `end - start` of the query parameters or if it should be driven by the actual data found.

Also if its driven by the data found, there multiple ways we could determine that. 

I guess the simplest would be take min(start) and max(end) of the data and subtract them.

kolesnikovae:
The `DurationNanos` field does not represent the time span over which profiles were collected. Instead, it indicates the time the profiler was actively collecting samples (i.e., the profile duration), and it is 0 in many cases.

References:
- https://github.com/google/pprof/blob/main/proto/profile.proto#L80
- https://github.com/google/pprof/blob/033d6d78b36af29b927e155f6d56cd62b9f949ac/profile/merge.go#L485
- https://github.com/grafana/pyroscope/blob/2c7e5971dc682cb442da4a5e234d01a2366a42c5/pkg/pprof/merge.go#L187

The proper solution is to read the actual values from the table and sum them up. For that reason, I've left the field blank for now, with the intention to implement this correctly in the future.

Leaving it blank is valid in some cases, whereas setting it to the query time range is never valid (from the pprof's perspective).

kolesnikovae:
Also, we need `DurationNanos` (along with `Period`, as sample rate is not fixed in all profilers) at the query time if we want to express CPU time via CPU cores (rate).

kolesnikovae:
The main challenge with `DurationNanos` is that, when a profile is split by sample labels, multiple rows can reference the same underlying value. If we sum these naively, we'll end up significantly overcounting.

[...]

PR #4255, opened 2025-06-16
fix: time series aggregation inconsistency across different time windows

marcsanmi:
## Problem

  Time series visualization showed inconsistent values when querying the same workload across different time ranges:

  - **1-hour queries**: Used 5-second aggregation windows → lower totals
  - **12-hour queries**: Used 30-second aggregation windows → higher totals

  This created misleading graphs where users thought their applications consumed more resources during longer time periods, when it was actually just an aggregation window artifact.

  ## Root Cause

  All profile types used simple sum aggregation regardless of their semantic meaning:
  - **Cumulative profiles** (CPU time, allocations) accumulated values over different window sizes without normalization
  - **Instant profiles** (current memory usage, goroutine counts) were summed instead of averaged
  
  ## Solution

  ### 1. **Profile Type Classification**
  Implemented profile type registry following frontend [schema](https://github.com/grafana/explore-profiles/blob/main/src/shared/infrastructure/profile-metrics/profile-metrics.json):

  **Cumulative Types** (rate normalized):
  - `cpu`, `alloc_*`, `delay`, `lock_time` - accumulated over time

  **Instant Types**:
  - **Gauges** (averaged): `inuse_*`, `goroutine` - current state snapshots
  - **Events** (summed): `samples`, `contentions`, `exceptions` - event counts

  ### 2. **Rate Normalization**
  Added `rateTimeSeriesAggregator` that divides cumulative values by step duration:

Solves https://github.com/grafana/pyroscope/issues/3587


pkg/model/time_series.go
```diff
@@ -69,14 +69,38 @@ type TimeSeriesAggregator interface {
...
+func NewTimeSeriesAggregator(labels []*typesv1.LabelPair, stepDurationSec float64, aggregation *typesv1.TimeSeriesAggregationType) TimeSeriesAggregator {
+	profileType := getProfileTypeName(labels)
+	profileInfo := GetProfileTypeInfo(profileType)
+
+	// If explicit aggregation type is provided, respect it for instant profiles only
+	if aggregation != nil && *aggregation == typesv1.TimeSeriesAggregationType_TIME_SERIES_AGGREGATION_TYPE_AVERAGE {
+		// For instant profiles, use average aggregation
+		if !profileInfo.IsCumulative {
+			return &avgTimeSeriesAggregator{ts: -1}
+		}
```
kolesnikovae: Let's delegate the decision about which aggregation to use to the caller – I believe it should be the client, specifically the frontend. IIRC, we already have some form of `ProfilesTypeRegistry` there (or at least we used to). Otherwise, the client wouldn't know what units to use or what the numbers represent. We also need to ensure compatibility with existing versions of the UI

pkg/model/profile_types.go
```diff
@@ -0,0 +1,163 @@
...
+		IsCumulative:    true,
+		AggregationType: typesv1.TimeSeriesAggregationType_TIME_SERIES_AGGREGATION_TYPE_RATE,
+	},
+	"inuse_objects": {
+		Name:            "inuse_objects",
+		Description:     "Number of objects currently in use",
+		Group:           "memory",
+		Unit:            "short",
+		IsCumulative:    false,
+		AggregationType: typesv1.TimeSeriesAggregationType_TIME_SERIES_AGGREGATION_TYPE_AVERAGE,
```
kolesnikovae: Please note that sum aggregation is enforced in v2 and is not exposed through the query API. The biggest difference is [here](https://github.com/grafana/pyroscope/blob/main/pkg/experiment/query_backend/query_time_series.go#L111C1-L140C2): we need to move the `RangeSeries` call to the frontend, somewhere [here](https://github.com/grafana/pyroscope/blob/main/pkg/frontend/read_path/query_frontend/query_select_time_series.go#L68) (`rate` also needs this), and make sure that we do not call `NewTimeSeriesMerger(true)` anywhere.

Another issue is that the denominator is not always correct for average. If we count "points" for an aggregate, we count rows in the profile table, not profiles themselves. It will break if a profile has multiple rows: e.g., Go `goroutines` profile uses avg. aggregation and it may have multiple points per profile (due to pprof labels). I can't see a simple solution to the problem.

I'm not sure if it works as intended in v1, since it's not currently in use. We need to make sure that we take into account that the average operation is not associative.

pkg/model/time_series.go
```diff
@@ -142,13 +166,47 @@ func (a *avgTimeSeriesAggregator) GetAndReset() *typesv1.Point {
...
+	a.annotations = a.annotations[:0]
+	return &typesv1.Point{
+		Timestamp:   tsCopy,
+		Value:       normalizedValue,
+		Annotations: annotationsCopy,
+	}
+}
+
+func (a *rateTimeSeriesAggregator) IsEmpty() bool       { return a.ts == -1 }
+func (a *rateTimeSeriesAggregator) GetTimestamp() int64 { return a.ts }
```
kolesnikovae, line 185: I'm not very sure this is what users expect from `rate` and normalization – in the test I see that reported rate changes depending on the step duration, which does not address the problem:

> Time series visualization showed inconsistent values when querying the same workload across different time ranges

By definition, `rate` measures how a metric is changing over time (ΔV/Δt, and the Δt is fixed to 1s in our case): https://sourcegraph.com/github.com/prometheus/prometheus/-/blob/promql/functions.go?L71-73 is a good example of how we can calculate this

pkg/model/time_series.go
simonswine: I think this PR is maybe not the right time, but I really would love to centralise this ProfileTypesRegistry in the backend, but until them it makes sense to shift this to the UI

pkg/model/time_series.go
simonswine: I am not too sure if we can do this much better, as we don't reliable have the duration (DurationNanos not set for every counter profile) of the profile and we also don't take it into account when selecting the time window.

I think fixing both of that is a massive change, that I don't think I want this tasl to turn into.

Let's say we have, we have 3 pods, two of them are using a scrape interval of 30s and one is using 15s. All 3 are using exactly one core all the time.

With normalising it to the step size, the result will be with step size 15s roughly this:

```
Pod    | 1 | 2 | 3 |
t+0s   | 0 | 2 | 1 | 
t+15s  | 2 | 0 | 1 |
t+30s  | 0 | 2 | 1 |
t+45s  | 2 | 0 | 1 |

total 3*4*15 = 180
```

And with step size 30s

```
Pod    | 1 | 2 | 3 |
t+0s   | 1 | 1 | 1 |
t+30s  | 1 | 1 | 1 |

total 3 * 2 * 30 = 180
```

I do think this looks very close to the correct data, it obviously is incorrect if you step size gets smaller than your collection window, that also the problem that prometheus has.

kolesnikovae: Yeah, I often use this approach to estimate average core usage – for example, if you get an aggregate profile worth 10 hours of CPU time and the query time range is 1 hour, that's essentially 10h/1h = 10 cores on average. The "total" itself isn't the issue. Problems arise when we have gaps in the data – which is common with ingest sampling. Some users also implement their own strategies, like the classic "10s every 60s" sampling.

I'm also curious how this works with very narrow query time ranges, like when a user tries to fetch an individual profile. I might be mistaken, but IIRC we can receive a step as small as 1ms or so.

This approach works in many cases and could already improve the situation. I'd say we should try it out and see – it's likely we'll need to go through a few iterations. However, as you mentioned, the proper solution lies much closer to the data itself, and can't really be implemented without `DurationNanos`, which is a must for delta profiles (otherwise, we're stuck calculating deltas ourselves).

I also find it interesting that many continuous profiling solutions don't even have a timeline. In many cases, it doesn't make much sense, it's hard to get right, and even harder to build correctly if sampling takes place (and it should take place).

pkg/model/time_series.go
kolesnikovae: It's cool if the registry is implemented in the backend and it's managed by the client/frontend – e.g., adding new types, altering existing ones, etc. It just should not be hardcoded in the backend.

pkg/model/time_series.go
simonswine: Fully agree with this we need to deploy this and find out 🙂 

Btw: The minimum step size is configured in Grafana data source settings and will 15s by default: https://grafana.com/docs/grafana/latest/datasources/pyroscope/configure-pyroscope-data-source/#querying.

marcsanmi: It makes sense. Let's test this out to see how it works and fix this separately https://github.com/grafana/pyroscope/issues/4192. Before that deploying this to dev, I want to address the other comment and remove from this PR the `ProfileTypesRegistry` in the backend for now.

pkg/model/time_series.go
marcsanmi: I think we're not doing any profile-type-aware aggregation... If I'm not wrong should it happen around here https://github.com/grafana/grafana/blob/ecf793ea05189af1c79bd87bbdba4cb7b5cba93d/pkg/tsdb/grafana-pyroscope-datasource/pyroscopeClient.go#L105?

simonswine: Yes you are right we are not passing any profile Type aggregation parameter today in Grafana and Drilldown, but the only way to modify our current behaviour without breaking existing API users is to make the UIs set that parameter, so we know they are new enough that they would also change the units in the Y axis at the same time.

https://github.com/grafana/pyroscope/blob/4542317b180b3510a161867ba25afee69a02cc5c/api/querier/v1/querier.proto#L162

marcsanmi:
@kolesnikovae I addressed your suggestions in 9a3ed0f779d4a6410e2db79946c1ffa9d25d124b in case you want to take a look. I'll deploy this to dev first to find out any unexpected behaviour.

**Edit:** I've deploy it to dev, you can check it in ops in the `grafanacloud-dev-profiles` data source. Right now everything is defaulting to rate aggregation for the sake of testing, but it seems that is consistently calculating the values... (e.g. take a look at the diff view) PLMKWYT 🙂

kolesnikovae:
I think it improves what we have right now!

But the timeline still looks quirky to me:

https://github.com/user-attachments/assets/521a3ce3-02e4-42ff-b5fb-973d717c8384

The diff problem isn't fully resolved yet. We'll likely get better results if both queries share the same step. We could pick the largest, the smallest, something in between, or use a smarter heuristic; it would be cool if users could choose the step. This logic seems best handled in the frontend.

Likewise, our current treatment of "rate" feels like a presentation detail. Storage only sums the data; calculating the rate just means dividing by step, which the frontend can do easily. Moving this to the UI would also let us switch between sum and "rate" views without another fetch.

---

The issue you referenced (https://github.com/grafana/pyroscope/issues/4192) is related in that it also involves `DurationNanos`, which we need for exporting pprof profiles. However, the scope is quite different and won't help with aggregation. I've explained the context in more detail in [this comment](https://github.com/grafana/pyroscope/issues/4192#issuecomment-2989908088) – please have a look when you get a chance.

 ---

Also, please take my comment on the average aggregation seriously – it isn't possible to implement it correctly at the moment; the current implementation will break on profiles that have sample-level (e.g., pprof) labels, such as Go `goroutines`:

```
// A single profile containing samples
1 time=1 v=1 {pod="a", endpoint="foo"} // Bock A
2 time=1 v=1 {pod="a", endpoint="bar"} // Bock B
3 time=1 v=1 {pod="a"} // Bock C

// Another profile
4 time=1 v=3 {pod="b"} // Block D
```

marcsanmi:
When implemented, I think it should help to better calculate aggregation rates, isn't it?
> The issue you referenced (https://github.com/grafana/pyroscope/issues/4192) is related in that it also involves DurationNanos, which we need for exporting pprof profiles. However, the scope is quite different and won't help with aggregation.

For now, I agree that moving rate calculation to the frontend makes sense. Here's what I plan to do next:

  1. **Backend changes**:
     - Remove rate calculation from the backend (`rateTimeSeriesAggregator`)
     - For all metrics, return raw sums (total values within each time window)

  2. **Frontend changes** (profiles-drilldown):
     - Implement rate calculation client-side (value / step_duration) for cumulative metrics
     - Handle step duration alignment in the backend for diff views
   
@kolesnikovae

kolesnikovae:
@marcsanmi sounds good to me! We also need to make sure the frontend shows appropriate units

marcsanmi:
As agreed let's handle this in the frontend. I'll create an issue to not forget about implementing average properly.

PR #4256, opened 2025-06-16
chore: Update offsets for python 3.13.0 (stable)

simonswine:
We appear to have used the a6 offset for a while, fixing them to the stable release. The offset also are valid for latest patch 3.13.5.


korniltsev (approved):
lgtm, thanks

I don't have access to the old dockerhub.
I suggest we use a personal dockerhub for the new test image
Alternatively @petethepig may want to put creds into 1password

PR #4261, opened 2025-06-18
feat(v2): request level write path overrides

kolesnikovae:


pkg/distributor/distributor.go
```diff
@@ -1111,3 +1115,13 @@ func exportSamples(e *pprof.SampleExporter, samples []*profilev1.Sample) *pprof.
 	e.ExportSamples(n.Profile, samplesCopy)
 	return n
 }
+
+func isAlloyEBPFRequest(series *distributormodel.PushRequest) bool {
+	for _, s := range series.Series {
```
simonswine: Probably worth having a quick path for returning `false`, when the request did not come in with a `pushv1`.

aleks-p (approved):
This makes sense to me. I tested it manually as well and it worked as expected.

simonswine (approved):
LGTM

PR #4265, opened 2025-06-18
Reapply "Fix: Handle duplicate validation correctly when sanitizing (#4238)"

simonswine:
This introduces #4238 again, it does break the OTLP validation integration tests.

@marcsanmi if you got the chance, please take a look at it.

Fixes https://github.com/grafana/pyroscope-squad/issues/480


marcsanmi:
@simonswine Please take a look when you're back. It seems the we weren't keeping the labels properly ordered after updating them.

pkg/validation/validate.go
```diff
@@ -128,32 +129,108 @@ func ValidateLabels(limits LabelValidationLimits, tenantID string, ls []*typesv1
...
+		} else {
+			// Move forwards - shift labels left
+			copy(ls[origIdx:insertIdx], ls[origIdx+1:insertIdx+1])
+			ls[insertIdx] = label
+			return ls, origIdx, nil
+		}
+	}
+
+	return ls, origIdx, nil
+}
```
simonswine: This is also my fault as I started this bit, but I think we a trying to much here and it gets quite complex. I think we can use some stdlib to make this easier:

* We copy the whole slice into a new without origIdx
* We use `slices.BinarySearchFunc` to find if a `newName` is already existing or if not at which position it is
* If its existing we return like the code does right now
* If not then `slices.Insert` the new one at the right place.

This will be less relying on our implementation, and likely more performant and easier to read.

Let me know if you have any questions.

pkg/validation/validate.go
```diff
@@ -128,32 +129,108 @@ func ValidateLabels(limits LabelValidationLimits, tenantID string, ls []*typesv1
...
 		}
 		lastLabelName = l.Name
+		idx += 1
 	}
 
 	return nil
 }
 
+// handleSanitizedLabel handles the case where a label name is sanitized. It ensures that the label name is unique and fails if the value is distinct.
+func handleSanitizedLabel(ls []*typesv1.LabelPair, origIdx int, origName, newName string) ([]*typesv1.LabelPair, int, error) {
```
simonswine, line 176: We are not using the int return, about the index to continue validating at. I can't reproduce a test case, as `.` is lexigraphically before `_`, so we will never move back wards into the already validate section. It would be better to handle that correctly. 

By setting it to the smaller of `origIdx` and `insertIdx`

simonswine (commented):
Thanks for picking that up, I do think it fairly hard to read and not entirely sure if its correct. Let's simplfy this as I mentioned in the comments.

marcsanmi:
@simonswine I addressed all the concerns, lmkwyt

pkg/validation/validate.go
```diff
@@ -130,32 +132,87 @@ func ValidateLabels(limits LabelValidationLimits, tenantID string, ls []*typesv1
...
+	origValue := ls[origIdx].Value
```
simonswine: Nit: You could create `newLabel` here then you don't need origValue and can use it in `BinarySearchFunc` and `Insert`.

pkg/validation/validate.go
```diff
@@ -130,32 +132,87 @@ func ValidateLabels(limits LabelValidationLimits, tenantID string, ls []*typesv1
...
+	insertIdx, found := slices.BinarySearchFunc(newSlice, &typesv1.LabelPair{Name: newName},
+		func(a, b *typesv1.LabelPair) int {
+			return strings.Compare(a.Name, b.Name)
+		})
+
+	if found {
+		if newSlice[insertIdx].Value == origValue {
+			// Same name and value - remove duplicate (already removed from newSlice)
+			copy(ls, newSlice)
+			return ls[:len(newSlice)], origIdx, nil
```
simonswine: It makes sense to return the new label slice instead

pkg/validation/validate.go
```diff
@@ -130,32 +132,87 @@ func ValidateLabels(limits LabelValidationLimits, tenantID string, ls []*typesv1
...
+				DuplicateLabelNamesAfterSanitizationErrorMsg,
+				phlaremodel.LabelPairsString(ls), newName, origName)
+		}
+	}
+
+	// Insert the new label at correct position
+	newLabel := &typesv1.LabelPair{Name: newName, Value: origValue}
+	newSlice = slices.Insert(newSlice, insertIdx, newLabel)
+
+	copy(ls, newSlice)
```
simonswine: Same: Probably would just return the new label slice instead

pkg/validation/validate.go
```diff
@@ -130,32 +132,87 @@ func ValidateLabels(limits LabelValidationLimits, tenantID string, ls []*typesv1
...
+	newSlice = append(newSlice, ls[origIdx+1:]...)
+
+	insertIdx, found := slices.BinarySearchFunc(newSlice, &typesv1.LabelPair{Name: newName},
+		func(a, b *typesv1.LabelPair) int {
+			return strings.Compare(a.Name, b.Name)
+		})
+
+	if found {
+		if newSlice[insertIdx].Value == origValue {
+			// Same name and value - remove duplicate (already removed from newSlice)
```
simonswine:
```suggestion
			// Same name and value we are done and can just return newSlice
```

pkg/validation/validate.go
```diff
@@ -130,32 +132,87 @@ func ValidateLabels(limits LabelValidationLimits, tenantID string, ls []*typesv1
...
+			copy(ls, newSlice)
+			return ls[:len(newSlice)], origIdx, nil
+		} else {
+			// Same name, different value - error
+			return ls, origIdx, NewErrorf(DuplicateLabelNames,
```
simonswine: Nit: I find it cleaner to return nothing

```suggestion
			return nil, 0, NewErrorf(DuplicateLabelNames,
```

simonswine (commented):
Can't approve as it's my PR, but I would approve if I could.

Thanks for revisiting this ❤️

pkg/validation/validate.go
marcsanmi: I think copy is the easiest way, otherwise it breaks ingest pprof integration tests.. Or we should look to add pointer to slice or modify the slice in-place.

alsoba13 (approved):

alsoba13:
LGTM,

I'm wondering how would be the performance impact of using `map[string]string` to handle duplicates, instead of doing slice tricks such as creating new slices and binary search them. I guess it's just fine in both ways as probably we are dealing with a relatively small amount of labels. Edit: also most common scenario would be 0 slice tricks (no renaming labels), so introducing maps here would be punishing

marcsanmi:
> I'm wondering how would be the performance impact of using map[string]string to handle duplicates, instead of doing slice tricks such as creating new slices and binary search them. I guess it's just fine in both ways as probably we are dealing with a relatively small amount of labels. Edit: also most common scenario would be 0 slice tricks (no renaming labels), so introducing maps here would be punishing

I think labels must stay sorted for consistent fingerprints (`Labels.Hash()` processes sequentially), so even with a map for duplicates, we'd still need binary search for correct insertion positioning. The current approach uses the sorted slice for both duplicate detection and positioning, eliminating an additional data structure, lmkwyt.

PR #4269, opened 2025-06-24
chore: Bump helm to v1.14.0

bryanhuhta:
In preparation for v1.14 release, bump the helm chart version


marcsanmi (commented):
I believe `Chart.lock` needs to be updated.

bryanhuhta:
> I believe `Chart.lock` needs to be updated.

I actually had to tag 1.14 so the 1.14 docker image would be published. It all looks green now

PR #4272, opened 2025-06-26
chore: Update envVar configuration to support custom envs

cesarokuti:
To support IPv6 clusters, we need to configure the MY_POD_IP environment variable using the valueFrom structure. Therefore, I suggest changing the extraEnvs field to allow an open configuration instead of using a predefined range.

Additionally, I've included the self-profile configuration to be managed as part of this update.


korniltsev:
Can you use `extraEnvFrom` instead?

Is this a breaking change? Can we make it non-breaking change ?

cesarokuti:
> Can you use `extraEnvFrom` instead?
> 
> Is this a breaking change? Can we make it non-breaking change ?

I created other parameter to configure this env. But I think we can deprecate in the future.

korniltsev:
Is it possible to use `extraEnvFrom` instead of modifying the chart to introduce new configuration options?

korniltsev:
could you submit `disableSelfProfile` as a separate change?

operations/pyroscope/helm/pyroscope/values.yaml
```diff
@@ -25,6 +28,11 @@ pyroscope:
     # The following environment variables are set by the Helm chart.
     # JAEGER_AGENT_HOST: jaeger-agent.jaeger.svc.cluster.local.
 
+  extraCustomEnvVars: {}
+    # The following environment variables raw form.
```
korniltsev, line 32: Please add an example how you could use this with `valueFrom` for example to generate something like this.
```yaml
        - name: MY_NODE_NAME
          valueFrom:
            fieldRef:
              fieldPath: spec.nodeName
```

korniltsev (changes requested):
Please ignore my other comments above.
The changes looks good to me. Please update an example and resolve conflicts and we will merge this

korniltsev-grafanista (approved):

PR #4278, opened 2025-07-02
chore: add ingest request body size limit

alsoba13:
Properly approaches https://github.com/grafana/pyroscope/pull/4275

### DIFF
--- a/pkg/ingester/pyroscope/ingest_handler.go
+++ b/pkg/ingester/pyroscope/ingest_handler.go
@@ -3,6 +3,7 @@ package pyroscope
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -15,6 +16,7 @@ import (
 
 	"github.com/grafana/pyroscope/pkg/tenant"
 	httputil "github.com/grafana/pyroscope/pkg/util/http"
+	"github.com/grafana/pyroscope/pkg/validation"
 
 	"github.com/grafana/pyroscope/pkg/og/convert/speedscope"
 
@@ -63,10 +65,22 @@ func (h ingestHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if err := readInputRawDataFromRequest(ctx, r, input); err != nil {
+		var status int
+		var maxBytesError *http.MaxBytesError
+		switch {
+		case errors.As(err, &maxBytesError):
+			err = fmt.Errorf("request body too large: %w", err)
+			status = http.StatusRequestEntityTooLarge
+			validation.DiscardedBytes.WithLabelValues(string(validation.BodySizeLimit), tenantID).Add(float64(maxBytesError.Limit))
+			validation.DiscardedProfiles.WithLabelValues(string(validation.BodySizeLimit), tenantID).Add(float64(1))
+		default:
+			status = http.StatusRequestTimeout
+		}
+
 		msg := "failed to read request body"
 		sp.LogFields(otlog.Error(err), otlog.String("msg", msg))
 		_ = h.log.Log("msg", msg, "err", err, "orgID", tenantID)
-		httputil.ErrorWithStatus(w, err, http.StatusRequestTimeout)
+		httputil.ErrorWithStatus(w, err, status)
 		return
 	}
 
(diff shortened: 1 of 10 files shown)


pkg/ingester/pyroscope/ingest_handler.go
```diff
@@ -63,10 +65,22 @@ func (h ingestHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
 	if err := readInputRawDataFromRequest(ctx, r, input); err != nil {
+		var status int
+		var maxBytesError *http.MaxBytesError
+		switch {
+		case errors.As(err, &maxBytesError):
+			err = fmt.Errorf("request body too large: %w", err)
+			status = http.StatusRequestEntityTooLarge
+			validation.DiscardedBytes.WithLabelValues(string(validation.BodySizeLimit), tenantID).Add(float64(maxBytesError.Limit))
+			validation.DiscardedProfiles.WithLabelValues(string(validation.BodySizeLimit), tenantID).Add(float64(1))
```
alsoba13, line 75: For DiscardedProfiles, we don't get to read the body, so I increase it by `1`, similarly to [ingester.go](https://github.com/grafana/pyroscope/blob/weekly-f125-16eebf27b/pkg/ingester/ingester.go#L352-L353).

For DiscardedBytes, only increased by `Limit`, we never get to read the rest of the message so we don't really know how big it was

pkg/validation/limits.go
```diff
@@ -139,6 +140,7 @@ func (e LimitError) Error() string {
 func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.Float64Var(&l.IngestionRateMB, "distributor.ingestion-rate-limit-mb", 4, "Per-tenant ingestion rate limit in sample size per second. Units in MB.")
 	f.Float64Var(&l.IngestionBurstSizeMB, "distributor.ingestion-burst-size-mb", 2, "Per-tenant allowed ingestion burst size (in sample size). Units in MB. The burst size refers to the per-distributor local rate limiter, and should be set at least to the maximum profile size expected in a single push request.")
+	f.Float64Var(&l.IngestionBodyLimitMB, "distributor.ingestion-body-limit-mb", 32, "Per-tenant ingestion body size limit in MB, before decompressing. 0 to disable.")
```
alsoba13, line 143: I opted to disable it by default, so we make sure no existing pyroscope starts failing after an update.

pkg/validation/limits.go
```diff
@@ -139,6 +140,7 @@ func (e LimitError) Error() string {
 func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.Float64Var(&l.IngestionRateMB, "distributor.ingestion-rate-limit-mb", 4, "Per-tenant ingestion rate limit in sample size per second. Units in MB.")
 	f.Float64Var(&l.IngestionBurstSizeMB, "distributor.ingestion-burst-size-mb", 2, "Per-tenant allowed ingestion burst size (in sample size). Units in MB. The burst size refers to the per-distributor local rate limiter, and should be set at least to the maximum profile size expected in a single push request.")
+	f.Float64Var(&l.IngestionBodyLimitMB, "distributor.ingestion-body-limit-mb", 0, "Per-tenant ingestion body size limit in MB, before decompressing. 0 to disable.")
```
alsoba13, line 143: I opted to disable it by default, so we make sure no existing pyroscope starts failing after an update.

alsoba13:
> Looks good overall.
> 
> Does an operator have enough information to act, in case the limit is breached? Would we easily know how often and by how much the limit is breached (e.g., will we see the actual values in the logs)?
> 
> Also, #4275 had real-world tests (exercising the actual /ingest path with profiling data). Is it possible we add that here to test the wiring itself?

I'll work on some tests in there, thank you!

With this solution, an operator will only know that the limit was reached. This issue tries to bound the memory and CPU usage for huge incoming messages (typically from malfunctioning clients). 
I can rework this to compute the original size by draining the rest of the message once we've hit the limit. Memory will be bounded, but it may still be CPU intensive.

aleks-p:
> With this solution, an operator will only know that the limit was reached. This issue tries to bound the memory and CPU usage for huge incoming messages (typically from malfunctioning clients).
> I can rework this to compute the original size by draining the rest of the message once we've hit the limit. Memory will be bounded, but it may still be CPU intensive.

Makes sense, lets not add extra work for logging purposes, especially if we don't intend to change this limit frequently.

aleks-p (approved):

PR #4286, opened 2025-07-06
perf: add searchHint to getSeriesIndex

liaol:
The `phlaredb.getSeriesIndex()` function is a performance bottleneck during "flushing", consuming approximately 12% of CPU.  

https://github.com/grafana/pyroscope/blob/4a2d0a8ed450c9b1e2bd249347b197efd723e99b/pkg/phlaredb/profiles.go#L43-L54
Since `rowRangesWithSeriesIndex` is strictly ordered and `rowNum` increments sequentially, implementing a `searchHint` is beneficial. This optimization reduced CPU usage of `getSeriesIndex` to 5% and reduced the "flushing" operation's duration from minutes to seconds in our tests.

Furthermore, if `seriesIDRowsRewriter.ReadRows()` has no data-race, adding `searchHint` to `seriesIDRowsRewriter`  struct would further improve efficiency.
https://github.com/grafana/pyroscope/blob/4a2d0a8ed450c9b1e2bd249347b197efd723e99b/pkg/phlaredb/profile_store.go#L538-L553

### DIFF
--- a/pkg/phlaredb/profiles.go
+++ b/pkg/phlaredb/profiles.go
@@ -40,13 +40,26 @@ type rowRangeWithSeriesIndex struct {
 // those need to be strictly ordered
 type rowRangesWithSeriesIndex []rowRangeWithSeriesIndex
 
-func (s rowRangesWithSeriesIndex) getSeriesIndex(rowNum int64) uint32 {
-	for _, rg := range s {
+func (s rowRangesWithSeriesIndex) getSeriesIndex(rowNum int64, searchHint *int) uint32 {
+	for i := *searchHint; i < len(s); i++ {
+		rg := s[i]
 		// it is possible that the series is not existing
 		if rg.rowRange == nil {
 			continue
 		}
 		if rg.rowNum <= rowNum && rg.rowNum+int64(rg.length) > rowNum {
+			*searchHint = i
+			return rg.seriesIndex
+		}
+	}
+	for i := 0; i < *searchHint; i++ {
+		rg := s[i]
+		// it is possible that the series is not existing
+		if rg.rowRange == nil {
+			continue
+		}
+		if rg.rowNum <= rowNum && rg.rowNum+int64(rg.length) > rowNum {
+			*searchHint = i
 			return rg.seriesIndex
 		}
 	}
--- a/pkg/phlaredb/profile_store.go
+++ b/pkg/phlaredb/profile_store.go
@@ -540,11 +540,12 @@ func (r *seriesIDRowsRewriter) ReadRows(rows []parquet.Row) (int, error) {
 	if err != nil {
 		return n, err
 	}
-
+	// searchHint for next call of getSeriesIndex
+	var searchHint int
 	for pos, row := range rows[:n] {
 		// actual row num
 		rowNum := r.pos + int64(pos)
-		row[colIdxSeriesIndex] = parquet.ValueOf(r.seriesIndexes.getSeriesIndex(rowNum)).Level(0, 0, colIdxSeriesIndex)
+		row[colIdxSeriesIndex] = parquet.ValueOf(r.seriesIndexes.getSeriesIndex(rowNum, &searchHint)).Level(0, 0, colIdxSeriesIndex)
 	}
 
 	r.pos += int64(n)


korniltsev:
Can you share a bit more about the size of your blocks? How many rowgroups, rows, series does it have? Something tells me you have too many series.

> Furthermore, if seriesIDRowsRewriter.ReadRows() has no data-race, adding searchHint to seriesIDRowsRewriter struct would further improve efficiency.

I don't see no concurrency in there during flushing (Except `(r *seriesIDRowsRewriter) SeekToRow` which is not used and imo should be removed). Let's implement this idea, should be way better as the buffer for the reader is only 64 .

liaol:
@korniltsev Yes, we have big series, the  size of `rowRangeWithSeriesIndex` is over 200K. So the seachHint will save over 63 * 200K operations.
I will move  `searchHint` to `seriesIDRowsRewriter struct` later.

pkg/phlaredb/profiles.go
```diff
@@ -40,13 +40,26 @@ type rowRangeWithSeriesIndex struct {
...
 		// it is possible that the series is not existing
 		if rg.rowRange == nil {
 			continue
 		}
 		if rg.rowNum <= rowNum && rg.rowNum+int64(rg.length) > rowNum {
+			*searchHint = i
+			return rg.seriesIndex
+		}
+	}
+	for i := 0; i < *searchHint; i++ {
```
korniltsev: Is the second loop needed?

liaol: Yes, for the corner case which not increments sequentially.

liaol:
@korniltsev I've moved `searchHint` to `seriesIDRowsRewriter`. Please review again.

pkg/phlaredb/profiles.go
korniltsev: Do you know how to reproduce this case? It's worth adding a comment what the corner case is and how to trigger it. Or maybe even a unit test so that we do not remove this in the future accidentally.

liaol: For example, consider row numbers 1, 2, 3, 4, 3, which may not actually exist. The second loop can maintain the consistency of `getSeriesIndex` before and after modification.

korniltsev: I don't understand. What is 1,2,3,4,3 and what does not exist. Can you add a test? Or at least source code comment explaining why the second loop needed.

korniltsev (changes requested):
While the second loop will not hurt, I'd like to have understanding myself and documented in the form of a comment or a test in the source code

liaol:
@korniltsev Sorry for not explaining clearly, I have submitted the unit tests and comments.

pkg/phlaredb/profile_test.go
```diff
@@ -279,3 +279,93 @@ func TestProfileIndex_Add_OutOfOrder(t *testing.T) {
...
+		{rowRange: &rowRange{rowNum: 0, length: 5}, seriesIndex: 1},
+		{rowRange: &rowRange{rowNum: 5, length: 5}, seriesIndex: 2},
+	}
+	searchHint := 0
+
+	idx := ranges.getSeriesIndex(6, &searchHint)
+	assert.Equal(t, uint32(2), idx)
+	assert.Equal(t, 1, searchHint)
+
+	// rowNum is not increasing, so it will not hit the first loop, and will hit the second loop
```
korniltsev: I don't get it. why would it decrease? Looking at the code it goes from zero to number of rows in rowgroup

https://github.com/grafana/pyroscope/blob/5b90569d50e869fce59e0784cee1044737e5ec06/pkg/phlaredb/profile_store.go#L546-L552

pkg/phlaredb/profile_test.go
```diff
@@ -279,3 +279,93 @@ func TestProfileIndex_Add_OutOfOrder(t *testing.T) {
...
+			searchHint: 1, // will skip 0 in main loop, hit in second loop
+			expectIdx:  0,
+		},
+		{
+			name: "searchHint out of range",
+			ranges: rowRangesWithSeriesIndex{
+				{rowRange: &rowRange{rowNum: 0, length: 5}, seriesIndex: 1},
+			},
+			rowNum:     2,
+			searchHint: 100, // should reset to 0
```
korniltsev: I don't get it. Why would searchHint be 100 if the rowgroup only have a single series?

liaol:
You're right, the second loop is unnecessary because rowNum is always incrementing.
I have removed the second loop and the unnecessary unit tests.

pkg/phlaredb/profile_test.go
```diff
@@ -279,3 +279,67 @@ func TestProfileIndex_Add_OutOfOrder(t *testing.T) {
...
+			ranges: rowRangesWithSeriesIndex{
+				{rowRange: &rowRange{rowNum: 0, length: 5}, seriesIndex: 1},
+				{rowRange: &rowRange{rowNum: 5, length: 5}, seriesIndex: 2},
+			},
+			rowNum:     6,
+			searchHint: 1,
+			expectIdx:  1,
+		},
+		{
+			name: "second loop hit",
```
korniltsev: there is no second loop

pkg/phlaredb/profile_test.go
```diff
@@ -279,3 +279,67 @@ func TestProfileIndex_Add_OutOfOrder(t *testing.T) {
...
+	}{
+		{
+			name: "main loop hit",
+			ranges: rowRangesWithSeriesIndex{
+				{rowRange: &rowRange{rowNum: 0, length: 5}, seriesIndex: 1},
+				{rowRange: &rowRange{rowNum: 5, length: 5}, seriesIndex: 2},
+			},
+			rowNum:     6,
+			searchHint: 1,
+			expectIdx:  1,
```
korniltsev: there is something wrong with the testate. rownum should point to row range with index 1 with seriesIndex 2

pkg/phlaredb/profile_test.go
```diff
@@ -279,3 +279,67 @@ func TestProfileIndex_Add_OutOfOrder(t *testing.T) {
...
 
 }
+
+func Test_rowRangesWithSeriesIndex_getSeriesIndex(t *testing.T) {
+	testCases := []struct {
+		name        string
+		ranges      rowRangesWithSeriesIndex
+		rowNum      int64
+		searchHint  int
+		expectIdx   int
```
korniltsev: we should rename this to  expectedSeriesIndex, to not confuse with any other indexes

pkg/phlaredb/profile_test.go
```diff
@@ -279,3 +279,67 @@ func TestProfileIndex_Add_OutOfOrder(t *testing.T) {
...
+		},
+		{
+			name: "second loop hit",
+			ranges: rowRangesWithSeriesIndex{
+				{rowRange: &rowRange{rowNum: 0, length: 5}, seriesIndex: 1},
+				{rowRange: &rowRange{rowNum: 5, length: 5}, seriesIndex: 2},
+			},
+			rowNum:     1,
+			searchHint: 1, // will skip 0 in main loop, hit in second loop
+			expectIdx:  0,
```
korniltsev: same here, rownum is in row range number 0 with seriesIndex 1. we should expect seriesIndex = 1 here, not 0

pkg/phlaredb/profile_test.go
korniltsev: we should expect seriesIndex2 here not 1

pkg/phlaredb/profile_test.go
```diff
@@ -279,3 +279,67 @@ func TestProfileIndex_Add_OutOfOrder(t *testing.T) {
...
+		},
+		{
+			name: "nil rowRange skipped",
+			ranges: rowRangesWithSeriesIndex{
+				{rowRange: nil, seriesIndex: 1},
+				{rowRange: &rowRange{rowNum: 10, length: 5}, seriesIndex: 2},
+			},
+			rowNum:     12,
+			searchHint: 0,
+			expectIdx:  1,
```
korniltsev: similar to above, the expected value should be 2

pkg/phlaredb/profile_test.go
```diff
@@ -279,3 +279,67 @@ func TestProfileIndex_Add_OutOfOrder(t *testing.T) {
...
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			searchHint := tc.searchHint
+			if tc.expectPanic {
+				assert.Panics(t, func() {
+					_ = tc.ranges.getSeriesIndex(tc.rowNum, &searchHint)
+				})
+			} else {
+				idx := tc.ranges.getSeriesIndex(tc.rowNum, &searchHint)
+				assert.Equal(t, tc.ranges[tc.expectIdx].seriesIndex, idx)
```
korniltsev: I think this assertion is incorrect. `getSeriesIndex` returns `seriesIndex`, not index into the ranges slice.  This should be fixed.

korniltsev (changes requested):

pkg/phlaredb/profile_test.go
liaol: all fixed.

korniltsev (approved):

PR #4292, opened 2025-07-10
fix: change `settings.v1.SettingsService` pathType to work with grafana

dowster:
Our grafana instance was not able to actually change the Profiles Settings in drilldown because traffic could not route to tenant settings. Our configuration is using an AWS ALB via the alb controller and it was configuring these paths as exact matches and putting them  _after_ the prefix path of `/*`. 

This change may need to be applied to the other routes in here which have `ImplementationSpecific` configured as their `PathType`.

### DIFF
--- a/operations/pyroscope/helm/pyroscope/Chart.yaml
+++ b/operations/pyroscope/helm/pyroscope/Chart.yaml
@@ -3,7 +3,7 @@ name: pyroscope
 description: 🔥 horizontally-scalable, highly-available, multi-tenant continuous profiling aggregation system
 home: https://grafana.com/oss/pyroscope/
 type: application
-version: 1.14.0
+version: 1.14.1
 appVersion: 1.14.0
 dependencies:
   - name: grafana-agent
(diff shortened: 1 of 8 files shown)


simonswine:
> This change may need to be applied to the other routes in here which have `ImplementationSpecific` configured as their `PathType`.

Yes I agree, we should do this to all routes. Surprising that you face no other problems with the other `ImplementationSpecific`. I will test this with at least nginx-ingress to figure out if this causes a breakage.

dowster:
> Surprising that you face no other problems with the other `ImplementationSpecific`.

I suspect we just haven't run into those issues yet, I'll push another commit to add it for the other routes.

dowster:
@simonswine was there something still needed here or is this good to go?

simonswine:
@dowster sorry this was waiting on me testing this with a few ingress controllers.

Unfortunately this breaks the nginx-ingress (which I tested first btw)

```
W0718 17:06:17.620862   89742 warnings.go:70] path /adhocprofiles.v1.AdHocProfileService/ cannot be used with pathType Prefix
Error: UPGRADE FAILED: failed to create resource: admission webhook "validate.nginx.ingress.kubernetes.io" denied the request: ingress contains invalid paths: path /querier.v1.QuerierService/ cannot be used with pathType Prefix
path /push.v1.PusherService/ cannot be used with pathType Prefix
path /settings.v1.SettingsService/ cannot be used with pathType Prefix
path /adhocprofiles.v1.AdHocProfileService/ cannot be used with pathType Prefix
```

This is as far as I understand because only alphanumeric characters and `/` are allowed in nginx paths, according to this:
https://kubernetes.github.io/ingress-nginx/faq/#validation-of-path

So we need to stick with ImplementationSpecific. What I suggest we have a new `.Values.ingress.pathType: ImplementationSpecific`, and then it informs the template and can then configure it for your ingress controller.

simonswine (approved):
Thank you for your patience reworking the approach!

LGTM.

kembox:
In helm chart 1.14.1, when all the paths changed to `ImplementationSpecific` , the `/` path is broken  in our case ( it was `Prefix` before )
Our alloy ( helm chart v1.1.1 )  couldn't push data to pyroscope with `"unimplemented: 404 Not Found"` error
```
ts=2025-08-29T01:54:18.578030828Z level=warn msg="final error sending to profiles to endpoint" component_path=/ component_id=pyroscope.write.pyroscope endpoint=http://our.internal err="unimplemented: 404 Not Found"
```
Changed the `pathType` for `/` back to `Prefix` works. Reverting to chart 1.14.0 also works. JFYI

PR #4298, opened 2025-07-14
WIP: Support avg time_series aggregation

simonswine:
This implements an avg time series aggregation. Achieving an avg
aggration is not trivial:

- On ingest we split each profile series into smaller smaller ones,
  by using the dynamic labels.

This PR will reaassemble profiles matching with matching ID/timestamp
and build an avg per series and the aggregate the averages per sum.

So lets say we have a 3 replica deployment with each using exactly 1GiB
of in use memory.

We would be expect each pods series to be a flat 1GiB (no matter how
many points fall into the step size) and when those 3 pods are combined,
we would expect a flat 3GiB line.


pkg/util/delayhandler/x
marcsanmi, line 1: Is this file needed?

PR #4301, opened 2025-07-15
chore: Update prometheus dependency to 3.5.0

simonswine:
Note: This moves labels.Labels from `slicelabels` to `stringlabels`, hence it requires a bit of work.

The most significant change here that the results of `label.Labels.Hash()` changes, so whenever we use that we should only rely on this for the runtime of the process and not inform the stored hash.


pkg/model/labels.go
```diff
@@ -149,7 +149,7 @@ func (ls Labels) ToPrometheusLabels() labels.Labels {
 	for i, l := range ls {
 		res[i] = labels.Label{Name: l.Name, Value: l.Value}
 	}
-	return res
+	return labels.New(res...)
```
aleks-p, line 152: we could probably use a builder here and in a few other places (might be more efficient), but this one is only used in `profilecli` so it is probably not worth it

pkg/metrics/observer.go
```diff
@@ -282,13 +279,13 @@ func (r *recording) initState(labelsMap map[string]string, externalLabels labels
 
 func newTimeSeries(exportedLabels labels.Labels, recordingTime int64) *prompb.TimeSeries {
 	// prompb.Labels don't implement sort interface, so we need to use labels.Labels and transform it later
-	pbLabels := make([]prompb.Label, 0, len(exportedLabels))
-	for _, label := range exportedLabels {
+	pbLabels := make([]prompb.Label, 0, exportedLabels.Len())
```
aleks-p: `Len()` is likely slower than the previous approach (described as "relatively slow" in the docs). Not sure if this could cause an issue here or not.

pkg/model/tree.go
```diff
@@ -334,10 +334,10 @@ func (t *Tree) MarshalTruncate(w io.Writer, maxNodes int64) (err error) {
 	for len(nodes) > 0 {
 		last := len(nodes) - 1
 		n, nodes = nodes[last], nodes[:last]
-		if _, _ = vw.Write(w, uint64(len(n.name))); err != nil {
+		if _, err = vw.Write(w, uint64(len(n.name))); err != nil {
 			return err
 		}
-		if _, _ = w.Write(unsafeStringBytes(n.name)); err != nil {
+		if _, err = w.Write(unsafeStringBytes(n.name)); err != nil {
```
aleks-p, line 340: good catch

aleks-p (approved):
Looks good to me. I left a few minor remarks / questions. Might be a good idea for @alsoba13 to take a look at the changes in the recording rules area.

pkg/metrics/observer.go
simonswine: I did a bit of refactoring, not too sure if this is really cheaper: [d6df499](https://github.com/grafana/pyroscope/pull/4301/commits/d6df499ad2f6addf48a79374a5b79e0c352d6b55)

I guess we will find it out once we deploy this 🙂

aleks-p: Thanks for taking a stab at it and yeah, lets give it a test and adjust later if needed.

pkg/metrics/observer.go
```diff
@@ -259,36 +259,30 @@ func (o *SampleObserver) Close() {
...
-func (r *recording) initState(labelsMap map[string]string, externalLabels labels.Labels, recordingTime int64) {
-	r.state.matches = r.matches(labelsMap)
+func (r *recording) initState(lb *labels.Builder, blockLabels labels.Labels, externalLabels labels.Labels, recordingTime int64) {
+	r.state.matches = r.matches(blockLabels)
 	if !r.state.matches {
 		return
 	}
 
-	exportedLabels := generateExportedLabels(labelsMap, r, externalLabels)
-	sort.Sort(exportedLabels)
```
alsoba13, line 272: I can't 100% remember why I was sorting this, but I think it's because of this constraint (from [Labels](https://prometheus.io/docs/specs/prw/remote_write_spec/#labels) in remote_write spec)
> MUST have label names sorted in lexicographical order.

Just to make sure, `lb.Labels()` does ensure this property, right?

alsoba13:
LGTM if we ensure labels are sorted when calling the builder Labels() func in the obsever. I see from the source code that they use some Sort functions, but it's weird to me that they don't mention if labels are sorted in the api

pkg/metrics/observer.go
simonswine: Yes that's an important comment:

The `LabelsBuilder` does sort automatically on `Labels()`, while the `ScratchBuilder` doesn't.

PR #4303, opened 2025-07-17
chore: Update golangci-lint and fix issues

simonswine:
- **Update golangci-lint to v2** `using migrate subcommand`
- **Use `--fix`**
- **Manually fix linting issues**


pkg/objstore/client/config.go
```diff
@@ -148,7 +148,7 @@ func validStoragePrefixPart(part string) bool {
 		return false
 	}
 	for i, b := range part {
-		if !((b >= 'a' && b <= 'z') || (b >= 'A' && b <= 'Z') || b == '_' || b == '-' || b == '.' || (b >= '0' && b <= '9' && i > 0)) {
+		if (b < 'a' || b > 'z') && (b < 'A' || b > 'Z') && b != '_' && b != '-' && b != '.' && (b < '0' || b > '9' || i == 0) {
```
simonswine, line 151: Please double check those linting issues, as I converted them by hand:

```
QF1001: could apply De Morgan's law (staticcheck)
```

marcsanmi: I could have missed something but it seems right 🙂

PR #4304, opened 2025-07-17
chore: Ensure tenant gets logged during ingest validation errors

simonswine:
Was noticing it is hard to match up the debug logs with the tenant


marcsanmi (approved):
LGTM

PR #4311, opened 2025-07-22
chore: upgrade dskit

aleks-p:
We are currently running a fairly old version of dskit (December 2023). This upgrades it to `20250718123059-f0494e388de2` (https://github.com/grafana/dskit/commit/f0494e388de2bbc4dee37b84e83c4576b1f26183). The upgrade brings in other transitive dependencies.

One notable change is that the prometheus dependency also got bumped, and the new version makes [string labels](https://github.com/bboreham/prometheus/commit/6136ae67e0acbe2592c47061e57ebaed93b8d4ab) the default behavior. More details here: https://github.com/prometheus/prometheus/pull/16436.

I've opted out of this via the `slicelabels` build flag, to reduce the blast radius of the change. I think we can remove the flag in a separate step, however in the meantime we will have the inconvenience when working with the repo locally (mostly in IDEs). Let me know if you prefer a different way of solving this.

Edit: The prometheus upgrade was done separately, in https://github.com/grafana/pyroscope/pull/4301


pkg/pyroscope/modules.go
```diff
@@ -507,9 +507,7 @@ func (f *Pyroscope) initServer() (services.Service, error) {
 		return nil, err
 	}
 	defaultHTTPMiddleware := []middleware.Interface{
-		middleware.Tracer{
-			RouteMatcher: f.Server.HTTP,
-		},
+		middleware.Tracer{},
```
aleks-p: not sure yet if this changes any behavior, need to confirm before merging

pkg/util/http.go
```diff
@@ -400,7 +400,6 @@ func NewHTTPMetricMiddleware(mux *mux.Router, namespace string, reg prometheus.R
 		}
 	}
 	return middleware.Instrument{
-		RouteMatcher:     mux,
```
aleks-p, line 403: not sure yet if this changes any behavior, need to confirm before merging

simonswine:
I have updated Prometheus to support string labels as well as slice label in #4301, so would like to get some reviews there.

aleks-p:
> I have updated Prometheus to support string labels as well as slice label in #4301, so would like to get some reviews there.

Sorry I missed that completely. I will take a look as there is definitely overlap.

pkg/pyroscope/modules.go
aleks-p: adding an explicit middleware for injecting the route, see https://github.com/grafana/dskit/pull/527 for more info

pkg/util/http.go
aleks-p: adding an explicit middleware for injecting the route, see https://github.com/grafana/dskit/pull/527 for more info

pkg/pyroscope/modules.go
aleks-p: confirmed that this works with a local test

pkg/util/http.go
aleks-p: confirmed this works by deploying in a dev environment

pkg/usagestats/seed_test.go
```diff
@@ -37,8 +37,10 @@ func createMemberlist(t *testing.T, port, memberID int) *memberlist.KV {
 	var cfg memberlist.KVConfig
 	flagext.DefaultValues(&cfg)
 	cfg.TCPTransport = memberlist.TCPTransportConfig{
-		BindAddrs: []string{"127.0.0.1"},
-		BindPort:  0,
+		BindAddrs:            []string{"127.0.0.1"},
+		BindPort:             0,
+		MaxConcurrentWrites:  3,
+		AcquireWriterTimeout: 250 * time.Millisecond,
```
aleks-p, line 43: This was a bit hidden and had to track it down with `git bisect` in dskit. It comes from https://github.com/grafana/dskit/pull/525 and works fine with the proper [default values](https://github.com/grafana/dskit/blob/ff33c5829b96e367e485ed3838ae48d5ed567e81/kv/memberlist/tcp_transport.go#L82).

Without this, it runs with a timeout of 0 which breaks things most of the time.

simonswine (approved):
Thanks for getting us up-to-date.


LGTM

PR #4312, opened 2025-07-23
Remove frontend package to fix the build

aleks-p:
A transitive frontend package was flagged for malicious code and its distributions removed. This breaks our builds. I've removed the package that brings the flagged one, as a temporary or potentially permanent workaround.

The UI is still functional in my limited tests, but I could be missing something.

Edit: More details at https://www.npmjs.com/package/stylus?activeTab=readme

### DIFF
--- a/og/package.json
+++ b/og/package.json
@@ -135,7 +135,6 @@
     "sass": "^1.26.10",
     "size-limit": "^6.0.3",
     "typescript": "^4.5.2",
-    "typescript-plugin-css-modules": "^3.4.0",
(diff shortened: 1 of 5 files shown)


korniltsev (approved):

marcsanmi (approved):

marcsanmi:
Maybe we can create a task to further investigate this

simonswine:
> Maybe we can create a task to further investigate this

Done in https://github.com/grafana/pyroscope/issues/4314

PR #4313, opened 2025-07-23
Fix panic when handling error in metastore snapshot compaction

aleks-p:
Fixes a panic in the error handling logic for metastore snapshot compaction.


simonswine (approved):
LGTM

PR #4320, opened 2025-07-25
chore: Add profilecli command to delete v1 blocks

simonswine:
As part of the v1 to v2 migration it is useful to clear out the bucket of old v1 blocks. This command will help to dry run this step and then also run the actual deletion.

The idea is based on a script from @aleks-p.

Generally ideas is you would run this using `kubectl exec` in a pod that has access to the bucket anyhow.

This is what it produced for my test example:

```
$ profilecli admin v2-migration bucket-cleanup --config.file ./cmd/pyroscope/pyroscope.yaml
DRY-RUN: If ran with --dry-run=false, this will delete everything under:
 -  adaptive_placement/phlaredb/
 -  anonymous/phlaredb/
 -  blocks/phlaredb/
 -  segments/phlaredb/
DRY-RUN: If ran with --dry-run=false, this will delete those object store keys:
 -  adaptive_placement/phlaredb/bucket-index.json.gz
 -  anonymous/phlaredb/bucket-index.json.gz
 -  blocks/phlaredb/bucket-index.json.gz
 -  segments/phlaredb/bucket-index.json.gz

$ profilecli admin v2-migration bucket-cleanup --config.file ./cmd/pyroscope/pyroscope.yaml --dry-run=false
This will delete everything under:
 -  adaptive_placement/phlaredb/
 -  anonymous/phlaredb/
 -  blocks/phlaredb/
 -  segments/phlaredb/
Last chance to cancel, waiting 3 seconds...
Deleted object store keys:
 -  adaptive_placement/phlaredb/bucket-index.json.gz
 -  anonymous/phlaredb/bucket-index.json.gz
 -  blocks/phlaredb/bucket-index.json.gz
 -  segments/phlaredb/bucket-index.json.gz
```

### DIFF
--- a/cmd/profilecli/v2-migration.go
+++ b/cmd/profilecli/v2-migration.go
@@ -92,7 +92,7 @@ func v2MigrationBucketCleanup(ctx context.Context, params *v2MigrationBucketClea
 			return nil
 		})
 		if err != nil {
-			return nil
+			return err
 		}
 
 		return nil
@@ -114,10 +114,9 @@ func v2MigrationBucketCleanup(ctx context.Context, params *v2MigrationBucketClea
 		fmt.Println(" - ", path)
 	}
 
-	threads := 16
 	if params.isDryRun() {
 		fmt.Println("DRY-RUN: If ran with --dry-run=false, this will delete those object store keys:")
-		return recurse(ctx, client, threads, func(key string) error {
+		return recurse(ctx, client, func(key string) error {
 			fmt.Println(" - ", key)
 			return nil
 		}, pathsToDelete)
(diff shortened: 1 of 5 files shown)


cmd/profilecli/v2-migration.go
```diff
@@ -0,0 +1,175 @@
...
+	if params.isDryRun() {
+		fmt.Println("DRY-RUN: If ran with --dry-run=false, this will delete everything under:")
+	} else {
+		fmt.Println("This will delete everything under:")
+	}
+	for _, path := range pathsToDelete {
+		fmt.Println(" - ", path)
+	}
+
+	threads := 16
```
marcsanmi: Should we move this to a constant? It'd also help to remove one argument in `recurse`

cmd/profilecli/v2-migration.go
```diff
@@ -0,0 +1,175 @@
...
+			return nil
+		}
+		err := client.Iter(ctx, name, func(name string) error {
+			if strings.HasSuffix(name, "phlaredb/") {
+				pathsToDelete = append(pathsToDelete, name)
+			}
+			return nil
+		})
+		if err != nil {
+			return nil
```
marcsanmi: Should we return the error?

marcsanmi (approved):

PR #4329, opened 2025-07-29
chore: Add tool for examples templating

simonswine:
Our examples/ contain a lot of duplication. This PR aims to make the maintenance a bit easier, by allowing to generate those thourgh a golang "script".

The first edit is to make `tempo` config work with the `Traces Drilldown` app.


aleks-p:
I like this idea, there is definitely a lot of duplication across examples, especially in the way Grafana, Pyroscope and Tempo are configured. Also, great that we are fixing the tempo drilldown app in the same go!

Do you think you have bandwidth to push this forward any time soon?

aleks-p:
closing this since #5248 landed

PR #4334, opened 2025-07-30
docs: Update architecture and os requirements for .NET

simonswine:


simonswine:
Here is the upstream issue btw: https://github.com/DataDog/dd-trace-dotnet/issues/3063#issuecomment-2969839362

korniltsev:
It simplifies development process - no need to do "if amd then download blobs else do nothing." The arm blobs print a message that arm is not supported(instead of silently doing nothing)

korniltsev (approved):

PR #4338, opened 2025-07-31
docs: Update ebpf supported languages content

tiffanyfay:
A few places use the same text so I consolidated it into a shared file. 

Also, several places mention ebpf, but unless someone is more familiar with ebpf if they're looking if Python/c/c++/rust is supported for auto instrumentation, it is several layers deep to find it so I bubbled it up. I also put that frame pointers are needed (except Python)

@Rperry2174 @korniltsev


korniltsev (approved):
<3

simonswine (approved):
Thank you for clarifying this ❤️ 

LGTM

Issue #4332, opened 2025-07-30
Partial profile ingestion


korniltsev:
A user may want to ingest subset of a pprof profile sample types and drop the rest.

One of the use cases is to "reduce cost" -  regular go memory profiles are known to be big due to cumulatie alloc_* sample types nature.  If we drop alloc values, the number of resulted samples would be drastically lower (usually with on disk size  of hundreds of kilobytes instead of multiple megabytes). One of the alternative ways to reduce the size of the memory profile is to use godeltaprf   https://github.com/grafana/pyroscope-go/tree/main/godeltaprof
This was requested by one of Grafana Labs customer(Note: they said they want only inuse_space , and even no inuse_objects)

Another use case is to drop "corrupted" values: for example ingesting delta memory - we may want to drop delta-ed inuse_* values . and when ingesting non delta memory profile we may want to  drop alloc_* values as they are cumulative . This has been drafted in the past. https://github.com/grafana/pyroscope/issues/3979 https://github.com/grafana/alloy/pull/2929 

Dumping ideas how we could achieve per sample type values dropping:
1. Add and expose configuration options to godeltaprof to allow selecting which sample types they want.
2. Modify alloy to allow dropping alloc_ * values instead of computing deltas on them - does not work for godeltaprof.
3. Modify alloy to allow dropping alloc_* values for godeltaprof profiles. 
4. Revive https://github.com/grafana/pyroscope/issues/3979 - modify alloy to include scaler configuration - allowing multiply some values by zero  )
5. introduce per tenant-service configuration to perform dropping on the server
6. ? I secretly hope I missed something and it is possible to achieve sample type dropping somehow without code modification, but I could not find a way yet.

simonswine:
I agree with your assessment, that this is not possible today, I wonder if we should do this in two places:

- On collection side (either alloy or godeltaprof), which will take a while to land everywhere
- Server side (something that is available to everyone who upgrade Pyroscope)

I do think this use case is good candidate for relabelling rules, which could also be something that we expose to alloy's `pyroscope.relabel`.

On the server side the main problem is that we do not split out the different types early on: We wait until we reach ingester/segment writer to do that. (eg. see [here](https://github.com/grafana/pyroscope/blob/f9c4f2cfdd083d0a7f8f79ab5b289835fba73fd6/pkg/segmentwriter/memdb/head.go#L75) for v2). If we would be doing this earlier (before we do the relabeling). We could simply drop them using a relabeling rules.

[...]

PR #4342, opened 2025-08-01
distributor: single profile series processing

korniltsev:
Currently our Push API has a double nested structure. It has multiple series and each series has multiple samples.
Naturally we try to process all of them at the same time.
Some of the callers are using this nesting.
- otel -> multiple pprof series 
- /ingest JFR -> multiple pprof series
- /ingest PPROF -> one pprof series
- Push API allows to have multiple series, multiple sampels, but AFAIK all of our sdks and alloy, uses just one series, one sample

This PR changes

- Refactored data model: Changed from batch processing (`PushRequest` with multiple `ProfileSeries` each containing multiple `ProfileSample`) to single-profile processing (`ProfileSeries` with a single `ProfileSample`)
- Per-profile sampling evaluation: Sampling rules are now evaluated independently for each profile instead of at the batch level, allowing more granular control and earlier rejection of profiles that don't match sampling criteria
- Concurrent processing: Added concurrent processing of individual profiles within a
  batch using goroutines, improving throughput for multi-profile pushes

The benefits of applying this change
- Usage group and throttling correctness (previously thee whole discarded request would be attributed to the one of the usage groups series). Please double check me on this.
- Partial ingestion success. (Imagine an otel profile has 5 profiles and one of them is invalid and previously we rejected all of them). This also aligns nice with https://github.com/open-telemetry/opentelemetry-proto/blob/6deb067376dc017240f940c252b8b116fd3becf1/opentelemetry/proto/collector/profiles/v1development/profiles_service.proto#L61 `ExportProfilesPartialSuccess partial_success = 1;` . Slightly related past discussion https://github.com/grafana/pyroscope/pull/4231#discussion_r2137910356 
- Code is slightly easier to read and reason - there is no double nested loop (no loop at all) in the main logic of the distributor.

I believe this change may simplify solving the https://github.com/grafana/pyroscope/issues/4332. I also believe this is a positive change on it's own.

This is an early draft, please send your opinions

### DIFF
--- a/pkg/ingester/pyroscope/ingest_handler_test.go
+++ b/pkg/ingester/pyroscope/ingest_handler_test.go
@@ -46,11 +46,11 @@ type MockPushService struct {
 func (m *MockPushService) PushBatch(ctx context.Context, req *model.PushRequest) error {
 	if m.Keep {
 		for _, series := range req.Series {
-			rawProfileCopy := make([]byte, len(series.Sample.RawProfile))
-			copy(rawProfileCopy, series.Sample.RawProfile)
+			rawProfileCopy := make([]byte, len(series.RawProfile))
+			copy(rawProfileCopy, series.RawProfile)
 			m.reqPprof = append(m.reqPprof, &flatProfileSeries{
 				Labels:     series.Labels,
-				Profile:    series.Sample.Profile.CloneVT(),
+				Profile:    series.Profile.CloneVT(),
 				RawProfile: rawProfileCopy,
 			})
 		}


pkg/distributor/model/push.go
```diff
@@ -77,69 +81,47 @@ func getProfileLanguageFromSpy(spyName string) string {
...
+		Annotations: req.Annotations,
 	}
 	return c
 }
 
-func (req *PushRequest) ClearAnnotations() {
-	for _, series := range req.Series {
-		series.Annotations = nil
-	}
+func (req *ProfileSeries) ClearAnnotations() {
```
aleks-p: I think we removed the usage of this and we likely don't need it anymore

pkg/distributor/distributor.go
```diff
@@ -241,192 +241,202 @@ func (d *Distributor) stopping(_ error) error {
...
+			itErr := util.RecoverPanic(func() error {
+				return d.pushSeries(ctx, s, req.RawProfileType, tenantID)
+			})
+			errorsMutex.Lock()
+			results[i] = itErr()
+			errorsMutex.Unlock()
+		}()
+	}
+	wg.Wait()
+	// todo return all errors
```
aleks-p: Did you want to use something like `multierror` here or are we good with returning the first error for now?

pkg/distributor/distributor.go
```diff
@@ -992,82 +986,62 @@ func newRingAndLifecycler(cfg util.CommonRingConfig, instanceCount *atomic.Uint3
...
+		usageGroups.CountDiscardedBytes(string(validation.DroppedByRelabelRules), int64(visitor.discardedBytes))
+	}
+
+	if s.DiscardedBytesRelabeling > 0 {
+		validation.DiscardedBytes.WithLabelValues(string(validation.DroppedByRelabelRules), s.TenantID).Add(float64(s.DiscardedBytesRelabeling))
+	}
+	if s.DiscardedProfilesRelabeling > 0 {
+		validation.DiscardedProfiles.WithLabelValues(string(validation.DroppedByRelabelRules), s.TenantID).Add(float64(s.DiscardedProfilesRelabeling))
+	}
+	// todo should we do normalization after relabeling?
```
aleks-p, line 1044: from a quick look I would say yes, perhaps we should create an issue for it?

pkg/distributor/model/push.go
```diff
@@ -35,12 +26,25 @@ type ProfileSample struct {
 	ID         string
 }
 
+// todo better name
```
aleks-p, line 23: Both this and `ProfileSample` are not great names imo. There are samples inside a `ProfileSample.Profile` which makes things extra confusing. 

If we would to start from scratch, we could do
- a `PushRequest` has a `Profiles []*PushProfileRequest` slice
- a `PushProfileRequest` has a `Profile *ProfileWrapper` field
- `ProfileSample` becomes `ProfileWrapper` (not super happy with this name, but I see wrapper to be more accurate than sample)

We could do this or a similar rename separately though.

korniltsev: I feel like we don't even need `ProfileWrapper` it could probably just be inlined.

pkg/distributor/model/push.go
korniltsev: At first I removed the use of the method because I had a slightly different structures

```
type ProfileSeries struct {
	// Caller provided, modified during processing
	Labels []*v1.LabelPair
	Sample *ProfileSample
}

type ProfileSeriesInflightRequest struct {
	// Caller provided, modified during processing
	Labels []*v1.LabelPair
	Sample *ProfileSample

	// todo split
	// Transient state
	TenantID string
	Language string

	Annotations []*v1.ProfileAnnotation

	// always 1
	TotalProfiles          int64
	TotalBytesUncompressed int64

	DiscardedProfilesRelabeling int64
	DiscardedBytesRelabeling    int64
}
```
And the intention was to use `ProfileSeries` as an external provided argument and `ProfileSeriesInflightRequest` as a transient state internally. I made this so that callers don't need to think of what they need to provide. And also the annotations would be no longer part of API.

But then I had to revert this change because the tests do assertions on the added Annotations and I was lazy to rewrite tests. I still think it would be nice to split the API .

korniltsev: Do you remember why we needed ClearAnnotations in the first place? if the `Push` does not pass any annotations down to Push~Parsed~Batch , then it looks like we don't need it at all?

aleks-p: It is there to be more explicit. The annotations field is in the proto, and you are right that it doesn't get propagated from Push to Push~Parsed~Batch right now but that could change and in theory is an external detail for the Push~Parsed~Batch function.

I don't have a strong opinion on this though, you can go ahead and remove the method entirely.

pkg/distributor/distributor.go
korniltsev: to be fair, I wanted to return `([]error, error)` which looks like a horrible API. or `PushResult{ errors[]error }, error`
But we don't need this now. We can do this later. I will remove the todo. The current approach preserves the current behavior by returning just one error. I will need to double check how `multierror` would work with the `connect` errors (what if there are two errors with different codes?)

pkg/distributor/distributor.go
```diff
@@ -241,196 +241,206 @@ func (d *Distributor) stopping(_ error) error {
...
 
 	for _, grpcSeries := range grpcReq.Msg.Series {
-		series := &distributormodel.ProfileSeries{
-			Labels:  grpcSeries.Labels,
-			Samples: make([]*distributormodel.ProfileSample, 0, len(grpcSeries.Samples)),
-		}
 		for _, grpcSample := range grpcSeries.Samples {
 			profile, err := pprof.RawFromBytes(grpcSample.RawProfile)
 			if err != nil {
 				return nil, connect.NewError(connect.CodeInvalidArgument, err)
```
korniltsev: I wonder if this should keep the error around and continue the loop and return the later after all other samples are pushd

pkg/distributor/distributor.go
```diff
@@ -241,196 +241,206 @@ func (d *Distributor) stopping(_ error) error {
...
+	wg := new(sync.WaitGroup)
+	for i, s := range req.Series {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			itErr := util.RecoverPanic(func() error {
+				return d.pushSeries(ctx, s, req.RawProfileType, tenantID)
+			})
+			errorsMutex.Lock()
+			results[i] = itErr()
```
korniltsev: This is a bug, this should not be invoked under lock

pkg/distributor/distributor.go
korniltsev: Let's try multierr https://github.com/grafana/pyroscope/pull/4342/commits/d615aef13117389ca18fd20e54c02d3aa7941da5

pkg/distributor/distributor.go
korniltsev: https://github.com/grafana/pyroscope/pull/4342/commits/d615aef13117389ca18fd20e54c02d3aa7941da5

pkg/distributor/distributor.go
korniltsev: https://github.com/grafana/pyroscope/pull/4342/commits/d615aef13117389ca18fd20e54c02d3aa7941da5

pkg/distributor/distributor.go
korniltsev: https://github.com/grafana/pyroscope/issues/4359

pkg/distributor/distributor.go
```diff
@@ -241,196 +242,204 @@ func (d *Distributor) stopping(_ error) error {
...
+			defer wg.Done()
+			itErr := util.RecoverPanic(func() error {
+				return d.pushSeries(ctx, s, req.RawProfileType, tenantID)
+			})()
+			errorsMutex.Lock()
+			res.Add(itErr)
+			errorsMutex.Unlock()
+		}()
+	}
+	wg.Wait()
```
simonswine: In case of a batch requests (they should be rare), its latency has a good chance to be actually slower than before, as the downstream requests might fall into two different "buffer windows" of segment writer. I don't thinks this is a real problem.

pkg/distributor/distributor.go
```diff
@@ -241,196 +242,204 @@ func (d *Distributor) stopping(_ error) error {
...
-		d.metrics.receivedCompressedBytes.WithLabelValues(string(profName), tenantID).Observe(float64(req.RawProfileSize))
+		d.metrics.receivedCompressedBytes.WithLabelValues(string(profName), tenantID).Observe(float64(req.ReceivedCompressedProfileSize))
+	}
+
+	res := multierror.New()
+	errorsMutex := new(sync.Mutex)
+	wg := new(sync.WaitGroup)
+	for _, s := range req.Series {
+		wg.Add(1)
+		go func() {
```
simonswine, line 318: As this can fully driven by customer input, I suggest we limit the maximum concurrency here

pkg/distributor/distributor.go
```diff
@@ -241,196 +242,204 @@ func (d *Distributor) stopping(_ error) error {
...
+			defer wg.Done()
+			itErr := util.RecoverPanic(func() error {
+				return d.pushSeries(ctx, s, req.RawProfileType, tenantID)
+			})()
+			errorsMutex.Lock()
+			res.Add(itErr)
```
simonswine, line 327: We loose the information which seriesIdx actually failed, maybe it make sense to pre allocate a `[]error slice`, this would also not need a lock.

One day if a client would implement retries, it is important to match the series up which should be retried, otherwise it leads to duplicates.

Also previously we would not error if there are no profiles in one series, but there are in the other, so I think that also needs to be taken care of.

pkg/distributor/distributor.go
korniltsev: In my opinion nothing changed here and we had the same situation before. The concurrency just moved from `sendRequestsToSegmentWriter` to `PushBatch`

pkg/distributor/distributor.go
korniltsev: I think this is a valid concern, however we had no concurrency limits before (in both v1 and v2 routes). I will create an issue and address this as a separate change.

korniltsev: https://github.com/grafana/pyroscope/issues/4366

pkg/distributor/distributor.go
korniltsev: > We loose the information which seriesIdx actually failed, maybe it make sense to pre allocate a []error slice

That was my initial implementation with the slice of errors. https://github.com/grafana/pyroscope/pull/4342#discussion_r2254261729 I then switched to the multierr as there are no users/callers who would benefit from this information (other than maybe otlp, but otlp only needs number of rejected profiles, not indexes) at the moment. Also we still need to return one error to the `Push`  client, maybe we can change that in `pushv2` api if needed. I'd prefer to keep the multierr and it will be easy to change that when/if needed

korniltsev: >Also previously we would not error if there are no profiles in one series, but there are in the other, so I think that also needs to be taken care of.

Are you referring to these nil checks? 
```
	if req.Sample == nil || req.Sample.Profile == nil {
		return noNewProfilesReceivedError()
	}
```
When `PushBatch` is invoked from `Push` - they will not be triggered. When `PushBatch` is invoked from other components (otel/ingest handler) it is considered a programmable error to pass nil there and we just forbid it. 


The only thing that is new is I think this check
```
	if len(req.Series) == 0 {
		return noNewProfilesReceivedError()
	}
```

Can you clarify what error exactly you refer to?

pkg/distributor/distributor.go
korniltsev: > they should be rare

They are actually not that rare. JFR explodes to multiple pprofs. Otel sends a profile per container.

pkg/distributor/distributor.go
simonswine: Sorry yes you are right, I was referring to the first case and missed how `Push` already filters them out.

```
	if req.Sample == nil || req.Sample.Profile == nil {
		return noNewProfilesReceivedError()
	}
```

simonswine: Happy to keep the `multierror`, while I agree there is API client that would programmatically make sense of it, I still think it would useful for human debugging, to know  which series has triggered it. You could maybe wrap errors with that information?

korniltsev: https://github.com/grafana/pyroscope/pull/4342/commits/d17e27fa77ceda80b884d7a6d2d3633b2e12cde0

I've added a series index and ID for easier debugging

korniltsev:
> I would like you to also tests this through ignores in ops before we merge and put it into next week's release.

Run in ops for couple hours with no issues observed

korniltsev:
> Run in ops for couple hours with no issues observed

Nevermind. I've added an error wrapping logic and it paniced with nil deref in tests. [I've made Sample not a pointer.](https://github.com/grafana/pyroscope/pull/4342/commits/d17e27fa77ceda80b884d7a6d2d3633b2e12cde0)

https://github.com/grafana/pyroscope/actions/runs/16823847754/job/47655867533?pr=4342
I will run the latest version in ops again.

pkg/distributor/model/push.go
korniltsev: I removed ProfileSample structure

PR #4344, opened 2025-08-01
chore: Reactivate update-contributors workflow

simonswine:


.github/workflows/update-contributors.yml
```diff
@@ -0,0 +1,53 @@
...
+      - uses: actions/setup-go@v5
+        with:
+          go-version: 1.23.11
+      - name: Update contributors
+        run: make update-contributors
+
+      - name: Commit and push changes
+        run: |
+          git config --local user.name 'Pyroscope Bot'
+          git config --local user.email 'dmitry+bot@pyroscope.io'
```
aleks-p: Looks like we are using the `pyroscope-development-app` above, should we also set this up to use it?

We do this in the Java SDK repo: https://github.com/grafana/pyroscope-java/blob/230eda066b9e20b54537a085d25da35e0001b38d/.github/workflows/release.yml#L100-L118

PR #4345, opened 2025-08-01
chore: Remove oauth2-proxy

simonswine:
The use of oauth2-proxy just for nonce handling is total overkill. Given there is a critical vulnerablility in oauth2-proxy, that will be flagged I rather remove it all together at this point.

https://github.com/oauth2-proxy/oauth2-proxy/blob/c0a928ededa40391baeedf8c5f3e104c047bfb6e/pkg/encryption/cipher.go#L101-L132


simonswine:
Replaces #4335

korniltsev (approved):

PR #4347, opened 2025-08-01
Improve distributor sampling rule evaluation

aleks-p:
Improves handling of overlapping usage groups, when performing sampling in distributors. A sampling rule defined by a more specific usage group now wins over one defined by a dynamic usage group.

A future improvement could be to also take label selectors into account, when deciding if one usage group is more specific than another.

### DIFF
--- a/pkg/validation/usage_groups.go
+++ b/pkg/validation/usage_groups.go
@@ -170,10 +170,7 @@ type UsageGroupMatchName struct {
 }
 
 func (m *UsageGroupMatchName) IsMoreSpecificThan(other *UsageGroupMatchName) bool {
-	if other == nil {
-		return true
-	}
-	return !strings.HasPrefix(m.ConfiguredName, dynamicLabelNamePrefix) && strings.HasPrefix(other.ConfiguredName, dynamicLabelNamePrefix)
+	return !strings.Contains(m.ConfiguredName, dynamicLabelNamePrefix) && strings.Contains(other.ConfiguredName, dynamicLabelNamePrefix)
 }
 
 func (m *UsageGroupMatchName) String() string {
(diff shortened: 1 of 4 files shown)


pkg/validation/usage_groups.go
```diff
@@ -169,6 +169,17 @@ type UsageGroupMatchName struct {
 	ResolvedName   string
 }
 
+func (m *UsageGroupMatchName) IsMoreSpecificThan(other *UsageGroupMatchName) bool {
+	if other == nil {
+		return true
+	}
+	return !strings.HasPrefix(m.ConfiguredName, dynamicLabelNamePrefix) && strings.HasPrefix(other.ConfiguredName, dynamicLabelNamePrefix)
```
korniltsev: should this be `strings.Contains` instead of `strings.HasPrefix` ? I saw usage group tests has `app/${labels.service_name}` test data.

pkg/validation/usage_groups.go
```diff
@@ -169,6 +169,17 @@ type UsageGroupMatchName struct {
 	ResolvedName   string
 }
 
+func (m *UsageGroupMatchName) IsMoreSpecificThan(other *UsageGroupMatchName) bool {
+	if other == nil {
```
korniltsev: Is this nil check needed? Does not look like we ever pass nil here.

korniltsev (approved):
LGTM with one nit regarding`IsMoreSpecificThan` for  `app/${labels.service_name}`

pkg/validation/usage_groups.go
aleks-p: good catch, I've fixed that and added a test

pkg/validation/usage_groups.go
aleks-p: it was needed in an earlier iteration but not anymore, thanks!

PR #4353, opened 2025-08-05
chore: remove e2e tests

korniltsev:
The tests are constantly failing https://github.com/grafana/pyroscope/actions/runs/16747902305/job/47413959084?pr=4342
Let's remove them altogether unless someone has capacity to look into this


simonswine:
Unsure how much value they really provide, I have found a fix for now in #4346. Let's maybe do it the next problem we hit.

Issue #4355, opened 2025-08-06
How to avoid gRPC ResourceExhausted error?


AntoxaBarin:
I have following setup:
- OTLP ebpf profiler
- OTLP Collector
- Pyroscope

When I set a high sampling frequency (10k Hz) in the ebpf profiler, OTLP profiles become large, and I get an error in OTLP Collector:

```
2025-08-06T10:40:28.031+0300    error   internal/queue_sender.go:51     
Exporting failed. Dropping data.       
{
  "resource": {
    "service.instance.id": "05673280-dfe5-4ed0-b039-3d14302ad37b",
    "service.name": "otelcontribcol",
    "service.version": "0.130.0-dev"
  },
  "otelcol.component.id": "otlp",
  "otelcol.component.kind": "exporter",
  "otelcol.signal": "profiles",
  "error": "not retryable error: Permanent error: rpc error: code = ResourceExhausted desc = grpc: received message after decompression larger than max 4194304",
  "dropped_items": 68057
}
```

I run Pyroscope with command:

```
./pyroscope  \
-self-profiling.disable-push=true \
-validation.max-profile-stacktrace-samples=200000 \
-validation.max-profile-size-bytes=33554432 \
-server.grpc-max-recv-msg-size-bytes=33554432 \
-server.grpc-max-send-msg-size-bytes=33554432 \
-distributor.ingestion-rate-limit-mb=100
```

I don't understand why I have 4Mb limit in error message, can I increase it?

korniltsev:
this should probably configured somewhere here as an option https://github.com/grafana/pyroscope/blob/85c16f57da2a00cc426192b06e1ba9866df51249/pkg/ingester/otlp/ingest_handler.go#L54

korniltsev:
Although it may be the otlp colector that you need to configure as well https://github.com/open-telemetry/opentelemetry-collector/blob/19e369e79b465c26601d74801917591a2742ce07/config/configgrpc/README.md?plain=1#L110
```
  receivers:
    otlp:
      protocols:
        grpc:
          endpoint: "0.0.0.0:4317"
          max_recv_msg_size_mib: 32
```

We would probably still need code changes in pyroscope to allow grpc server configuration

korniltsev:
let me know if configuring the collector helps, and I will update pyroscope later

AntoxaBarin:
@korniltsev, thank you for response! Here is my Collector's config:

```yaml
receivers:
  otlp:
    protocols:
      grpc:
[...]

PR #4358, opened 2025-08-07
otel: use server.Config grpc server options

korniltsev:
Should fix https://github.com/grafana/pyroscope/issues/4355


simonswine (approved):
LGTM

PR #4362, opened 2025-08-07
chore: Fix reference in grafana-alloy/ebp/setup-linux.md

Gisson:
There was a link missing from  https://grafana.com/docs/pyroscope/latest/configure-client/grafana-alloy/ebpf/setup-linux/#before-you-begin for the free trial

### DIFF
--- a/docs/sources/configure-client/grafana-alloy/ebpf/setup-linux.md
+++ b/docs/sources/configure-client/grafana-alloy/ebpf/setup-linux.md
@@ -23,7 +23,7 @@ Before you begin, you need:
 - Access to Grafana with the [Grafana Pyroscope data source][pyroscope-ds] provisioned.
 
 {{% admonition type="note" %}}
-If you don't have a Grafana or a Pyroscope server, you can use the [Grafana Cloud][gcloud] free plan to get started.
+If you don't have a Grafana or a Pyroscope server, you can use the [Grafana Cloud][https://grafana.com/auth/sign-up/create-user?pg=pricing&plcmt=free&cta=create-free-account] free plan to get started.
 {{% /admonition %}}
 
 ## Verify system meets the requirements


korniltsev (approved):

docs/sources/configure-client/grafana-alloy/ebpf/setup-linux.md
```diff
@@ -23,7 +23,7 @@ Before you begin, you need:
 - Access to Grafana with the [Grafana Pyroscope data source][pyroscope-ds] provisioned.
 
 {{% admonition type="note" %}}
-If you don't have a Grafana or a Pyroscope server, you can use the [Grafana Cloud][gcloud] free plan to get started.
+If you don't have a Grafana or a Pyroscope server, you can use the [Grafana Cloud][https://grafana.com/auth/sign-up/create-user?pg=pricing&plcmt=free&cta=create-free-account] free plan to get started.
```
aleks-p: I don't think this is the link we intended to have. We do have a link already (defined at the bottom of the page) but it doesn't work due to the limitation described at https://grafana.com/docs/writers-toolkit/write/shortcodes/#admonition

We should either move the reference style link inside the block here, or just use the link directly.

@knylander-grafana is there a preference?

aleks-p: also, we should apply the proper fix to the docker and kubernetes pages

knylander-grafana: The reference link is at the bottom of the file. You can change the link there for the reference if you want to. Otherwise, I'd probably use the link below since it goes to the account creation. 
```suggestion
If you don't have a Grafana or a Pyroscope server, you can use the [Grafana Cloud](https://grafana.com/auth/sign-up/create-user?pg=pricing&plcmt=free&cta=create-free-account) free plan to get started.
```

knylander-grafana:
I fixed the other links to change them from refs to proper links.

PR #4365, opened 2025-08-08
fix issue in pprof split when using `relabel.LabelDrop`

aleks-p:
When we split profiles by sample labels and we have relabeling rules, we first perform the split and then process the relabelling rules. Because relabelling rules can change the groups, at the end we process the groups again. In case of relabeling rules that drop labels, the expected outcome is that we merge samples that were split back together.

This PR fixes a subtle issue where we don't preserve the internal state between the processing of individual groups during the merge, resulting in extra processing and inconsistent state.

Ideally we should avoid splitting samples by labels that are getting dropped, however this is an edge case and handling it more efficiently might be too complex (relabelling rules can do modifications that are hard to anticipate).

TODO: Add a test and a benchmark that covers this scenario


bryanhuhta (approved):
LGTM

korniltsev:
wow, I would never spot this issue just by looking at the code :0

aleks-p:
> wow, I would never spot this issue just by looking at the code :0

Yeah, it is quite hidden and I only noticed it during debugging.

PR #4367, opened 2025-08-11
distributor: add new metrics

korniltsev:
Suggesting new metrics for these issues

https://github.com/grafana/pyroscope-squad/issues/557
https://github.com/grafana/pyroscope-squad/issues/346

1. `distributor_received_decompressed_bytes_total` - incremented for each received profile as soon we try processing the profile, before any ratelimit/sampling checks
2. `distributor_processed_decompressed_bytes` - incremented after ratelimit/sampling checks at the exit of the processing routine with `defer` (note : potentially after normalization if the profile is valid)

TODO: add tests

### DIFF
--- a/pkg/distributor/metrics.go
+++ b/pkg/distributor/metrics.go
@@ -17,6 +17,9 @@ type metrics struct {
 	receivedSamplesBytes      *prometheus.HistogramVec
 	receivedSymbolsBytes      *prometheus.HistogramVec
 	replicationFactor         prometheus.Gauge
+
+	receivedDecompressedBytesTotal *prometheus.HistogramVec
+	processedDecompressedBytes     *prometheus.HistogramVec
 }
 
 func newMetrics(reg prometheus.Registerer) *metrics {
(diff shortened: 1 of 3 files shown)


pkg/distributor/metrics.go
```diff
@@ -71,6 +74,24 @@ func newMetrics(reg prometheus.Registerer) *metrics {
 			},
 			[]string{"type", "tenant"},
 		),
+		receivedDecompressedBytesTotal: prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Namespace: "pyroscope",
+				Name:      "distributor_received_decompressed_bytes_total",
```
aleks-p, line 80: I expect this to be confused with the existing metric because they have an overlapping name, but I can't think of a better name. Perhaps once we validate that this metric works the way we want, we can remove or rename the existing one (potentially breaking dashboards in the process).

aleks-p (approved):
LGTM

PR #4370, opened 2025-08-12
feat: Set querier.max-flamegraph-nodes-max default to 1,000,000

bryanhuhta:
closes https://github.com/grafana/pyroscope-squad/issues/560

This config value previously defaulted to 0 (unlimited). In retrospect, it's valuable for it to be set to some (quite high) limit to stop unmanageably large values of `maxNodes` to be sent in query requests.

This will apply to any endpoint which supports `maxNodes` EXCEPT `/SelectMergeProfile`. See 

https://github.com/grafana/pyroscope/blob/b201c2c3954cd19b5be3036967ba239f25c22540/pkg/frontend/readpath/queryfrontend/query_select_merge_profile.go#L32-L34

### DIFF
--- a/pkg/validation/limits.go
+++ b/pkg/validation/limits.go
@@ -180,7 +180,7 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&l.MaxProfileSymbolValueLength, "validation.max-profile-symbol-value-length", 65535, "Maximum length of a profile symbol value (labels, function names and filenames, etc...). Profiles are not rejected instead symbol values are truncated. 0 to disable.")
 
 	f.IntVar(&l.MaxFlameGraphNodesDefault, "querier.max-flamegraph-nodes-default", 8<<10, "Maximum number of flame graph nodes by default. 0 to disable.")
-	f.IntVar(&l.MaxFlameGraphNodesMax, "querier.max-flamegraph-nodes-max", 0, "Maximum number of flame graph nodes allowed. 0 to disable.")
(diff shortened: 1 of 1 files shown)


cmd/pyroscope/help.txt.tmpl
```diff
@@ -160,7 +160,7 @@ Usage of ./pyroscope:
   -querier.max-flamegraph-nodes-default int
     	Maximum number of flame graph nodes by default. 0 to disable. (default 8192)
   -querier.max-flamegraph-nodes-max int
-    	Maximum number of flame graph nodes allowed. 0 to disable.
+    	Maximum number of flame graph nodes allowed. 0 to disable. (default 1000000)
```
simonswine: Nit: Feels slightly inconsistent: Let's make it consistent either decimal or power of twos everywhere.  So `1<<20` for the max if we mimic the default with`1<<13`)

simonswine (approved):
LGTM

cmd/pyroscope/help.txt.tmpl
bryanhuhta: Sure, I can update that. I initially chose an even 1,000,000 because my personal experience has been that its easier to quickly identify 1000000 being "about a million" as opposed to 1048576 being "about a million" and that 1,000,000 indicates an arbitrary maximum and not a carefully selected number better.

fixed in https://github.com/grafana/pyroscope/pull/4370/commits/0246d6315ef124c09e804fe70cb4f13f7590ae1c

PR #4371, opened 2025-08-13
fix(update): Insecure TLS configuration

ptrgits:
fix the problem, update the TLS configuration in the `NewInClusterK8sClient` function so that the minimum TLS version is set to `tls.VersionTLS12` instead of `tls.VersionTLS10`. This change should be made directly in the assignment to the `MinVersion` field of the `tls.Config` struct on line 89. No other changes are required, as the rest of the TLS configuration is appropriate. The fix only requires changing the constant used for the minimum version.

#### References
Wikipedia: [Transport Layer Security](https://en.wikipedia.org/wiki/Transport_Layer_Security)
Mozilla: [Security/Server Side TLS](https://wiki.mozilla.org/Security/Server_Side_TLS)
OWASP: [Transport Layer Protection Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Transport_Layer_Protection_Cheat_Sheet.html)


korniltsev:
image
Thank you

PR #4375, opened 2025-08-14
feat: annotate sampled profiles

aleks-p:
Annotating profiles that are subject to distributor sampling. Similar to #3956, with a bit of refactoring to make things cleaner. The annotations will allow sampling to be visualized in profiling UIs and in the future adjust metrics generated from profiles (cc @alsoba13).


pkg/distributor/annotation/throttling.go
```diff
@@ -1,12 +1,23 @@
...
 )
 
-func CreateTenantAnnotation(c *Config) ([]byte, error) {
-	annotation := &ProfileAnnotation{
+type ThrottledAnnotation struct {
+	PeriodType        string `json:"periodType"`
+	PeriodLimitMb     int    `json:"periodLimitMb"`
+	LimitResetTime    int64  `json:"limitResetTime"`
+	SamplingPeriodSec int    `json:"samplingPeriodSec"`
+	SamplingRequests  int    `yaml:"samplingRequests"`
```
korniltsev: yaml?

korniltsev (approved):
lgtm

pkg/distributor/annotation/throttling.go
korniltsev:
```suggestion
	SamplingRequests  int    `json:"samplingRequests"`
```

aleks-p: whoops, good thing is it hasn't been used yet :)

PR #4376, opened 2025-08-15
feat(distributor): sample type relabeling

korniltsev:
- Update validation, to treat nil profile, a profile with no sample types, and a profile with nil sample types as invalid.
- Return `ValidatedProfile` form the validation routine - an attempt to pass the information that the profile is validated and  string references (and other references ) are safe to do with no bounds check
- Add sample type relabeling. In one of our heap testdata goes from 847138 to 46178,

Implements https://github.com/grafana/pyroscope/issues/4332


pkg/model/sampletype/relabel.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/grafana/pyroscope/pkg/validation"
+
+	"github.com/prometheus/prometheus/model/relabel"
+
+	googlev1 "github.com/grafana/pyroscope/api/gen/proto/go/google/v1"
+	phlaremodel "github.com/grafana/pyroscope/pkg/model"
+	phlarerelabel "github.com/grafana/pyroscope/pkg/model/relabel"
+)
+
+func Relabel(p validation.ValidatedProfile, rules []*relabel.Config, series []*typesv1.LabelPair) {
```
aleks-p: nit, `series` -> `labels`?

docs/sources/configure-server/reference-configuration-parameters/index.md
```diff
@@ -2221,6 +2221,23 @@ distributor_usage_groups:
...
+#   received with an label 'environment="secrets"' and the second rule will add
+#   a label 'powered_by="Grafana Labs"' to all profile series.
+#   sample_type_relabeling_rules:
+#       - action: drop
+#         regex: secret
+#         source_labels:
+#           - environment
+#       - action: replace
+#         replacement: grafana-labs
+#         target_label: powered_by
```
simonswine: I think we should find a better example here. As this is not very relevant for the problem at hand.

pkg/validation/limits.go
```diff
@@ -63,6 +63,8 @@ type Limits struct {
 	IngestionRelabelingRules                RelabelRules         `yaml:"ingestion_relabeling_rules" json:"ingestion_relabeling_rules" category:"advanced"`
 	IngestionRelabelingDefaultRulesPosition RelabelRulesPosition `yaml:"ingestion_relabeling_default_rules_position" json:"ingestion_relabeling_default_rules_position" category:"advanced"`
 
+	SampleTypeRelabelingRules RelabelRules `yaml:"sample_type_relabeling_rules" json:"sample_type_relabeling_rules" category:"advanced"`
```
simonswine: We should maybe have a validation check when unmarshalling to restrict relabelling rules of actions other than `drop` and `keep`.

simonswine (approved):
Once the two comments are addressed, happy for this to go ahead

aleks-p (approved):
looks good to me as well

PR #4380, opened 2025-08-19
feat: Support stacktrace selector for `SelectMergeStacktraces` requests

bryanhuhta:
closes https://github.com/grafana/pyroscope-squad/issues/561


simonswine:
Taking this work over from @bryanhuhta.

This adds support for the `stack_trace_selector` we already support  on `SelectMergeProfiles` in the `SelectMergeStacktraces` call.

The resolving of symbols can happen in two ways:

- Take the full symbol parent pointer tree and add samples to them
- Symbolise as you go through the samples

This PR implements the selector for both variants.

bryanhuhta (commented):
I can't approve since this is my PR, but LGTM! I think this approach is reasonable and the added tests are a nice touch to bolster confidence in correctness.

simonswine (approved):
LGTM

PR #4391, opened 2025-08-25
jetbrains: commit run configurations for v1 and v2

korniltsev:
I recently started using git work trees more and now for each corktree I create new run configurations.

I usually do not commit run configurations to not confuse non-jet brains users, but this time thinking it may be a good idea.

Jetbrains themselves seem to commit some of their common run configurations https://github.com/JetBrains/intellij-community/tree/master/.idea/runConfigurations so it dos not look like something completely crazy.

Curious what other jetbrains users think and what running configurations they have

This PR adds v1 and v2 run configurations with embedassets go tag.

This is how to save one, btw
Screenshot 2025-08-25 at 2 13 50 PM

### DIFF
--- a/.idea/runConfigurations/help.xml
+++ b/.idea/runConfigurations/help.xml
@@ -1,0 +1,13 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="help" type="GoApplicationRunConfiguration" factoryName="Go Application">
+    <module name="pyroscope" />
+    <working_directory value="$PROJECT_DIR$" />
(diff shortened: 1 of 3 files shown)


aleks-p (approved):
I think these can be helpful and I don't see any harm in adding them either.

Personally I copy `.idea/` when I use worktrees, as I have multiple configurations with different sets of arguments.

korniltsev:
Commit yours and update mine please. I will be happy to save a bunch of clicks :)

Issue #4388, opened 2025-08-23
Document push api RawSample.ID field


korniltsev:
1. It seems it's currently undocumented if the ID should be UUID. Later down the ingestion pipeline we seem to expect it to be UUID. We should document if the ID field should be UUID and what happens if it is empty or not UUID.

2. It looks like we currently ignore (do not pas further then infesters) the user provided ID field and only maybe include in the distributor errors.
We generate new UUID when pass a profile down the ingestion pipeline, both for ingesters and segment writers. I think it makes sense to use the user provided ID for at least one case when we send profiles to segment writers and there is no pprof split.

korniltsev-grafanista:
https://github.com/grafana/pyroscope-squad/issues/648

PR #4393, opened 2025-08-25
Extract signal correlation information from OTEL profiles

fandreuz:
In this PR I add code to extract `profile_id` and `span_id` from OTEL profiles when available. This makes features from grafana/grafana#77537 available for OpenTelemetry too in Pyroscope.


pkg/ingester/otlp/convert.go
```diff
@@ -43,8 +56,13 @@ func ConvertOtelToGoogle(src *otelProfile.Profile, dictionary *otelProfile.Profi
 			}
 			svc2Profile[svc] = p
 		}
-		if _, err := p.convertSampleBack(sample, dictionary); err != nil {
+		if gs, err := p.convertSampleBack(sample, dictionary); err != nil {
 			return nil, err
+		} else if applyProfileId {
+			gs.Label = append(gs.Label, &googleProfile.Label{
+				Key: p.addstr(pprof.ProfileIDLabelName),
+				Str: p.addstr(hex.EncodeToString(src.GetProfileId())),
```
korniltsev: 1. I am not sure I understand why we apply profile Id to each sample, can we just return the profile id along with the profile?
2. if we still need this code, then can we rewrite the code to do the hex encoding once, not for each sample?

korniltsev: https://github.com/grafana/pyroscope/blob/cbf9983a98a24389d7c9853817b942264a0ab472/pkg/pprof/pprof.go#L966 
I think `profile_id` label name is just a legacy name for the `span_id`. and we only need just one the latter one. Have you tested your change? Did you need to add this?

pkg/ingester/otlp/convert.go
```diff
@@ -26,8 +28,19 @@ func at[T any](arr []T, i int32) (T, error) {
 	return zero, fmt.Errorf("index %d out of bounds", i)
 }
 
+func allZeros(arr []byte) bool {
```
korniltsev: nit: maybe bytes.Equal(src.GetProfileId(), zeroProfileId[:]) is slightly better

pkg/ingester/otlp/convert.go
```diff
@@ -377,6 +395,18 @@ func (p *profileBuilder) convertSampleAttributesToLabelsBack(os *otelProfile.Sam
...
 	}
+
+	if os.GetLinkIndex() != 0 {
+		link, err := at(dictionary.LinkTable, os.GetLinkIndex())
+		if err != nil {
+			return fmt.Errorf("could not access link at index %d: %w", os.GetLinkIndex(), err)
+		}
+		gs.Label = append(gs.Label, &googleProfile.Label{
+			Key: p.addstr(pprof.SpanIDLabelName),
+			Str: p.addstr(hex.EncodeToString(link.GetSpanId())),
```
korniltsev, line 390: nit note (nothing to change here): usually, we only add a root span only, and this is slightly different. At the same time It is too late and there is no way to find the root span id. We may need to solve/change this in the future.

pkg/ingester/otlp/convert.go
fandreuz: In Async-Profiler we only set `span_id` and `trace_id`, and that's enough to make it work. I added this snippet to handle `profile_id` too to avoid discarding the information, I'll remove it if you feel that's unnecessary.

> I am not sure I understand why we apply profile Id to each sample, can we just return the profile id along with the profile?

From what I understand, only `Sample` can have `Label`s in pprof, is that right?

korniltsev: > From what I understand, only Sample can have Labels in pprof, is that right?
Yes. 
I was referring to this https://github.com/grafana/pyroscope/blob/9b7d9dbf788b5e66abd78bffe385189aab5f9400/api/push/v1/push.proto#L43 or https://github.com/grafana/pyroscope/blob/9b7d9dbf788b5e66abd78bffe385189aab5f9400/pkg/distributor/model/push.go#L31

In practice the fields are ignored(regenerated by the distributors at the moment https://github.com/grafana/pyroscope/issues/4388 ) so I think we can just ignore it here as well and remove it.

fandreuz: 5f5c6f6fb27191cbade60b21c00893408d84fb03

pkg/ingester/otlp/convert.go
fandreuz: 877d1e7bee267b1e44edb4cf96e5793a37bc4de9

korniltsev (approved):

korniltsev:
it would be nice to add a test though

fandreuz:
> it would be nice to add a test though

Hi @korniltsev, I tried adding something to the other OTLP integration tests (test data is in one of the [old commits](https://github.com/grafana/pyroscope/pull/4393/commits/b393cdf353ad1f94b2ae9ce98ba2b578bbce4f8b)). Unfortunately, I wasn't able to user `labelSelector` to extract `span_id` labels from ingested spans. I got empty queries with all the tests I made.

However, I tested this branch at multiple levels:
- debug log statements, the label seems to be correctly set
- signal correlation with Tempo and Pyroscope also works

If you are open to help, I could push my attempts and we see what's wrong.

PR #4397, opened 2025-08-27
frontend: remove legacy frontend

korniltsev:
There Is a `isGrafanaFlamegraphEnabled` which is always `true`
And if it is not set to true it used `@pyroscope/legacy/flamegraph` .

PR: Remove the flag and the legacy renderers.


simonswine (approved):
LGTM

PR #4399, opened 2025-08-28
fix: Log errors when hitting ingest limits

bryanhuhta:
When investigating a rate limit, I realized we don't log errors that cause the rate limiter to be tripped. These errors contain some helpful info like what was the value of the limit hit and when it will reset.


korniltsev (approved):

bryanhuhta:
Superseded by https://github.com/grafana/pyroscope/pull/4401

PR #4404, opened 2025-09-01
Cherry pick Aleksandar readpath patches

korniltsev:
cherry-picking P1 and P2 as Aleksandar suggested https://github.com/grafana/pyroscope-squad/issues/568#issuecomment-3238278899


simonswine (approved):
LGTM.

I would be surprised the swissmaps do anything in 1.24. But I guess we can compare and remove afterwards

PR #4406, opened 2025-09-02
chore: Validate recording rule's metric name (expect prefix)

alsoba13:
We will check in backend that metric name field in recording rules do start with the `pyroscope_exported_metric_` prefix.

### DIFF
--- a/pkg/model/recording_rule.go
+++ b/pkg/model/recording_rule.go
@@ -18,7 +18,7 @@ type RecordingRule struct {
 	FunctionName   string
 }
 
-const metricNamePrefix = "pyroscope_exported_metric_"
+const metricNamePrefix = "profiles_recorded_"
 
 func NewRecordingRule(rule *settingsv1.RecordingRule) (*RecordingRule, error) {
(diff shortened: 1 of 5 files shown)


pkg/model/recording_rule.go
```diff
@@ -17,15 +18,17 @@ type RecordingRule struct {
 	FunctionName   string
 }
 
+const metricNamePrefix = "pyroscope_exported_metric_"
```
alsoba13: I'm happy to hear proposals here. I feel we lack some reference to recording rules, maybe `pyroscope_recorded_metric_` makes more sense

simonswine: I think tempo uses `traces_` as a prefix. I also don't like the word metric in there, as everyhting stored in Prometheus/Mimir is a metric.

So I am proposing:

```
profiles_recorded_
```

Which could result in things like:

```
profiles_recorded_cpu_usage_total
profiles_recorded_memory_allocations_total
```

simonswine (approved):
LGTM

PR #4408, opened 2025-09-03
WIP: Generate API docs for connect-go API

simonswine:


api/gen/proto/go/push/v1/push.pb.go
oleg-kozlyuk-grafana, line 1: Generic question: why is a generated file committed to the repo?

api/gen/proto/go/push/v1/push.pb.go
```diff
@@ -180,7 +183,7 @@ type RawSample struct {
 	state protoimpl.MessageState `protogen:"open.v1"`
 	// raw_profile is the set of bytes of the pprof profile
 	RawProfile []byte `protobuf:"bytes,1,opt,name=raw_profile,json=rawProfile,proto3" json:"raw_profile,omitempty"`
-	// unique ID of the profile
+	// unique UUID of the profile
```
oleg-kozlyuk-grafana: nit: second U in UUID already stands for Unique

tools/api-docs-generator/main.go
oleg-kozlyuk-grafana, line 1: Suggestion: an explanation for what this tool does would not hurt, either as command-line `help` param, or a README

api/go.mod
```diff
@@ -2,10 +2,11 @@ module github.com/grafana/pyroscope/api
 
 go 1.24.0
 
-toolchain go1.24.8
```
oleg-kozlyuk-grafana, line 5: Is downgrade intended?

docs/sources/reference-server-api/index.md
```diff
@@ -1,15 +1,13 @@
 ---
 description: Learn about the Pyrocope server API
```
marcsanmi:
```suggestion
description: Learn about the Pyroscope server API
```

api/connect-openapi/gen/segmentwriter/v1/push.openapi.yaml
```diff
@@ -0,0 +1,185 @@
...
+      additionalProperties: false
+    types.v1.ProfileAnnotation:
+      type: object
+      properties:
+        key:
+          type: string
+          title: key
+        value:
+          type: string
+          title: value
```
marcsanmi, line 183: Is it worth adding description for these properties? or they are already self-explanatory enough?

api/gen/proto/go/push/v1/push.pb.go
simonswine: In it is a very common pattern in Go, basically when doing dependency management using `go mod`, not all repo files are included (tests non go files are removed). So often it is impossible to execute the code generation commands then.

There for it is common to commit the generated files (With all of the problems this causes)

api/go.mod
simonswine: Thanks for spotting, fixed in https://github.com/grafana/pyroscope/pull/4408/commits/ba833b1ac806d560045a5f736b524ed25614e1ed

api/connect-openapi/gen/segmentwriter/v1/push.openapi.yaml
simonswine: They actually can't be used, made that clearer by hiding them from the API docs bd2e5c4bc5279f240a53ef0df503e66a829b0ab1 and adding comments

tools/api-docs-generator/main.go
simonswine: Done both

api/gen/proto/go/push/v1/push.pb.go
simonswine: Fixed everywhere I could find it

PR #4413, opened 2025-09-04
fix: correct HasFunctions flags for mixed symbolization profiles

marcsanmi:
### Summary

Fixes a flamegraph truncation bug in profiles where enabling symbolization would cause some flamegraph blocks to disappear.

### Root Cause

The issue occurred with mixed symbolization profiles - profiles containing both symbolized and unsymbolized locations within the same mapping. The symbolizer would leave `HasFunctions=true` on these mappings (inherited from upstream), which is incorrect for mixed symbolization.

During profile normalization, `clearAddresses()` removes addresses from locations when `HasFunctions=true`, assuming all locations are symbolized. For mixed mappings, this incorrectly cleared addresses from unsymbolized locations, making them indistinguishable during profile merging (identical LocationKeys: `{address: 0, lines: [], mappingId: X}`).

### Solution

Added `HasFunctions` flag correction logic that:

  1. **Collects symbolization stats** for each mapping by counting total vs symbolized locations
  2. **Corrects HasFunctions flags** based on actual symbolization state:
     - `HasFunctions=true` only when ALL locations in a mapping are symbolized
     - `HasFunctions=false` for mixed or unsymbolized mappings

Mixed mappings now preserve addresses during normalization, ensuring correct deduplication during merge operations.

### DIFF
--- a/pkg/symbolizer/symbolizer.go
+++ b/pkg/symbolizer/symbolizer.go
@@ -70,53 +70,67 @@ func (s *Symbolizer) SymbolizePprof(ctx context.Context, profile *googlev1.Profi
 		}
 		mappingsToSymbolize[uint64(i+1)] = true
 	}
-	if len(mappingsToSymbolize) == 0 {
-		return nil
-	}
-
-	locationsByMapping, err := s.groupLocationsByMapping(profile, mappingsToSymbolize)
-	if err != nil {
-		return fmt.Errorf("grouping locations by mapping: %w", err)
-	}
-
+	var allSymbolizedLocs []symbolizedLocation
 	stringMap := make(map[string]int64, len(profile.StringTable))
 	for i, str := range profile.StringTable {
 		stringMap[str] = int64(i)
 	}
 
-	var allSymbolizedLocs []symbolizedLocation
-
-	for mappingID, locations := range locationsByMapping {
-		mapping := profile.Mapping[mappingID-1]
-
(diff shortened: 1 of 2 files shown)


marcsanmi:
> I also think we need to consider changing the address clearing as part of normalizatoin only to happen for nodes that are in fact symbolised.

I'll create a task for this 👍

korniltsev-grafanista:
I'd like to have a test (integration test?) that actually reproduces the issue with the `clearAddress` invocation - I don't understand the order of the invocations - to me it looks like `clearAddress` and `Normalize` is only called by distributors, right? and symbolization is happening later after the profile is normalized and stored in the database, right?  and in the test `TestFlamegraphTruncationIntegration` it happens in the other order.
I am not saying the fix is incorrect, I just don't understand what's broken and how we fixed it. I am probably missing something.

korniltsev-grafanista:
https://github.com/grafana/pyroscope/pull/4444

marcsanmi:
### ⚠️ PR scope clarification

This PR only fixes `HasFunctions` flags for profiles processed by **query-time symbolization** - it ensures accurate metadata when symbolized profiles are exported or reprocessed. It does **NOT** fix profiles damaged during ingestion (the most common flamegraph truncation case) since symbolization runs after normalization. Might be still valuable for data integrity and future-proofing if symbolization moves to ingestion time.

PR #4416, opened 2025-09-05
fix: `pyroscope/build` and `profilecli/build` had undefined rules in Makefile

jkramer-52:
```
$ make pyroscope/build
make: *** No rule to make target `go/bin-pyroscope', needed by `pyroscope/build'.  Stop.
$ make profilecli/build
make: *** No rule to make target `go/bin-profilecli', needed by `profilecli/build'.  Stop.
```


simonswine (approved):
LGTM

PR #4417, opened 2025-09-05
feat: Helm support for v1/v2 storage

simonswine:
This PR intends to add initial support for deploying the v2 storage layer using helm.

While a goal is to use this to migrate from v1/v2, this has only worked for the single-binary migration for me v1-v2.

I do think this should not impact users that only use v1, so I think this has reached a point where I would like to merge it.

Note: I also had to upgrade helm-docs to a later version in order to ignore some settings from the README.md, this in turn meant had to modify the helm-lint to be included in this repo.  (I also couldn't see the value provided by super-linter and therefore removed it)

Subtasks:

- [x] Has flags for microservices yes/no
- [x] Has the ability to deploy v1 and/or v2.
- [x] Setup the folders and service discovery for v2.
- [x] Ideally those flags should support migration from v1 to v2, (do for single binary)

Next steps:

- [ ] Fix `make deploy-monitoring` (separate PR)
- [ ] Update documentation about microservices part of values.yaml (separate PR)
- [ ] Update documentation about how to deploy and migrate v2 (separate PR)


docs/internal/V2-MIGRATION.md
```diff
@@ -0,0 +1,95 @@
...
+  ./operations/pyroscope/helm/pyroscope \
+  --install \
+  --set architecture.storage.v1=true \
+  --set architecture.storage.v2=false \
+  --set pyroscope.persistence.enabled=true
+```
+
+### Step: 2: Now deploy v2 and enable dual ingest
+
+- Note: The python command will switch to the v2 write path 10 minutes after it is run.
```
alsoba13: What about this?
```suggestion
- Note: The python command will switch to the v2 read path for data ingested 10 minutes after it is run.
```

docs/internal/V2-MIGRATION.md
```diff
@@ -0,0 +1,95 @@
...
+helm upgrade \
+  pyroscope \
+  ./operations/pyroscope/helm/pyroscope \
+  --set architecture.storage.v1=true \
+  --set architecture.storage.v2=true \
+  --set pyroscope.persistence.enabled=true \
+  --set architecture.storage.migration.queryBackendFrom=$(python3 -c "import datetime; print((datetime.datetime.now(datetime.UTC)+ datetime.timedelta(minutes = 10)).strftime('%Y-%m-%dT%H:%M:%SZ'))")
+```
+
+### Step 3: Now remove v1 components, this will loose all data before Step 2
```
alsoba13: "Now" sounds too immediately after step 2. I would reword to something like this:
```suggestion
### Step 3: Remove v1 components
Once data before Step 2 is no longer relevant, we can get rid of the v1 components. This will loose all data before Step 2.
```

docs/internal/V2-MIGRATION.md
```diff
@@ -0,0 +1,95 @@
...
+  --install \
+  --set architecture.microservices.enabled=true \
+  --set architecture.overwriteResources.requests.cpu=10m \
+  --set minio.enabled=true \
+  --set architecture.storage.v1=true \
+  --set architecture.storage.v2=false \
+  --set pyroscope.persistence.enabled=true
+```
+
+### Step: 2: Now deploy v2 and enable dual ingest
```
alsoba13: both previous comments apply here as well

marcsanmi:
Do you think it'd be worth adding helm template validation to ensure incompatible configurations aren't allowed? I was thinking in a separate `_validation.tpl` file, but this might be too much for now.

simonswine:
> Do you think it'd be worth adding helm template validation to ensure incompatible configurations aren't allowed? I was thinking in a separate `_validation.tpl` file, but this might be too much for now.

Yes I do think it is worth, but I think it is tricky what to add and what to leave, what have you thought off?

Please add it to https://github.com/grafana/pyroscope/issues/4473

PR #4421, opened 2025-09-09
optimization: avoid merging single profiles

aleks-p:
There are code paths where we end up merging a single profile. While in most cases the overhead is low, this is not the case when we are dealing with hundreds of thousands of samples. 

Example:

```
var pm pprof.ProfileMerge
profile := xxx // can come from different sources
pm.merge(profile)
return p.Profile()
```

We end up doing such merges when
- resolving symbols, and there is a single partition
- aggregating query reports, and there is a single report

End-to-end benchmark, with individual profiles having ~100-500k samples, ~50-100k locations, ~10-20k unique strings:

```
goos: darwin
goarch: arm64
pkg: github.com/grafana/pyroscope/pkg/querybackend
cpu: Apple M2 Pro
                                      │ reference.txt │           optimized.txt           │
                                      │    sec/op     │   sec/op    vs base               │
_Invoke_QueryPprof/with_raw_output-10     8.485 ± 14%   7.861 ± 4%  -7.35% (p=0.004 n=10)

                                      │ reference.txt │            optimized.txt            │
                                      │     B/op      │     B/op      vs base               │
_Invoke_QueryPprof/with_raw_output-10    14.84Gi ± 2%   13.48Gi ± 3%  -9.18% (p=0.000 n=10)

                                      │ reference.txt │            optimized.txt            │
                                      │   allocs/op   │  allocs/op   vs base                │
_Invoke_QueryPprof/with_raw_output-10     73.24M ± 0%   65.79M ± 0%  -10.17% (p=0.000 n=10)
```

### DIFF
--- a/pkg/phlaredb/symdb/resolver.go
+++ b/pkg/phlaredb/symdb/resolver.go
@@ -257,6 +257,24 @@ func (r *Resolver) Tree() (*model.Tree, error) {
 func (r *Resolver) Pprof() (*googlev1.Profile, error) {
 	span, ctx := opentracing.StartSpanFromContext(r.ctx, "Resolver.Pprof")
 	defer span.Finish()
+
+	if len(r.p) == 1 {
+		// skip the merge path when there is a single partition to avoid the overhead
+		var p *googlev1.Profile
+		err := r.withSymbols(ctx, func(symbols *Symbols, appender *SampleAppender) error {
+			resolved, err := symbols.Pprof(ctx, appender, r.maxNodes, SelectStackTraces(symbols, r.sts))
+			if err != nil {
+				return err
+			}
+			p = resolved
+			return nil
+		})
+		if err != nil {
+			return nil, err
+		}
+		return p, nil
+	}
+
 	var p pprof.ProfileMerge
 	err := r.withSymbols(ctx, func(symbols *Symbols, appender *SampleAppender) error {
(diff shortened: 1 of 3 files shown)


pkg/phlaredb/symdb/resolver.go
```diff
@@ -286,6 +304,18 @@ func (r *Resolver) withSymbols(ctx context.Context, fn func(*Symbols, *SampleApp
...
 }
 
+func (r *Resolver) canSkipProfileMerge() bool {
+	if len(r.p) > 1 {
+		return false
+	}
+	if r.sts != nil && r.sts.GoPgo != nil && r.sts.GoPgo.AggregateCallees {
+		// we rely on merges to implement GoPgo.AggregateCallees
+		return false
+	}
```
aleks-p, line 314: We might be able to avoid this by changing how aggregateCallees is implemented. To keep the blast radius small, I'm leaving that for a future iteration.

simonswine: Probably best to capture in an issue if you not creating the PR right away 🙂

simonswine (approved):
LGTM

Tried this e2e locally with the pgo optimzed profiles and the full pprof. I also noticed it makes the integration tests a bit faster.

pkg/phlaredb/symdb/resolver.go
aleks-p: good idea, https://github.com/grafana/pyroscope/issues/4424

PR #4422, opened 2025-09-10
Add ability to use flat query plan for pprof queries

aleks-p:
**Background**

When we are dealing with large amounts of data, the current nested query plan increases the overhead for serialization. Each query backend that does merges has to send and receive data over the wire. Because of the nested nature of the plan, the same data gets serialized/deserialized multiple times.

I am experimenting with flattening out the query plan, particularly for `SelectMergeProfile` which is the one endpoint that can deal with large amounts of data (for it to really matter).

**Idea**

Query frontend nodes fan out individual block-level queries to query backend nodes. Query backend replicas only do reads (and local merges as needed). The query frontend then merges all the data back together. This increases the resource requirements for query frontends, though we are overall saving on serialization/deserialization costs. We could be creating many routines for long queries, so we might need to consider adding a concurrency limit.

**Testing**

I've verified it produces the same results with the integration test, as well as manual tests in our internal environments. 

Keeping it as a draft to do some more stress testing, but feel free to let me know why this might be a bad idea :)

### DIFF
--- a/pkg/frontend/readpath/queryfrontend/query_flat_plan.go
+++ b/pkg/frontend/readpath/queryfrontend/query_flat_plan.go
@@ -125,10 +125,7 @@ func (q *QueryFrontend) queryFlatPlan(
 		return nil, err
 	}
 
-	aggregated, err := reportAggregator.response()
-	if err != nil {
-		return nil, err
-	}
+	aggregated := reportAggregator.response()
 
 	if aggregated == nil || len(aggregated.Reports) == 0 {
 		return nil, nil
--- a/pkg/frontend/readpath/queryfrontend/report_aggregator.go
+++ b/pkg/frontend/readpath/queryfrontend/report_aggregator.go
@@ -64,18 +64,6 @@ func newAggregator(request *queryv1.QueryRequest) *reportAggregator {
 	}
 }
 
-func (ra *reportAggregator) aggregateResponse(resp *queryv1.InvokeResponse, err error) error {
-	if err != nil {
-		return err
-	}
-	for _, r := range resp.Reports {
-		if err = ra.aggregateReport(r); err != nil {
-			return err
-		}
-	}
-	return nil
-}
-
 func (ra *reportAggregator) aggregateReport(r *queryv1.Report) (err error) {
 	if r == nil {
 		return nil
@@ -116,7 +104,7 @@ func (ra *reportAggregator) aggregateReportNoCheck(report *queryv1.Report) (err 
 	return a.aggregate(report)
 }
 
-func (ra *reportAggregator) response() (*queryv1.QueryResponse, error) {
+func (ra *reportAggregator) response() *queryv1.QueryResponse {
 	// if there are staged reports, we can just add them, no need to aggregate because there is one per type
 	reports := make([]*queryv1.Report, 0, len(ra.staged))
 	for _, st := range ra.staged {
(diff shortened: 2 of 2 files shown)


pkg/frontend/readpath/queryfrontend/report_aggregator.go
```diff
@@ -0,0 +1,157 @@
+package queryfrontend
```
aleks-p, line 1: Note, this is copied from query backend and adapted. We might benefit from a "report aggregation" package that is shared between the 2, though there are minor differences in the types used.

pkg/frontend/readpath/queryfrontend/query_flat_plan.go
```diff
@@ -0,0 +1,142 @@
...
+			})
+			if err != nil {
+				return err
+			}
+			for _, report := range resp.Reports {
+				if report.ReportType == queryType {
+					sp.LogFields(
+						log.String("msg", "got report"),
+						log.String("report_type", queryType.String()),
+						log.String("msg_size", fmt.Sprint(resp.SizeVT())),
```
aleks-p, line 114: `resp.SizeVT()` here can be quite expensive to calculate, see if we can do it in a different way or skip if not

pkg/frontend/readpath/queryfrontend/query_flat_plan.go
```diff
@@ -0,0 +1,142 @@
...
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	metastorev1 "github.com/grafana/pyroscope/api/gen/proto/go/metastore/v1"
+	queryv1 "github.com/grafana/pyroscope/api/gen/proto/go/query/v1"
+	"github.com/grafana/pyroscope/pkg/querybackend"
+)
+
+func (q *QueryFrontend) queryFlatPlan(
```
korniltsev-grafanista, line 20: is it hard to put the logic for query plan creation into `queryplan` package?

korniltsev-grafanista:
I wonder if instead of a boolean flag we could have parametrized knobs passing to the plan building routine

https://github.com/grafana/pyroscope/blob/0920249772c3215b7c6642522ac7d94051a1d6c0/pkg/querybackend/queryplan/query_plan.go#L23

i.e. with an override set maxReads=1 maxMerges=1000000000 

Noticed there is a todo, not sure if it meant to allow overrides or maybe something more sophisticated
```
// TODO(kolesnikovae): Should be dynamic.
```

aleks-p:
> I wonder if instead of a boolean flag we could have parametrized knobs passing to the plan building routine
> 
> https://github.com/grafana/pyroscope/blob/0920249772c3215b7c6642522ac7d94051a1d6c0/pkg/querybackend/queryplan/query_plan.go#L23
> 
> i.e. with an override set maxReads=1 maxMerges=1000000000
> 
> Noticed there is a todo, not sure if it meant to allow overrides or maybe something more sophisticated
> 
> ```
> // TODO(kolesnikovae): Should be dynamic.
> ```

We could reshape the query plan based on the amount of data queried, number of replicas and other similar heuristics, I think the TODO refers to that.

The issue I was tackling is that for large payloads the nested query plan creates too much overhead and we can improve things by not using a query plan at all. It is similar with a plan with a single merge and unlimited reads, however that would add one more hop in the process, compared to doing the merge in the query frontend which this PR does.

This needs further considerations and testing though, so I will close it for now.

PR #4427, opened 2025-09-11
chore: Updated workflows to use self-hosted agents

oleg-kozlyuk-grafana:
* Agent size guessed approximately based on workflow contents. Feedback on incorrect sizing is welcome
* Note: after merge, a change in "required checks" in repo settings is needed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,7 @@ jobs:
         run: make generate check/unstaged-changes
   test:
     strategy:
+      fail-fast: false
       matrix:
         arch: [amd64, arm64]
     runs-on: ubuntu-${{ matrix.arch }}-large


.github/workflows/ci.yml
```diff
@@ -45,13 +45,10 @@ jobs:
         run: make generate check/unstaged-changes
   test:
     strategy:
+      fail-fast: false
```
simonswine, line 48: This is  so that it doens't skip a test for the other architecture anymore when one fails?
❤️

oleg-kozlyuk-grafana: Yep, helps quite a bit esp. if you have many configs

oleg-kozlyuk-grafana:
> No idea how small and large relates to the previous sizes, but we will find out

Default GitHub runners have 4vCPU and 16 GB now, so default (no suffix) self-hosted agent like `ubuntu-x64` matches it. `-small` has 2vCPU (like GitHub runners before 2023) and `-large` - 8vCPU

Ref: 
https://docs.github.com/en/actions/reference/runners/github-hosted-runners

simonswine:
Seems all sensible, let me know if you need someone to edit the required checks?

oleg-kozlyuk-grafana:
I won't have required access for a bit, so it would be appreciated; otherwise, I'll merge & adjust the settings once permissions come through

PR #4430, opened 2025-09-11
Reduce memory allocations when rendering diff flamegraphs

aleks-p:
When we render large diff flamegraphs, we create quite a bit of allocations due to prepending to a slice instead of appending. We also use an older and less efficient way of truncating the diff tree, I've replaced that with the one used for truncating regular (non-diff) trees

Screenshot 2025-09-11 at 18 07 44

Screenshot 2025-09-11 at 18 07 55

```
goos: darwin
goarch: arm64
pkg: github.com/grafana/pyroscope/pkg/model
cpu: Apple M2 Pro
                      │ reference-diff.txt │         optimized-diff.txt          │
                      │       sec/op       │   sec/op     vs base                │
_NewFlamegraphDiff-10         3956.7m ± 3%   252.0m ± 7%  -93.63% (p=0.000 n=10)

                      │ reference-diff.txt │          optimized-diff.txt          │
                      │        B/op        │     B/op      vs base                │
_NewFlamegraphDiff-10       18313.8Mi ± 0%   749.1Mi ± 0%  -95.91% (p=0.000 n=10)

                      │ reference-diff.txt │         optimized-diff.txt          │
                      │     allocs/op      │  allocs/op   vs base                │
_NewFlamegraphDiff-10          1.489M ± 0%   1.053M ± 0%  -29.33% (p=0.000 n=10)
```


pkg/model/flamegraph_diff.go
```diff
@@ -259,18 +265,29 @@ func combineMinValues(leftTree, rightTree *Tree, maxNodes int) uint64 {
...
+			h = minheap.Push(h, maxVal)
+		}
+		return true
 	})
-	return c.MinValue()
+
+	if len(h) < maxNodes {
+		return 0
+	}
+	return uint64(h[0])
```
aleks-p, line 290: this is based on `model.Tree.minValue()`, adapted for 2 trees

aleks-p:
> Wonder how "good" your benchmarks were, but might make sense to add them.

I can't add the data I used for the benchmark, but I can likely capture the improvement with synthetic data as well. I will add the benchmark code before merging.

Edit: done

PR #4437, opened 2025-09-16
chore: TODO cleanup: use buffer pool instead of generic

oleg-kozlyuk-grafana:


pkg/segmentwriter/segment.go
```diff
@@ -274,7 +274,7 @@ func (s *segment) flushBlock(stream flushStream) (*bufferpool.Buffer, *metastore
 		Datasets:        make([]*metastorev1.Dataset, 0, len(stream.heads)),
 	}
 
-	blockFile := bufferpool.GetBuffer(len(stream.heads) * 2 / 1024) // rough estimate, 2kb per head
+	blockFile := bufferpool.GetBuffer(2 * 1024) // reserve extra to avoid resizes
```
korniltsev-grafanista, line 277: Should we maybe remember the size of the buffer between flushes?

Also 2k is likely too short

Not sure if you have access https://ops.grafana-ops.net/goto/bey8h9q72mu4gd?orgId=1

image

Do you know why was it divided by 1024? Was it a bug?

oleg-kozlyuk-grafana: The division was a typo - sorry about that, already fixed before comment. 

Good point about buffer size - I tried to take a safe route, but it looks like - esp. considering how `bufferpool` works - it won't work. Will rethink

PR #4438, opened 2025-09-16
20250916 adhoc use stacktrace tree

simonswine:
- **chore: Implement adhoc upload through model.StacktraceTree**
- **chore: Ensure adhoc API request get logged**
- **Keep state between loops per type**
- **Add benchmark**

### DIFF
--- a/pkg/og/structs/flamebearer/convert/convert.go
+++ b/pkg/og/structs/flamebearer/convert/convert.go
@@ -5,15 +5,15 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	profilev1 "github.com/grafana/pyroscope/api/gen/proto/go/google/v1"
 	"path"
 	"reflect"
 	"strconv"
 	"strings"
-	"time"
 	"unicode"
 
-	"github.com/grafana/pyroscope/pkg/og/agent/spy"
+	profilev1 "github.com/grafana/pyroscope/api/gen/proto/go/google/v1"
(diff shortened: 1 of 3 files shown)


pkg/og/structs/flamebearer/convert/convert.go
```diff
@@ -187,37 +187,96 @@ func JSONToProfile(b []byte, name string, maxNodes int) ([]*flamebearer.Flamebea
...
 		return nil, fmt.Errorf("parsing pprof: %w", err)
 	}
+
+	t := model.NewStacktraceTree(int(maxNodes * 2))
+	stack := make([]int32, 0, 64)
+	m := make(map[uint64]int32)
+
 	fbs := make([]*flamebearer.FlamebearerProfile, 0)
-	for _, stype := range tree.SampleTypes(p) {
-		sampleRate := uint32(100)
```
marcsanmi, line 197: The new implementation uses hardcoded sample rates in `ExportToFlamebearer` instead of reading the actual rate and calculate it dynamically. Not sure whether this could cause significant differences compared to how it was done until now.

marcsanmi (approved):
LGTM

pkg/og/structs/flamebearer/convert/convert.go
simonswine: I do think in both cases the values are not correct, even the default of 100 is also wrong, as the UI uses it as 1/s, while the backend as Hz.

I have captured this here #4440

PR #4439, opened 2025-09-17
Update OpenTelemetry protocol to 1.8.0

oleg-kozlyuk-grafana:


pkg/ingester/otlp/convert.go
```diff
@@ -420,13 +410,32 @@ func (p *profileBuilder) convertMappingBack(om *otelProfile.Mapping, dictionary
...
 		Filename:        p.addstr(filenameLabel),
 		BuildId:         p.addstr(buildID),
-		HasFunctions:    om.HasFunctions,
-		HasFilenames:    om.HasFilenames,
-		HasLineNumbers:  om.HasLineNumbers,
-		HasInlineFrames: om.HasInlineFrames,
+		HasFunctions:    true,
+		HasFilenames:    true,
+		HasLineNumbers:  true,
+		HasInlineFrames: true,
```
simonswine, line 416: This feels okay to me (after #4413), but @marcsanmi please double check

pkg/test/integration/testdata/README.md
```diff
@@ -0,0 +1,124 @@
+## Data generation
```
simonswine, line 1: ❤️  Thanks for adding this

Eventually after this merged, we might wanna extend this to support https://github.com/grafana/pyroscope/issues/4450

pkg/ingester/otlp/convert.go
marcsanmi: It seems reasonable to me because of the braking change. However, to make it clear, my fix didn't really address the issue about mixed symbolziation at ingest time, I added a clarifying comment in my closed [PR](https://github.com/grafana/pyroscope/pull/4413#issuecomment-3323409682)

The main change here is that we'll now always flag OTLP profiles as `HasFunctions=true` at ingest time, which means they'll always go through symbolization at query time. If OTLP profiles contain
mixed symbolization (some locations symbolized, others not), this change could cause flamegraph truncation since addresses get cleared during normalization. So, making the current issue more visible.

@korniltsev I'd appreciate your thoughts on this, correct me whether I'm wrong regarding the otel profiler.

marcsanmi: I've tested in dev and ops with the [fix](https://github.com/grafana/pyroscope/pull/4413) and with the same Alloy versions and symbolization works. I'd say, to test this in dev (maybe it's already done) but for me this is a LGTM. I'll prepare a revert for that fix.

The next step would be to change our no longer valid logic of `HasUnsymbolizedProfiles` on ingest path and the corresponding logic in the symbolizer (read path).

marcsanmi: Created this task https://github.com/grafana/pyroscope-squad/issues/580

marcsanmi:
LGTM as per https://github.com/grafana/pyroscope/pull/4439#discussion_r2375110890

pkg/ingester/otlp/convert.go
korniltsev-grafanista: I would like us to understand the issue (disappeared flamegraphs) and have a [test](https://github.com/grafana/pyroscope/pull/4444)  reproducing issue. 

- I do understand the `clearAddress` issue https://github.com/grafana/pyroscope-squad/issues/570 
- I don't understand what caused flamegraphs to **completely** disappear here https://raintank-corp.slack.com/archives/C048VSWPZ4P/p1756900469097559
- I don't understand how https://github.com/grafana/pyroscope/pull/4413 fixes anything.

marcsanmi (dismissed):
LGTM as per https://github.com/grafana/pyroscope/pull/4439#discussion_r2375110890. 

Thanks for taking care of this 👍

As @korniltsev-grafanista pointed out, if we merge this we won't symbolize any profiles coming from otel. See https://github.com/grafana/pyroscope/blob/487f757a25d3371a3360aabc3216c0223a95ca1f/pkg/symbolizer/symbolizer.go.

pkg/ingester/otlp/convert.go
marcsanmi: > I don't understand what caused flamegraphs to completely disappear here https://raintank-corp.slack.com/archives/C048VSWPZ4P/p1756900469097559

Me neither. I cannot reproduce it right now, but I'll further investigate whether it's related to timeouts when query public debuginfodserver...

> I don't understand how https://github.com/grafana/pyroscope/pull/4413 fixes anything.

I tested it and this didn't fix the original issue. It tricked me since once I applied my fix it started working back in dev, but apparently was unrelated to symbolization... I'll revert it.

korniltsev-grafanista: > I don't understand what caused flamegraphs to completely disappear here https://raintank-corp.slack.com/archives/C048VSWPZ4P/p1756900469097559

I still don't know the exact root cause, but I found the combination of these flags reprodcues the problem `-querier.typed-pprof-enabled=true -symbolizer.enabled=true`

> and have a test https://github.com/grafana/pyroscope/pull/4444 reproducing issue.

Ignore this comment regarding having a test since the flag that breaks the query is not in main.

pkg/ingester/otlp/convert.go
```diff
@@ -420,13 +410,32 @@ func (p *profileBuilder) convertMappingBack(om *otelProfile.Mapping, dictionary
 		FileOffset:      om.FileOffset,
 		Filename:        p.addstr(filenameLabel),
 		BuildId:         p.addstr(buildID),
-		HasFunctions:    om.HasFunctions,
-		HasFilenames:    om.HasFilenames,
-		HasLineNumbers:  om.HasLineNumbers,
-		HasInlineFrames: om.HasInlineFrames,
+		HasFunctions:    true,
```
korniltsev-grafanista, line 413: I think this break our server symbolizer, specifically here https://github.com/grafana/pyroscope/blob/fac6eccc13a4b2d0bc8ead4287133d9b182f3bc9/pkg/symbolizer/symbolizer.go#L68

If we set `HasFunctions:    true` then these profiles will be exempt from symbolization according to the current logic.

Some potential fixes:
- Set HasFunctions to true, only if every single location referencing the mapping has Lines
- Change the symbolizer to not look into the `HasFunctions` field and symbolize those mappings that have no Lines
- smth else?

korniltsev-grafanista (changes requested):
We need to address symbolization problem with `HasFunctions=true`
Otherwise looks good to me.

pkg/ingester/otlp/convert.go
marcsanmi: I've created a [task](https://github.com/grafana/pyroscope-squad/issues/580) this morning kind of reflecting this. Feel free to extend it as needed.

oleg-kozlyuk-grafana:
@korniltsev-grafanista @marcsanmi I have tested latest upstream `ebpf-profiler` with 1.8.0 and current state (`HasFunctions=true`) and see no issues with frames not being symbolized. Please take a look again

image
Screenshot 2025-10-01 at 10 30 33

pkg/ingester/otlp/convert.go
oleg-kozlyuk-grafana: I've tested latest eBPF profiler and the symbols are correctly provided by upstream. Unless there is a specific scenario where we expect this not to be the case, I propose we merge this as-is and execute the HasFunctions refactoring separately?

oleg-kozlyuk-grafana: I've tested profiling of Go, JavaSrcipt and native code and I don't see any issues with symbolization. Unless there is a specific scenario I'm missing, shall we merge this as-is and execute the HasFunction refactoring separately?

marcsanmi: Yes, I imagine the issue might come when we receive unsymbolized profiles. If everything is symbolized it works, but at the time we got some unsymbolized profile from otel, they won't be symbolized becuase of the `hasFunctions=true`.

PR #4442, opened 2025-09-17
DRAFT feat: Allow utf-8 characters in label names

jake-kramer:
This feature is a relaxation in previous validation for label names. Given the backward incompatible nature of this change (for example, PromQL queries must be [updated to support utf-8 label names](https://prometheus.io/docs/guides/utf8/#querying)), it is gated behind a "client capability": a key/val that a client must set in the `Accept:` header.


pkg/featureflags/client_capability.go
```diff
@@ -0,0 +1,52 @@
+package featureflags
```
jake-kramer, line 1: Wasn't sure if this was the most appropriate place for this logic or not -- open to suggestions

pkg/featureflags/client_capability.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"fmt"
+	"mime"
+	"net/http"
+)
+
+const (
+	// Capability names
+	AllowUtf8LabelNamesCapabilityName = "allow-utf8-labelnames"
```
jake-kramer: Presumably want to document all these -- where would be best place for that?

pkg/featureflags/client_capability.go
```diff
@@ -0,0 +1,52 @@
...
+func ParseClientCapabilities(header http.Header) ([]*ClientCapability, error) {
+	acceptHeader := header.Get("Accept")
+	if acceptHeader != "" {
+		if _, params, err := mime.ParseMediaType(acceptHeader); err != nil {
+			return nil, err
+		} else {
+			capabilities := make([]*ClientCapability, 0, len(params))
+			seenCapabilityNames := make(map[string]struct{})
+			for k, v := range params {
+				// Check for duplicates
```
jake-kramer: Realizing now this logic is silly given key names in a map are unique 🤦

jake-kramer (commented):
Still have tests todo, but wanted to get this draft out for comment sooner rather than later

pkg/featureflags/client_capability.go
```diff
@@ -0,0 +1,52 @@
...
+	AllowUtf8LabelNamesCapabilityName = "allow-utf8-labelnames"
+)
+
+type ClientCapability struct {
+	Name  string
+	Value string
+}
+
+func ParseClientCapabilities(header http.Header) ([]*ClientCapability, error) {
+	acceptHeader := header.Get("Accept")
```
simonswine: There can be multiple values for a single header key, so you should use `header.Values(...)` instead

pkg/distributor/distributor.go
```diff
@@ -248,7 +249,14 @@ func (d *Distributor) Push(ctx context.Context, grpcReq *connect.Request[pushv1.
 		Series:         make([]*distributormodel.ProfileSeries, 0, len(grpcReq.Msg.Series)),
 		RawProfileType: distributormodel.RawProfileTypePPROF,
 	}
+
 	allErrors := multierror.New()
+
+	clientCapabilities, parseErr := featureflags.ParseClientCapabilities(grpcReq.Header())
```
simonswine: I don't think you can use that logic on the ingest path. 

The API clients (allloy, profilecli upload, SDK) need to figure out if the backend side support utf8. For them to know that, they need to at least one round trip to figure it out. 

For now we should just accept them no matter what. Then when it comes to implement this in the client we need to figure out how we can implement this backward compatbile.

Always try with utf8, if you get a validation error, no longer send utf8 for the rest of the session.

I don't think this should be part of this issue (and if only for porfilecli). 

We should focus on the query path: There the Pyroscope backend needs to know if the client supports utf8, that works well with the accept header approach.

pkg/featureflags/client_capability.go
simonswine: This is kind of overlapping with #4408, let's do it in there once this is merged.

pkg/distributor/distributor.go
jake-kramer: Ok, my understanding of changes required:

- Write path should be updated to be permissive to all (valid) utf-8 label names
- Read path should:
    * Return raw label name if client indicates utf-8 support via header
    * Rewrite label name (with same logic we used to do on write) if client does not indicate utf-8 support

> Always try with utf8, if you get a validation error, no longer send utf8 for the rest of the session.

I'm not understanding what components you're referring to here.. which session and what is raising a validation error?

jake-kramer:
### I've verified

- Writes work as expected
- HTTP and gRPC middleware both work as expected
- Reads work with the OG UI
- Writes work with profilecli

### Breaking changes
1. If a client writes a utf-8 label name with this new logic, but reads the label name with the old logic, they will see utf-8 label names.
2. Previously, writing `[labelName_1: val, labelName.1: val]` would resolve to just `[labelName_1: val]` on write [perhaps this was a logic bug?]. With new query logic, but ut8 label names disabled, `labelName.1` would convert to `labelName_1` and then an error would be returned for duplicate label names.

### Open questions cc @simonswine @bryanhuhta 
1. Is implementing this logic for only the v2 `LabelNames` and `Series` APIs sufficient?
3. Any issues with the breaking changes above this PR introduces?

### TODOs
- Testing
- Documentation
- Integrating against other clients

alsoba13:
TL;DR: Worth mentioning that recording rules are not 100% compatible with UTF-8. We should validate exported label names and metric name.

Extended:
So recording rules may target some profiles to be recorded. Those are aggregated and exported through [remote_write v1](https://prometheus.io/docs/specs/prw/remote_write_spec/). There's an experimental remote write v2 version that can handle that, but we are uncertain about mimir's plans to support this. I think best approach right now is to not support utf8 in recording rules.

From recording rules model https://github.com/grafana/pyroscope/blob/main/api/settings/v1/recording_rules.proto#L56-L92:
* `metric_name` should still be `[a-zA-Z_:]([a-zA-Z0-9_:])*`
* `matchers` should be able to handle utf8 `{"service.name"="foo"}` should be okay.
* `group_by` should still be `[a-zA-Z_:]([a-zA-Z0-9_:])*`. Those are label names to export, so `["service.name"]` must NOT be supported.
* `external_labels` label names should still be `[a-zA-Z_:]([a-zA-Z0-9_:])*`.

I propose to validate those so we make sure our rules are ok. We have 2 places to validate this, first at [recording rule creating time](https://github.com/grafana/pyroscope/blob/main/pkg/settings/recording/recording.go#L251-L303) (UI sending a POST request to tenant-settings), and compaction workers fetching rules from tenant-settings at compaction time, basically wrapped in [this constructor](https://github.com/grafana/pyroscope/blob/main/pkg/model/recording_rule.go#L36-L90). Maybe you find that some of the requirements already meet (I remember using some validations for the metric_names, for example).

jake-kramer:
## Current State of the PR

- **Write behavior**:  
  - All writes (v1 and v2) now only perform UTF-8 label name validation.  
  - Previously, label names were restricted to `[a-zA-Z_.:]([a-zA-Z0-9_.:])*` (with `"."` transformed to `"_"`).  

- **Read behavior (v2 APIs)**:  
  - Behavior depends on the `allow-utf8-labelnames` client capability set in the `Accept` header:  
    - **Enabled** → Return label names as written.  
    - **Disabled** → Fallback to the legacy logic (reject invalid names, transform `"."` → `"_"`).  

---

## Known Issues

1. **Mixed read/write semantics during rollout**  
   - If a client writes a UTF-8 label name (new logic) but reads it with the old logic, they will see UTF-8 label names.  
   - This is expected during deployment and can be mitigated by feature flags that delay enabling new writes until rollout is complete.  

2. **Not implemented for v1 APIs**  
   - Assumption: these APIs are not actively supported—can we confirm?  

3. **Label selectors and compatibility**  
   - Any API that uses a `label_selector` may break if queries assume `"."` were transformed to `"_"`, regardless of if the client enabled the `allow-utf8-labelnames` client capability.  

---

## Proposed Modification

**Change writes to also respect the `allow-utf8-labelnames` client capability.**. This would imply existing/old write logic would remain unless this client capability was enabled.

- **Pros**:  
  - Seamless deployment path without need for a runtime feature flag.  
  - Avoids breaking queries (e.g., `label_selector`) for clients that haven’t yet updated their capability setting.  

- **Cons**:  
  - Expands the compatibility matrix (UTF-8 on/off × read/write paths).  
  - Increases the risk of confusion if clients inconsistently set the capability across different calls.

marcsanmi:
My two cents about it:

While client capabilities provide granular control, I think we should keep a server-side feature flag as a kill switch ideally to allow us to:
- Emergency rollback without code changes

> Assumption: these APIs are not actively supported—can we confirm?

I think the v1 APIs should be supported because:
- OSS users may still be using the traditional architecture (V2=false)
- AFAIK, no deprecation timeline has been communicated for V1 frontend

> Label selectors and compatibility

I'm not sure about the best approach in a mixed-data scenario... I see:
1. Dual query approach: "Temporarily" query both formats
2. Data migration: Background job to normalize historical label names
3. Query translation: Automatically try both formats in label selectors
... Or we could just break

I believe the key decision is whether we want a clean migration (store UTF-8 only, handle compatibility at read/query time) or safer migration (store based on client capability, accept permanent mixed data); do we want to support both formats coexisting at the same time for a given user...?

pkg/frontend/readpath/queryfrontend/query_label_names.go
```diff
@@ -46,5 +47,16 @@ func (q *QueryFrontend) LabelNames(
...
 	}
-	return connect.NewResponse(&typesv1.LabelNamesResponse{Names: report.LabelNames.LabelNames}), nil
+
+	labelNames := report.LabelNames.LabelNames
+	if capabilities, ok := featureflags.GetClientCapabilities(ctx); !ok || !capabilities.AllowUtf8LabelNames {
+		// Use legacy label name sanitization if utf8 label names not enabled
+		if labelNames, err := validation.SanitizeLabelNames(labelNames); err != nil {
+			return nil, connect.NewError(connect.CodeInvalidArgument, err)
+		} else {
+			return connect.NewResponse(&typesv1.LabelNamesResponse{Names: labelNames}), nil
```
simonswine: Instead of sanitising we should filter out all label names, that are not passing the "legacy validation".

pkg/frontend/readpath/queryfrontend/query_series_labels.go
```diff
@@ -54,5 +55,30 @@ func (q *QueryFrontend) Series(
 	if report == nil {
 		return connect.NewResponse(&querierv1.SeriesResponse{}), nil
 	}
-	return connect.NewResponse(&querierv1.SeriesResponse{LabelsSet: report.SeriesLabels.SeriesLabels}), nil
+
+	seriesLabels := report.SeriesLabels.SeriesLabels
```
simonswine: Instead of sanitising on the result of the query, which will be complicated to do correctly (label collisions, merging of already existing series).  

We should filter on the request (so before `q.querySingle`).

When c.MsgLabelName is empty (meaning show me all series). We should do a LabelNames call and filter out any non legacy compliant names and then pass them on. 

When there `c.MsgLabelNames` given we should filter them so they only contain legacy compliant ones.

public/app/components/TagsBar.tsx
```diff
@@ -230,7 +230,7 @@ function LabelsSubmenu({
 
 // Identifies whether a label is in a query or not
 function isLabelInQuery(query: string, label: string, labelValue: string) {
-  return query.includes(`${label}="${labelValue}"`);
+  return query.includes(`"${label}=${labelValue}"`);
```
simonswine: I would do this in separate PR to keep the scope as tight as possible.

I also think this would be the correct way to double-quote utf8 names in selectors:

```suggestion
  return query.includes(`"${label}"="${labelValue}"`);
```

pkg/featureflags/client_capability.go
```diff
@@ -0,0 +1,119 @@
...
+		// Convert metadata to http.Header for reuse of existing parsing logic
+		httpHeader := make(http.Header)
+		for key, values := range md {
+			// gRPC metadata keys are lowercase, HTTP headers are case-insensitive
+			httpHeader[http.CanonicalHeaderKey(key)] = values
+		}
+
+		// Reuse existing HTTP header parsing
+		// TODO add metrics = # requests like this and # clients [need
+		//  labels for requests and clients/tenet and user agent(?)]
```
simonswine: Metrics like this are likely very costly (high cardinality), so I would advise against it. If we need anything like this we already should log those headers. (Maybe doublecheck that works as exptected, but we are setting `-server.log-request-headers`)

cmd/profilecli/client_test.go
```diff
@@ -0,0 +1,79 @@
...
+		ClientCapabilities []string
+		Want               []string
+	}{
+		{
+			Name:   "empty header adds capability",
+			Header: http.Header{},
+			ClientCapabilities: []string{
+				"allow-utf8-labelnames=true",
+			},
+			Want: []string{"*/*; allow-utf8-labelnames=true"},
```
simonswine: According to the examples in https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept, I think it might be more common to have no space:

```suggestion
			Want: []string{"*/*;allow-utf8-labelnames=true"},
```

cmd/profilecli/client_test.go
```diff
@@ -0,0 +1,79 @@
...
+		},
+		{
+			Name: "existing header appends capability",
+			Header: http.Header{
+				"Accept": []string{"application/json"},
+			},
+			ClientCapabilities: []string{
+				"allow-utf8-labelnames=true",
+			},
+			Want: []string{"application/json", "*/*; allow-utf8-labelnames=true"},
```
simonswine: I do think it would be better to join the given content-types with the client capabilities.
```suggestion
			Want: []string{"application/json;allow-utf8-labelnames=true"},
```

cmd/profilecli/client_test.go
```diff
@@ -0,0 +1,79 @@
...
+		},
+		{
+			Name: "multiple existing values appends capability",
+			Header: http.Header{
+				"Accept": []string{"application/json", "text/plain"},
+			},
+			ClientCapabilities: []string{
+				"allow-utf8-labelnames=true",
+			},
+			Want: []string{"application/json", "text/plain", "*/*; allow-utf8-labelnames=true"},
```
simonswine: Not too sure if I am miss understanding https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept:

```suggestion
				"Accept": []string{"application/json,text/plain"},
			},
			ClientCapabilities: []string{
				"allow-utf8-labelnames=true",
			},
			Want: []string{"application/json;allow-utf8-labelnames=true,text/plain;allow-utf8-labelnames=true"
```

jake-kramer:
Taking this PR out of draft mode 🎉 

This PR implements a client capability framework, with `allow-utf8-labelnames` as the first capability. Capabilities are flags clients set in the `Accept` header to inform the API about specific client support. This is useful for cross-API features (like utf-8 support).

**This PR does not provide full utf-8 label name support.** Instead, it offers the first pass at what will be full support, with follow up PRs to provide full support (expanded on below).

This PR respects the `allow-utf8-labelnames` client capability in the read path (both v1 and v2 `LabelNames` and `Series` APIs). If the capability is not set or not set to `true`, these APIs will filter out "invalid" label names (i.e. with chars outside of `[a-zA-Z0-9_.]`) from the response. This logic is currently a no-op, given it's not yet possible to write label names outside of this charset.

A future PR will disable the write path label name sanitization if the `allow-utf8-labelnames` client capability (gated behind a per-tenant feature flag).

Another future PR will update the OG UI and profilecli clients to take advantage of this client capability.

cc @simonswine @bryanhuhta

pkg/featureflags/client_capability.go
jake-kramer: I'm considering adding a counter metric with low cardinality labels (tenant + capability name).. any concerns?

pkg/frontend/readpath/queryfrontend/query_series_labels.go
```diff
@@ -9,9 +9,54 @@ import (
...
+
+	filtered := make([]string, 0, len(labelNames))
+	for _, name := range toFilter {
+		if _, _, ok := validation.SanitizeLegacyLabelName(name); !ok {
+			level.Debug(q.logger).Log("msg", "filtering out label", "label_name", name)
+			continue
+		}
+		filtered = append(filtered, name)
+	}
+	return filtered, nil
```
simonswine, line 56: Is there any reason why we can't call `LabelNames()` and rely on the implementation there?

pkg/querier/querier_test.go
```diff
@@ -211,6 +416,130 @@ func Test_Series(t *testing.T) {
...
+			allowUtf8LabelNames: false,
+			setCapabilities:     true,
+			requestLabelNames:   []string{},
+			expectedLabelNames:  []string{"bar", "foo"},
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			foobarlabels := phlaremodel.NewLabelsBuilder(nil).Set("foo", "bar")
```
simonswine: Ideally this should contain utf8 label names

And the we should have a test case that tests it:
```
		{
			[...]
            allowUtf8LabelNames: true,
			requestLabelNames:   []string{},
			expectedLabelNames:  []string{"bar", "foo","日本語"},
		},
```

simonswine (approved):
I do think this is very close to LGTM, a couple of suggestions, after that I am happy for you to merge this.

pkg/featureflags/client_capability.go
simonswine: Yes that's makes a lot of sense, probably a good follow up PR

pkg/frontend/readpath/queryfrontend/query_series_labels.go
jake-kramer: I thought I saw examples of other APIs using this approach for internal calls, but now I can't find that. Happy to use the frontdoor if it works.

jake-kramer: Also curious: any potential issues with passing a potentially unbounded list of labelNames here?https://github.com/grafana/pyroscope/blob/b2888fae7257f2264df1ca1c739dda132b9b1c9c/pkg/frontend/readpath/queryfrontend/query_series_labels.go#L91-L101

jake-kramer:
> TL;DR: Worth mentioning that recording rules are not 100% compatible with UTF-8. We should validate exported label names and metric name.
> 
> Extended: So recording rules may target some profiles to be recorded. Those are aggregated and exported through [remote_write v1](https://prometheus.io/docs/specs/prw/remote_write_spec/). There's an experimental remote write v2 version that can handle that, but we are uncertain about mimir's plans to support this. I think best approach right now is to not support utf8 in recording rules.
> 
> From recording rules model https://github.com/grafana/pyroscope/blob/main/api/settings/v1/recording_rules.proto#L56-L92:
> 
> * `metric_name` should still be `[a-zA-Z_:]([a-zA-Z0-9_:])*`
> * `matchers` should be able to handle utf8 `{"service.name"="foo"}` should be okay.
> * `group_by` should still be `[a-zA-Z_:]([a-zA-Z0-9_:])*`. Those are label names to export, so `["service.name"]` must NOT be supported.
> * `external_labels` label names should still be `[a-zA-Z_:]([a-zA-Z0-9_:])*`.
> 
> I propose to validate those so we make sure our rules are ok. We have 2 places to validate this, first at [recording rule creating time](https://github.com/grafana/pyroscope/blob/main/pkg/settings/recording/recording.go#L251-L303) (UI sending a POST request to tenant-settings), and compaction workers fetching rules from tenant-settings at compaction time, basically wrapped in [this constructor](https://github.com/grafana/pyroscope/blob/main/pkg/model/recording_rule.go#L36-L90). Maybe you find that some of the requirements already meet (I remember using some validations for the metric_names, for example).

https://github.com/grafana/pyroscope/blob/e0e688bc615c1b1ac83173e9e16dbca67ae0cac1/pkg/settings/recording/recording.go#L279-L290

@alsoba13 If I'm understanding the code correctly, it seems like label names from recording rules are [currently utf-8 validated](https://github.com/prometheus/common/blob/v0.65.0/model/labels.go#L103-L115) ([because of this code](https://github.com/prometheus/common/blob/v0.65.0/model/metric.go#L30-L49)), and not validated by Prometheus' legacy validation. I can address this issue, but in a separate PR since it's orthogonal to label name compatibility for profiles.

pkg/frontend/readpath/queryfrontend/query_series_labels.go
jake-kramer: Resolved offline with @bryanhuhta. Plenty of mitigations in different layers to be ok with this issue.

pkg/querier/querier.go
```diff
@@ -378,6 +405,42 @@ func (q *Querier) blockSelect(ctx context.Context, start, end model.Time) (block
...
+	}
+
+	for _, name := range toFilter {
+		if _, _, ok := validation.SanitizeLegacyLabelName(name); !ok {
+			level.Debug(q.logger).Log("msg", "filtering out label", "label_name", name)
+			continue
+		}
+		filtered = append(filtered, name)
+	}
+	return filtered, nil
```
simonswine, line 441: This is not necessary, as `LabeNames` should already filter them with the same logic

pkg/featureflags/client_capability.go
jake-kramer: Followed up here: https://github.com/grafana/pyroscope/pull/4498

pkg/querier/querier.go
jake-kramer: Logic is required when `len(req.Msg.LabelNames) != 0`; will add for that case

jake-kramer: https://github.com/grafana/pyroscope/pull/4502

PR #4445, opened 2025-09-18
chore(.idea): add module

korniltsev-grafanista:


simonswine:
Unsure what it does, but hopefully the XML doesn't steal my private keys

korniltsev-grafanista:
> Unsure what it does, but hopefully the XML doesn't steal my private keys

:D :D :D 

Basically we have runConfigurations commited already and they reference a `module` called `pyroscope`

And so if I do `git worktree add ../pyroscope-task-foo` the module will be called `pyroscope-task-foo` and run configurations do not work.

 And this PR commits the module and hopefully run configurations start working in worktrees as well.

PR #4449, opened 2025-09-20
Fix: strengthen Input validation and prevent Integer overflows

odaysec:
#### Summary of Description
This request introduces several security and robustness improvements across multiple components of the Pyroscope codebase. The changes focus on strengthening input validation, preventing integer overflows, and ensuring safer error handling.

1. **Safe Parsing of Sample Rate (`pkg/ingester/pyroscope/ingest_handler.go`)**

   * Replaced `strconv.Atoi` with `strconv.ParseUint(sr, 10, 32)` to ensure that only values safely representable as `uint32` are accepted.
   * If parsing fails, a default value is used as before.
   * Prevents unsafe conversions and potential overflows.

2. **Bounds Check for ByteSize (`pkg/og/util/bytesize/bytesize.go`)**

   * Added a check to ensure that parsed values do not exceed `math.MaxInt64` before converting to `ByteSize (int64)`.
   * Returns a parse error if the value is out of range.
   * Imported `math` where required.

3. **Validation for Shards (`pkg/querier/replication.go`)**

   * Introduced validation to ensure that `shards` is greater than 0 and does not exceed `math.MaxInt`.
   * Added checks to confirm that `shardIdx` is within bounds before being used as an index.
   * Prevents invalid slice operations and potential panics.

4. **Safe Handling of Split Count (`pkg/storegateway/gateway_blocks_http.go`)**

   * Replaced `strconv.Atoi` with `strconv.ParseInt(sc, 10, 32)` and added validation.
   * Ensured that `splitCount` is within `[0, math.MaxUint32]`; otherwise, reset to a safe default (0).
   * Prevents unsafe casting and integer overflows.

5. **Validation of Log Entry Buffer Size (`pkg/metastore/fsm/log_entry.go`)**

   * Introduced a maximum allowed size constant (64 MB) for `len(raw)`.
   * If `len(raw)` exceeds this threshold, return an error instead of allocating a large buffer.
   * Ensures safe and bounded memory allocation.

#### Rationale

These changes improve the overall security and stability of Pyroscope by:

* Preventing unsafe integer conversions.
* Avoiding integer overflows and out-of-bounds slice usage.
* Enforcing safer defaults for untrusted input.
* Reducing the risk of excessive memory allocation.

All modifications are scoped to the mentioned files and functions. No external dependencies were added, and the fixes are fully compatible with existing functionality.

[Go Language Integer overflow](https://golang.org/ref/spec#Integer_overflow)
[Making slices, maps and channels](https://golang.org/ref/spec#Making_slices_maps_and_channels)

### DIFF
--- a/pkg/ingester/pyroscope/ingest_handler.go
+++ b/pkg/ingester/pyroscope/ingest_handler.go
@@ -124,7 +124,7 @@ func (h ingestHandler) parseInputMetadataFromRequest(_ context.Context, r *http.
 	}
 
 	if sr := q.Get("sampleRate"); sr != "" {
-		sampleRate, err := strconv.Atoi(sr)
+		sampleRate, err := strconv.ParseUint(sr, 10, 32)
 		if err != nil {
 			_ = h.log.Log(
 				"err", err,
--- a/pkg/metastore/fsm/log_entry.go
+++ b/pkg/metastore/fsm/log_entry.go
@@ -58,6 +58,10 @@ func marshal(v proto.Message) ([]byte, error) {
 	if err != nil {
 		return raw, err
 	}
+	maxRawSize := 64 * 1024 * 1024 // 64 MB guard
+	if len(raw) > maxRawSize {
+		return nil, fmt.Errorf("marshaled message too large: %d bytes", len(raw))
+	}
 	buf := make([]byte, 4+len(raw))
 	copy(buf[4:], raw)
 	return buf, err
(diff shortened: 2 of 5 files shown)


aleks-p:
Thanks for the contribution. These changes seem to be mostly defensive guards on internal trust boundaries rather than reachable inputs. The bytesize overflow check is incomplete (the multiplier can still overflow afterward), and the 64 MB cap in the metastore FSM could break legitimate large entries. Given that, plus no accompanying tests, we're going to close this.

Happy to look at a focused fix if you can demonstrate a concrete reachable issue.

PR #4451, opened 2025-09-22
feat(helm): add option to add annotations to headless service

wind0r:
It would also be great to be able to add extra annotations to the headless service.
We want to use this to add external dns settings to the headless service for ingest path.

tested with:
Single Binary:
```
helm template pyroscope . \
  --set pyroscope.components.ingester.service.headlessAnnotations."istio\.io/rev"="stable"
```
Microservices
```
helm template pyroscope . \
  --values values-micro-services.yaml \
  --set pyroscope.components.ingester.service.headlessAnnotations."istio\.io/rev"="stable"
```


simonswine (approved):
LGTM

Thanks for your contribution

PR #4453, opened 2025-09-23
feat: add health check/warmup to segmentwriter to reduce first-write latency

oleg-kozlyuk-grafana:
* Adds dual-purpose check that:
  * Establishes connection to object store before reporting ready to reduce latency
  * Logs a warning if there is an issue with credentials

### DIFF
--- a/pkg/segmentwriter/service.go
+++ b/pkg/segmentwriter/service.go
@@ -193,7 +193,7 @@ func (i *SegmentWriterService) starting(ctx context.Context) error {
 	// Perform bucket health check before ring registration to warm up the connection
 	// and avoid slow first requests affecting p99 latency
 	// On error, will emit a warning but continue startup
-	i.performBucketHealthCheck(ctx)
+	_ = i.performBucketHealthCheck(ctx)
 
 	if err := services.StartManagerAndAwaitHealthy(ctx, i.subservices); err != nil {
 		return err


pkg/segmentwriter/service.go
```diff
@@ -155,7 +159,42 @@ func New(
...
+	if !i.config.BucketHealthCheckEnabled {
+		return nil
+	}
+
+	level.Debug(i.logger).Log("msg", "starting bucket health check", "timeout", i.config.BucketHealthCheckTimeout.String())
+
+	healthCheckCtx, cancel := context.WithTimeout(ctx, i.config.BucketHealthCheckTimeout)
+	defer cancel()
+
+	err := i.storageBucket.Iter(healthCheckCtx, "", func(string) error {
```
simonswine, line 175: Probably quicker to do an Exists()/Get()/Attribures() and expect a not found.

What I am also unsure about if the bucket name is wrong, do we get a "not found" or another error?

pkg/segmentwriter/service.go
```diff
@@ -155,7 +159,42 @@ func New(
...
+
+	level.Warn(i.logger).Log("msg", "bucket health check failed", "err", err)
+	return nil // Don't fail startup, just warn
+}
+
 func (i *SegmentWriterService) starting(ctx context.Context) error {
+	// Perform bucket health check before ring registration to warm up the connection
+	// and avoid slow first requests affecting p99 latency
+	// On error, will emit a warning but continue startup
+	_ = i.performBucketHealthCheck(ctx)
```
simonswine, line 196: I guess the bigger question is how would we know that something is wrong, if we mark ourselves as ready/healthy anyhow. Maybe the right thing here is to exit with an error code.

pkg/segmentwriter/service.go
oleg-kozlyuk-grafana: My idea here is to try and not depend on implementation specifics and whether the bucket is empty. To my best understanding, bucket not being present should emit a `not found` error indeed, while any other scenario would result in successful request.

Speaking frankly, I am also not sure if the error is same, but this approach feels the most robust out of alternatives I've considered.

pkg/segmentwriter/service.go
oleg-kozlyuk-grafana: I'd like to take a cautious approach here to make sure I don't create an incident accidentally because I didn't consider something. As discussed over call, I'll add a TODO comment and an issue into board, so we can change it into a fatal error once we're sure this doesn't break PROD.

pkg/segmentwriter/service.go
simonswine: I think the reasoning make sense, let's keep it like that.

In case of the bucket list being very long/expensive to gather, we could consider adding a prefix that likely would never exist, then it will always be expecting an empty reply.

Also for the functioning of the segment writer it is critical that we have the write permission, so another alternative, could be testing if an Upload is allowed.

simonswine (approved):
LGTM, let's add the TODO comment/issue and get this in

PR #4455, opened 2025-09-24
chore: add dependabot config

oleg-kozlyuk-grafana:


korniltsev-grafanista:
For our examples, can we upgrade only pyroscope sdks? I don't really care about ruby on rails updates or node express updates or otel updates.

Can we update only minor/patches versions, not major versions?

Can we not upgrade frontend dockerfile node hashsum (because I like docker caches to work and want (controversial, less "secure"?) )?

Can we upgrade packages separately? not as just one huge upgrade? https://github.com/oleg-kozlyuk-grafana/pyroscope/pull/4/files  Some dependencies are more important than the others, for example prometheus dependencies - they recently switched labels implementation and we had to apply extra care

oleg-kozlyuk-grafana:
> For our examples, can we upgrade only pyroscope sdks? I don't really care about ruby on rails updates or node express updates or otel updates.

My concern is that the samples will become outdated and we would not catch compatibility issues. For example, pyroscope SDK depending on an old version of package that is incompatible with something new user would like to use. We might also cause ourselves unexpected issues if we update only dependencies of pyroscope itself but not samples

I'm happy to exclude it and revisit at later date, however.

> Can we update only minor/patches versions, not major versions?

Sure, it's possible, will add an exclusion

> Can we not upgrade frontend dockerfile node hashsum (because I like docker caches to work and want (controversial, less "secure"?) )?

Build caches would work after a single checkout

> Can we upgrade packages separately? not as just one huge upgrade? https://github.com/oleg-kozlyuk-grafana/pyroscope/pull/4/files Some dependencies are more important than the others, for example prometheus dependencies - they recently switched labels implementation and we had to apply extra care

I can remove grouping but keep in mind it would produce 42 different PRs instead of 1 (or rather, batch of 5 every week). It would get exhausting soon

korniltsev-grafanista:
> Build caches would work after a single checkout

How would they work if the hashsum is updated every week?

oleg-kozlyuk-grafana:
> How would they work if the hashsum is updated every week?

We can update things monthly instead of weekly if that's a concern. The primary idea is to avoid having a very old vulnerable image

PR #4457, opened 2025-09-26
fix: revert HasFunctions flags for mixed symbolization

marcsanmi:
Reverts 7ee3759f88dada1adcb833c30b249d5542df31c2 since it doesn't fix the real issue. 

Moreover, we shouldn't rely anymore in `hasFunctions` and add fixes in top of that since it won't work with new otlp versions (see https://github.com/grafana/pyroscope/pull/4439)


korniltsev-grafanista (approved):
lgtm if this is just a revert

Issue #4383, opened 2025-08-20
After upgrade to the latest version (bump from otlp v1.4.0 to v1.7.0), unsymbolized frames disappear from query


ading1977:
#### Describe the bug
I used to run pyroscope `1.13.4` which is using otlp `v1.4.0`. For unsymbolized frames, it will show the binary name and raw address info. However, after upgrading to `weekly-f126-4a2d0a8ed` to use otlp `v1.7.0`, all unsymbolized frames disappear from the query and UI.

#### To Reproduce

1. Start Pyroscope `weekly-f126-4a2d0a8ed`
2. Perform query on my own program which has symbols stripped. All I am seeing now are kernel level system calls.

#### Expected behavior

Unsymbolized frames should show binary name + address.

#### Environment

- Infrastructure: [k8s]
- Deployment tool: [yaml]

#### Additional Context
I wonder if there is a newly introduced configuration that disable this on purpose?

ading1977:
Just to provide a piece of evidence, when deploying the following build: `main-837db07`, the unsymbolized frames can show up properly.

This is the specific commit:

```
commit 837db078953e69cd676290575980efd881cbbf5d (temp)
Author: Francesco Andreuzzi <andreuzzi.francesco@gmail.com>
Date:   Mon Jun 2 03:42:52 2025 +0200

    Bump `go.opentelemetry.io/proto/otlp` to `v1.7.0` (#4212)
```

I suspect this might be broken by the following commit:

```
commit ac947a1d6f1d7ce092a424f00670322587f2b7bc
Author: Marc Sanmiquel <marcsanmiquel@gmail.com>
Date:   Tue Jun 10 08:58:06 2025 +0200

    feat(symbolization): Add system libraries symbolization in the read path (#3799)

    * feat(symbolization): Add system libraries symbolization in the read path
```

korniltsev-grafanista:
I am adding an undocumented option to add symbolization stub to the alloy ebpf profiler here https://github.com/grafana/pyroscope/issues/4383

Still worth adding/fixing it in the database as well

korniltsev-grafanista:
The symbolizer creates such stubs, should be closed

PR #4470, opened 2025-10-02
feat: Speculative improvement of handling of unsymbolized profiles

oleg-kozlyuk-grafana:


pkg/ingester/otlp/convert.go
```diff
@@ -394,11 +394,21 @@ func (p *profileBuilder) convertSampleAttributesToLabelsBack(os *otelProfile.Sam
...
 
+	hasLines := false
+	hasFunctions := false
+	hasInlineFrames := false
+	for i, line := range ol.Line {
+		hasFunctions = hasFunctions || line.FunctionIndex > 0
+		hasLines = hasLines || line.Line > 0
+		hasInlineFrames = hasInlineFrames || i >= 1
+	}
+	hasFilenames := hasLines
```
simonswine: `hasFilenames`, shouldalso check if there is a non 0 filename attribute on the `function` object:

https://github.com/google/pprof/blob/436353cc1ee64cd00b8975ad18397ca49131f673/proto/profile.proto#L229-L230

simonswine (approved):
LGTM

pkg/ingester/otlp/convert.go
```diff
@@ -394,11 +394,21 @@ func (p *profileBuilder) convertSampleAttributesToLabelsBack(os *otelProfile.Sam
 }
 
 // convertMappingsBack converts a slice of OpenTelemetry Mapping entries to Google Mapping entries.
-func (p *profileBuilder) convertMappingBack(om *otelProfile.Mapping, dictionary *otelProfile.ProfilesDictionary) (uint64, error) {
+func (p *profileBuilder) convertMappingBack(ol *otelProfile.Location, om *otelProfile.Mapping, dictionary *otelProfile.ProfilesDictionary) (uint64, error) {
 	if i, ok := p.mappingMap[om]; ok {
 		return i, nil
 	}
 
+	hasLines := false
```
marcsanmi: These lines are just going to be called once per mapping because of the `om` cache key.
```go
if i, ok := p.mappingMap[om]; ok {
  return i, nil
}
```
I'm wondering whether we should rely on the first location to flag the whole mapping with those values 🤔

marcsanmi: I accidentally removed the previous message where I said that we could take the chance to address this issue [here](https://github.com/grafana/pyroscope/issues/4383#issuecomment-3296029025) by adding back stub creation.

**Update:** Forget about the stub creation (it was removed [here](https://github.com/grafana/pyroscope/commit/ac947a1d6f1d7ce092a424f00670322587f2b7bc#diff-30b0385316ef719e3cf6b51bae2226ac4d6afd84b5861f2d3a67abbccbefb762L221-L225)), I'm working on a separate PR adding it. See https://github.com/grafana/pyroscope/pull/4494

oleg-kozlyuk-grafana: Updated the mapping logic to better handle mixed cases

marcsanmi:
I think the current logic doesn't handle locations with empty Line arrays (unsymbolized). When len(ol.Line) == 0, the inner loop never runs, so hasFunctions stays true. This prevents blocks from being tagged as unsymbolized, causing the symbolizer to never run - when enabled. This affects both fully unsymbolized profiles and mixed-symbolization cases (where some locations are symbolized, some aren't). We could check if len(ol.Line) == 0 before the loop and set all flags to false

oleg-kozlyuk-grafana:
> I think the current logic doesn't handle locations with empty Line arrays (unsymbolized). When len(ol.Line) == 0, the inner loop never runs, so hasFunctions stays true. This prevents blocks from being tagged as unsymbolized, causing the symbolizer to never run - when enabled. This affects both fully unsymbolized profiles and mixed-symbolization cases (where some locations are symbolized, some aren't). We could check if len(ol.Line) == 0 before the loop and set all flags to false

Fair point point. I don't think `opentelemetry-collector` currently produces frames like that, but we might encounter clients with different interpretation of standard. Raised #4517

PR #4474, opened 2025-10-03
feat: metastore auto-join

aleks-p:
When we want to increase the number of metastore replicas (e.g., to increase availability), currently we need to rely on the admin UI and manually add the new replicas to the Raft cluster.

This change adds a `-metastore.raft.auto-join` flag. When set to `true` **and** the application has no raft state (is a freshly created replica) - we will attempt to add the node to an existing Raft cluster using the same API endpoint used by the admin UI. If auto-join fails, we fallback to the existing behavior and the replica will need to be manually added.

Tested in a local cluster, as well as with the provided integration test.

### DIFF
--- a/pkg/metastore/raftnode/node.go
+++ b/pkg/metastore/raftnode/node.go
@@ -30,6 +30,7 @@ type Config struct {
 
 	BootstrapPeers       []string `yaml:"bootstrap_peers"`
 	BootstrapExpectPeers int      `yaml:"bootstrap_expect_peers"`
+	AutoJoin             bool     `yaml:"auto_join"`
 
 	ServerID         string `yaml:"server_id"`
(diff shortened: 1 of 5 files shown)


pkg/metastore/raftnode/node.go
```diff
@@ -176,7 +182,19 @@ func (n *Node) Init() (err error) {
 		return fmt.Errorf("failed to check for existing state: %w", err)
 	}
 	if !hasState {
-		level.Warn(n.logger).Log("msg", "no existing state found, trying to bootstrap cluster")
+		level.Warn(n.logger).Log("msg", "no existing state found")
```
simonswine: Nit: Not too sure this message provides a lot of value, maybe we incorporate it into the ones on line 188/192?

PR #4476, opened 2025-10-06
chore: Allow to configure subPaths for metastore and shared

simonswine:
This allows to adapt those variables, without fullly patching them out.

That's used GL internal, as we started of with a subPath of "metastore".

### DIFF
--- a/operations/pyroscope/helm/pyroscope/README.md
+++ b/operations/pyroscope/helm/pyroscope/README.md
@@ -71,6 +71,8 @@
 | pyroscope.persistence.accessModes[0] | string | `"ReadWriteOnce"` |  |
 | pyroscope.persistence.annotations | object | `{}` |  |
 | pyroscope.persistence.enabled | bool | `false` |  |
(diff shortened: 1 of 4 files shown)


aleks-p (approved):

operations/pyroscope/helm/pyroscope/README.md
```diff
@@ -71,6 +71,8 @@
 | pyroscope.persistence.accessModes[0] | string | `"ReadWriteOnce"` |  |
 | pyroscope.persistence.annotations | object | `{}` |  |
 | pyroscope.persistence.enabled | bool | `false` |  |
+| pyroscope.persistence.metastore.subPath | string | `".metastore"` |  |
+| pyroscope.persistence.shared.subPath | string | `".shared"` |  |
```
simonswine, line 75: This was introduced in #4417, in order to avoid conflicts between tenant storage for the `target=all`.

While this is not 100% avoiding conflicts, a `.metastore` tenant is a lot less likely than a `metastore` tenant.

PR #4486, opened 2025-10-08
chore: Add release notes and update helm release for v1.15.0

marcsanmi:
I've recently released v1.15.0.

### DIFF
--- a/operations/pyroscope/helm/pyroscope/Chart.yaml
+++ b/operations/pyroscope/helm/pyroscope/Chart.yaml
@@ -3,7 +3,7 @@ name: pyroscope
 description: 🔥 horizontally-scalable, highly-available, multi-tenant continuous profiling aggregation system
 home: https://grafana.com/oss/pyroscope/
 type: application
-version: 1.15.0
+version: 1.15.1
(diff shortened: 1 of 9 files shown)


operations/pyroscope/helm/pyroscope/Chart.yaml
```diff
@@ -4,7 +4,7 @@ description: 🔥 horizontally-scalable, highly-available, multi-tenant continuo
 home: https://grafana.com/oss/pyroscope/
 type: application
 version: 1.15.0
```
simonswine: To get a release you need to raise it.

simonswine (approved):
LGTM

operations/pyroscope/helm/pyroscope/Chart.yaml
marcsanmi: Yes, the issue is that it was already bumped to 1.15 and we tried as much as possible to keep it up with Pyroscope release version. As talked, I've to updated the patch version in 6ffe52d96637d365610510e701d56f5c52cc3672.

simonswine (approved):
LGTM

PR #4488, opened 2025-10-08
feat: add OTLP canary query

oleg-kozlyuk-grafana:


cmd/profilecli/canary_exporter_probes.go
```diff
@@ -74,6 +82,230 @@ func (ce *canaryExporter) testIngestProfile(ctx context.Context, now time.Time)
...
+	}
+
+	// Marshal to protobuf
+	protoData, err := proto.Marshal(req)
+	if err != nil {
+		return fmt.Errorf("failed to marshal OTLP profile to protobuf: %w", err)
+	}
+
+	// Create HTTP request using the instrumented client
+	url := ce.params.URL + "/v1/profiles"
```
korniltsev-grafanista: Is this really how the collector sends data? just encodes the proto? If so, then can we just support it in the ingest handler? Otherwise just remove this if it is unused.

cmd/profilecli/canary_exporter_probes.go
```diff
@@ -74,6 +82,230 @@ func (ce *canaryExporter) testIngestProfile(ctx context.Context, now time.Time)
...
+							Name: "pyroscope-canary-exporter",
+						},
+						Profiles: []*otlpprofiles.Profile{profile},
+					},
+				},
+			},
+		},
+	}
+
+	return req, nil
```
korniltsev-grafanista: err is always nil? remove error from return values if so.

cmd/profilecli/canary_exporter_probes.go
oleg-kozlyuk-grafana: Yes: https://opentelemetry.io/docs/specs/otlp/#binary-protobuf-encoding

I plan to add support in next PR, but given lint complains about unused function, will readd the probe there

cmd/profilecli/canary_exporter_probes.go
oleg-kozlyuk-grafana: That's one of aspects of Go I'm less excited about. Removed

PR #4489, opened 2025-10-08
feat: Allow utf-8 label names in write path

jake-kramer:
Utf-8 label names are valid in the write path if the config `client-capability.allow_utf_8_label_names=true` is set, and the client capability `allow-utf8-labelnames=true` is passed in the `Accept` header for API calls.

Legacy label name validation and sanitization will continue to be enforced on write path, otherwise.

Follow up to https://github.com/grafana/pyroscope/pull/4442

@simonswine @bryanhuhta

### DIFF
--- a/pkg/validation/validate.go
+++ b/pkg/validation/validate.go
@@ -148,8 +148,7 @@ func ValidateLabels(clientCapabilities *featureflags.ClientCapabilities, limits 
 			return NewErrorf(LabelValueTooLong, LabelValueTooLongErrorMsg, phlaremodel.LabelPairsString(ls), l.Value)
 		}
 		if clientCapabilities != nil && clientCapabilities.AllowUtf8LabelNames {
-			// Using Prometheus utf-8 label name validation
-			if !model.LabelName(l.Name).IsValid() {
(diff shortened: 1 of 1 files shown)


jake-kramer:
> The write path should not use `allow-utf8-labelnames=true`. The capability is a read path only capability.
> 
> When profile producers (SDK, profilecli upload, Alloy pyroscope.write) support utf8 label names, they should:
> 
> Send the labels in utf8, if they have any.
> 
> Then there are two options:
> 
> * Backend support utf8 label names/has enabled support: All is fine
> * Backend doens't support: The first request fails and the client falls back to filtering/translating label names.
> 
> So basically you should create a tenant override maybe `IngestionAllowUtf8LabelNames`, around here:
> 
> https://github.com/grafana/pyroscope/blob/019bac278eb3f627f7ee08dcc86ce4eec44d5e09/pkg/validation/limits.go#L65
> 
> Then the ingest path should not consider client flags, they should just call `limits.IngestionAllowUtf8LabelNames()`, and the act accordingly.

Adding a metric to determine scope and impact of removing write path sanitization by default https://github.com/grafana/pyroscope/pull/4501

PR #4493, opened 2025-10-08
feat: Use `allow-utf8-labelnames` client capability in UI

jake-kramer:
Also updates to support local UI dev

Follow up from https://github.com/grafana/pyroscope/pull/4442
Depends on https://github.com/grafana/pyroscope/pull/4496


jake-kramer:
Bug discovered when trying to get `"key"="val"` selectors to work with utf-8 chars; resolved in https://github.com/grafana/pyroscope/pull/4496.

PR #4494, opened 2025-10-09
fix: add read-path stub creation for unsymbolized profiles

marcsanmi:
Fixes https://github.com/grafana/pyroscope/issues/4383

### Problem
After commit ac947a1d6 removed stub creation, unsymbolized OTLP profiles  were skipped when symbolizer is disabled. If the symbolizer is enabled, symbolizer itself will populate those stubs.

### Fix
Create stubs at **query time**. When symbolizer is disabled and blocks have
unsymbolized profiles, generate stubs functions like `"libjvm.so 0xcafebabe"`.

**Why read-path?** Preserves raw profiles in storage, allowing future symbolization when users toggle the symbolizer on.

⚠️ The fix is only added to v2. Adding it to v1 would require either:
  - Checking every profile for unsymbolized locations on every query (performance overhead), or
  - Adding equivalent to block metadata checking to v1's querier to avoid checking every profile at read time.

### Considerations
We could also try adding those stubs at ingest time. However, if symbolizer is enabled, the system would skip stub functions preventing real symbolization unless we re-process data or implement stub detection logic


pkg/ingester/otlp/convert.go
```diff
@@ -260,6 +283,18 @@ func (p *profileBuilder) convertLocationBack(ol *otelProfile.Location, dictionar
 		}
 	}
 
+	// Create symbolization stub for unsymbolized locations. These stubs will be replaced
+	// with real symbols during symbolization if available.
```
korniltsev-grafanista: > These stubs will be replaced with real symbols during symbolization if available.

I don't understand how would symbolizer do that? Does not it skip location if `loc.Line` is not empty?

marcsanmi: Yes, the second part of the comment should be removed. It's here since first I tried to update the symbolizer to detect stub functions to try it to (re)symbolize them... Removed in bfe6c6e94443c22164d193efe2f6af0b7cc44a79

oleg-kozlyuk-grafana:
Check #4470 - this case may be already covered

marcsanmi:
> Check https://github.com/grafana/pyroscope/pull/4470 - this case may be already covered

https://github.com/grafana/pyroscope/pull/4470 fixed the `HasFunctions` flag so blocks are properly tagged as unsymbolized. This PR adds the stub creation logic that runs when symbolizer is disabled. Both changes were needed, but I'd say address slightly different things.

pkg/frontend/readpath/queryfrontend/query_frontend_test.go
```diff
@@ -477,3 +478,225 @@ func Test_QueryFrontend_Series_WithLabelNameFiltering(t *testing.T) {
...
+				Location: []*profilev1.Location{
+					{Id: 1, MappingId: 1, Address: 0xcafebabe, Line: nil},
+					{Id: 2, MappingId: 1, Address: 0xdeadbeef, Line: nil},
+				},
+				Function: []*profilev1.Function{},
+				Sample: []*profilev1.Sample{
+					{LocationId: []uint64{1, 2}, Value: []int64{100}},
+				},
+			},
+			queries: []*queryv1.Query{{QueryType: queryv1.QueryType_QUERY_PPROF}},
```
simonswine, line 507: Nit: For unsymbolized locations without an address (0x0), I want to avoid that we output `<file> 0x0`, we should do `<file>` instead.

marcsanmi: Done in e8f51b370de56140655021014abe790535680482

korniltsev-grafanista:
I've run the branch, but I see no stubs
image

korniltsev-grafanista:
does this only fix pprof query? and not tree query?

marcsanmi:
> does this only fix pprof query? and not tree query?

Yes, you're right. It needs to be applied to tree as well to update the flamegraph. It'll just work for pprof (export)... Going to update it

marcsanmi:
Closing this PR since this needs to be done at the query backend level since we skip the locations there; so we don't have them available in the query frontend. Since this would imply a bigger change, I'm closing this and creating an issue to address it in the future.

Closing this PR since this needs to be done at the query backend level, where we skip the locations; so we don't have them available in the query frontend. Since this would imply a bigger change, I'm closing this and creating an issue to properly address it.

PR #4495, opened 2025-10-09
Add OTLP HTTP/json and HTTP/protobuf ingestion

oleg-kozlyuk-grafana:


pkg/api/api.go
```diff
@@ -201,8 +201,7 @@ func (a *API) RegisterDistributor(d *distributor.Distributor, limits *validation
 	})
 
 	a.RegisterRoute("/opentelemetry.proto.collector.profiles.v1development.ProfilesService/Export", otlpHandler, writePathOpts...)
-	// TODO(@petethepig): implement http/protobuf and http/json support
-	// a.RegisterRoute("/v1/profiles", otlpHandler, true, true, "POST")
+	a.RegisterRoute("/v1/profiles", otlpHandler, writePathOpts...)
```
korniltsev-grafanista: should this be `/v1development/profiles` ? which path does collector use now?

oleg-kozlyuk-grafana: There are:
* `/ingest`
* `/pyroscope/ingest/`
* `/opentelemetry.proto.collector.profiles.v1development.ProfilesService/Export` for gRPC

Good point though, `/otlp/v1development` would make more sense

korniltsev-grafanista: does collector send to `/otlp/v1development` ?

korniltsev-grafanista: TIL about `/pyroscope/ingest` :)

oleg-kozlyuk-grafana: Ah, I see what you mean now. Yes, OTLP exporter uses `/v1development/profiles` [by default](https://github.com/open-telemetry/opentelemetry-collector/blob/33fa41381d07a4e359c7e7bb6a3c10dad9b0e294/exporter/otlphttpexporter/config.go#L61C61-L61C84). Even though it's configurable, it would make sense to work with default values. Fixed

PR #4499, opened 2025-10-10
Add Renovate config

oleg-kozlyuk-grafana:
Adding Renovate config as first step to check the PRs created. After confirming everything works, will remove Dependabot and enable automerge

### DIFF
--- a/.github/workflows/renovate-config-validator.yml
+++ b/.github/workflows/renovate-config-validator.yml
@@ -26,4 +26,4 @@ jobs:
           persist-credentials: false
 
       - name: Validate Renovate Config
-        uses: grafana/shared-workflows/actions/validate-renovate-config@main # TODO: pin to tag once tag is created
+        uses: grafana/shared-workflows/actions/validate-renovate-config@validate-renovate-config/v0.1.0


renovate.json
```diff
@@ -0,0 +1,93 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:best-practices"
```
marcsanmi, line 4: Should we use `"github>grafana/grafana-renovate-config//presets/base"`?

.github/workflows/renovate-config-validator.yml
```diff
@@ -0,0 +1,34 @@
+name: Validate Renovate config
```
dsotirakis, line 1: for this one you can also use https://github.com/grafana/shared-workflows/tree/main/actions/validate-renovate-config 😁

.prettierignore
```diff
@@ -16,4 +16,6 @@ cmd/**
 pkg/test/**
 examples/**
 .devcontainer/**
-.vscode/**
\ No newline at end of file
+.vscode/**
+renovate.json
+renovate.json5
```
dsotirakis, line 21: since you're only adding a json file I don't think you need a `.json5` extension here?

renovate.json
```diff
@@ -0,0 +1,93 @@
...
+  "packageRules": [
+    {
+      "description": "Ignore major updates for all dependencies",
+      "matchUpdateTypes": ["major"],
+      "enabled": false
+    },
+    {
+      "description": "Group all Go module updates (main project)",
+      "matchManagers": ["gomod"],
+      "matchPaths": [
```
marcsanmi, line 21: I might be wrong, but `matchPaths` is a valid config in renovate? Maybe `matchFileNames`?

dsotirakis: @marcsanmi 👋 
in theory, should the config validation github action runs and passes, the config should be correct!

renovate.json
dsotirakis: so by default your config will also extend the one you mention, unless you ignore it using [ignorePresets](https://docs.renovatebot.com/configuration-options/#ignorepresets)

.prettierignore
oleg-kozlyuk-grafana: It's not detrimental to have both in `.prettierignore`, and will save someone a cycle in future if we switch

.github/workflows/renovate-config-validator.yml
oleg-kozlyuk-grafana: The tag validate-renovate-config-v0.1.0 referenced in README does not exist

dsotirakis: oh - gimme one second

dsotirakis: it's there now @oleg-kozlyuk-grafana 😄 
https://github.com/grafana/shared-workflows/tree/validate-renovate-config/v0.1.0!

PR #4501, opened 2025-10-10
feat: Add `sanitized_label_names_total` counter metric

jake-kramer:
This metric will be used to determine how often (if at all) clients are running in to label sanitization.


pkg/validation/validate.go
```diff
@@ -147,6 +157,7 @@ func ValidateLabels(limits LabelValidationLimits, tenantID string, ls []*typesv1
 			return NewErrorf(LabelValueTooLong, LabelValueTooLongErrorMsg, phlaremodel.LabelPairsString(ls), l.Value)
 		}
 		if origName, newName, ok := SanitizeLegacyLabelName(l.Name); ok && origName != newName {
+			sanitizedLabelNames.WithLabelValues(tenantID).Inc()
```
simonswine: Can you move that down, after `handleSanitizedLabel`, we should not count the error case.

PR #4502, opened 2025-10-10
fix: Skip duplicated label name filtering in some cases

jake-kramer:


pkg/querier/querier.go
```diff
@@ -428,17 +427,19 @@ func (q *Querier) filterLabelNames(
...
 		}
-		toFilter = response.Msg.Names
-	}
-
-	for _, name := range toFilter {
-		if _, _, ok := validation.SanitizeLegacyLabelName(name); !ok {
-			level.Debug(q.logger).Log("msg", "filtering out label", "label_name", name)
-			continue
+		return response.Msg.Names, nil
+	} else {
```
simonswine: Nit: Find it easier to read without the else branch

PR #4514, opened 2025-10-13
chore: Automatically select self-hosted/github agents

oleg-kozlyuk-grafana:
Rationale: allows to correctly run CI for PRs from forks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,11 +48,12 @@ jobs:
       fail-fast: false
       matrix:
         arch: [amd64, arm64]
-    runs-on: > # Using this long expression here to avoid having agent type as build parameter
-      ${{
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -48,7 +48,7 @@ jobs:
       fail-fast: false
       matrix:
         arch: [amd64, arm64]
-    runs-on: ubuntu-${{ matrix.arch }}-large
+    runs-on: ${{ github.repository_owner == 'grafana' && format('ubuntu-{0}-large', matrix.arch) || 'ubuntu-latest' }}
```
simonswine: For the fallback in the arch==arm64 case it should be: [`ubuntu-24.04-arm`](https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-24-image.md) instead of `ubuntu-latest`

PR #4519, opened 2025-10-13
fix: panic when handling unsafe tenant name

oleg-kozlyuk-grafana:


pkg/tenant/interceptor.go
```diff
@@ -46,7 +46,10 @@ func (i *authInterceptor) WrapUnary(next connect.UnaryFunc) connect.UnaryFunc {
 		if !i.enabled {
 			return next(InjectTenantID(ctx, DefaultTenantID), req)
 		}
-		_, ctx, _ = ExtractTenantIDFromHeaders(ctx, req.Header())
+		_, ctx, err := ExtractTenantIDFromHeaders(ctx, req.Header())
+		if err != nil {
+			return nil, err
+		}
```
simonswine: This will change the observed http error code for clients, as we no longer wrap it with `connect.Error` and `connect.CodeUnauthenticated` (so clients will get a 500 with this PR rather than a 401). And have a few more subtle side effects I can't explicitly name. 

I think the inconsistency is really here in L92 which returns ctx, while L98 doesn't:

https://github.com/grafana/pyroscope/blob/2af1eeaa418da6dc5910a84f06ca6c289c8d5616/pkg/tenant/interceptor.go#L89-L102

Also the same panic is likely possible in this line https://github.com/grafana/pyroscope/blob/2af1eeaa418da6dc5910a84f06ca6c289c8d5616/pkg/tenant/interceptor.go#L75

oleg-kozlyuk-grafana: Thanks for detailed comments - I wasn't sure whether to go with idiomatic Go way as-recently-learned or to simply patch the missing return. Updated as per suggestion

PR #4521, opened 2025-10-14
feat: Log sanitized label names

jake-kramer:
Gives more visibility in to which label names are
being written that are being rewritten.


pkg/validation/validate.go
```diff
@@ -162,6 +162,10 @@ func ValidateLabels(limits LabelValidationLimits, tenantID string, ls []*typesv1
 			if err != nil {
 				return err
 			}
+			level.Debug(util.Logger).Log(
```
simonswine, line 165: Suggestions: Worth logging the service_name, to better understand where those metrics coming from.

PR #4522, opened 2025-10-14
fix: Add stricter validation for metrics from profiles fields

jake-kramer:
Since metrics from profiles is using Prometheus remote_write 1.0, the legacy format `^[a-zA-Z_][a-zA-Z0-9_]*$` for metric and label names must be enforced.


alsoba13:
Looks great, thank you for taking care of it.

LGTM

Issue #3624, opened 2024-10-14
publish compiled dashboards, alert rules and recording rules


BramSchutte:
#### Is your feature request related to a problem? Please describe.

The source code files for the dashboards, alert rules and recording rules are published in the repository. Can the build dashboards, alert rules and recording rules be published as well?

#### Describe the solution you'd like

A compiled, ready to use/download dashboards, alert rules and recording rules that are available in this repository.

#### Describe alternatives you've considered

Building the dashboards, alert rules and recording roles, but with constraints on my machine, i cannot install dependencies. Thus not compile the dashboards, alert rules and recording rules

#### Additional context

not needed.

maikel-09:
+1

wessp01:
+1

kolesnikovae:
Thank you for creating the issue! We do have a set of dashboards, alert rules, and recording rules; however, these are very tailored to Grafana's infrastructure. I do believe we could extract the pieces that can be shared, though. We will discuss this internally and prioritize accordingly.

HujinoKun:
Hello,

+1

Could you publish the dashboards on grafana or at least offer a json version in the repo? @kolesnikovae

BramSchutte:
@kolesnikovae Do you have an update regarding releasing a compiled list of dashboards, alert rules, and recording rules. In our case it does not matter that they work out of the box. But a nice base dashboard that can be used as a guideline is really something i could use.

xocasdashdash:
hey!
how do you go about building the dashboards locally?

michaeldmitry:
+1

simonswine:
I have started with four internal dashboards and the rules rely relies on in #4525. 

Let me know if those are usable for your purposes and which areas you lack insights.

PR #4525, opened 2025-10-15
feat: Add pyroscope-monitoring helm chart

simonswine:
This contains a fully fledged observability deployment for Pyroscope OSS monitoring

Over time I would like to consolidate our internal dashboards and alerts into this helm chart. This also allows us to test/debug Pyroscope OSS in a local kind deployment.

In order to test this setup in your local kind:
```
$ make deploy-micro-services deploy-monitoring
# Once everything is up
$ kubectl port-forward deployment/pyroscope-monitoring 3000:3000
# Open grafana on http://localhost:3000
```

### DIFF
--- a/operations/pyroscope/helm/pyroscope-monitoring/README.md
+++ b/operations/pyroscope/helm/pyroscope-monitoring/README.md
@@ -2,7 +2,18 @@
 
 ![Version: 0.1.0](https://img.shields.io/badge/Version-0.1.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.0.0](https://img.shields.io/badge/AppVersion-0.0.0-informational?style=flat-square)
 
-A Helm chart for monitoring Grafana Pyroscope
+A Helm chart for monitoring Grafana Pyroscope. This helm chart uses otel-lgtm to monitor the health of the Grafana Pyroscope backend.
+
+> ⚠️ **Note:** This monitoring setup is not production grade and is intended for development and testing purposes only.
+
+## Dashboards
+
+This chart provisions the following Grafana dashboards under the "Pyroscope" folder:
+
+- **operational** - General operational metrics dashboard
+- **v2-metastore** - Metastore-related metrics dashboard (only relevant with v2 storage layer)
+- **v2-read-path** - Read path performance dashboard (focusing on v2 storage layer)
+- **v2-write-path** - Write path performance dashboard (focusing on v2 storage layer)
 
 ## Requirements
 
--- a/operations/pyroscope/helm/pyroscope-monitoring/values.yaml
+++ b/operations/pyroscope/helm/pyroscope-monitoring/values.yaml
@@ -221,7 +221,7 @@ monitoring:
   alloy-metrics:
     enabled: true
     image:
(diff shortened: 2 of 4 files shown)


simonswine:
@manab-pr please have a look, this PR also adds the dashboards/rules used into a subfolder: https://github.com/grafana/pyroscope/tree/7227037426b3f6bc4acf77df7c287e8026ce96ed/operations/monitoring

manab-pr:
Thanks for the context, @simonswine! I've reviewed #4525 and understand the approach now.

  The pyroscope-monitoring Helm chart is great for Kubernetes users who want a complete monitoring stack. However, I want to make sure we're still addressing
  the original issue #3624. 
  The users in that issue were asking for pre-compiled dashboard files because they:
  - Don't necessarily use Kubernetes/Helm
  - Have existing Grafana instances and just want to import dashboards
  - Can't/won't install jsonnet/jb to compile the mixin

For non-helm users: Will there be a way to extract/download the dashboard JSON files from the helm chart for manual import into Grafana? Or would it make sense to keep a small set of pre-compiled dashboards specifically for users who need standalone files, while the helm  chart serves K8s users?

simonswine:
@manab-pr Yes if you look here, you will find the rules and dashboards: https://github.com/grafana/pyroscope/tree/20251015_helm-chart-monitoring/operations/monitoring

CI will make sure that those files are in sync, when someone changes the helm chart.

Obviously when you just use dashboards/rules, no one guarantees that you run kube-state-metrics/cadvisor/node-exporter. Like we do through the helm chart.

operations/pyroscope/helm/pyroscope-monitoring/values.yaml
```diff
@@ -0,0 +1,250 @@
+# Default values for pyroscope-monitoring.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
+
+image:
+  repository: grafana/otel-lgtm
```
aleks-p, line 8: This embeds and runs Prometheus, Loki, Tempo and Pyroscope binaries. Do we want a way to configure and prevent this, or are we ok with running Pyroscope and other applications inside the deployment for monitoring Pyroscope? :)

operations/pyroscope/helm/pyroscope-monitoring/README.md
```diff
@@ -0,0 +1,43 @@
...
+| podAnnotations | object | `{}` |  |
+| podLabels | object | `{}` |  |
+| podSecurityContext | object | `{}` |  |
+| replicaCount | int | `1` |  |
+| resources | object | `{}` |  |
+| securityContext | object | `{}` |  |
+| service.deployStaticName | bool | `true` |  |
+| service.type | string | `"ClusterIP"` |  |
+| tolerations | list | `[]` |  |
+
```
marcsanmi, line 54: Nit: Is it worth adding a section listing the dashboards that will be created automatically? Maybe even integrate it with helm docs by adding it as an annotation in `values.yaml` ?

marcsanmi (approved):
LGTM! needed to clean my laptop memory, but run like a charm 😁

operations/pyroscope/helm/pyroscope-monitoring/values.yaml
simonswine: This is exactly why I have chosen it, I don't want to maintain that gang of 4. I don't expect anyone to really do that to a production grade with this, will mention all of that in the README

PR #4527, opened 2025-10-16
fix: create fallback symbols for mappings without BuildIDs

marcsanmi:
Fixes https://github.com/grafana/pyroscope-squad/issues/596

@alsoba13 noticed while testing that we aren't creating proper fallback symbols for mappings with empty buildid; now we do. See the difference between both flamegraphs where e.g `a58` now is `linux-vdso.1.so!0xa58`

Screenshot 2025-10-16 at 12 26 11
Screenshot 2025-10-16 at 13 48 25


pkg/symbolizer/symbolizer.go
```diff
@@ -99,18 +99,25 @@ func (s *Symbolizer) SymbolizePprof(ctx context.Context, profile *googlev1.Profi
...
+			} else {
+				symLoc = req.locations[i]
+			}
+
 			allSymbolizedLocs = append(allSymbolizedLocs, symbolizedLocation{
 				loc:     loc,
-				symLoc:  req.locations[i],
+				symLoc:  symLoc,
 				mapping: mapping,
 			})
```
simonswine: This branching looks awfully complex. Could  we maybe add that logic to `symbolize()`, as it's basically the same outcome if buildID is empty or we can't find the buildID. 

I also think these parts could be worth having in their own function, so we DRY.

https://github.com/grafana/pyroscope/blob/d28afcdfb0645b6bb03749140f83e3efaa9062a4/pkg/symbolizer/symbolizer.go#L263-L266

https://github.com/grafana/pyroscope/blob/d28afcdfb0645b6bb03749140f83e3efaa9062a4/pkg/symbolizer/symbolizer.go#L274-L276

simonswine: Also might be worth having a metrics for empty build id mappings

marcsanmi: Good points, thanks! Addressed both in 13edd15eba4178a211384fb38652ba45dab344a1

PR #4529, opened 2025-10-16
fix: Replace global nop logger with working logger

jake-kramer:


simonswine (commented):
Would prefer not to have any globals like this. Can we instead pass the logger via arguments to fix https://github.com/grafana/pyroscope/pull/4521

Can you update the comment before Logger to make it clear this you should expected this to be a noplogger.

We probably should midterm remove it or move it to logger_test.go, so it's only available for tests.

jake-kramer:
Closing in favor of https://github.com/grafana/pyroscope/pull/4531

PR #4531, opened 2025-10-17
fix: Replace nop logger used in label validation

jake-kramer:


jake-kramer:
Added an issue for `util.Logger` https://github.com/grafana/pyroscope/issues/4532

simonswine (approved):
LGTM

PR #4533, opened 2025-10-17
fix: Docs release workflow

simonswine:
Apparently the regex syntax `\d` may or maybe not supported.

I have replaced it with a standardized `[0-9]` instead

https://stackoverflow.com/questions/18514135/bash-regular-expression-cant-seem-to-match-any-of-s-s-d-d-w-w-etc


jdbaldry:
Oops sorry that changed in https://github.com/grafana/writers-toolkit/pull/898 and corresponded with the v2 release of that workflow.

Bash only supports ERE and not PCRE

PR #4537, opened 2025-10-17
[fix] Don't 500 on bad profile types

bryanhuhta:
This PR adds explicit validation for profile types in queries. Failing the validation will return a 400. Previously, we would fail with a 500, which messed with our alerts.


alsoba13 (approved):

simonswine:
Merging to get into f141

Issue #4524, opened 2025-10-15
pyroscope.ebpf collect interval under 15s issue


luweglarz:
#### Describe the bug
When setting pyroscope.ebpf component collect_interval under 15s on alloy Pyroscope doesn't show similar interval on UI, I am not sure if it is the expected behavior.
here is with 30s collect_interval
<img width="1873" height="140" alt="Image" src="https://github.com/user-attachments/assets/b5eac3a6-faa5-44bc-b20b-adc09a04da53" />
and here is with 5s collect_interval
<img width="1814" height="144" alt="Image" src="https://github.com/user-attachments/assets/f0c7fea3-b83b-469a-8e97-d5f2a487106c" />

#### To Reproduce
1.Set pyroscope.ebpf collect_interval to a value under 5s
2.Check pyroscope UI intervals

#### Expected behavior

I expect to see profiles on the UI correlated to the collect_interval

#### Environment

- Infrastructure: docker containers
- Deployment tool: docker-compose

simonswine:
We currently hardcode 15s as the lowest step interval, that's why you values 15s seconds apart, maybe you can make this configurable? 

https://github.com/grafana/pyroscope/blob/38a40a93f921d7e66b0ef346d59e5d9dc28632bf/pkg/querier/timeline/calculator.go#L11

luweglarz:
That sounds doable, should it be configurable through the Querier config ?

manab-pr:
Yes, I've implemented it as a configurable option in the Querier config!

  The changes add a `--querier.min-step-duration` flag (defaults to 15s for backward compatibility) that allows users to configure the minimum
  step duration.

  Key changes:
  - Added `MinStepDuration` field to querier Config
  - Created `CalcPointIntervalWithMinInterval()` function that accepts custom minimum intervals
  - Updated all HTTP handlers to use the configurable value
  - Added tests covering eBPF use cases (1s, 5s intervals)

  Users can now run: `--querier.min-step-duration=5s` for fast eBPF profiling intervals.

PR #4540, opened 2025-10-20
feat: Make the minStepDuration of the querier configurable per tenant from limits configuration

luweglarz:
For issue https://github.com/grafana/pyroscope/issues/4524

Add a way to configure the minInterval in the querier configuration for series graph between each point


luweglarz:
I am not sure with the naming of the field, maybe something more clear would be better like seriesMinInterval, what do you think @simonswine ?

pkg/querier/querier.go
```diff
@@ -47,12 +47,14 @@ import (
...
 	PoolConfig      clientpool.PoolConfig `yaml:"pool_config,omitempty"`
 	QueryStoreAfter time.Duration         `yaml:"query_store_after" category:"advanced"`
+	MinInterval     time.Duration         `yaml:"min_interval" category:"advanced"`
 }
 
 // RegisterFlags registers distributor-related flags.
 func (cfg *Config) RegisterFlags(fs *flag.FlagSet) {
 	cfg.PoolConfig.RegisterFlagsWithPrefix("querier", fs)
 	fs.DurationVar(&cfg.QueryStoreAfter, "querier.query-store-after", 4*time.Hour, "The time after which a metric should be queried from storage and not just ingesters. 0 means all queries are sent to store. If this option is enabled, the time range of the query sent to the store-gateway will be manipulated to ensure the query end is not more recent than 'now - query-store-after'.")
+	fs.DurationVar(&cfg.MinInterval, "querier.min-interval", 5*time.Second, "The minimum interval between points in a query.")
```
simonswine: In the API we call this value "step", so maybe this is a better name:

```suggestion
	fs.DurationVar(&cfg.MinInterval, "querier.min-step-duration", 5*time.Second, "The minimum interval between points in a query.")
```

simonswine: The default should stay 15s, for backwards compatibility:

```suggestion
	fs.DurationVar(&cfg.MinInterval, "querier.min-interval", 15*time.Second, "The minimum interval between points in a query.")
```

simonswine (commented):
Thank you for your PR.

I think it would be preferred, if we could configure this a per tenant basis. 

For this we typically add a field to the "limit" struct, this has the benefit, that we can change the value on a per tenant basis and also update it without restarting the pyroscope process, here is the related struct:

https://github.com/grafana/pyroscope/blob/5b02fdd218d5230221629f2a69b2abf639362e70/pkg/validation/limits.go#L37

Let me know if you have any questions

luweglarz:
I applied your suggestions @simonswine

luweglarz:
Hi @simonswine  could this be merged anytime soon? Thanks !

luweglarz:
Hi @simonswine sorry to ping you again on this one, is there any chance to get it merged ? Thanks !

PR #4544, opened 2025-10-21
Reenable OTLP HTTP canary exporter

oleg-kozlyuk-grafana:
Now that support for new endpoints have been added to cloud-backend-gateway, these should work. I had also to change other probes slightly as OTLP profile has different structure

### DIFF
--- a/cmd/profilecli/canary_exporter.go
+++ b/cmd/profilecli/canary_exporter.go
@@ -332,7 +332,6 @@ func (ce *canaryExporter) testPyroscopeCell(ctx context.Context) error {
 	// }
 
 	// ingest via OTLP HTTP/JSON
-	// Note: HTTP endpoints are not yet implemented in Pyroscope (see pkg/api/api.go:204-205)
 	if err := ce.runProbe(ctx, "ingest-otlp-http-json", func(rCtx context.Context) error {
 		return ce.testIngestOTLPHttpJson(rCtx, now)
 	}); err != nil {


cmd/profilecli/canary_exporter.go
```diff
@@ -333,19 +333,19 @@ func (ce *canaryExporter) testPyroscopeCell(ctx context.Context) error {
 
 	// ingest via OTLP HTTP/JSON
 	// Note: HTTP endpoints are not yet implemented in Pyroscope (see pkg/api/api.go:204-205)
```
simonswine: I think that line is outdated now

cmd/profilecli/canary_exporter.go
```diff
@@ -160,7 +160,7 @@ func newCanaryExporter(params *canaryExporterParams) *canaryExporter {
 	}
 
 	ce.queryProbes = append(ce.queryProbes, &queryProbe{name: "query-select-merge-profile", f: ce.testSelectMergeProfile})
-	// ce.queryProbes = append(ce.queryProbes, &queryProbe{name: "query-select-merge-otlp-profile", f: ce.testSelectMergeOTLPProfile})
+	ce.queryProbes = append(ce.queryProbes, &queryProbe{name: "query-select-merge-otlp-profile", f: ce.testSelectMergeOTLPProfile})
 
 	if params.QueryProbeSet == "all" {
```
simonswine, line 165: Idea[optional]: We have that query probe set CLI parameter, maybe we define a new one: `all-without-otlp`

In case we detect another problem we could easily roll it back without having to rebuild the canary-exporter

PR #4545, opened 2025-10-21
fix: fix recording rules on functions by relaxing ObserveSymbols condition

alsoba13:
We have this setup where SampleObserver will sequentially observe profiles while compacted, and will detect context switch such as tenant, dataset (service_name), or series (changing fingerprint between consecutive profiles). During the compaction, at the rewrite symbols step, SampleObserver also has the opportunity to observe symbols.

Symbols are written within the context of a dataset, that means that on a dataset switch, new symbols are considered and the previous ones are dispatched. When we detect a dataset switch, we find all recording rules that will be needed for the dataset and keep track of all the symbols (function names) we need to observe, which is fine.

We had this condition where we don't observe symbols if the current profile does not match any rule with a function, which appears to be wrong. The reason for this is that Symbols are only read once, so we can have this run condition where some symbols are written, but not observed (no matching rules for the current profile), but those symbols are important in a following row, but they are skipped forever.

## Example

A simple example of this is when we process CPU go profiles. Nanoseconds and samples are an exact same tree but different values. Same for memory bytes/objects, they share the topology of the tree but not the totals. 

* **rule:** record function "foo" on cpu-samples

Sample observer sequence:
| Row                     | SymbolRewriter                          | SymbolObserver                          |
|-------------------------|-----------------------------------------|-----------------------------------------|
| ...                     | ...                                     | ...                                     |
| profile_cpu_nanoseconds | rewriting new symbols                  | (no rule matching - no symbols observed) |
| profile_cpu_samples     | (no new symbols - no symbols rewritten) | (skipped)                               |

There are more complex corner cases where this can happen.

## Solution:
The proposed solution is to observe symbols for every rule on the whole dataset, no matter whether the current row matches or not. We were doing dataset-wide work here already (once we observe symbols, we observe for every rule in that dataset, not only for the current row), but we were not always observing.

There's a risk here where we observe symbols that will be never match ANY row. Solution for that would need to know beforehand if rows match, by reading them twice or reimplementing a more intelligent way of fetching symbols on demand.

### DIFF
--- a/pkg/block/compaction_test.go
+++ b/pkg/block/compaction_test.go
@@ -1,18 +1,28 @@
-package block
+package block_test
 
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"os"
 	"path/filepath"
+	"sort"
+	"strings"
 	"testing"
 
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/prompb"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/encoding/protojson"
 
 	metastorev1 "github.com/grafana/pyroscope/api/gen/proto/go/metastore/v1"
+	"github.com/grafana/pyroscope/pkg/block"
+	"github.com/grafana/pyroscope/pkg/metrics"
+	phlaremodel "github.com/grafana/pyroscope/pkg/model"
 	"github.com/grafana/pyroscope/pkg/objstore/testutil"
+	"github.com/grafana/pyroscope/pkg/test/mocks/mockmetrics"
 )
 
 func Test_CompactBlocks(t *testing.T) {
@@ -26,12 +36,12 @@ func Test_CompactBlocks(t *testing.T) {
 	require.NoError(t, err)
 
 	dst, tempdir := testutil.NewFilesystemBucket(t, ctx, t.TempDir())
-	compactedBlocks, err := Compact(ctx, resp.Blocks, bucket,
-		WithCompactionDestination(dst),
-		WithCompactionTempDir(tempdir),
-		WithCompactionObjectOptions(
-			WithObjectDownload(filepath.Join(tempdir, "source")),
-			WithObjectMaxSizeLoadInMemory(0)), // Force download.
+	compactedBlocks, err := block.Compact(ctx, resp.Blocks, bucket,
+		block.WithCompactionDestination(dst),
+		block.WithCompactionTempDir(tempdir),
+		block.WithCompactionObjectOptions(
+			block.WithObjectDownload(filepath.Join(tempdir, "source")),
+			block.WithObjectMaxSizeLoadInMemory(0)), // Force download.
 	)
 
 	require.NoError(t, err)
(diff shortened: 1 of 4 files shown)


pkg/block/testdata/profiles_recorded_shadowed.txt
alsoba13, line 1: Before the fix, this set had empty (value = 0) series. 
This is the example in the description, where cpu-samples are shadowed by cpu-nanoseconds, and memory-xxx-bytes are shadowed by memory-xxx-objects

pkg/block/compaction_test.go
```diff
@@ -60,3 +70,272 @@ func Test_CompactBlocks(t *testing.T) {
 		require.Len(t, compactedBlocks[0].Datasets, 4)
 	})
 }
+
+func Test_CompactBlocks_recordingRules(t *testing.T) {
```
alsoba13, line 74: good opportunity to have an integration test with real blocks and some relevant rules

pkg/block/testdata/profiles_recorded.txt
alsoba13, line 1: FYI, I queried the resulting compacted block, and manually checked for these numbers.

pkg/metrics/observer.go
```diff
@@ -177,11 +177,14 @@ func (o *SampleObserver) initSeriesState(row block.ProfileEntry) {
...
-// ObserveSymbols will skip observation if no rule evaluated true for matchers.
+// ObserveSymbols will observe symbols as soon as there's a function rule targeting this dataset.
+// Symbols are observed only once, and the current rule may have symbols that a future matching rule needs.
+// This is suboptimal as we may read more symbols than needed. However, the current interface does not let us do better,
+// and a bigger refactor may be needed to address this issue.
 // At the end of this process we'll have a map stacktraceId -> matching rule, so later we can get stacktraces from the
 // row and quickly look up for matching rules
 func (o *SampleObserver) ObserveSymbols(strings []string, functions []schemav1.InMemoryFunction, locations []schemav1.InMemoryLocation, stacktraceValues [][]int32, stacktraceIds []uint32) {
-	if !o.state.recordSymbols {
+	if len(o.state.targetStrings) == 0 {
```
alsoba13, line 187: Instead of observing only if the current row need to recordSymbols, we observe if there's any target string at all in the context of the dataset

simonswine, line 187: Optional: Do you think we can add a metric a hit/miss metric, how often we fetch the symbols to later not have a match in the profile?

alsoba13:
> LGTM
> 
> Thanks for finding, fixing adding a test. I do think the "proof is in the pudding", so let's see how much overhead this incurs.

I have a diff already, I ran this in dev until today's ~8:30 UTC, now main branch version is running again:
- No big deal in ObserveSymbols: +5,33% CPU [[diff](https://dev.grafana-dev.net/a/grafana-pyroscope-app/explore?searchText=&panelType=time-series&layout=grid&hideNoData=off&explorationType=diff-flame-graph&var-serviceName=fire-dev-001%2Fcompaction-worker&var-profileMetricId=process_cpu:cpu:nanoseconds:cpu:nanoseconds&var-spanSelector=undefined&var-dataSource=grafanacloud-profiles&var-filters=&var-filtersBaseline=&var-filtersComparison=&var-groupBy=all&maxNodes=16384&from=now-24h&to=now&diffFrom=2025-10-22T06:00:11.759Z&diffTo=2025-10-22T08:00:57.854Z&diffFrom-2=2025-10-22T05:07:17.379Z&diffTo-2=2025-10-22T11:07:17.379Z&from-3=now-6h&to-3=now&from-2=now-6h&to-2=now)]
- No big deal in total time spent in compaction: +1,27% p50 in compaction time [[viz](https://ops.grafana-ops.net/goto/ef1ta6fadlx4wd?orgId=1)]

This of course depends on the recording rules defined. Let's see how it looks in bigger envs.

pkg/metrics/observer.go
alsoba13: it seems interesting to have, let's consider this in the future:
https://github.com/grafana/pyroscope-squad/issues/612

PR #4548, opened 2025-10-22
fix: use github app to update-make-docs workflow

simonswine:
The update-docs workflow no longer has the correct permissions for the repo.

This will use our github app to push the necessary changes to the docs workflows

cc @jdbaldry


simonswine:
This PR was created correctly from the latest versions of the workflow: https://github.com/grafana/pyroscope/pull/4550#issuecomment-3431663756

PR #4552, opened 2025-10-22
fix: Ensure to build the same architecture image, as the current machine

simonswine:
On darwin/arm64 or linux/arm64 hosts, we currently  build a linux/amd64 image with a linux/arm64 binary.

This fixes this behaviour and aligns the image to the host's architecture.


jake-kramer (approved):
LGTM thanks!

PR #4559, opened 2025-10-23
fix: incorrect multitenancy header handling in new HTTP OTLP endpoint

oleg-kozlyuk-grafana:
New HTTP OTLP endpoints used `ExtractFromGRPCRequest` instead of `ExtractOrgIDFromHTTPRequest` by accident, which caused Pyroscope to throw `error 500` in multitenant environment


oleg-kozlyuk-grafana:
@simonswine added a set of tests

PR #4562, opened 2025-10-23
Simplify canary-exporter label check to "at least these labels"

oleg-kozlyuk-grafana:


marcsanmi (approved):
LGTM

cmd/profilecli/canary_exporter_probes.go
```diff
@@ -510,15 +504,24 @@ func (ce *canaryExporter) testLabelNames(ctx context.Context, now time.Time) err
 		model.LabelNameServiceNamePrivate,
 		model.LabelNameType,
 		model.LabelNameUnit,
-		"ingestion_method",
-		"instance",
-		"job",
```
simonswine, line 515: Why in particular do we want to exclude those labels? These are user set external labels, while all the other ones are labels that are required or set by the ingest path. I think at least one of them is essential to have

cmd/profilecli/canary_exporter_probes.go
```diff
@@ -493,12 +493,6 @@ func (ce *canaryExporter) testLabelNames(ctx context.Context, now time.Time) err
...
 	}
 
-	expectedLabelCount := 12
-	if len(respQuery.Msg.Names) != expectedLabelCount {
-		level.Error(logger).Log("msg", "received an invalid number of labels", "expected", expectedLabelCount, "received", len(respQuery.Msg.Names), "labels", strings.Join(respQuery.Msg.Names, ","))
-		return fmt.Errorf("expected %d label names, got %d", expectedLabelCount, len(respQuery.Msg.Names))
-	}
-
 	labelNames := respQuery.Msg.Names
 	slices.Sort(labelNames)
```
simonswine: No longer needed

cmd/profilecli/canary_exporter_probes.go
```diff
@@ -510,15 +504,24 @@ func (ce *canaryExporter) testLabelNames(ctx context.Context, now time.Time) err
...
 		"service.name",
 		model.LabelNameServiceName,
 	}
 
-	if !slices.Equal(labelNames, expectedLabelNames) {
-		return fmt.Errorf("expected label names to be %s, got %s", expectedLabelNames, labelNames)
+	// Use map as set for O(1) lookups
+	labelNamesSet := make(map[string]struct{}, len(labelNames))
+	for _, label := range labelNames {
+		labelNamesSet[label] = struct{}{}
```
simonswine, line 512: I think it would be good to check for duplicate label names in the output (unlikely), but the previous version would have caught it.

cmd/profilecli/canary_exporter_probes.go
oleg-kozlyuk-grafana: Not "want to exclude" per se, but I've minimized the amount of checks to one thing per fact rather than just checking for full match.

For that matter, `service.name` is preserved, it is an external label

cmd/profilecli/canary_exporter_probes.go
oleg-kozlyuk-grafana: Good point, removed

cmd/profilecli/canary_exporter_probes.go
oleg-kozlyuk-grafana: I don't think generic output correctness checks are part of scope of canary-exporter

PR #4564, opened 2025-10-24
chore(helm): replace minio with seaweedfs for object storage

### LINKED ISSUE #4557: Replace minio with alternatives
While our helm chart uses a fairly old version of minio anyhow, we likely want to move on from minio, given they stopped providing helm charts and docker images for the OSS community.

Likely alternatives we should verify are:

- https://git.deuxfleurs.fr/Deuxfleurs/garage
- https://github.com/versity/versitygw/
- https://github.com/seaweedfs/seaweedfs

ArshiyaDev:
## Summary
Replace MinIO with SeaweedFS as the object storage backend in Pyroscope Helm chart.

Fixes #4557

## Motivation
SeaweedFS offers several advantages for production deployments:
- **Production-ready**: Battle-tested in large-scale environments with proven reliability
- **S3-compatible API**: Drop-in replacement for S3, ensuring compatibility with existing tooling
- **Active community**: Strong community support and regular updates
- **Performance**: Optimized for handling large volumes of profiling data efficiently

## Changes
- Removed `minio` dependency from Chart.yaml
- Added `seaweedfs` (v4.0.393) as the storage backend
- Updated Chart.lock and dependency artifacts

## Testing
- [x] `make lint` passes
- [x] `make go/test` passes
- [x] Helm chart validates successfully


ArshiyaDev:
Hi @simonswine
I’ve implemented the changes for Issue #4557. 
Whenever you get a chance, please take a look and let me know if anything needs improvement.  
Thanks!

operations/pyroscope/helm/pyroscope/Chart.yaml
```diff
@@ -16,11 +16,11 @@ dependencies:
     version: "1.3.1"
     repository: https://grafana.github.io/helm-charts
     condition: alloy.enabled
-  - name: minio
```
simonswine, line 19: Ideally we would keep minio in here and just switch the usages from `minio.enabled` to `seaweedfs.enabled`. So in case anyone using it, can decide when to migrate at their own pace.

ArshiyaDev: okay i will handle it .

ArshiyaDev:
Thanks for the detailed feedback.

I agree with the points raised (incomplete wiring, Helm version requirement, and the MinIO random password issue).
I also agree that we should keep MinIO for backward compatibility and introduce SeaweedFS as an opt-in alternative (seaweedfs.enabled), allowing users to migrate at their own pace.

I’m happy to continue with this approach and update Helm values, docs, and checks.
If you prefer RustFS (#4681) instead, please let me know and assign to me.

PR #4567, opened 2025-10-24
chore: Update dependencies pt2

oleg-kozlyuk-grafana:


Makefile
```diff
@@ -10,7 +10,7 @@ BIN := $(CURDIR)/.tmp/bin
 COPYRIGHT_YEARS := 2021-2022
 LICENSE_IGNORE := -e /testdata/
 GO_TEST_FLAGS ?= -v -race -cover
-GO_MOD_VERSION := 1.24.0
+GO_MOD_VERSION := 1.24.6
```
aleks-p: I wonder if we should remove this altogether and let `go mod tidy` determine it automatically. 

`1.24.6` currently comes from [github.com/google/gnostic v0.7.1](https://github.com/google/gnostic/blob/39aedcfe9f00c4336420c8ed927fb0f803f921f2/go.mod#L3). If a newer version of a dependency starts asking for something higher, running `go mod tidy` manually would produce different results compared to `make go/mod` or the make target would start failing. 

I think @marcsanmi and @simonswine have touched this in the past and might have opinions

marcsanmi: That's a good point. Do we prefer an error or a working update?Dependencies already dictate the minimum version. Maybe it's a good time to just rely on `go mod tidy`.

marcsanmi (approved):

Makefile
oleg-kozlyuk-grafana: Removed GO_MOD_VERSION altogether - pending more input

Issue #3989, opened 2025-03-11
[bug][pyroscope] Pyroscope does not ingest Speedscope files with multiple profiles


lagutrop:
#### Describe the bug
I've been using Pyroscope version 0.37.2 for a couple of months and decided to upgrade to the new version of Pyroscope (1.12). In the above version it was possible to call the ingest endpoint (POST localhost:4040/ingest?name=MyServiceName&from=1741682023&until=1741682083&format=speedscope) with multiple profiles like the example that you provide in your TestData folder.
Now with the upgrade to 1.12.0, I'm getting the following error

```
{
    "code": "unknown",
    "message": "invalid_argument: pyroscopeIngesterAdapter failed to convert metadata: unknown profile type: seconds"
}
```

#### To Reproduce
Use the last release build 1.12.0 and call the following endpoint in Postman for example

POST localhost:4040/ingest?name=MyServiceName&from=1741682023&until=1741682083&format=speedscope

In body select Raw and insert the following Speedscope

```
{
  "exporter": "speedscope@0.6.0",
  "$schema": "https://www.speedscope.app/file-format-schema.json",
  "name": "Two Samples",
  "activeProfileIndex": 1,
  "profiles": [
    {
      "type": "sampled",
      "name": "one",
      "unit": "seconds",
      "startValue": 0,
      "endValue": 14,
      "samples": [
        [0, 1, 2],
        [0, 1, 2],
        [0, 1, 3],
        [0, 1, 2],
        [0, 1]
      ],
      "weights": [1, 1, 4, 3, 5]
    },
    {
      "type": "sampled",
      "name": "two",
      "unit": "seconds",
      "startValue": 0,
      "endValue": 14,
      "samples": [
        [0, 1, 2],
        [0, 1, 2],
        [0, 1, 3],
        [0, 1, 2],
        [0, 1]
      ],
      "weights": [1, 1, 4, 3, 5]
    }
  ],
  "shared": {
    "frames": [
      { "name": "a" },
      { "name": "b" },
      { "name": "c" },
      { "name": "d" }
    ]
  }
}
```

The following error will appear, with status code 422 Unprocessable Entity.

```
{
    "code": "unknown",
    "message": "invalid_argument: pyroscopeIngesterAdapter failed to convert metadata: unknown profile type: seconds"
}
```

#### Expected behavior

[...]

PR #4568, opened 2025-10-24
fix: Handle Speedscope sample types in ingestion

jake-kramer:
Fixes https://github.com/grafana/pyroscope/issues/3989

### DIFF
--- a/pkg/test/integration/ingest_speedscope_test.go
+++ b/pkg/test/integration/ingest_speedscope_test.go
@@ -42,6 +42,14 @@ var (
 				{"wall:wall:nanoseconds:cpu:nanoseconds", nil, 0},
 			},
 		},
+		{
+			name:           "multi profile samples bytes units",
+			speedscopeFile: testdataDirSpeedscope + "/two-sampled-bytes.speedscope.json",
+			expectStatus:   200,
+			expectedMetrics: []expectedMetric{
+				{"memory:samples:bytes::", nil, 0},
+			},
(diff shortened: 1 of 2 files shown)


pkg/test/integration/helper.go
```diff
@@ -244,7 +244,7 @@ func (p *PyroscopeTest) NewRequestBuilder(t *testing.T) *RequestBuilder {
 }
 
 func (p *PyroscopeTest) TempAppName() string {
-	return fmt.Sprintf("pprof.integration.%d",
+	return fmt.Sprintf("pprof-integration-%d",
```
jake-kramer, line 247: The dots were messing with downstream sample type parsing: https://github.com/grafana/pyroscope/blob/8db0333af86d5d8dafa557aea1f9c30535eef4c3/pkg/ingester/pyroscope/ingest_adapter.go#L202-L209

pkg/test/integration/ingest_speedscope_test.go
```diff
@@ -0,0 +1,72 @@
...
+	expectStatus    int
+	expectedMetrics []expectedMetric
+}
+
+const (
+	testdataDirSpeedscope = repoRoot + "pkg/og/convert/speedscope/testdata"
+)
+
+var (
+	speedscopeTestData = []speedscopeTestDataStruct{
```
marcsanmi, line 25: Could we add a case for the `bytes` unit?

PR #4580, opened 2025-10-29
Improve metastore observability

aleks-p:
Adds tracing to the metastore public endpoints. This should help troubleshooting cases where the metastore latency goes up. Fixes https://github.com/grafana/pyroscope-squad/issues/603.

Here is an example for the IndexService.AddBlock RPC:

Screenshot 2025-10-29 at 14 27 42

### DIFF
--- a/pkg/metastore/fsm/fsm.go
+++ b/pkg/metastore/fsm/fsm.go
@@ -275,8 +275,6 @@ func (fsm *FSM) applyCommand(cmd *raft.Log) any {
 		return errResponse(cmd, fmt.Errorf("unknown command type: %d", e.Type))
 	}
 
-	fsm.metrics.contextRegistrySize.Set(float64(fsm.contextRegistry.Size()))
-
 	// Apply is never called concurrently with Restore, so we don't need
 	// to lock the FSM: db.boltdb is guaranteed to be in a consistent state.
 	rawTx, err := fsm.db.boltdb.Begin(true)
--- a/pkg/metastore/raftnode/node.go
+++ b/pkg/metastore/raftnode/node.go
@@ -313,7 +313,7 @@ func (n *Node) TransferLeadership() (err error) {
 // Propose makes an attempt to apply the given command to the FSM.
 // The function returns an error if node is not the leader.
 func (n *Node) Propose(ctx context.Context, t fsm.RaftLogEntryType, m proto.Message) (resp proto.Message, err error) {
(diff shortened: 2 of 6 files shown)


pkg/metastore/tracing/context_registry.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/grafana/pyroscope/pkg/util"
+)
+
+// ContextRegistry maintains a mapping of IDs to contexts for tracing purposes.
+// This allows us to propagate tracing context from HTTP handlers down to BoltDB transactions
+// without persisting the context in Raft logs.
+//
+// The registry is only used on the leader node where Propose() is called. Followers will
+// not have contexts available and will use context.Background() instead.
+type ContextRegistry struct {
```
aleks-p, line 19: This will add some memory overhead but I think it is worth it for gaining more visibility. If anyone has alternative ideas, I'd be happy to rework this part.

pkg/metastore/fsm/fsm.go
```diff
@@ -234,6 +246,18 @@ func (fsm *FSM) applyCommand(cmd *raft.Log) any {
 	if err := e.UnmarshalBinary(cmd.Data); err != nil {
 		return errResponse(cmd, err)
 	}
+
+	ctx := context.Background()
+	if ctxID := string(cmd.Extensions); ctxID != "" {
```
aleks-p, line 251: We use the extension field of the raft Log type, to carry the context identifiers

https://github.com/hashicorp/raft/blob/c4fb904a43d299a2b9ba1000ae18a5b3f60994ab/log.go#L94

pkg/metastore/raftnode/node.go
```diff
@@ -300,18 +312,44 @@ func (n *Node) TransferLeadership() (err error) {
...
+	span.LogFields(otlog.String("msg", "log entry marshalled"))
 	timer := prometheus.NewTimer(n.metrics.apply)
 	defer timer.ObserveDuration()
-	future := n.raft.Apply(raw, n.config.ApplyTimeout)
+
+	span.LogFields(otlog.String("msg", "applying log entry"))
+
+	future := n.raft.ApplyLog(raft.Log{
+		Data:       raw,
+		Extensions: []byte(ctxID),
```
aleks-p, line 342: We use the extension field of the raft Log type, to carry the context identifiers (writing point)

https://github.com/hashicorp/raft/blob/c4fb904a43d299a2b9ba1000ae18a5b3f60994ab/log.go#L94

PR #4586, opened 2025-10-31
fix(api): add ingest limits to connect push requests

korniltsev-grafanista:
There are ingestion body size limits in the `/ingest` handler. This PR brings the same limit to the `Push` api.

It does not support per-tenant overrides for now, only defaults.
I could not find how to implement per tenant limits as the connect interceptor seem to be late and already read and deserialized the message.
We can probably try to do something like this https://github.com/grafana/pyroscope/commit/ba4c2a66940a010d2007961d2c983898ac1a1fd7 , but I don't like this option much, so for now not use it.

### DIFF
--- a/pkg/api/connect_options.go
+++ b/pkg/api/connect_options.go
@@ -2,12 +2,12 @@ package api
 
 import (
 	"connectrpc.com/connect"
-	"github.com/grafana/pyroscope/pkg/tenant"
-	"github.com/grafana/pyroscope/pkg/validation"
 
 	connectapi "github.com/grafana/pyroscope/pkg/api/connect"
+	"github.com/grafana/pyroscope/pkg/tenant"
 	"github.com/grafana/pyroscope/pkg/util"
 	"github.com/grafana/pyroscope/pkg/util/delayhandler"
+	"github.com/grafana/pyroscope/pkg/validation"
 )
 
 func connectInterceptorRecovery() connect.HandlerOption {


korniltsev-grafanista:
I've tested this change manually with this snippet

```
func TestIngestLarge(t *testing.T) {
	builder := testhelper.NewProfileBuilder(int64(1)).
		CPUProfile().
		WithLabels(
			"job", "test",
			"service_name", "svc1",
		)
	builder.ForStacktraceString("foo", "bar", "baz").AddSamples(1)
	for i := 0; i < 20*1024; i++ {
		builder.ForStacktraceString(
			fmt.Sprintf("foojasfdljasdklfjas;lkd%d", i),
			fmt.Sprintf("barasdfjaosdhfadhf%d", i),
			fmt.Sprintf("bazashdifoahsdfohasdf%d", i),
		).AddSamples(3)
	}
	rawProfile, err := builder.MarshalVT()
	require.NoError(t, err)
	require.Greater(t, len(rawProfile), 2*1024*1024)
	pusher := pushv1connect.NewPusherServiceClient(
		http.DefaultClient,
		"http://localhost:4040", // -distributor.ingestion-body-limit-mb=2
		connectapi.DefaultClientOptions()...,
	)
	_, err = pusher.Push(context.Background(), connect.NewRequest(&pushv1.PushRequest{
		Series: []*pushv1.RawProfileSeries{{
			Labels:  builder.Labels,
			Samples: []*pushv1.RawSample{{RawProfile: rawProfile}},
		}},
	}))
	require.Error(t, err)
	fmt.Println(err) // resource_exhausted: message size 3386424 is larger than configured max 2097152
}
```
I will not add any tests in this PR.

marcsanmi:
A random idea, which I didn't verified: To support per-tenant limits, we could move auth to HTTP middleware (instead of Connect interceptor) for this endpoint, then wrap as: auth middleware → body limit middleware → Connect handler. This would ensure tenant ID is extracted before body size is checked. This would make the Pusher service different from other Connect endpoints, similar to the reverted commit...

Issue #4584, opened 2025-10-30
Speedscope profiles dropped by compaction


jake-kramer:
#### Describe the bug

Uploading a Speedscope profile via the `/ingest` endpoint will not fully persist. The initial upload is correct, but series (observed by seeing missing symbols in the UI) will slowly drop over time.

#### To Reproduce

1. Start Pyroscope on localhost:4040
2.
```
$ curl -X POST "localhost:4040/ingest?name=MyServiceName&from=$(gdate -d '6 minutes ago' +%s)&until=$(gdate -d '15 minutes ago' +%s)&format=speedscope" \
> -H "Content-Type: application/json" \
> -d @speedscope.json
```
3. Observe the `Thread (*)` symbols at `localhost:4040` being dropped over time as compaction runs

[1] [speedscope.json](https://github.com/user-attachments/files/23244803/speedscope.json)

#### Additional Context

Originally reported: https://github.com/grafana/pyroscope/issues/3989#issuecomment-3457559250

This diff applied to the [speedscope-ingestion-fix branch](https://github.com/grafana/pyroscope/pull/4568/commits/f1ecaad9f64d7d2569f88d9fe06946ab8a274eb7) is a hack that resolves the issue, showing the root cause.
```diff
diff --git a/pkg/og/convert/speedscope/parser.go b/pkg/og/convert/speedscope/parser.go
index 5e0214912..a183c858d 100644
--- a/pkg/og/convert/speedscope/parser.go
+++ b/pkg/og/convert/speedscope/parser.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"time"
 
 	"github.com/grafana/pyroscope/pkg/og/ingestion"
 	"github.com/grafana/pyroscope/pkg/og/storage"
@@ -52,8 +53,8 @@ func parseAll(rawData []byte, md ingestion.Metadata) ([]*storage.PutInput, error
 		LabelSet:   md.LabelSet,
 	}
 
-	for _, prof := range file.Profiles {
-		putInput, err := parseOne(&prof, input, file.Shared.Frames, len(file.Profiles) > 1)
+	for i, prof := range file.Profiles {
+		putInput, err := parseOne(i, &prof, input, file.Shared.Frames, len(file.Profiles) > 1)
 		if err != nil {
 			return nil, err
 		}
[...]

PR #4588, opened 2025-10-31
fix: Add `profile_name` label to Speedscope profile ingestion

jake-kramer:
This label is important to prevent all speedscope profiles from the same ingestion upload being deduped during compaction

Fixes https://github.com/grafana/pyroscope/issues/4584


jake-kramer:
Has an additional bonus effect of allowing for filtering by thread for [this specific](https://github.com/grafana/pyroscope/issues/3989#issuecomment-3457559250) Speedscope profile.

image

korniltsev-grafanista (approved):
lgtm

While this fix fixes the issue with the problematic profile in the linked issue, there still may be a problem of wrongly deduplicated profile if the profile names are not unique. I think it makes sense (maybe in a followup) to explicitly merge the needed profiles before returning/forwarding them to ingesters.

PR #4592, opened 2025-11-03
fix: Implement Speedscope profile merging

jake-kramer:
Profile merging is used to prevent distinct profiles from being compacted/deduped downstream.

Follow up to https://github.com/grafana/pyroscope/pull/4588

### DIFF
--- a/pkg/og/convert/speedscope/parser.go
+++ b/pkg/og/convert/speedscope/parser.go
@@ -1,9 +1,11 @@
 package speedscope
 
 import (
+	"cmp"
 	"context"
 	"encoding/json"
 	"fmt"
+	"slices"
 
 	"github.com/grafana/pyroscope/pkg/og/ingestion"
 	"github.com/grafana/pyroscope/pkg/og/storage"
@@ -75,7 +77,6 @@ func mergeProfiles(profiles []profile) []profile {
 		t          string
 		unit       unit
 		startValue float64
-		endValue   float64
 	}
 
 	merged := make(map[mergeKey]profile)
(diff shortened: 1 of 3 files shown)


pkg/og/convert/speedscope/parser.go
```diff
@@ -62,6 +65,42 @@ func parseAll(rawData []byte, md ingestion.Metadata) ([]*storage.PutInput, error
...
+			mergedProf.Samples = append(mergedProf.Samples, prof.Samples...)
+			mergedProf.Events = append(mergedProf.Events, prof.Events...)
+			mergedProf.Weights = append(mergedProf.Weights, prof.Weights...)
+			merged[k] = mergedProf
+		} else {
+			merged[k] = prof
+		}
+	}
+
+	return maps.Values(merged)
```
simonswine: Nit: Not too sure if a big deal, but this randomizes the order of the results, maybe we order it by startValue

pkg/og/convert/speedscope/parser.go
```diff
@@ -62,6 +65,42 @@ func parseAll(rawData []byte, md ingestion.Metadata) ([]*storage.PutInput, error
...
+// This prevents situations downstream where two different
+// profiles are deduped during congestion for having the
+// same label set and timestamp.
+func mergeProfiles(profiles []profile) []profile {
+	type mergeKey struct {
+		name       string
+		t          string
+		unit       unit
+		startValue float64
+		endValue   float64
```
simonswine: I think endValue (not super familiar) but I think this is the end timestamp, will not be represented in pprof, so it will still conflict as long as all the other values are matching.

This also need to be handled as part of the merge

jake-kramer: Yes, good call out!

PR #4594, opened 2025-11-05
chore: parallelize mapping symbolization to improve query performance

marcsanmi:
Aims to solve/improve https://github.com/grafana/pyroscope-squad/issues/632

Parallelise symbolizer mapping processing to fix timeouts with eBPF profiles containing multiple unsymbolized modules. Previously `symbolize` was processed sequentially for each mapping; now it's done concurrently with max 10 parallel requests.

Tested this in the ops service causing timeout and it reduced considerably the amount of time, firs time loading around 7-8s subsequent calls around 3-4 due to caching.

### DIFF
--- a/docs/sources/release-notes/v1-14.md
+++ b/docs/sources/release-notes/v1-14.md
@@ -2,7 +2,7 @@
 title: Version 1.14 release notes
 menuTitle: V1.14
 description: Release notes for Grafana Pyroscope 1.14
-weight: 250
+weight: 200
 ---
 
 ## Version 1.14.1 release notes
--- a/docs/sources/release-notes/v1-15.md
+++ b/docs/sources/release-notes/v1-15.md
@@ -2,7 +2,7 @@
 title: Version 1.15 release notes
 menuTitle: V1.15
 description: Release notes for Grafana Pyroscope 1.15
-weight: 250
+weight: 150
 ---
 
 ## Version 1.15.0 release notes
--- a/docs/sources/release-notes/v1-16.md
+++ b/docs/sources/release-notes/v1-16.md
@@ -1,0 +1,61 @@
+---
+title: Version 1.16 release notes
+menuTitle: V1.16
+description: Release notes for Grafana Pyroscope 1.16
+weight: 100
+---
+
+## Version 1.16.0 release notes
+
+The Pyroscope team is excited to present Grafana Pyroscope 1.16.0
+
+This release contains enhancements, fixes, improves stability & performance.
+
+Notable changes are listed below. For more details, check out the [1.16.0 changelog](https://github.com/grafana/pyroscope/compare/v1.15.0...v1.16.0).
+
+### Enhancements
+* Add OTLP HTTP/json and HTTP/protobuf ingestion (#4495)
+* Add a basic metric for metastore DLQ recovery (#4563)
+* Improve metastore observability (#4580)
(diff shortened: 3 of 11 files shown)


pkg/symbolizer/symbolizer.go
```diff
@@ -84,36 +85,100 @@ func (s *Symbolizer) SymbolizePprof(ctx context.Context, profile *googlev1.Profi
...
+}
+
+// symbolizeMappingsConcurrently symbolizes multiple mappings concurrently with a concurrency limit.
+func (s *Symbolizer) symbolizeMappingsConcurrently(
+	ctx context.Context,
+	profile *googlev1.Profile,
+	locationsByMapping map[uint64][]*googlev1.Location,
+) ([]symbolizedLocation, error) {
+	// Concurrency limit to avoid overwhelming debuginfod server
+	const maxConcurrency = 10
```
bryanhuhta: Could be nice to have this as a config value so it can be tweaked without a recompile or building a new image

pkg/symbolizer/symbolizer.go
```diff
@@ -84,36 +85,99 @@ func (s *Symbolizer) SymbolizePprof(ctx context.Context, profile *googlev1.Profi
...
 
-	return nil
+	if err != nil {
+		return nil, err
+	}
+
+	var allSymbolizedLocs []symbolizedLocation
+	for result := range results {
+		allSymbolizedLocs = append(allSymbolizedLocs, result.locations...)
+	}
```
bryanhuhta: This might be overengineering so I won't make it blocking, but I wonder if we could use an atomic int to calculate the size of `allSymbolizedLocs` ahead of time? I'm not sure how large this slice might get, but we could theoretically be doing a lot of allocs here. It would be nice to do a single alloc all at once.

Also, I assume the order of the symbolized locations doesn't matter here? The synchronous approach would maintain a deterministic ordering, while this approach will be nondeterministic.

marcsanmi: > This might be overengineering so I won't make it blocking, but I wonder if we could use an atomic int to calculate the size of allSymbolizedLocs ahead of time? I'm not sure how large this slice might get, but we could theoretically be doing a lot of allocs here. It would be nice to do a single alloc all at once.

Good point. I've pre-allocated `allSymbolizedLocs` by calculating `totalLocs` from `locationsByMapping` before starting concurrent work, which avoids multiple reallocations. I'd say an atomic counter isn't needed since we already have the exact count synchronously from the input data - using atomics would add unnecessary complexity for the same result.

> Also, I assume the order of the symbolized locations doesn't matter here? The synchronous approach would maintain a deterministic ordering, while this approach will be nondeterministic.

It shouldn't matter since we [use](https://github.com/grafana/pyroscope/blob/0069df714ed03f97105dd96c4dffe3690f4a3cb1/pkg/symbolizer/symbolizer.go#L266) `locIdx := loc.Id - 1` in `updateAllSymbolsInProfile`; we use ids and not positions

pkg/symbolizer/symbolizer.go
marcsanmi: Yes, since I don't expect this needed to be changed often I decided to leave it here but we can add it whenever we want.

marcsanmi: Done it in 0e360590c453e3f3167f7f4636262ced6363eace

PR #4597, opened 2025-11-06
chore: minor symbolizer metrics improvements

marcsanmi:


pkg/symbolizer/symbolizer.go
```diff
@@ -384,11 +384,12 @@ func (s *Symbolizer) symbolizeWithTable(table *lidia.Table, req *request) {
...
 func (s *Symbolizer) getLidiaBytes(ctx context.Context, buildID string) ([]byte, error) {
 	if client, ok := s.client.(*DebuginfodHTTPClient); ok {
-		if found, _ := client.notFoundCache.Get(buildID); found {
-			s.metrics.cacheOperations.WithLabelValues("not_found", "get", statusSuccess).Inc()
-			return nil, buildIDNotFoundError{buildID: buildID}
+		if sanitizedBuildID, err := sanitizeBuildID(buildID); err == nil {
+			if found, _ := client.notFoundCache.Get(sanitizedBuildID); found {
+				s.metrics.cacheOperations.WithLabelValues("not_found", "get", statusSuccess).Inc()
+				return nil, buildIDNotFoundError{buildID: buildID}
+			}
```
alsoba13, line 391: I might fail to understand the whole picture, but shouldn't you increment the miss count if `found == false`?
```suggestion
		if sanitizedBuildID, err := sanitizeBuildID(buildID); err == nil {
			if found, _ := client.notFoundCache.Get(sanitizedBuildID); found {
				s.metrics.cacheOperations.WithLabelValues("not_found", "get", statusSuccess).Inc()
				return nil, buildIDNotFoundError{buildID: buildID}
			}
			s.metrics.cacheOperations.WithLabelValues("not_found", "get", "miss").Inc()
```

marcsanmi: Yes... This is tricky.  We might have 2 cache lookups for the same buildID in the same request flow:
- First lookup at line 388 → miss
- Second lookup at [debuginfod client](https://github.com/grafana/pyroscope/blob/5d9ff7f125a7d9cd8cc6181292ab4e1e0e36f2c0/pkg/symbolizer/debuginfod_client.go#L120) → miss

Semantically, I'd say is 1 miss because:
  - It's the same cache
  - Same buildID
  - Same request
  -  We check twice as a design pattern: first to skip expensive object storage lookups, then as a final check before HTTP (catches cases where concurrent requests finish between the checks and populate the cache; very narrow but might happen...)
  
On my mind this is how it would be:  
  - Count hits at both locations (they're mutually exclusive)
  - Count misses at only one location (the authoritative one where we actually proceed to HTTP)
  
  LMWKY; we might just want to count double misses here

alsoba13: I see now, thank you!

LGTM

PR #4605, opened 2025-11-10
feat: add proto definitions for individual profile retrieval API

marcsanmi:
Add protobuf definitions for individual profile retrieval API

### Proto Definitions
 
 1. Exemplar Support in Timeseries (api/types/v1/types.proto)

- Added Exemplar message containing metadata for individual profile samples:
  - profile_id: Profile UUID for retrieval
  - span_id: Span ID if profile was split by span during ingestion
  - timestamp: Capture time (ms since epoch)
  - value: Total sample value (e.g., CPU ns, bytes)
  - labels: Series labels NOT included in group_by query parameter

- Added exemplars field to Point message for timeseries.

 2. Profile ID Selector (api/querier/v1/querier.proto)
- Extended `SelectMergeStacktracesRequest` with `profile_id_selector` field to query individual profiles by UUID. This will allow users to click on an exemplar and view the specific profile as a flamegraph.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -306,7 +306,7 @@ docker-image/pyroscope/dlv:
 	# dlv is not intended for local use and is to be installed in the
 	# platform-specific docker image together with the main Pyroscope binary.
 	@mkdir -p $(@D)
-	GOPATH=$(CURDIR)/.tmp GOAMD64=v2 CGO_ENABLED=0 $(GO) install -ldflags "-s -w -extldflags '-static'" github.com/go-delve/delve/cmd/dlv@v1.23.0
+	GOPATH=$(CURDIR)/.tmp GOAMD64=v2 CGO_ENABLED=0 $(GO) install -ldflags "-s -w -extldflags '-static'" github.com/go-delve/delve/cmd/dlv@v1.25.2
 	mv $(CURDIR)/.tmp/bin/$(GOOS)_$(GOARCH)/dlv $(CURDIR)/.tmp/bin/dlv
 
 .PHONY: clean
--- a/api/connect-openapi/gen/vcs/v1/vcs.openapi.yaml
+++ b/api/connect-openapi/gen/vcs/v1/vcs.openapi.yaml
@@ -409,6 +409,10 @@ components:
           type: string
           title: rootPath
           description: the root path where the project lives inside the repository
+        functionName:
+          type: string
+          title: functionName
+          description: the function name as provided by the symbols
       title: GetFileRequest
       additionalProperties: false
     vcs.v1.GetFileResponse:
@@ -435,8 +439,8 @@ components:
           type: string
           title: clientID
           description: |-
-            clientID must be propagated when calling https://github.com/login/oauth/authorize
-             in the client_id query parameter.
+            clientID must be propagated when calling
+             https://github.com/login/oauth/authorize in the client_id query parameter.
         callbackURL:
           type: string
           title: callbackURL
@@ -462,10 +466,11 @@ components:
           title: cookie
           description: |-
             Deprecated
-             In future version, this cookie won't be sent. Now, old cookie is sent alongside the new expected
-             data (token, token_expires_at and refresh_token_expires_at). Frontend will be responsible of computing
-             its own cookie from the new data.
-             Remove after completing https://github.com/grafana/explore-profiles/issues/187
+             In future version, this cookie won't be sent. Now, old cookie is sent
+             alongside the new expected data (token, token_expires_at and
+             refresh_token_expires_at). Frontend will be responsible of computing its
+             own cookie from the new data. Remove after completing
+             https://github.com/grafana/explore-profiles/issues/187
         token:
           type: string
           title: token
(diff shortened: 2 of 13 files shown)


api/types/v1/types.proto
```diff
@@ -133,3 +135,22 @@ message GetProfileStatsResponse {
...
+
+// Exemplar represents metadata for an individual profile sample.
+// Exemplars allow users to drill down from aggregated timeline views
+// to specific profile instances.
+message Exemplar {
+  // Milliseconds since epoch when the profile was captured.
+  int64 timestamp = 1 [(gnostic.openapi.v3.property).example = {yaml: "1730000023000"}];
+
+  // Unique identifier for the profile (UUID).
+  string id = 2 [(gnostic.openapi.v3.property).example = {yaml: "7c9e6679-7425-40de-944b-e07fc1f90ae7"}];
```
simonswine: With span profiles there can be more than a single exemplar per timestamp/id.

I think the span ID needs to be promoted to an attribute here.

```suggestion
  string profile_id = 2 [(gnostic.openapi.v3.property).example = {yaml: "7c9e6679-7425-40de-944b-e07fc1f90ae7"}];
  string span_id = 2.5 [(gnostic.openapi.v3.property).example = {yaml: "deadbeef"}];```

File: api/types/v1/types.proto [hunk 2]
```diff
@@ -133,3 +135,22 @@ message GetProfileStatsResponse {
...
+
+  // Total sample value for this profile (e.g., CPU nanoseconds, bytes allocated).
+  uint64 value = 3 [(gnostic.openapi.v3.property).example = {yaml: "2450000000"}];
+
+  // Optional dynamic labels that were used to split the profile during ingestion
+  // (e.g., trace_id, span_id). Static series labels are not included as they're
+  // already present in the parent Series.
+  repeated LabelPair labels = 4;
```
simonswine: Not too sure if I fully understand that comment, but AFAIK the parent series only contains the labels that you specify in `groupBy`,  so would this contain all the labels that have been common for each profile id?

api/types/v1/types.proto
marcsanmi: Is 2.5 a valid number?

api/types/v1/types.proto
marcsanmi: Description wasn't accurate. Updated as talked in the meeting ecc872c4fd010e273811eb6785c7d64622e99105

api/types/v1/types.proto
marcsanmi: Done in da300889e8f8d6bef0cd1a1494bf78ef8100bf30

api/querier/v1/querier.proto
```diff
@@ -217,6 +219,7 @@ message SelectSeriesRequest {
   optional types.v1.StackTraceSelector stack_trace_selector = 8;
   // Select the top N series by total value.
   optional int64 limit = 9;
+  optional bool include_exemplars = 10;
```
alsoba13: Having this optional generates a pointer to a bool:
```
    	...
    	IncludeExemplars *bool  `protobuf:"varint,10,opt,name=include_exemplars,json=includeExemplars,proto3,oneof" json:"include_exemplars,omitempty"`
    	...
```

Can we have this non-optional? I **assume** it would generate a regular bool (defaulted to false if not present, which works great)

simonswine: What @alsoba13 said and I also think `bool`s are really inflexible, I would prefer to have an enum exemplarType:

```
- NONE
- INDIVIDUAL
- SPAN
```

Which could be extended even further. Eventually I think we will also need parameters that say the method we truncate the exemplars and something like `maxExemplars`

marcsanmi: Adde `ExemplarType` enum in this commit f133b642fe36dfcdbf17d8b491a368a5c1a968d5. Even it might feel a bit redundant I added it as `optional` as well as with the `UNSPECIFIED` field for consistency with the rest of our api enums and as a good practise - lmkwyt.

api/querier/v1/querier.proto
```diff
@@ -217,6 +219,8 @@ message SelectSeriesRequest {
   optional types.v1.StackTraceSelector stack_trace_selector = 8;
   // Select the top N series by total value.
   optional int64 limit = 9;
+  // Type of exemplars to include in the response.
+  optional types.v1.ExemplarType exemplar_type = 10;
```
alsoba13: What about this?
```suggestion
  types.v1.ExemplarType exemplar_type = 10;
```
we already have `EXEMPLAR_TYPE_UNSPECIFIED` for undefined, and having to deal with nil check is annoying and error prone:
```
if c.Msg.ExemplarType != nil && *c.Msg.ExemplarType == typesv1.ExemplarType_EXEMPLAR_TYPE_INDIVIDUAL {
```

alsoba13: Okay, ignore this I just saw your other comment. Let's keep as is for consistency

alsoba13: I just want to give a bit more of my opinion so maybe you can reconsider this:
1. I see most of the enums being used without `optional`. There's just one exception: `types.v1.TimeSeriesAggregationType` at `SelectProfilesRequest`.
2. It's not only tedious use it in the server side, but also in the client side:
```
	// define it
	exemplarType := typesv1.ExemplarType_EXEMPLAR_TYPE_INDIVIDUAL,
	req := connect.NewRequest(&querierv1.SelectSeriesRequest{
		...
		ExemplarType:  &exemplarType, // reference it
		...
	})
```
would be just great to consume it like this:
```
	req := connect.NewRequest(&querierv1.SelectSeriesRequest{
		...
		ExemplarType:  typesv1.ExemplarType_EXEMPLAR_TYPE_INDIVIDUAL,
		...
	})
```

marcsanmi: Yep, that's the counterside.. but since we have `UNSPECIFIED` maybe we can just get rid of the `optional`; I'll update it.

marcsanmi: Done in 98b88e595b0bd6bc78fcc4b753dd64c0eee16fdd

PR #4606, opened 2025-11-11
feat: Add `.pyroscope.yaml` with source code mapping support

simonswine:
This PR adds support for `.pyroscope.yaml` configuration files for custom source code mappings and adds logic to use those mappings to find Go and Java code

  - Added .pyroscope.yaml configuration support for custom source code location mappings
  - Supports both local paths and GitHub repository mappings for Go and Java
  - Implements prefix-based matching with longest-prefix-wins strategy
  - Enhanced Go standard library resolution with version detection and configurable source
  - Added function_name parameter to file lookups, so java profiles (that don't contain file paths), can be looked up.
  - Improved error tracing with OpenTracing tags in GitHub client

Configuration Example (also in the PR):

This will map a special path prefix called `$GOROOT` to a particular go version. This signals to the backend the go version used to compile this should be kept update to date with the version of Pyroscope used to build.

```
  source_code:
    mappings:
      - path:
          - prefix: $GOROOT/src
        language: go
        source:
          github:
            owner: golang
            repo: go
            ref: go1.24.8
            path: src
```

### DIFF
--- a/examples/language-sdk-instrumentation/java/rideshare/.pyroscope.yaml
+++ b/examples/language-sdk-instrumentation/java/rideshare/.pyroscope.yaml
@@ -1,3 +1,4 @@
+version: v1
 source_code:
   mappings:
     - function_name:
--- a/.pyroscope.yaml
+++ b/.pyroscope.yaml
@@ -1,3 +1,4 @@
+version: v1
 source_code:
   mappings:
     - path:


api/connect-openapi/gen/vcs/v1/vcs.openapi.yaml
```diff
@@ -409,6 +409,10 @@ components:
           type: string
           title: rootPath
           description: the root path where the project lives inside the repository
+        functionName:
```
jake-kramer, line 412: Looks like currently logic is: use localPath if `go`, use functionName if `java`. May lead to confusion down the road if we start seeing java profiles with the file paths properly set.

examples/language-sdk-instrumentation/java/rideshare/.pyroscope.yaml
```diff
@@ -0,0 +1,36 @@
+source_code:
```
jake-kramer, line 2: This config file is great. Wondering if there's also an opportunity to auto-detect Java files w/o a config file too and support automatic 3rd party path resolution like we do with go? I imagine at Uber scale it would be a large lift to make a config file that captures everything they'd like.

api/connect-openapi/gen/vcs/v1/vcs.openapi.yaml
simonswine: Yes I guess the profiling data you send needs to match with the `.pyroscope.yaml` you set. I do think this makes sense and is both in customer control.

examples/language-sdk-instrumentation/java/rideshare/.pyroscope.yaml
simonswine: I think the only opportunity we have is setting it on the client site (as a parameter to GetFile).

Then we could try to match a path in the repo with the package path. Still fairly tricky to detect the source code prefix (like `src/main` or `<package>/src/main`

pkg/frontend/vcs/config/config.go
```diff
@@ -0,0 +1,190 @@
...
+
+var validLanguages = []Language{
+	LanguageGo,
+	LanguageJava,
+}
+
+// PyroscopeConfig represents the structure of .pyroscope.yaml configuration file
+type PyroscopeConfig struct {
+	SourceCode SourceCodeConfig `yaml:"source_code"`
+}
```
bryanhuhta, line 38: I think we also want a version tag here so we can continue to support older versions easily.

pkg/frontend/vcs/config/config.go
```diff
@@ -0,0 +1,190 @@
...
+	var bestMatch *MappingConfig
+	var bestMatchLen = -1
+	for _, m := range c.SourceCode.Mappings {
+		if result := m.Match(file); result > bestMatchLen {
+			bestMatch = &m
+			bestMatchLen = result
+		}
+	}
+	return bestMatch
+}
```
bryanhuhta, line 186: It'll be interesting to see how expensive this linear search is time-wise for very large projects. I suspect there could easily be many thousands of these mappings.

It's out of scope of this PR, but later we may need to improve this search heuristic.

pkg/frontend/vcs/source/find.go
```diff
@@ -22,52 +26,119 @@ type VCSClient interface {
...
+		content, err := ff.fetchRepoFile(ctx, ff.file.Path, ff.ref)
 		if err != nil {
 			return nil, err
 		}
 		return newFileResponse(content.Content, content.URL)
 	}
 }
 
+// loadConfig attempts to load .pyroscope.yaml from the repository root
+func (ff *FileFinder) loadConfig(ctx context.Context) {
```
bryanhuhta, line 100: Again a future improvement, but it would be nice to cache the config + git commit (at least in memory, maybe even in object storage) so we only need to make this call once.

examples/language-sdk-instrumentation/java/rideshare/.pyroscope.yaml
jake-kramer: Noting here that we discussed offline a tool for programmatically generating this config that I'm working on. Has the added benefit of moving the expensive logic of translating function names to git repos outside of the `GetFile` call path.

pkg/frontend/vcs/config/config.go
```diff
@@ -0,0 +1,190 @@
...
+
+	if err := m.Source.Validate(); err != nil {
+		errs = append(errs, err)
+	}
+
+	return errors.Join(errs...)
+}
+
+// Validate checks if a source configuration is valid
+func (m *Source) Validate() error {
```
jake-kramer, line 132: Would be helpful error message to surface if the supplied source is not supported (i.e. not github or local)

pkg/frontend/vcs/source/find.go
```diff
@@ -22,52 +26,119 @@ type VCSClient interface {
...
+		return ff.findJavaFile(ctx, mapping)
+	// todo: add more languages support
+	default:
+		return ff.findFallback(ctx)
+	}
+}
+
+func (ff FileFinder) findFallback(ctx context.Context) (*vcsv1.GetFileResponse, error) {
+	switch filepath.Ext(ff.file.Path) {
+	case ExtGo, ExtAsm:
```
jake-kramer: I think we should add a comment about the `ExtAsm` case being updated when other compiled languages are supported

pkg/frontend/vcs/source/find.go
```diff
@@ -22,52 +26,119 @@ type VCSClient interface {
...
+		Path:  configPath,
+		Ref:   ff.ref,
+	})
+	if err != nil {
+		// Config is optional, so just log and continue
+		level.Debug(ff.logger).Log("msg", "no .pyroscope.yaml found", "path", configPath)
+		return
+	}
+	sp.SetTag("config.url", file.URL)
+	sp.SetTag("config", file.Content)
```
jake-kramer, line 121: Isn't this too large to tag in span?

pkg/frontend/vcs/source/find.go
```diff
@@ -22,52 +26,119 @@ type VCSClient interface {
...
+
+	cfg, err := config.ParsePyroscopeConfig([]byte(file.Content))
```
jake-kramer, line 123: nit: I still think it would be useful to load up this config when the frontend loads up and `WARN` any errors just as a useful sanity check. OK to instead rely on config tooling I'm working on.

pkg/frontend/vcs/source/find.go
```diff
@@ -22,52 +26,119 @@ type VCSClient interface {
...
+	ff.config = cfg
+	level.Debug(ff.logger).Log("msg", "loaded .pyroscope.yaml", "url", file.URL, "mappings", len(cfg.SourceCode.Mappings))
+	sp.SetTag("config.source_code.mappings_count", len(cfg.SourceCode.Mappings))
+
+}
+
 // fetchRepoFile fetches the file content from the configured repository.
 func (arg FileFinder) fetchRepoFile(ctx context.Context, path, ref string) (*vcsv1.GetFileResponse, error) {
+	if arg.rootPath != "" {
+		path = filepath.Join(arg.rootPath, path)
```
jake-kramer, line 138: nit: There's sort of a rough edge here where this root path is defined in the UI at repo level and then the mapping config can contain local paths that are function-level root paths. I can see as a user being confused playing with these two configs.

Maybe at some point a helper tooltip in the UI can clarify how these configs interact.

pkg/frontend/vcs/source/find.go
simonswine: Also needs the token or userid, some users might not have access to the particular repo.

pkg/frontend/vcs/source/find_java.go
```diff
@@ -0,0 +1,60 @@
...
+// findJavaFile finds a java file in a vcs repository.
+func (ff FileFinder) findJavaFile(ctx context.Context, mappings ...*config.MappingConfig) (*vcsv1.GetFileResponse, error) {
+	sp, ctx := opentracing.StartSpanFromContext(ctx, "findJavaFile")
+	defer sp.Finish()
+	sp.SetTag("file.function_name", ff.file.FunctionName)
+	sp.SetTag("file.path", ff.file.Path)
+
+	javaPath := convertJavaFunctionNameToPath(ff.file.FunctionName)
+	for _, m := range mappings {
+		return ff.fetchMappingFile(ctx, m, javaPath)
```
jake-kramer: This doesn't work as expected for multiple mappings. Not sure I see the benefit of passing multiple mappings to these functions, though.

pkg/frontend/vcs/source/find.go
bryanhuhta: Yeah there might be some more details to iron out. I think caching just the config should be fine because eventually we will need to hit the GitHub API to fetch an actual file, which permissions will apply then.

pkg/frontend/vcs/config/config.go
simonswine: Not fully sure, if I am understanding this comment: There are two parts:

Not surfacing the error: I agree nothing surfaces this right now to a user, I think we will need first CLI support like `profilescli pyroscope-yaml validate`, but it should also be possible to surface them in Drilldown, when function details is used.

"if the supplied source is not supported (i.e. not github or local)": This will trigger the error in Line 136. Not too sure what else we could do.

pkg/frontend/vcs/config/config.go
simonswine: That would be a good problem to have and we could find many ways to improve the speed here. 

Right now request to GitHub is at best served in 125ms, so I think it will be a while before this becomes an issue.

pkg/frontend/vcs/source/find.go
simonswine: Don't think this is possible, we need an end user to supply us an encrypted token and the owner/repo/root-path information before we can fetch a config.

pkg/frontend/vcs/source/find.go
simonswine: Not aware of any particular limits, infact this might be the only way for us to see those files, as we don't really have access to them if they are in a private repo.

pkg/frontend/vcs/source/find.go
simonswine: Yes I agree, in order to make the UX workable, we need to show the parsed config in context to what else is configured. I don't think though this will be a good investment of time just yet.

At least of the paths in `.pyroscope.yaml` are relative to that file, and the only location we are reading `.pyroscope.yaml` from is the "rootPath".

pkg/frontend/vcs/source/find_java.go
simonswine: Thanks for pointing this out, in case of a not found error we should use the next mappings like we had done in go already.


Fixed in 604937076fde81ea339d3e244001de3ddb5d454b

pkg/frontend/vcs/config/config.go
simonswine: 1d441c26ac193db5373198034107ebfb13272520

pkg/frontend/vcs/source/find.go
simonswine: See fba41036cc27eda0c2f4b65a5cace6afc3b761e4

pkg/frontend/vcs/config/config.go
jake-kramer: By the "supplied source is not supported" I meant specifically this scenario:

```yaml
mappings:
  - function_name:
    - prefix: org/example/rideshare
  language: java
  source:
     local:
       path: src
  - function_name:
    - prefix: org/example/foo
  language: java
  source:
     gitlab:
       path: src
```

will return the error `"no source type supplied, you need to supply exactly one source type"` which does not let the user know that `gitlab` is an unsupported source

pkg/frontend/vcs/source/find.go
jake-kramer: It seems like this [`max_attribute_bytes` config](https://grafana.com/docs/tempo/latest/configuration/#set-max-attribute-size-to-help-control-out-of-memory-errors) which defaults to 2048 is related?

PR #4610, opened 2025-11-12
fix: Set file name and line number in symbolization

jake-kramer:


simonswine (approved):
LGTM

I think lidia doesn't know yet about file name and line numbers, but once they are there this should work as expected.

PR #4615, opened 2025-11-13
 feat: add exemplar support to timeseries responses for individual profile retrieval

marcsanmi:
This PR implements backend support for exemplars in timeseries queries.

### Protobuf API

  - Add Exemplar message with profile_id, timestamp, value, and labels fields
  - Extend Point message with exemplars field to embed exemplar metadata in timeseries responses
  - Add ExemplarType enum to control exemplar inclusion (NONE, INDIVIDUAL)

 ### Query Pipeline

  - Thread profileID through v2 query backend via `ProfileEntry.ID`
  - Read profile UUID from Parquet ID column during timeseries queries
  - Use `withFetchProfileIDs(true)` option to enable profile ID fetching only when exemplars are requested
  - Maintain v1 compatibility by passing empty `profileID` in legacy query paths

### Exemplar Management

  - ExemplarBuilder: New dedicated component for per-series exemplar tracking
    - Deduplicates exemplars by fingerprint (label set hash)
    - Merges exemplars with same (profileID, timestamp) by intersecting labels
    - Handles dynamic labels (e.g., pod changes during migration)
  - Early label filtering: Filter out groupBy labels at ingestion time using efficient `WithoutLabels()` method
  - Three-level label intersection:
    a. ExemplarBuilder: Within single block
    b. TimeSeriesMerger: Across blocks at same timestamp
    c. RangeSeries: Across timestamps within time bucket

###  Merging & Aggregation

  - Single-block path: Exemplars attached to points with matching timestamps
  - Multi-block path: `TimeSeriesMerger` merges exemplars across blocks, preserving highest values
  - `sumTimeSeriesAggregator` and `avgTimeSeriesAggregator` preserve and merge exemplars during re-aggregation
  - Top-N selection per point (default: 1, configurable via maxExemplarsPerPoint)

 ### Performance Optimizations

  - Efficient sorted-list attachment algorithm (O(E+P) instead of O(E*P))
  - Conditional ID column fetching only when exemplars requested
  - In-place label filtering using Subtract() method
  - Early filtering eliminates redundant aggregator-level filtering

  ### Implementation Flow

  1. v2 single-block: queryTimeSeries → TimeSeriesBuilder.Add() → filter labels → ExemplarBuilder.Add() → BuildWithExemplars()
  2. v2 multi-block: TimeSeriesBuilder → TimeSeriesMerger.MergeTimeSeries() → RangeSeries() re-aggregates with exemplars
  3. v1 paths: Empty profileID for backward compatibility (no exemplars)

### DIFF
--- a/pkg/frontend/readpath/queryfrontend/query_frontend.go
+++ b/pkg/frontend/readpath/queryfrontend/query_frontend.go
@@ -120,7 +120,7 @@ func (q *QueryFrontend) Query(
 		if shouldSymbolize && originalQuery.QueryType == queryv1.QueryType_QUERY_TREE {
 			modifiedQueries[i].QueryType = queryv1.QueryType_QUERY_PPROF
 			modifiedQueries[i].Pprof = &queryv1.PprofQuery{
-				MaxNodes: 0,
+				MaxNodes: originalQuery.Tree.GetMaxNodes(),
 			}
 			modifiedQueries[i].Tree = nil
 		}
@@ -299,7 +299,7 @@ func (q *QueryFrontend) processAndSymbolizeProfiles(
 
 		// Convert back to tree if originally a tree
 		if i < len(originalQueries) && originalQueries[i].QueryType == queryv1.QueryType_QUERY_TREE {
-			treeBytes, err := model.TreeFromBackendProfile(&prof, originalQueries[i].Tree.MaxNodes)
+			treeBytes, err := model.TreeFromBackendProfile(&prof, originalQueries[i].Tree.GetMaxNodes())
 			if err != nil {
 				return fmt.Errorf("failed to build tree: %w", err)
 			}
--- a/pkg/pyroscope/modules_experimental.go
+++ b/pkg/pyroscope/modules_experimental.go
@@ -369,6 +369,7 @@ func (f *Pyroscope) initQueryBackendClient() (services.Service, error) {
 	c, err := querybackendclient.New(
 		f.Cfg.QueryBackend.Address,
 		f.Cfg.QueryBackend.GRPCClientConfig,
+		f.Cfg.QueryBackend.ClientTimeout,
 	)
 	if err != nil {
 		return nil, err
--- a/pkg/querybackend/backend.go
+++ b/pkg/querybackend/backend.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"time"
 
 	"github.com/go-kit/log"
 	"github.com/grafana/dskit/grpcclient"
@@ -20,10 +21,12 @@ import (
 type Config struct {
 	Address          string            `yaml:"address"`
 	GRPCClientConfig grpcclient.Config `yaml:"grpc_client_config" doc:"description=Configures the gRPC client used to communicate between the query-frontends and the query-schedulers."`
+	ClientTimeout    time.Duration     `yaml:"client_timeout"`
 }
 
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.Address, "query-backend.address", "localhost:9095", "")
+	f.DurationVar(&cfg.ClientTimeout, "query-backend.client-timeout", 30*time.Second, "Timeout for query-backend client requests.")
 	cfg.GRPCClientConfig.RegisterFlagsWithPrefix("query-backend.grpc-client-config", f)
 }
 
--- a/pkg/querybackend/client/client.go
+++ b/pkg/querybackend/client/client.go
@@ -2,6 +2,8 @@ package querybackendclient
 
 import (
 	"context"
+	"fmt"
+	"time"
 
 	"github.com/grafana/dskit/grpcclient"
 	"github.com/grafana/dskit/services"
@@ -15,8 +17,8 @@ type Client struct {
 	grpcClient queryv1.QueryBackendServiceClient
 }
 
-func New(address string, grpcClientConfig grpcclient.Config, dialOpts ...grpc.DialOption) (*Client, error) {
-	conn, err := dial(address, grpcClientConfig, dialOpts...)
+func New(address string, grpcClientConfig grpcclient.Config, timeout time.Duration, dialOpts ...grpc.DialOption) (*Client, error) {
+	conn, err := dial(address, grpcClientConfig, timeout, dialOpts...)
 	if err != nil {
 		return nil, err
 	}
@@ -26,14 +28,15 @@ func New(address string, grpcClientConfig grpcclient.Config, dialOpts ...grpc.Di
 	return &c, nil
 }
 
-func dial(address string, grpcClientConfig grpcclient.Config, dialOpts ...grpc.DialOption) (*grpc.ClientConn, error) {
+func dial(address string, grpcClientConfig grpcclient.Config, timeout time.Duration, dialOpts ...grpc.DialOption) (*grpc.ClientConn, error) {
 	options, err := grpcClientConfig.DialOption(nil, nil, nil)
 	if err != nil {
 		return nil, err
 	}
 	// TODO: https://github.com/grpc/grpc-proto/blob/master/grpc/service_config/service_config.proto
+	serviceConfig := fmt.Sprintf(grpcServiceConfigTemplate, timeout.Seconds())
 	options = append(options,
-		grpc.WithDefaultServiceConfig(grpcServiceConfig),
+		grpc.WithDefaultServiceConfig(serviceConfig),
 		grpc.WithMaxCallAttempts(500),
 	)
 	options = append(options, dialOpts...)
@@ -48,7 +51,7 @@ func (b *Client) Invoke(ctx context.Context, req *queryv1.InvokeRequest) (*query
 	return b.grpcClient.Invoke(ctx, req)
 }
 
-const grpcServiceConfig = `{
+const grpcServiceConfigTemplate = `{
     "loadBalancingPolicy":"round_robin",
     "methodConfig": [{
         "name": [{"service": ""}],
@@ -63,6 +66,6 @@ const grpcServiceConfig = `{
               "RESOURCE_EXHAUSTED"
             ]
         },
-        "timeout": "20s"
+        "timeout": "%.0fs"
     }]
 }`
--- a/pkg/querybackend/client/client_test.go
+++ b/pkg/querybackend/client/client_test.go
@@ -94,7 +94,7 @@ func Test_Concurrency(t *testing.T) {
 	resolver.Register(&multiResolverBuilder{targets: addresses})
 	backendAddress := "multi:///"
 
-	cl, err := New(backendAddress, grpcClientCfg, dialOpt)
+	cl, err := New(backendAddress, grpcClientCfg, 30*time.Second, dialOpt)
 	require.NoError(t, err)
 
 	for i := 0; i < nServers; i++ {


pkg/querybackend/query_profile_entry.go
```diff
@@ -39,22 +41,33 @@ func profileEntryIterator(q *queryContext, groupBy ...string) (iter.Iterator[Pro
...
+	buf := make([][]parquet.Value, 4)
 	entries := iter.NewAsyncBatchIterator[*parquetquery.IteratorResult, ProfileEntry](
 		results, bigBatchSize,
 		func(r *parquetquery.IteratorResult) ProfileEntry {
 			buf = r.Columns(buf,
 				schemav1.SeriesIndexColumnName,
 				schemav1.TimeNanosColumnName,
-				schemav1.StacktracePartitionColumnName)
+				schemav1.StacktracePartitionColumnName,
+				schemav1.IDColumnName)
```
simonswine: This unnecessarily fetches IDs for `query_pprof` and `query_tree`. I would prefer if we could move this logic to `query_time_series` only.

Once we want to implement the profileIDSelector, it needs to happen here, but I don't think we wan to fetch the full column every time.

pkg/model/time_series_builder.go
```diff
@@ -55,6 +60,7 @@ func (s *TimeSeriesBuilder) Add(fp model.Fingerprint, lbs Labels, ts int64, valu
 					},
 				},
 			}
+			s.trackExemplar(labelsByString, ts, profileID, value, lbs)
```
simonswine: I think those labels in `lbs` are incomplete, they will only contain the "groupBy" labels. So if groupBy is empty, this label set is empty, don't think this is what we want want from exemplars, they should contain the whole label set attached to them.

pkg/model/time_series_builder.go
```diff
@@ -64,10 +70,106 @@ func (s *TimeSeriesBuilder) Add(fp model.Fingerprint, lbs Labels, ts int64, valu
 		Value:       value,
 		Annotations: pAnnotations,
 	})
+	s.trackExemplar(labelsByString, ts, profileID, value, lbs)
+}
+
+// trackExemplar tracks a profile as a potential exemplar for a specific Point.
+// Keeps the top-N highest-value profiles per point (N = maxExemplarsPerPoint).
+func (s *TimeSeriesBuilder) trackExemplar(seriesKey string, ts int64, profileID string, value float64, fullLabels Labels) {
```
simonswine: One profileID might be split into multiple rows (and onto different shards), so in order to know that some particular profile ids are the "smallest" you need to aggregate those first before you can drop them.

As part of the time series builder, you likely will need to keep everything and only select the ones to keep as part of the aggregator.

pkg/model/time_series_builder.go
```diff
@@ -10,12 +10,19 @@ import (
...
 type TimeSeriesBuilder struct {
 	labelsByFingerprint map[model.Fingerprint]string
 	labelBuf            []byte
 	by                  []string
 
 	series seriesByLabels
+
+	exemplarCandidates    map[string]map[int64][]exemplarCandidate
+	maxExemplarCandidates int
 }
```
simonswine, line 21: The current implementation is not handling labels correctly

The labelset received by `.Add`, is cut down to the groupBy labels here: https://github.com/grafana/pyroscope/blob/091abe53785f171c8fabe92a4c2b82a7ecf00ec0/pkg/model/time_series_builder.go#L52

Also exemplars with the same profileID and timestamp, must be merged. But still their label set might be different because of dynamic labels, so we also need to intersect their labels to only keep matching labels.

In general with this PR `TimeSeriesBuilder` is becoming very complex, so I would say let's refactor this and address a shortcomings issues on the way:

How about per series we have an exempalrBuilder, roughly with this look:

```
type ExemplarBuilder struct {
  fingerPint map[fingerPrint]labelSetsIndex // finger print of the original full series to the current labelSet used
  labelSets []model.Labels

  exemplars []struct {
     timestamp 
     profileID string
     seriesRef # pointer into 
     value
  }  // this should be ordered by timestamp and then profileID
}
```

The Add then should check for fingerprint, if it exists use the label set right away and the do a sorted append into the exemplars slice.

pkg/querybackend/query_time_series.go
```diff
@@ -63,30 +68,43 @@ func queryTimeSeries(q *queryContext, query *queryv1.Query) (r *queryv1.Report,
...
 		}
+		var profileID string
 		for _, e := range row.Values {
 			if e[0].Column() == annotationKeysColumn.ColumnIndex && e[0].Kind() == parquet.ByteArray {
 				annotations.Keys = append(annotations.Keys, e[0].String())
 			}
 			if e[0].Column() == annotationValuesColumn.ColumnIndex && e[0].Kind() == parquet.ByteArray {
 				annotations.Values = append(annotations.Values, e[0].String())
 			}
+			if e[0].Column() == idColumn.ColumnIndex && e[0].Kind() == parquet.ByteArray {
```
simonswine: This doesn't work for me, needs: `parquet.UUID().Type().Kind()`

pkg/querybackend/query_time_series.go
```diff
@@ -53,6 +57,7 @@ func queryTimeSeries(q *queryContext, query *queryv1.Query) (r *queryv1.Report,
 		column.ColumnIndex,
 		annotationKeysColumn.ColumnIndex,
 		annotationValuesColumn.ColumnIndex,
+		idColumn.ColumnIndex,
```
simonswine: We should only do this when necessary aka, when exemplar type is set to `individual`

pkg/querybackend/query_time_series.go
```diff
@@ -63,30 +68,43 @@ func queryTimeSeries(q *queryContext, query *queryv1.Query) (r *queryv1.Report,
...
 			float64(row.Values[0][0].Int64()),
 			annotations,
+			profileID,
 		)
 	}
 	if err = rows.Err(); err != nil {
 		return nil, err
 	}
 
+	fullLabelsByFingerprint, err := getFullLabelsForExemplars(q, builder)
```
simonswine: I would prefer to incorporate that into the `profileEntryIterator`

pkg/querybackend/query_time_series.go
```diff
@@ -125,16 +143,88 @@ func (a *timeSeriesAggregator) build() *queryv1.Report {
...
+	}
+	defer func() {
+		_ = postings.Close()
+	}()
+
+	chunks := make([]index.ChunkMeta, 1)
+	lbs := make(phlaremodel.Labels, 0, 10)
+
+	for postings.Next() {
+		fp, err := q.ds.Index().SeriesBy(postings.At(), &lbs, &chunks)
```
simonswine: I think this should always give you the empty label set, you need Series instead;
```suggestion
		fp, err := q.ds.Index().Series(postings.At(), &lbs, &chunks)
```

simonswine (commented):
Just focused on the query-backend for now

pkg/querybackend/query_time_series.go
marcsanmi: I'll remove them as we talked. They will come from the `profileEntryIterator`

pkg/model/time_series_builder.go
```diff
@@ -25,59 +26,126 @@ func NewTimeSeriesBuilder(by ...string) *TimeSeriesBuilder {
...
+				{
+					Timestamp:   ts,
+					Value:       value,
+					Annotations: pAnnotations,
 				},
-			}
-			return
+			},
 		}
+	} else {
```
simonswine: Nit: I would find that a lot more readable, when we would get rid of the `else` and always append the point. (and obviously not append the point in `if !exists`.

pkg/model/time_series_builder.go
```diff
@@ -25,59 +26,126 @@ func NewTimeSeriesBuilder(by ...string) *TimeSeriesBuilder {
...
+		for _, ex := range exemplars {
+			exemplarsByTimestamp[ex.Timestamp] = append(exemplarsByTimestamp[ex.Timestamp], ex)
+		}
+
+		for _, point := range ser.Points {
+			if exs, found := exemplarsByTimestamp[point.Timestamp]; found {
+				point.Exemplars = exs
+			}
+		}
+	}
```
simonswine: As exemplars are already time ordered, I think we could get rid of that extra map.

pkg/querybackend/query_time_series.go
```diff
@@ -31,7 +32,7 @@ func init() {
 }
 
 func queryTimeSeries(q *queryContext, query *queryv1.Query) (r *queryv1.Report, err error) {
-	entries, err := profileEntryIterator(q, query.TimeSeries.GroupBy...)
+	entries, err := profileEntryIterator(q)
```
simonswine: Now that #4621 has merged, we should be able to deal with groupBy's again.

pkg/querybackend/query_time_series.go
```diff
@@ -64,30 +77,45 @@ func queryTimeSeries(q *queryContext, query *queryv1.Query) (r *queryv1.Report,
...
 				annotations.Keys = append(annotations.Keys, e[0].String())
 			}
 			if e[0].Column() == annotationValuesColumn.ColumnIndex && e[0].Kind() == parquet.ByteArray {
 				annotations.Values = append(annotations.Values, e[0].String())
 			}
+			if includeExemplars && e[0].Column() == idColumn.ColumnIndex && e[0].Kind() == parquet.UUID().Type().Kind() {
+				var u uuid.UUID
+				copy(u[:], e[0].ByteArray())
+				profileID = u.String()
+			}
```
simonswine: This can also now use the profileEntryIterator's parsing #4621

pkg/querybackend/query_time_series.go
```diff
@@ -126,16 +154,40 @@ func (a *timeSeriesAggregator) build() *queryv1.Report {
...
 		},
 	}
 }
+
+// filterLabels filters both series labels and exemplar labels based on groupBy.
+// Series labels are filtered to only include groupBy labels.
+// Exemplar labels are filtered to exclude groupBy labels.
+func (a *timeSeriesAggregator) filterLabels(series []*typesv1.Series, groupBy []string) []*typesv1.Series {
+	for _, s := range series {
+		s.Labels = phlaremodel.Labels(s.Labels).WithLabels(groupBy...)
```
simonswine: Isn't that already done here: https://github.com/grafana/pyroscope/blob/e260b6ccaf2e4bf19301a5c6468c412af4fdb563/pkg/model/time_series_builder.go#L52

pkg/querybackend/query_time_series.go
```diff
@@ -126,16 +154,40 @@ func (a *timeSeriesAggregator) build() *queryv1.Report {
...
+		for _, point := range s.Points {
+			for _, exemplar := range point.Exemplars {
+				exemplar.Labels = phlaremodel.FilterNonGroupedLabels(
```
simonswine: A label builder with `Delete` could be a lot more allocation efficient than this function.

I also think this might make sense to do this in the `timeseriesBuilder.Add` rather than at this stage.

pkg/querybackend/query_time_series.go
```diff
@@ -34,8 +34,9 @@ func queryTimeSeries(q *queryContext, query *queryv1.Query) (r *queryv1.Report,
 	opts := []profileIteratorOption{
 		withFetchPartition(false), // Partition data not needed, as we don't access stacktraces at all
 	}
-	exemplarsEnabled := false // TODO: This will be enabled as part of #4615
-	if exemplarsEnabled {
+
+	includeExemplars := query.TimeSeries.ExemplarType == typesv1.ExemplarType_EXEMPLAR_TYPE_INDIVIDUAL
```
simonswine, line 38: [Can also be done in a follow up PR]

 Let's error here with unimplemented when the type is anything else than `UNKOWN/NONE/INIDIVIDUAL`.

In the same way the query-backend path doesn't error when `profileIDSelector` is not empty.

simonswine (approved):
LGTM

Great work, thanks for going through all my comments

pkg/querybackend/query_time_series.go
marcsanmi: I'll do it in follow up PR

PR #4619, opened 2025-11-17
feat: Add metric to track unsymbolized profile blocks at ingestion

marcsanmi:
Adds a new metric to track whether flushed profile blocks contain unsymbolized profiles.


simonswine:
Merging and backporting this

PR #4621, opened 2025-11-17
chore: Make profileEntryIterator more configurable

simonswine:
As part of #4615, we change quite a bit what data we pull, this is an attempt at refactoring this iterator, so it can be more flexibly allow us to select columns. This refactors profileEntryIterator to use functional options pattern, allowing selective column fetching and more flexible query configuration.

  Changes

  - Introduced options-based API for profileEntryIterator and getSeries functions
  - Added configurable column selection with priority-based join ordering
  - Implemented profile ID filtering and retrieval capabilities
  - Added IDColumnName constant to schema v1
  - Optimized queryTimeSeries to skip fetching partition data when not needed

### DIFF
--- a/pkg/querybackend/query_time_series.go
+++ b/pkg/querybackend/query_time_series.go
@@ -33,7 +33,7 @@ func queryTimeSeries(q *queryContext, query *queryv1.Query) (r *queryv1.Report, 
 	opts := []profileIteratorOption{
 		withFetchPartition(false), // Partition data not needed, as we don't access stacktraces at all
 	}
-	exemplarsEnabled := true
+	exemplarsEnabled := false // TODO: This will be enabled as part of #4615
 	if exemplarsEnabled {
 		opts = append(opts,
 			withAllLabels(),


pkg/querybackend/query_profile_entry.go
```diff
@@ -23,39 +24,142 @@ type ProfileEntry struct {
...
-				schemav1.StacktracePartitionColumnName)
+			buf = r.Columns(buf, columns...)
 			x := series[buf[0][0].Uint32()]
-			return ProfileEntry{
+			e := ProfileEntry{
 				RowNum:      r.RowNumber[0],
 				Timestamp:   model.TimeFromUnixNano(buf[1][0].Int64()),
 				Fingerprint: x.fingerprint,
 				Labels:      x.labels,
 				Partition:   buf[2][0].Uint64(),
```
marcsanmi: I think this can fail now. Maybe we cant set it in the processor? This conditionally depends on `opts.fetchPartition`

pkg/querybackend/query_profile_entry.go
```diff
@@ -23,39 +24,142 @@ type ProfileEntry struct {
...
+			b := buf[offset][0].Bytes()
+			if len(b) != 16 {
+				return
+			}
+			copy(u[:], b)
+			e.ID = u.String()
+		})
+	}
+
 	buf := make([][]parquet.Value, 3)
```
marcsanmi: Nit: Can we use columns as size?

simonswine: Yes totally!

pkg/querybackend/query_profile_entry.go
simonswine: Yes that was an oversight

PR #4625, opened 2025-11-19
Introduce additional metastore compaction metrics

aleks-p:
Adds a few more metrics that should help with understanding compaction job planning, scheduling and processing. These can be used for improved observability, as well as for autoscaling of the compaction worker workload.

### `metastore_compaction_global_queue_blocks_current`

A gauge giving us the total number of blocks across all queues, per compaction level. This is a low cardinality equivalent of `pyroscope_metastore_compaction_block_queue_blocks`, without the shard and tenant labels.

---

### `metastore_compaction_global_queue_queues_current`

A gauge giving us the total number of compaction queues, per compaction level.

---

### `metastore_compaction_global_queue_batches_current`

A gauge giving us the number of jobs that could be created from all queues, per compaction level. This is an indicator of compaction work that is ready to be picked up but is not materialized into jobs.

### DIFF
--- a/pkg/metastore/compaction/compactor/metrics.go
+++ b/pkg/metastore/compaction/compactor/metrics.go
@@ -2,7 +2,6 @@ package compactor
 
 import (
 	"strconv"
-	"sync"
 	"sync/atomic"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -70,23 +69,20 @@ func (b *queueStatsCollector) Collect(m chan<- prometheus.Metric) {
 }
 
 type globalQueueStats struct {
-	blocksPerLevel       []atomic.Int32
-	queuesPerLevel       []atomic.Int32
-	blocksPerLevelAndKey []sync.Map
+	blocksPerLevel  []atomic.Int32
+	queuesPerLevel  []atomic.Int32
+	batchesPerLevel []atomic.Int32
 }
 
 func newGlobalQueueStats(numLevels int) *globalQueueStats {
 	return &globalQueueStats{
-		blocksPerLevel:       make([]atomic.Int32, numLevels),
-		queuesPerLevel:       make([]atomic.Int32, numLevels),
-		blocksPerLevelAndKey: make([]sync.Map, numLevels),
+		blocksPerLevel:  make([]atomic.Int32, numLevels),
+		queuesPerLevel:  make([]atomic.Int32, numLevels),
+		batchesPerLevel: make([]atomic.Int32, numLevels),
 	}
 }
 
 func (g *globalQueueStats) AddBlocks(key compactionKey, delta int32) {
-	b, _ := g.blocksPerLevelAndKey[key.level].LoadOrStore(key, &atomic.Int32{})
-	counter := b.(*atomic.Int32)
-	counter.Add(delta)
 	g.blocksPerLevel[key.level].Add(delta)
 }
 
@@ -94,12 +90,16 @@ func (g *globalQueueStats) AddQueues(key compactionKey, delta int32) {
 	g.queuesPerLevel[key.level].Add(delta)
 }
 
+func (g *globalQueueStats) AddBatches(key compactionKey, delta int32) {
+	g.batchesPerLevel[key.level].Add(delta)
+}
+
 type globalQueueStatsCollector struct {
 	compactionQueue *compactionQueue
 
-	blocks      *prometheus.Desc
-	queues      *prometheus.Desc
-	backlogJobs *prometheus.Desc
+	blocks  *prometheus.Desc
+	queues  *prometheus.Desc
+	batches *prometheus.Desc
 }
 
 const globalQueueMetricsPrefix = "compaction_global_queue_"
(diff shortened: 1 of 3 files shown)


pkg/metastore/compaction/compactor/metrics.go
```diff
@@ -66,3 +66,79 @@ func (b *queueStatsCollector) Collect(m chan<- prometheus.Metric) {
...
+		if levelIdx < len(c.compactionQueue.levels) && c.compactionQueue.levels[levelIdx] != nil {
+			queue := c.compactionQueue.levels[levelIdx]
+
+			maxBlocks := queue.config.maxBlocks(uint32(levelIdx))
+			if maxBlocks == 0 {
+				// This is likely a misconfiguration, we'll just skip it.
+				continue
+			}
+
+			for _, staged := range queue.staged {
```
kolesnikovae: There are race conditions: access to `c.compactionQueue.levels` and ranging sub-queues via `queue.staged`.

I recommend counting items as they are added and removed: add a global counterpart of the `queueStats` struct and update it together with the fine-grained stat counters (4 places).

To track the number of queues per level, we want to catch the moment when `staged` objects are created and deleted (2 places).

aleks-p: Thanks @kolesnikovae, I've reworked this now.

pkg/metastore/compaction/compactor/metrics.go
```diff
@@ -66,3 +68,97 @@ func (b *queueStatsCollector) Collect(m chan<- prometheus.Metric) {
...
+				}
+				return true
+			})
+		}
+
+		ch <- prometheus.MustNewConstMetric(c.blocks, prometheus.GaugeValue, float64(blocksAtLevel), levelLabel)
+		ch <- prometheus.MustNewConstMetric(c.queues, prometheus.GaugeValue, float64(queuesAtLevel), levelLabel)
+		ch <- prometheus.MustNewConstMetric(c.backlogJobs, prometheus.GaugeValue, float64(backlogJobs), levelLabel)
+	}
+}
```
kolesnikovae, line 151: I'm curious whether the metrics need any dimensions.

From our discussion, I understood that we want to add some metrics _without any labels_ apart from those added at the host/pod level – something very specific to our _internal_ requirements. In that case, we indeed need to add metrics that would aggregate the numbers across dimensions.

Otherwise, if there's no such goal, we already have all the numbers we're collecting here, and the new metrics seem redundant.

---

`compaction_global_queue_queues_current` (how many TSL queues exist in the system):
```
count by (level) (pyroscope_metastore_compaction_block_queue_blocks)
count by (level) (pyroscope_metastore_compaction_block_queue_batches)
```

`compaction_global_queue_backlog_blocks_current`:
```
sum by (level) (pyroscope_metastore_compaction_block_queue_blocks)
```

`compaction_global_queue_backlog_jobs_current`:
```
sum by (level) (pyroscope_metastore_compaction_block_queue_batches)
```

pkg/metastore/compaction/compactor/metrics.go
```diff
@@ -66,3 +68,97 @@ func (b *queueStatsCollector) Collect(m chan<- prometheus.Metric) {
...
+			c.compactionQueue.globalStats.blocksPerLevelAndKey[levelIdx].Range(func(k, v interface{}) bool {
+				blocks := v.(*atomic.Int32).Load()
+				if blocks == 0 {
+					c.compactionQueue.globalStats.blocksPerLevelAndKey[levelIdx].Delete(k)
+				} else {
+					backlogJobs += int(blocks) / int(maxBlocks)
+				}
+				return true
+			})
+		}
```
kolesnikovae: I must be missing something but I don't fully understand why we'd need the map here.

* Would `blocksAtLevel/maxBlocks` give us the sought for value?
* Should we count batches (that translate to jobs) instead of estimating it here?

pkg/metastore/compaction/compactor/metrics.go
aleks-p: > we want to add some metrics without any labels apart from those added at the host/pod level

I think a level label can be useful and after checking it doesn't cause any issues in the way the metric would be collected and used. The existing metrics you suggested have tenant and shard labels which increase the cardinality quite a bit, which is the main reason this PR exists.

I wasn't aware of the meaning of the `_batches` metric. It seems to be similar to `compaction_global_queue_backlog_jobs_current`, but with the shard and tenant labels, while also taking the max age into account, making it more accurate. Unfortunately we still can't use it, due to the high cardinality labels, but I can make the equivalent metric without those labels.

pkg/metastore/compaction/compactor/metrics.go
aleks-p: > Would blocksAtLevel/maxBlocks give us the sought for value?

Not really (unless I am misunderstanding), we can have thousands of blocks at a given level, but 0 jobs because blocks belong to different sets of tenants and shards.

> Should we count batches (that translate to jobs) instead of estimating it here?

Yes, I'll give this a try. I wasn't aware of the semantic meaning of batches but I am now (thank you for that). What I'd be after is batches per level and that should be fairly easy to collect, without the map.

kolesnikovae: Thank you, Aleks. Just to make it clear – I do believe the new low-cardinality metrics are needed and the PR is valuable

PR #4629, opened 2025-11-20
fix: limit max nodes in symbolization queries

marcsanmi:
Use validated `Tree.MaxNodes` directly when converting TREE to PPROF queries for symbolization. This should prevent unbounded resource usage during symbolization while still allowing processing of large profiles.

Also increased query backend client timeout from 20s to 30s.

### DIFF
--- a/examples/tracing/ruby/Gemfile.lock
+++ b/examples/tracing/ruby/Gemfile.lock
@@ -74,7 +74,7 @@ GEM
     pyroscope-otel (0.1.4)
       opentelemetry-api (~> 1.1)
       pyroscope (>= 0.5.1, < 1.0)
-    rack (3.1.16)
+    rack (3.1.18)
     rack-protection (4.1.1)
       base64 (>= 0.1.0)
       logger (>= 1.6.0)


pkg/frontend/readpath/queryfrontend/query_frontend.go
```diff
@@ -118,9 +119,14 @@ func (q *QueryFrontend) Query(
...
 		if shouldSymbolize && originalQuery.QueryType == queryv1.QueryType_QUERY_TREE {
+			maxNodes := int64(validation.SmallestPositiveNonZeroIntPerTenant(tenants, q.limits.MaxFlameGraphNodesMax))
+			if maxNodes == 0 {
+				maxNodes = int64(validation.SmallestPositiveNonZeroIntPerTenant(tenants, q.limits.MaxFlameGraphNodesDefault))
+			}
+
 			modifiedQueries[i].QueryType = queryv1.QueryType_QUERY_PPROF
 			modifiedQueries[i].Pprof = &queryv1.PprofQuery{
-				MaxNodes: 0,
+				MaxNodes: maxNodes,
```
simonswine: Shouldn't we take the `MaxNodes` from `queryv1.TreeQuery`?

pkg/querybackend/client/client.go
```diff
@@ -63,6 +63,6 @@ const grpcServiceConfig = `{
               "RESOURCE_EXHAUSTED"
             ]
         },
-        "timeout": "20s"
```
simonswine, line 66: I am not quite sure what other timeouts on our whole stack are, but I agree 30s than 20s is better, maybe we could make this a CLI flag?

pkg/frontend/readpath/queryfrontend/query_frontend.go
marcsanmi: You're right. Also validation isn't needed here since it's already done in `ValidateMaxNodes`.

marcsanmi: Done in f33cb0929a337d4744d37abd09f5512e0897eca6

pkg/frontend/readpath/queryfrontend/query_frontend.go
```diff
@@ -120,7 +120,7 @@ func (q *QueryFrontend) Query(
 		if shouldSymbolize && originalQuery.QueryType == queryv1.QueryType_QUERY_TREE {
 			modifiedQueries[i].QueryType = queryv1.QueryType_QUERY_PPROF
 			modifiedQueries[i].Pprof = &queryv1.PprofQuery{
-				MaxNodes: 0,
+				MaxNodes: originalQuery.Tree.MaxNodes,
```
simonswine: nit: Maybe a nil check of `originalQuery.Tree` might be sensible

simonswine (approved):
LGTM

marcsanmi:
Can you quickly (re)review @simonswine? I've just used `GetMaxNodes` that already does the nil type check internally.

PR #4637, opened 2025-11-25
fix: preserve original timestamps in exemplars during aggregation

marcsanmi:
Closes https://github.com/grafana/pyroscope-squad/issues/650

 
Exemplar timestamps were being overwritten with step-aligned timestamps during aggregation, losing the original profile capture time needed for accurate drill-down.

Screenshot 2025-11-25 at 12 28 09


simonswine (approved):
LGTM! Thanks for fixing this

PR #4640, opened 2025-11-25
feat: Implement `java-jar-mapper` tool

jake-kramer:
This tool is used to modify a `.pyroscope.yaml` file with source code mappings for 3rd party libraries from a given jar.

It uses heuristics and 3rd party APIs to try to resolve discovered 3rd party library functions in to git repos.

There is also a hard coded mapping in `jar-mappings.json` for any popular 3rd party libraries that the tool's heuristics are unable to resolve.

### DIFF
--- a/tools/source-code-mapping-generators/java-jar-mapper/jar.go
+++ b/tools/source-code-mapping-generators/java-jar-mapper/jar.go
@@ -619,7 +619,7 @@ func (e *JARExtractor) extractSpringBootJARs(files []*zip.File) ([]string, strin
 	if len(jarFiles) == 0 {
 		// This shouldn't happen if isSpringBootFatJAR returned true,
 		// but handle it gracefully
-		return nil, "", nil, fmt.Errorf("Spring Boot fat JAR contains no nested JARs in BOOT-INF/lib/")
+		return nil, "", nil, fmt.Errorf("spring boot fat JAR contains no nested JARs in BOOT-INF/lib/")
 	}
 
 	sort.Strings(jarFiles)
--- a/tools/source-code-mapping-generators/java-jar-mapper/services.go
+++ b/tools/source-code-mapping-generators/java-jar-mapper/services.go
@@ -173,7 +173,8 @@ func (s *GitHubTagService) FindTagForVersion(owner, repo, version string) (strin
 		resp, err := makeRequest(refURL)
 		if err == nil {
 			defer resp.Body.Close()
-			if resp.StatusCode == http.StatusOK {
+			switch resp.StatusCode {
+			case http.StatusOK:
 				var refData struct {
 					Ref string `json:"ref"`
 				}
@@ -182,7 +183,7 @@ func (s *GitHubTagService) FindTagForVersion(owner, repo, version string) (strin
 						return strings.TrimPrefix(refData.Ref, "refs/tags/"), nil
 					}
 				}
-			} else if resp.StatusCode == http.StatusForbidden {
+			case http.StatusForbidden:
 				return "", fmt.Errorf("GitHub API rate limited. Set GITHUB_TOKEN environment variable")
 			}
 		}


tools/source-code-mapping-generators/java-jar-mapper/jar.go
```diff
@@ -0,0 +1,667 @@
...
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strings"
+)
+
+type CommandRunner interface {
```
korniltsev-grafanista, line 15: what is the purpose of this interface?

tools/source-code-mapping-generators/java-jar-mapper/services.go
```diff
@@ -0,0 +1,270 @@
...
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+	"time"
+)
+
+type HTTPClient interface {
```
korniltsev-grafanista, line 14: what is the purpose of this interface? Can't we use http.Client?

tools/source-code-mapping-generators/java-jar-mapper/services.go
```diff
@@ -0,0 +1,270 @@
...
+		},
+	}
+}
+
+func (c *DefaultHTTPClient) Get(url string) (*http.Response, error) {
+	return c.client.Get(url)
+}
+
+// MavenService handles downloading JAR files from Maven Central.
+type MavenService struct {
```
korniltsev-grafanista, line 35: MavenService is not used?

tools/source-code-mapping-generators/java-jar-mapper/jdk.go
```diff
@@ -0,0 +1,194 @@
...
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/grafana/pyroscope/pkg/frontend/vcs/config"
+)
+
+// extractJDKVersionFromJAR extracts the JDK version from a JAR file by analyzing
+// the class file major version numbers. Returns the major version (e.g., "8", "11", "17").
+func extractJDKVersionFromJAR(jarPath string) (string, error) {
```
korniltsev-grafanista, line 15: The class version in the class file header is not the version of the JDK used to run the program. I am not 100% sure but it's likely the `javac -target <TARGET>` used at compile time. Furthermore libraries are often compiled with low target to allow running on many JDKs, a lot of them target 8 or 11 JDKs. Running this logic against different libraries would produce different results. It should only run against non libraries classes, but even then it's not reliable as app devs can still compile to a different target (they have less reasons than the lib devs, but still)

tools/source-code-mapping-generators/java-jar-mapper/jar-mappings.json
```diff
@@ -0,0 +1,124 @@
+{
```
korniltsev-grafanista, line 1: why is  the main config yaml, and this config is json?

tools/source-code-mapping-generators/java-jar-mapper/jar-mappings.json
```diff
@@ -0,0 +1,124 @@
...
+  "mappings": [
+    {
+      "jar": "tomcat-embed-core",
```
korniltsev-grafanista, line 4: If this field is compared against artifact id, then maybe rename the field to `artifact_id`

tools/source-code-mapping-generators/java-jar-mapper/jar-mappings.json
```diff
@@ -0,0 +1,124 @@
...
+      "path": "spring-boot-project/spring-boot/src/main/java"
+    },
+    {
+      "jar": "spring-boot-autoconfigure",
+      "owner": "spring-projects",
+      "repo": "spring-boot",
+      "path": "spring-boot-project/spring-boot-autoconfigure/src/main/java"
+    },
+    {
+      "jar": "agent",
```
korniltsev-grafanista, line 118: does this mean any artifact with the name `agent` will resolve to `pyroscope-java` ? this does not seem right

tools/source-code-mapping-generators/java-jar-mapper/main.go
```diff
@@ -0,0 +1,86 @@
...
+	"flag"
+	"fmt"
+	"os"
+
+	"github.com/grafana/pyroscope/pkg/frontend/vcs/config"
+)
+
+func main() {
+	var (
+		jarPath    = flag.String("jar", "", "Path to the Java JAR file to analyze")
```
korniltsev-grafanista, line 13: Am I expected to call this tool on every dependency manually? This is not very convenient. We should at least allow passing multiple jars or a directory(ies?) of jars.

tools/source-code-mapping-generators/java-jar-mapper/jar.go
```diff
@@ -0,0 +1,667 @@
...
+	}
+	defer mainJAR.Close()
+
+	// Check if this is a Spring Boot fat JAR
+	if isSpringBootFatJAR(mainJAR.File) {
+		return e.extractSpringBootJARs(mainJAR.File)
+	}
+
+	// Check for Bazel runfiles directory (e.g., ProjectRunner.runfiles for ProjectRunner.jar)
+	bazelJARs := e.findBazelRunfileJARs(jarPath)
```
korniltsev-grafanista, line 518: does the customer we build this for have these files and does it work for them?

PR #4642, opened 2025-11-26
feat: Add new `source-code coverage` option to profilecli

jake-kramer:
This option has a few modes:
1. `list-functions`: will list all functions found in a provided profile
2. `function`: will determine if a provided profile and provided .pyroscope.yaml config file will resolve to a source code file
3. full coverage analysis (default): will detail which functions in a provided profile resolve to a source code file given a .pyroscope.yaml

A Github token is required if your .pyroscope.yaml contains github source mappings

### DIFF
--- a/cmd/profilecli/source_code_coverage.go
+++ b/cmd/profilecli/source_code_coverage.go
@@ -76,11 +76,11 @@ func addSourceCodeCoverageParams(cmd commander) *sourceCodeCoverageParams {
 	params := new(sourceCodeCoverageParams)
 	cmd.Flag("profile", "Path to pprof profile file").Required().StringVar(&params.ProfilePath)
 	cmd.Flag("config", "Path to .pyroscope.yaml file").StringVar(&params.ConfigPath)
-	cmd.Flag("github-token", "GitHub token for API access (or use GITHUB_TOKEN env var)").StringVar(&params.GithubToken)
 	cmd.Flag("output", "Output format: text or detailed").Default("text").StringVar(&params.OutputFormat)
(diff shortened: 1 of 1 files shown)


cmd/profilecli/main.go
```diff
@@ -105,6 +105,9 @@ func main() {
 	readyCmd := app.Command("ready", "Check Pyroscope health.")
 	readyParams := addReadyParams(readyCmd)
 
+	sourceCodeCoverageCmd := app.Command("source-code-coverage", "Measure the coverage of .pyroscope.yaml source code mappings for translating function names/paths from a pprof profile to VCS source files.")
```
simonswine: It might be more extensible to use a hierachical `profilecli source-code coverage`, as we can e.g. also add `profilecli source-code validate`

cmd/profilecli/source_code_coverage.go
```diff
@@ -0,0 +1,560 @@
...
+	}
+
+	return cfg, configData, profile, nil
+}
+
+func setupVCSClient(ctx context.Context, configData []byte, githubToken string) (source.VCSClient, *http.Client, error) {
+	fmt.Fprintf(os.Stderr, "Setting up GitHub client...\n")
+	tokenStr := githubToken
+	if tokenStr == "" {
+		tokenStr = os.Getenv("GITHUB_TOKEN")
```
simonswine: I would prefer to use the build in way of overriding envVars:

https://github.com/grafana/pyroscope/blob/14097c26c2fd3ba3f9f7cda77608e6d56eef378e/cmd/profilecli/client.go#L123

PR #4646, opened 2025-11-27
perf: Merge duplicate annotations

simonswine:
As part of a time series merge this will remove duplicate annotations.

If we think this information is valuable, we could add a field how often we saw the particular annotations being repeated, but I don't think that data is that valuable.


pkg/model/time_series_merger.go
```diff
@@ -134,14 +135,49 @@ func (m *TimeSeriesMerger) mergePoints(points []*typesv1.Point) int {
 			points[j].Value += points[i].Value
 			// Duplicate annotations are semantically correct and provide useful information.
 			// Users of the data can decide whether to discard or make use of duplicates.
```
marcsanmi: Nit: This comment no longer applies

PR #4647, opened 2025-11-27
chore: Update java images from openjdk to sapmachine

simonswine:
openjdk imags have been retired: https://github.com/docker-library/openjdk


marcsanmi:
Out of curiosity. Did you pick `sapmachine` over proposed other alternatives for any particular reason?

simonswine:
> Out of curiosity. Did you pick `sapmachine` over proposed other alternatives for any particular reason?

I looked for a ubuntu/debian one (so that I don't have to change `apt-get`s into `dnf install`s  and it was the first one I found.

PR #4657, opened 2025-12-01
feat: Add profile_id_selector for individual profile retrieval

marcsanmi:
Enable querying specific profiles by ID in `SelectMergeStacktraces`, `SelectMergeProfile`, and `Diff` API endpoints for the v2 architecture.

This builds on the exemplar support added in #4615, which exposes `profileIDs` in timeseries responses. With profile IDs now available in the timeline, users need a way to retrieve those specific profiles.

### DIFF
--- a/api/querier/v1/querier.proto
+++ b/api/querier/v1/querier.proto
@@ -199,6 +199,8 @@ message SelectMergeProfileRequest {
   optional int64 max_nodes = 5;
   // Select stack traces that match the provided selector.
   optional types.v1.StackTraceSelector stack_trace_selector = 6;
+  // List of Profile UUIDs to query
+  repeated string profile_id_selector = 7 [(gnostic.openapi.v3.property).example = {yaml: "['7c9e6679-7425-40de-944b-e07fc1f90ae7']"}];
 }
 
 message SelectSeriesRequest {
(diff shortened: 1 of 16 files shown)


api/querier/v1/querier.proto
```diff
@@ -199,6 +199,8 @@ message SelectMergeProfileRequest {
   optional int64 max_nodes = 5;
   // Select stack traces that match the provided selector.
   optional types.v1.StackTraceSelector stack_trace_selector = 6;
+  // List of Profile UUIDs to query
+  repeated string profile_id_selector = 7 [(gnostic.openapi.v3.property).example = {yaml: "['7c9e6679-7425-40de-944b-e07fc1f90ae7']"}];
```
alsoba13, line 203: I'm pretty sure this is the way to go, the opportunity cost of doing a list over a single id is just free. But do we have a pros/cons list on why we want this to be a list of ids? (having this is just cool, but I struggle to imagine a use case where user filters+aggregates by multiples)

alsoba13 (approved):

api/querier/v1/querier.proto
simonswine: For me this is also about consistency with `repeated string span_selector = 2;`

The only real use case I can see might be, you have a script spitting out profile ids and  you want to query them together.

marcsanmi: Exactly that, even without a relevant use case for now...
> For me this is also about consistency with repeated string span_selector = 2;

PR #4662, opened 2025-12-03
fix: Return modified labels after validation sanitization

jake-kramer:
This fixes a bug where the logic in `handleSanitizedLabel` was not properly executed. The code would modify the slice itself, but this new modified slice was never returned up the call stack.

### DIFF
--- a/pkg/validation/validate_test.go
+++ b/pkg/validation/validate_test.go
@@ -163,10 +163,12 @@ func TestValidateLabels_SanitizedLabelsReturned(t *testing.T) {
 			name: "single dotted label is sanitized",
 			inputLabels: []*typesv1.LabelPair{
 				{Name: model.MetricNameLabel, Value: "cpu"},
-				{Name: "service.name", Value: "my-svc"},
+				{Name: "service_name", Value: "my-svc"},
+				{Name: "label.dot", Value: "val"},
 			},
 			expectedLabels: []*typesv1.LabelPair{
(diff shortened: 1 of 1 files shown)


marcsanmi (approved):
AFAIU, the modified length wasn't being propagated back to the callers, was it?

LGTM

marcsanmi:
cc @simonswine  you might want to take a look as well.

jake-kramer:
> AFAIU, the modified length wasn't being propagated back to the callers, was it?
> 
> LGTM

Yes exactly, here: https://github.com/grafana/pyroscope/blob/2e303da3c36bd6f670a781d0454e7f249230ecee/pkg/validation/validate.go#L207-L228

jake-kramer:
It will soon be just validation since the sanitization logic will be removed with the utf8 label name work

pkg/model/pprofsplit/pprof_split.go
```diff
@@ -85,10 +88,11 @@ func VisitSampleSeries(
 	for _, idx := range groupsKept.order {
 		for _, group := range groupsKept.m[idx] {
 			if len(group.sampleGroup.Samples) > 0 {
-				if err := visitor.ValidateLabels(group.labels); err != nil {
+				validatedLabels, err := visitor.ValidateLabels(group.labels)
```
simonswine, line 91: This should be:
```
var err error
group.labels, err = visitor.ValidateLabels(group.labels)
[...]
```

PR #4666, opened 2025-12-05
Update golang version to 1.24.11

korniltsev-grafanista:


korniltsev-grafanista:
I've used the script. I probably did not commit/reverted it by mistake

PR #4679, opened 2025-12-09
chore: Run update-contributors workflow once a month

jake-kramer:
Current schedule is too often: the first week of every month and every Sunday. This modification is just for the 1st of the month at 00:00.


korniltsev-grafanista (approved):

korniltsev-grafanista:
Another option could be once after each release

jake-kramer:
@korniltsev-grafanista I like that too. If this is still too noisy, let's plan on that.

PR #4683, opened 2025-12-10
chore: update rideshare examples

aleks-p:
This updates the rideshare example apps that we use to produce internal images we use for demo purposes. The changes aim to address security vulnerabilities, once merged we can run `make rideshare/docker/push` to produce the actual images.

This is a stop-gap solution, in the future we could consider expanding `make tools/update_examples` to do similar upgrades (in addition to what it currently does).


jake-kramer:
Can you verify `Test Examples` still works? Unfortunately it's broken in `main` too 😞

simonswine:
Here is the test run: https://github.com/grafana/pyroscope/actions/runs/20128634602/job/57764240900

The python errors are because of a broken release pipeline to python, I am trying to fix this: 

https://github.com/grafana/pyroscope-rs/pull/269

aleks-p:
> Can you verify `Test Examples` still works? Unfortunately it's broken in `main` too 😞

I only updated the examples that we use to produce images and from what I can see those passed in the run @simonswine mentioned (I downgraded the broken dependency to a working one in the flask example). I will do a follow up at a future time to update the rest of the examples as well.

PR #4684, opened 2025-12-11
fix: Assign appropriate HTTP status codes to errors

bryanhuhta:
We were incorrectly getting SLO read budget burns because some of our validation was being incorrectly returned as 5xx when they should have been 4xx. The read budget burns were specifically impacting our `/render` endpoint. I went through and cleaned up as many of the errors as I could, assigning them appropriate status codes.

This also removes some usages of `github.com/gogo/status` in favor of connect error codes (ultimately everything will be resolved to an HTTP status code).

### DIFF
--- a/pkg/querier/http.go
+++ b/pkg/querier/http.go
@@ -10,13 +10,11 @@ import (
 	"strings"
 
 	"connectrpc.com/connect"
-	"github.com/gogo/status"
 	"github.com/google/pprof/profile"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/labels"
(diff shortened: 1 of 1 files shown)


korniltsev-grafanista (approved):
it would be nice to have a test, so that we don't accidentally turn them into 5xx again later

bryanhuhta:
I have https://github.com/grafana/pyroscope/pull/4685 open to add more thorough tests for all public read path endpoints. I'm going to merge this PR as-is so the changes here don't get lost in the much larger set of changes in https://github.com/grafana/pyroscope/pull/4685.

PR #4686, opened 2025-12-12
feat: Add trace spans for exemplars and profile ID selectors

marcsanmi:


simonswine (approved):
LGTM, thanks for adding.

Btw: Tried a bit to test this locally, but I can't seem to get traces working right now.

PR #4691, opened 2025-12-15
lidia: dynsym symbols

korniltsev-grafanista:
this allows extractin at least some symbols on stripped shared libraries

### DIFF
--- a/lidia/lidia.go
+++ b/lidia/lidia.go
@@ -162,7 +162,7 @@ func CreateLidiaFromELF(elfFile *elf.File, output io.WriteSeeker, opts ...Option
 		o(&rc.opt)
 	}
 	var (
-		symErr, dynSymErr, err error
+		symErr, dynSymErr error
 	)
 	symbols, symErr := elfFile.Symbols()
 	if symErr != nil {
@@ -190,7 +190,7 @@ func CreateLidiaFromELF(elfFile *elf.File, output io.WriteSeeker, opts ...Option
 
 	rb.sort()
 
-	err = rc.write(output)
+	err := rc.write(output)
 	if err != nil {
 		return fmt.Errorf("failed to write lidia file: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


lidia/lidia.go
```diff
@@ -161,13 +161,21 @@ func CreateLidiaFromELF(elfFile *elf.File, output io.WriteSeeker, opts ...Option
 	for _, o := range opts {
 		o(&rc.opt)
 	}
-
-	symbols, err := elfFile.Symbols()
-	if err != nil {
-		return fmt.Errorf("failed to read symbols from ELF file: %w", err)
+	var (
+		symErr, dynSymErr, err error
```
marcsanmi: Maybe we can remove `err` and just use `err := rc.write(output)` below.

lidia/lidia.go
```diff
@@ -161,13 +161,21 @@ func CreateLidiaFromELF(elfFile *elf.File, output io.WriteSeeker, opts ...Option
...
+	symbols, symErr := elfFile.Symbols()
+	if symErr != nil {
+		symbols, dynSymErr = elfFile.DynamicSymbols()
+		if dynSymErr != nil {
+			return fmt.Errorf("failed to read symbols from ELF file: %w, %w", symErr, dynSymErr)
+		}
 	}
 
 	for _, symbol := range symbols {
+		if elf.ST_TYPE(symbol.Info) != elf.STT_FUNC || symbol.Name == "" {
```
marcsanmi, line 176: Since you correctly added this, the test started failing since it relied on object-type symbols, not function ones. I did a bit of "readelfing" and this should make them pass for the `TestCreateReadLookup` test:
```go
{
	name:           "Known function address",
	addr:           0x3c85d0,
	expectFunction: "github.com/prometheus/client_model/go.init",
	expectFound:    true,
},
{
	name:           "Known function address",
	addr:           0x3c8b70,
	expectFunction: "github.com/prometheus/client_model/go.MetricType.Enum",
	expectFound:    true,
},
```

PR #4699, opened 2025-12-17
fix panic in pprof split when using `relabel.LabelDrop`

aleks-p:
Fixes a rare issue where having ingest relabelling rules configured to drop labels can cause panics in segment writers. The root cause is a slice aliasing bug where multiple `lazyGroups` share the same underlying sample array. When one group appends samples, it can corrupt other groups leading to incorrect data in resulting split profiles, and occasionally raise panics.

The issue surfaced after #4365 got merged. Before #4365, samples were never merged into existing groups because of another problem. Solving that exposed a new and more subtle issue that we are solving here.

I wasn't able to write a test that captures the panic behavior. I see no regression in tests or the benchmark added in #4365.

### DIFF
--- a/pkg/model/pprofsplit/pprof_split.go
+++ b/pkg/model/pprofsplit/pprof_split.go
@@ -153,8 +153,7 @@ type lazyGroup struct {
 func (g *lazyGroup) addSampleGroup(stringTable []string, sg pprof.SampleGroup) {
 	if len(g.sampleGroup.Samples) == 0 {
 		g.sampleGroup.Labels = sg.Labels
-		// defensive copy, we don't want to modify the original slice
-		g.sampleGroup.Samples = append([]*profilev1.Sample(nil), sg.Samples...)
+		g.sampleGroup.Samples = append(g.sampleGroup.Samples, sg.Samples...)
 		return
 	}
 
--- a/pkg/model/pprofsplit/pprof_split_test.go
+++ b/pkg/model/pprofsplit/pprof_split_test.go
@@ -485,6 +485,51 @@ func Test_VisitSampleSeries(t *testing.T) {
 				},
 			},
 		},
+		{
+			description: "merging groups after label drop should preserve all samples",
+			rules: []*relabel.Config{
+				{
+					Action: relabel.LabelDrop,
+					Regex:  relabel.MustNewRegexp("^drop_me$"),
+				},
+			},
+			labels: []*typesv1.LabelPair{},
+			profile: &profilev1.Profile{
+				StringTable: []string{"", "do_not_drop_me", "drop_me", "a"},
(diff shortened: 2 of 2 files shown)


pkg/model/pprofsplit/pprof_split.go
```diff
@@ -152,7 +152,9 @@ type lazyGroup struct {
 
 func (g *lazyGroup) addSampleGroup(stringTable []string, sg pprof.SampleGroup) {
 	if len(g.sampleGroup.Samples) == 0 {
-		g.sampleGroup = sg
+		g.sampleGroup.Labels = sg.Labels
```
jake-kramer, line 155: Still new to Go, but can't `append` also modify the underlying slice?

https://github.com/grafana/pyroscope/blob/f455e251d5b0db6791c2455dc96eeba9424a49ba/pkg/model/pprofsplit/pprof_split.go#L172

marcsanmi: Yes, but it should be safe now since we're creating a defensive copy; it has its own private backing array.

pkg/model/pprofsplit/pprof_split.go
```diff
@@ -207,8 +209,12 @@ func (g *groupsWithFingerprints) add(stringTable []string, lbls phlaremodel.Labe
...
 	// add the labels to the list
 	g.m[fp] = append(g.m[fp], &lazyGroup{
-		sampleGroup: group,
-		labels:      lbls,
+		sampleGroup: pprof.SampleGroup{
+			Labels: group.Labels,
+			// defensive copy, we don't want to modify the original slice
+			Samples: append([]*profilev1.Sample(nil), group.Samples...),
+		},
+		labels: lbls,
```
simonswine, line 216: Nit: In theory you call `addSampleGroup` here and reuse the line 155

aleks-p:
> Another theory I have is: As pprof.GroupSamplesWithoutLabels tries to order samples so this in place modifying has no impact, but I think we might have the problem, when dropped sample label names modify the order of sample labels.

Yes, this is the core problem. Grouping is built and optimized for the regular scenario where groups are not modified after creation. When using `labeldrop` we change the label structure and therefore have to modify groups and we can't do this in place anymore. 

> Still a bit of unsatisfying that we can't reproduce it (yet).

We can't reproduce the panic, but I think we can still reproduce the group sample corruption, I'll spend a bit of time on this today.

pkg/model/pprofsplit/pprof_split.go
aleks-p: I am thinking of removing the other defensive copy (L157). All lazy groups are created via this line and will therefore have their own backing sample array so we shouldn't need to do this again. I was probably a bit paranoid :)

PR #4702, opened 2025-12-19
chore: Update helm chart and release notes for v1.17

marcsanmi:
cc @alsoba13


docs/sources/release-notes/v1-17.md
```diff
@@ -0,0 +1,45 @@
+---
```
simonswine: Can you keep the docs change separate?

simonswine (approved):
Thanks

PR #4708, opened 2025-12-23
[release/v1.16] Backport go 1.24.11 and x/crypto update

simonswine:
- **Update golang version to 1.24.11**
- **chore: Patch golang.org/x/crypto/ssh**


marcsanmi:
Just seen this, sorry. Should we update the toolchain as well? https://github.com/grafana/pyroscope/blob/6555bf9c9164ff22ace380f9ffb933aa907379c5/go.mod#L5

simonswine:
Toolchain doesn't make a difference to what we release, if you care then let's PR it into main and backport: https://github.com/grafana/pyroscope/blob/main/go.mod

simonswine:
Probably best to include it into: https://github.com/grafana/pyroscope/blob/main/tools/upgrade-go-version.sh

PR #4717, opened 2025-12-30
test: Prevent test flake in pkg/phlaredb/bucketindex

simonswine:
Poll until it conditions succeed


pkg/phlaredb/bucketindex/loader_test.go
```diff
@@ -433,26 +433,26 @@ func TestLoader_ShouldCacheIndexNotFoundOnBackgroundUpdates(t *testing.T) {
 	require.NoError(t, DeleteIndex(ctx, bkt, "user-1", nil))
 
 	// Wait until the next index load attempt occurs.
```
jake-kramer: Nit: this comment can be removed

PR #4719, opened 2025-12-30
Update goreleaser and use it to push weekly builds

simonswine:
This updated goreleaser and make the config more flexible so we can use it directly for pushing weekly images and theoretically building and pushing main images (I don't want to change this now, though).

I am also purposely ignoring those warnings for now, as I would like to avoid increasing the scope of this change further.

```
  • dockers and docker_manifests are being phased out and will eventually be replaced by dockers_v2, check https://goreleaser.com/deprecations#dockers for more info
```

Changes:

- **Push weekly release images through goreleaser**
- **Allow linux only builds to speed up weekly builds**
- **Migrate from builds to ids, as field is deprecated**

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -7,8 +7,8 @@ before:
 env:
   # Strip debug information from the binary by default, weekly builds will have debug information
   - GORELEASER_DEBUG_INFO_FLAGS={{ if and (index .Env "GORELEASER_STRIP_DEBUG_INFO") (eq .Env.GORELEASER_STRIP_DEBUG_INFO "false")  }}{{ else }}-s -w{{ end }}
-  # Use a custom image prefix (registry) or fallback to grafana/
-  - IMAGE_PREFIX={{ or (index .Env "IMAGE_PREFIX") "grafana/" }}
+  # Use a custom image prefix (registry) or fallback to docker.io/grafana/
+  - IMAGE_PREFIX={{ or (index .Env "IMAGE_PREFIX") "docker.io/grafana/" }}
   # Allow a custom image tag to be used
   - IMAGE_TAG={{ or (index .Env "IMAGE_TAG") .Version }}
   # Publish latest image tag
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ GOPRIVATE=github.com/grafana/frostdb
 
 # Boiler plate for building Docker containers.
 # All this must go at top of file I'm afraid.
-IMAGE_PREFIX ?= grafana/
+IMAGE_PREFIX ?= docker.io/grafana/
 
 IMAGE_TAG ?= $(shell ./tools/image-tag)
 GIT_REVISION := $(shell git rev-parse --short HEAD)


Makefile
```diff
@@ -391,7 +391,7 @@ $(BIN)/mockery: Makefile go.mod
 # Note: When updating the goreleaser version also update .github/workflow/release.yml and .git/workflow/weekly-release.yaml
 $(BIN)/goreleaser: Makefile go.mod
 	@mkdir -p $(@D)
-	GOBIN=$(abspath $(@D)) $(GO) install github.com/goreleaser/goreleaser/v2@v2.7.0
+	GOTOOLCHAIN=go1.25.5 GOBIN=$(abspath $(@D)) $(GO) install github.com/goreleaser/goreleaser/v2@v2.13.2
```
jake-kramer, line 394: Should toolchain instead be updated here? So we don't forget about this val down the road.

https://github.com/grafana/pyroscope/blob/ecb1d3499c18933f82a54e47241d9e5879ae3785/go.mod#L5

.goreleaser.yaml
```diff
@@ -33,6 +42,31 @@ builds:
         -X "github.com/grafana/pyroscope/pkg/util/build.Revision={{ .ShortCommit }}"
         -X "github.com/grafana/pyroscope/pkg/util/build.BuildDate={{ .CommitDate }}"
     id: pyroscope
+  - env:
```
jake-kramer, line 45: optional: YAML anchors and aliases are a good tool for reducing duplication and improving readability. Up to you if you think it's worthwhile here.

.goreleaser.yaml
```diff
@@ -167,12 +229,13 @@ release:
     - [grafana/pyroscope](https://hub.docker.com/r/grafana/pyroscope/tags)
 
     ```bash
-      docker pull docker.io/grafana/pyroscope:{{ trimprefix .Tag "v" }}
+      docker pull {{ .Env.IMAGE_PREFIX }}{{ .ProjectName }}:{{ .Env.IMAGE_TAG }}
```
jake-kramer, line 232: Will this work when missing the `docker.io` prefix?

.goreleaser.yaml
simonswine: I thought about this, but after seeing https://github.com/goreleaser/goreleaser-action/issues/397 I decided against it, did not actually test it if it's still affected.

.goreleaser.yaml
simonswine: It will work for the docker command, not necessarily for other implementations like `podman`, I think we might want to switch the default IMAGE_PREFIX to include it, let met try that first

Makefile
simonswine: I think this one should follow along with what go-releaser demands, while our go.mod file should make their own demands to what Pyroscope expects

.goreleaser.yaml
simonswine: Ok that works: https://github.com/grafana/pyroscope/actions/runs/20616364461/job/59209803166

PR #4720, opened 2025-12-31
chore: Publish Github releases immediately during release process

jake-kramer:
Follow up to https://github.com/grafana/pyroscope/pull/4719

Few changes:
- Disable Github draft releases in GoReleaser to solve Homebrew release breaking
- Update `release.yml` to skip Homebrew release when not "latest" release
- Update `weekly-release.yml` to set as not "latest" release
- Updated `RELEASE.md` with new manual patch process changes
- Fix some typos when reading `$GORELEASER_LINUX_ONLY`


jake-kramer:
Not sure the best way to test this outside of a real release + patch release. Any ideas?

simonswine (approved):
LGTM

I do think this is quite hard to test without a proper release.

We also would need to backport this PR and #4719 to the "active" release branches v1.17/v1.16, before we can test it.

I do think we likely might take the CVE scan dependency update as a reason to release patch releases.

PR #4726, opened 2026-01-02
feat: Add support for Python source code integration

jake-kramer:
Enables source code mappings and stdlib lookups for Python

### DIFF
--- a/pkg/frontend/vcs/source/find_python_test.go
+++ b/pkg/frontend/vcs/source/find_python_test.go
@@ -99,9 +99,9 @@ func Test_isPythonStdlibPath(t *testing.T) {
 			expectedOk:      true,
 		},
 		{
-			name:            "Multiple versions in path",
-			path:            "/opt/python3.8/lib/python3.9/email/mime/text.py",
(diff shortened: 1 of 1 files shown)


pkg/frontend/vcs/source/find_python_test.go
```diff
@@ -0,0 +1,111 @@
...
+		{
+			name:            "Ends at python version - with trailing slash only",
+			path:            "/lib/python3.12/",
+			expectedPath:    "",
+			expectedVersion: "",
+			expectedOk:      false,
+		},
+		{
+			name:            "Nested stdlib path",
+			path:            "/opt/python/lib/python3.9/email/mime/text.py",
```
simonswine, line 89: Maybe it is worth taking the last match, as this case is not that unlikely

```suggestion
			path:            "/opt/python3.9/lib/python3.9/email/mime/text.py",
```

jake-kramer: https://github.com/grafana/pyroscope/pull/4726/commits/66fbb5d91912f2acd8b6d7f490f6b0824bc87353 addresses this

PR #4732, opened 2026-01-07
fix: Properly map local file paths for Python source code integration

jake-kramer:
Was missing the analogous logic present in the Go implementation: https://github.com/grafana/pyroscope/blob/eb8902e49bab6c7c2b1b10ac12bc296113c36893/pkg/frontend/vcs/source/find_go.go#L37-L42


simonswine (approved):
LGTM

PR #4735, opened 2026-01-08
chore: log exemplars query parameters

alsoba13:


pkg/util/spanlogger/query_log.go
```diff
@@ -101,6 +101,7 @@ func (l LogSpanParametersWrapper) SelectMergeStacktraces(ctx context.Context, c
 		"profile_type", c.Msg.ProfileTypeID,
 		"format", c.Msg.Format,
 		"max_nodes", c.Msg.GetMaxNodes(),
+		"profile_id_selector", c.Msg.ProfileIdSelector,
```
simonswine: Should join them otherwise you won't see it
```suggestion
		"profile_id_selector", lazyJoin(c.Msg.ProfileIdSelector),
```

PR #4736, opened 2026-01-09
Allow retrieving heatmap from spans/individual profiles

simonswine:
This PR implements a complete heatmap query API for visualizing profile data distribution over time and value dimensions, with support for both individual profiles and span-level data.

  Key Changes:

  - Heatmap Query Infrastructure: Adds new SelectHeatmap endpoint with proto definitions (HeatmapQuery, HeatmapReport, HeatmapSeries), query backend implementation for INDIVIDUAL and SPAN query types, and data aggregation pipeline with builder/merger components for processing parquet data into time/value buckets.
  - Exemplar Support: Each Y-axis bucket includes optional exemplar metadata (profile ID, span ID, timestamp, labels) representing the highest-value profile in that cell, enabling drill-down from aggregated visualizations to individual profiles with little-endian span ID encoding for query consistency.
  - Series Limiting & Validation: Implements TopSeries to return top-N series by total value, with input validation for step, limit, and time range parameters.

### DIFF
--- a/pkg/frontend/frontend_stubs.go
+++ b/pkg/frontend/frontend_stubs.go
@@ -10,7 +10,7 @@ import (
 	querierv1 "github.com/grafana/pyroscope/api/gen/proto/go/querier/v1"
 )
 
-var errNotAvailableInV1Frontend = connect.NewError(connect.CodeUnimplemented, errors.New("This endpoint not available in v1 frontend"))
+var errNotAvailableInV1Frontend = connect.NewError(connect.CodeUnimplemented, errors.New("This endpoint is not available in v1 frontend"))
 
 func (f *Frontend) SelectHeatmap(
 	ctx context.Context,


pkg/frontend/frontend_select_time_series.go
```diff
@@ -78,3 +78,10 @@ func (f *Frontend) SelectSeries(
 	series := m.Top(int(c.Msg.GetLimit()))
 	return connect.NewResponse(&querierv1.SelectSeriesResponse{Series: series}), nil
 }
+
+func (f *Frontend) SelectHeatmap(
+	ctx context.Context,
+	c *connect.Request[querierv1.SelectHeatmapRequest],
+) (*connect.Response[querierv1.SelectHeatmapResponse], error) {
+	return nil, connect.NewError(connect.CodeUnimplemented, nil)
```
aleks-p: should we communicate somehow (reference server API docs) that the API only works in v2?

pkg/model/heatmap/merger.go
```diff
@@ -0,0 +1,251 @@
...
+		} else {
+			remapped[i] = ref
+		}
+	}
+	return remapped
+}
+
+// remapProfileId remaps a profile ID (which is a reference into the attribute table)
+func (m *Merger) remapProfileId(profileId int64, table *queryv1.AttributeTable, refMap map[int64]int64) int64 {
+	if table == nil || refMap == nil {
```
aleks-p: Is `table` needed here and just for the nil check?

pkg/model/heatmap/range.go
```diff
@@ -0,0 +1,344 @@
...
+}
+
+// yBucket represents a Y-axis bucket
+type yBucket struct {
+	min uint64
+	max uint64
+}
+
+// createYAxisBuckets creates evenly spaced Y-axis buckets
+func createYAxisBuckets(minValue, maxValue uint64, numBuckets int) []yBucket {
```
aleks-p: Can we use []int64 for Y buckets like for time buckets or does tracking min and max make some things easier?

pkg/model/heatmap/points_builder.go
```diff
@@ -0,0 +1,105 @@
...
+type exemplar struct {
+	timestamp   int64
+	profileID   string
+	spanID      uint64
+	labelSetRef int
+	value       uint64
+}
+
+// newPointsBuilder creates a new ExemplarBuilder.
+func newPointsBuilder() *pointsBuilder {
```
aleks-p: Nit, was this renamed from exemplar builder? Maybe we should rename receivers and any outdated comments.

aleks-p (approved):
Looks good to me!

pkg/model/heatmap/merger.go
```diff
@@ -0,0 +1,251 @@
...
+	return refMap
+}
+
+// remapRefs remaps a slice of attribute refs using the provided mapping
+func (m *Merger) remapRefs(refs []int64, refMap map[int64]int64) []int64 {
+	if len(refs) == 0 || refMap == nil {
+		return refs
+	}
+
+	remapped := make([]int64, len(refs))
```
marcsanmi: Would a sync.Pool or pre-allocated buffers help to avoid creating allocations in the hot path?

pkg/model/heatmap/points_builder.go
```diff
@@ -0,0 +1,105 @@
...
+		e.labelSetRef = labelSetIdx
+		eb.labelSets[e.labelSetRef] = eb.labelSets[e.labelSetRef].Intersect(labels)
+	}
+
+	eb.exemplars = slices.Insert(eb.exemplars, pos, e)
+}
+
+func cmpExemplar(a, b exemplar) int {
+	if r := a.timestamp - b.timestamp; r != 0 {
+		return int(r)
```
marcsanmi: Could integer overflow cause incorrect sorting? e.g
```go
 func cmpExemplar(a, b exemplar) int {
        if a.timestamp < b.timestamp {
                return -1
        }
        if a.timestamp > b.timestamp {
                return 1
        }
        if a.spanID < b.spanID {
                return -1
        }
        if a.spanID > b.spanID {
                return 1
        }
        return strings.Compare(a.profileID, b.profileID)
  }
```

pkg/model/heatmap/range.go
```diff
@@ -0,0 +1,344 @@
...
+	var buckets []int64
+	for t := start; t <= end; t += step {
+		buckets = append(buckets, t)
+	}
+	return buckets
+}
+
+// findTimeBucket finds the index of the time bucket for a given timestamp
+func findTimeBucket(timestamp int64, buckets []int64) int {
+	for i, bucket := range buckets {
```
marcsanmi: Would a binary search be faster for the already sorted buckets?

pkg/model/heatmap/heatmap.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/grafana/pyroscope/pkg/model"
+)
+
+type attributeKey struct {
+	key   unique.Handle[string]
+	value unique.Handle[string]
+}
+
+// This attribute table is used to store the attribute values for the heatmap. It might be reused for other queries at a later time.
+type attributeTable struct {
```
marcsanmi, line 20: I created a specific attributetable model in my [PR](https://github.com/grafana/pyroscope/pull/4756/). If it makes sense, we might want to reuse it once merged.

simonswine: Yes let's do that whatever merges first

pkg/model/heatmap/merger.go
simonswine: Not required, form an older iteration, refactored this a fair bit.

pkg/model/heatmap/merger.go
simonswine: I had not profiled this bit at all, it could likely benefit from it, also it should be safe to chage it in place. (not planning to optimize this further, but it made sense in this place)

pkg/frontend/frontend_select_time_series.go
simonswine: Good point added it to the error and the docs

pkg/model/heatmap/points_builder.go
simonswine: Done

pkg/model/heatmap/points_builder.go
simonswine: Yes it could: on negative timestamps or on 32bit architectures

pkg/model/heatmap/range.go
simonswine: The mistake was made in the Exemplar proto API, not too sure this should either be int64 or straight a double: int64 is used in parquet and that is using it because of pprof. Our API Repsonse is using doubles (so is SelectSeries.

I would argue those values should be an int64. Will provide a PR in main after this merges

simonswine: Will provide a PR in main first, before fixing it here

pkg/model/heatmap/range.go
simonswine: Depends on the amount of points, it should really be calcuated

PR #4749, opened 2026-01-12
fix: pass `$IMAGE_TAG` directly to Goreleaser

jake-kramer:
Goreleaser tried to infer `$IMAGE_TAG` from `.Version` which isn't present/set when the `env` block is being evaluated. Instead, require this env var to be set by callers explicitly.


aleks-p (approved):
LGTM

PR #4756, opened 2026-01-15
feat: implement AttributeTable with string interning for exemplar labels

marcsanmi:
## Summary

Implements string interning for exemplar labels using AttributeTable in the internal query API, reducing network bandwidth when exemplars are requested. 

The user-facing API remains unchanged - AttributeTable is only used for internal communication between query backends and the frontend.


api/types/v1/types.proto
```diff
@@ -159,10 +159,6 @@ message Exemplar {
...
   uint64 value = 4 [(gnostic.openapi.v3.property).example = {yaml: "2450000000"}];
 
-  // Series labels that are NOT included in the group_by query parameter.
-  // These labels complete the full series identity of this exemplar's profile.
-  // For example, if group_by=["service"], this would contain labels like "pod",
-  // "namespace", "region" that were omitted from grouping. This allows identifying
-  // which specific series instance this profile sample came from.
-  repeated LabelPair labels = 5;
+  // Labels as references into AttributeTable
+  repeated int64 attribute_refs = 5;
```
simonswine: I don't think we should change the user facing API. I think the only the report the query backends use between each other should be using this interning and they should also use it for series labels, profile and span ids.

For this we should define a separate Exemplar type in `query/v1`, which is used for the TimeSeries report.

I want to avoid the extra complexity for API clients for now and really only handle the retrieving of points, exemplars that way.

If we decide to do this on the client facing API, it should be opt-in and we probably would have to do it for all endpoints returning attributes.

api/query/v1/query.proto
```diff
@@ -157,6 +161,31 @@ message SeriesLabelsReport {
   repeated types.v1.Labels series_labels = 2;
 }
 
+message Exemplar {
+  int64 timestamp = 1;
+  string profile_id = 2;
+  string span_id = 3;
+  uint64 value = 4;
```
simonswine: Like stated in https://github.com/grafana/pyroscope/pull/4779 this should be int64 instead.

api/query/v1/query.proto
```diff
@@ -157,6 +161,31 @@ message SeriesLabelsReport {
...
+
+message Point {
+  double value = 1;
+  int64 timestamp = 2;
+  repeated types.v1.ProfileAnnotation annotations = 3;
+  repeated Exemplar exemplars = 4;
+}
+
+message Series {
+  repeated types.v1.LabelPair labels = 1;
```
simonswine: To align better with Heatmap, might make sense to also intern those?

pkg/model/attribute_table.go
```diff
@@ -0,0 +1,103 @@
...
+	typesv1 "github.com/grafana/pyroscope/api/gen/proto/go/types/v1"
+)
+
+type attributeKey struct {
+	key   unique.Handle[string]
+	value unique.Handle[string]
+}
+
+// Table stores unique label keys and values for efficient serialization.
+type Table struct {
```
simonswine: I think the `AttributeTable` is something fairly central, so it should go into `pkg/model` instead

simonswine (commented):
Looks good, do you mind rebasing that after my heatmap PR and with https://github.com/grafana/pyroscope/pull/4779

api/types/v1/types.proto
```diff
@@ -159,11 +159,8 @@ message Exemplar {
...
   int64 value = 4 [(gnostic.openapi.v3.property).example = {yaml: "2450000000"}];
 
-  // Series labels that are NOT included in the group_by query parameter.
-  // These labels complete the full series identity of this exemplar's profile.
-  // For example, if group_by=["service"], this would contain labels like "pod",
-  // "namespace", "region" that were omitted from grouping. This allows identifying
-  // which specific series instance this profile sample came from.
+  // Additional labels specific to this exemplar (e.g., pod name, node name).
+  // These labels are not included in the series group_by labels.
   repeated LabelPair labels = 5;
```
simonswine: This could be explained a bit better:

> Labels specific to this exemplar (e.g., pod name, node name). When an exemplar label overlaps with the series label, it is not included here.

api/query/v1/query.proto
```diff
@@ -229,3 +233,29 @@ message HeatmapReport {
...
+
+message Point {
+  double value = 1;
+  int64 timestamp = 2;
+  repeated types.v1.ProfileAnnotation annotations = 3;
+  repeated Exemplar exemplars = 4;
+}
+
+message Series {
+  repeated int64 attribute_refs = 1;
```
simonswine, line 290: Can you add a comment here

api/query/v1/query.proto
```diff
@@ -229,3 +233,29 @@ message HeatmapReport {
...
+  string profile_id = 2;
+  string span_id = 3;
+  int64 value = 4;
+  repeated int64 attribute_refs = 5;
+}
+
+message Point {
+  double value = 1;
+  int64 timestamp = 2;
+  repeated types.v1.ProfileAnnotation annotations = 3;
```
simonswine: I would love to see annotations to be in the attribute table as well. 

If not in this PR, remove annotations for now, so we don't have two annotation fields

api/query/v1/query.proto
```diff
@@ -229,3 +233,29 @@ message HeatmapReport {
   repeated HeatmapSeries heatmap_series = 2;
   AttributeTable attribute_table = 3;
 }
+
+message Exemplar {
+  int64 timestamp = 1;
```
simonswine, line 270:
```
// Milliseconds unix timestamp
```

api/query/v1/query.proto
```diff
@@ -229,3 +233,29 @@ message HeatmapReport {
...
+  string profile_id = 2;
+  string span_id = 3;
+  int64 value = 4;
+  repeated int64 attribute_refs = 5;
+}
+
+message Point {
+  double value = 1;
+  int64 timestamp = 2;
```
simonswine, line 280:
```
// Milliseconds unix timestamp
```

api/query/v1/query.proto
```diff
@@ -124,6 +126,7 @@ message Report {
   TreeReport tree = 6;
   PprofReport pprof = 7;
   HeatmapReport heatmap = 8;
+  TimeSeriesWithAttributeTableReport time_series_with_attribute_table = 9;
```
simonswine: I have second thoughts on the naming

```suggestion
  TimeSeriesInterned time_series_interned = 9;
```

```suggestion
  TimeSeriesCompact time_series_compact = 9;
```

pkg/frontend/readpath/queryfrontend/query_select_time_series.go
```diff
@@ -46,26 +47,146 @@ func (q *QueryFrontend) SelectSeries(
...
+	if err != nil {
+		return nil, err
+	}
+
+	series = phlaremodel.TopSeries(series, int(c.Msg.GetLimit()))
+	return connect.NewResponse(&querierv1.SelectSeriesResponse{Series: series}), nil
+}
+
+// queryStandard queries using the standard path without exemplar optimization.
+func (q *QueryFrontend) query(
```
simonswine: comment not matching symbol

pkg/model/attribute_table_merger.go
```diff
@@ -0,0 +1,405 @@
...
+	"strconv"
+	"strings"
+	"sync"
+
+	queryv1 "github.com/grafana/pyroscope/api/gen/proto/go/query/v1"
+	typesv1 "github.com/grafana/pyroscope/api/gen/proto/go/types/v1"
+)
+
+// SeriesMerger merges query.v1.Series with AttributeTable optimization.
+type SeriesMerger struct {
```
simonswine: This adds a lot of duplication of https://github.com/grafana/pyroscope/blob/main/pkg/model/heatmap/merger.go. maybe we can solve this using generics?

pkg/querybackend/query_time_series_with_attribute_table.go
```diff
@@ -0,0 +1,240 @@
...
+		newTimeSeriesWithAttributeTableAggregator,
+		true,
+		[]block.Section{
+			block.SectionTSDB,
+			block.SectionProfiles,
+		}...,
+	)
+}
+
+func queryTimeSeriesWithAttributeTable(q *queryContext, query *queryv1.Query) (r *queryv1.Report, err error) {
```
simonswine: Duplicating the whole file will be very difficult to maintain longer term.

marcsanmi: N/A https://github.com/grafana/pyroscope/pull/4793

pkg/model/attribute_table_merger.go
marcsanmi: Added in separate PR https://github.com/grafana/pyroscope/pull/4793

pkg/frontend/readpath/queryfrontend/query_select_time_series.go
marcsanmi: N/A since the file has changed.

pkg/querybackend/query_time_series.go
```diff
@@ -179,11 +242,242 @@ func (a *timeSeriesAggregator) build() *queryv1.Report {
...
+	})
+	a.merger.merge(r)
+	return nil
+}
+
+func (a *timeSeriesCompactAggregator) build() *queryv1.Report {
+	sum := typesv1.TimeSeriesAggregationType_TIME_SERIES_AGGREGATION_TYPE_SUM
+	stepMilli := time.Duration(a.query.GetStep() * float64(time.Second)).Milliseconds()
+
+	typesV1 := a.merger.expand()
```
simonswine: This appears to be very unncessary

pkg/querybackend/query_time_series.go
```diff
@@ -179,11 +242,242 @@ func (a *timeSeriesAggregator) build() *queryv1.Report {
...
+				}
+				if len(p.Exemplars) > 0 {
+					pt.Exemplars = make([]*queryv1.Exemplar, len(p.Exemplars))
+					for i, ex := range p.Exemplars {
+						pt.Exemplars[i] = &queryv1.Exemplar{
+							Timestamp:     ex.Timestamp,
+							ProfileId:     ex.ProfileId,
+							SpanId:        ex.SpanId,
+							Value:         ex.Value,
+							AttributeRefs: remap.Refs(slices.Clone(ex.AttributeRefs)),
```
simonswine: Not too sure why this is cloned?

pkg/querybackend/query_time_series.go
```diff
@@ -179,11 +242,242 @@ func (a *timeSeriesAggregator) build() *queryv1.Report {
...
+	return nil
+}
+
+func (a *timeSeriesCompactAggregator) build() *queryv1.Report {
+	sum := typesv1.TimeSeriesAggregationType_TIME_SERIES_AGGREGATION_TYPE_SUM
+	stepMilli := time.Duration(a.query.GetStep() * float64(time.Second)).Milliseconds()
+
+	typesV1 := a.merger.expand()
+	iter := timeseries.NewTimeSeriesMergeIterator(typesV1)
+	ranged := timeseries.RangeSeries(iter, a.startTime, a.endTime, stepMilli, &sum)
```
simonswine: While this currently works, it feels like quite complex. Can[t we do instead of `expand()` provide a `func (m *timeSeriesMerger) newIterator() iter.Iterator[model.Value]`, which provides the data in the format that RangeSeries expects? (similar to `timeseries.NewTimeSeriesMergeIterator`.

You will also have to define a `newCompactSeriesIterator(.) *compactSeriesIterator` analog to `timeseries.NewSeriesIterator`.

pkg/querybackend/query_time_series.go
```diff
@@ -179,11 +242,242 @@ func (a *timeSeriesAggregator) build() *queryv1.Report {
...
+			}
+		}
+
+		result[i] = &typesv1.Series{Labels: labels, Points: points}
+	}
+
+	return result
+}
+
+func (m *timeSeriesMerger) mergePoints(points []*queryv1.Point) int {
```
simonswine:
```suggestion
func (m *timeSeriesCompactMerger) mergePoints(points []*queryv1.Point) int {
```

simonswine (commented):
Focused first on the querybackend changes, this still feels like why to complex. Maybe I am missing something.

pkg/querybackend/query_time_series.go
marcsanmi: It shouldn't be, updated.

pkg/querybackend/query_time_series.go
marcsanmi: Added in 53860ad318f3c4f19ce75e078f83d0687e398095

marcsanmi:
@simonswine You can omit commit (7455a1e36e607c836d4953ceb4ed4221c908c211) and focus on this one 53860ad318f3c4f19ce75e078f83d0687e398095 since it suppresses previous implementation. I'm adding the iterator and creating a logical folder separation for `timeseriescompact`.

pkg/frontend/readpath/queryfrontend/query_select_time_series.go
```diff
@@ -47,26 +48,120 @@ func (q *QueryFrontend) SelectSeries(
...
+
+func expandQuerySeries(series []*queryv1.Series, table *queryv1.AttributeTable) []*typesv1.Series {
+	if len(series) == 0 {
+		return nil
+	}
+	if table == nil {
+		table = &queryv1.AttributeTable{}
+	}
+
+	labelMap := make(map[int64]*typesv1.LabelPair, len(table.Keys))
```
simonswine: This map is not needed, attribute table slices can be accessed directly.

I also want you share code from the `heatmap` after moving it into `model/attributetable`.

Labels: https://github.com/grafana/pyroscope/blob/main/pkg/model/heatmap/range.go#L192 into 
Exemplars: https://github.com/grafana/pyroscope/blob/1ed705c70b78fc1f85d0ba216517e248cfeb5026/pkg/model/heatmap/range.go#L300

pkg/model/timeseries/merger.go
```diff
@@ -183,7 +183,12 @@ func mergeExemplars(a, b []*typesv1.Exemplar) []*typesv1.Exemplar {
 	for _, group := range byProfileID {
 		ex := group.exemplar
 		if len(group.labelSets) > 1 {
-			ex.Labels = phlaremodel.IntersectAll(group.labelSets)
+			intersected := phlaremodel.IntersectAll(group.labelSets)
```
simonswine: I am a bit confused by than, can you explain this?

I suspect this is the LLM trying to pass a test

pkg/frontend/readpath/queryfrontend/query_select_time_series.go
```diff
@@ -47,26 +48,120 @@ func (q *QueryFrontend) SelectSeries(
...
-		return connect.NewResponse(&querierv1.SelectSeriesResponse{}), nil
+	if report == nil || report.TimeSeries == nil {
+		return nil, nil
 	}
-	series := timeseries.TopSeries(report.TimeSeries.TimeSeries, int(c.Msg.GetLimit()))
-	return connect.NewResponse(&querierv1.SelectSeriesResponse{Series: series}), nil
+	return report.TimeSeries.TimeSeries, nil
+}
+
+func (q *QueryFrontend) queryCompact(ctx context.Context, start, end int64, labelSelector string, req *querierv1.SelectSeriesRequest) ([]*typesv1.Series, error) {
```
simonswine, line 96: Can you comment here and in the `pkg/model/timeseries*`, that currently we only use timeseriescompact for the examplar retrieval and the legacy path for all other time series queries.

Over time we want to move from timeseries to timeseriescompact and get rid of timeseries.

pkg/frontend/readpath/queryfrontend/query_select_time_series.go
```diff
@@ -47,26 +48,120 @@ func (q *QueryFrontend) SelectSeries(
 	if err != nil {
 		return nil, connect.NewError(connect.CodeInvalidArgument, err)
 	}
+
+	var series []*typesv1.Series
+	if c.Msg.GetExemplarType() == typesv1.ExemplarType_EXEMPLAR_TYPE_INDIVIDUAL {
```
simonswine, line 55: Can you adda todo here, that we would like to use queryCompact for everything, once this has been rolled out and remove queryStandard?

pkg/model/attributetable/attributetable.go
```diff
@@ -70,3 +83,21 @@ func (t *Table) Build(res *queryv1.AttributeTable) *queryv1.AttributeTable {
 
 	return res
 }
+
+// ResolveRefs converts attribute references to label pairs.
+func ResolveRefs(refs []int64, table *queryv1.AttributeTable) []*typesv1.LabelPair {
```
simonswine:
```suggestion
func ResolveLabelPairs(refs []int64, table *queryv1.AttributeTable) []*typesv1.LabelPair {
```

pkg/frontend/readpath/queryfrontend/query_select_time_series.go
```diff
@@ -47,26 +49,116 @@ func (q *QueryFrontend) SelectSeries(
...
+			if len(p.AnnotationRefs) > 0 {
+				points[j].Annotations = make([]*typesv1.ProfileAnnotation, 0, len(p.AnnotationRefs))
+				for _, ref := range p.AnnotationRefs {
+					if ref >= 0 && ref < int64(len(table.Keys)) {
+						points[j].Annotations = append(points[j].Annotations, &typesv1.ProfileAnnotation{
+							Key:   table.Keys[ref],
+							Value: table.Values[ref],
+						})
+					}
+				}
```
simonswine: This should be `attributetable.ResovleAnnotations`

pkg/frontend/readpath/queryfrontend/query_select_time_series.go
```diff
@@ -47,26 +49,116 @@ func (q *QueryFrontend) SelectSeries(
...
+				for k, ex := range p.Exemplars {
+					points[j].Exemplars[k] = &typesv1.Exemplar{
+						Timestamp: ex.Timestamp,
+						ProfileId: ex.ProfileId,
+						SpanId:    ex.SpanId,
+						Value:     ex.Value,
+						Labels:    attributetable.ResolveRefs(ex.AttributeRefs, table),
+					}
+				}
+			}
```
simonswine: This should be `attributetable.ResolveExemplars()`.

It also should replace `pointsToExemplar` in here:  https://github.com/grafana/pyroscope/blob/20b549ff8e9fd69ef0e40f9608d0348ca1f0b750/pkg/model/heatmap/range.go#L300

simonswine (approved):
I give you the LGTM, but I ideally want to move a bit more code to attributetable

pkg/frontend/readpath/queryfrontend/query_select_time_series.go
marcsanmi: This involved a bit of refactoring since `groupBy` was implemented in different ways in exemplars and heatmaps. You can check it in this commit 89184cbb9f5d7fe067397ecd48f842d9bab8842e. I've found easier and more readable to have 2 functions reusing `resolveExemplar` instead of creating a callback that fmpov, wouldn't be adding any advantage... lmkwyt

simonswine: Very likely `queryv1.HeatmapPoint` shouldn't exist and just be using `queryv1.Point`

Let's do this in a follow up PR.

simonswine (approved):

PR #4757, opened 2026-01-15
Partially revert function name modification from `pyspy`-based profiles

jake-kramer:
This change reverts the logic from https://github.com/grafana/pyroscope/pull/4730 for `pyspy` profiles that don't have absolute paths in functions. This is to maintain the existing UX for profiles pre-`pyroscope-io v0.8.15` which now passes absolute paths for function names.


simonswine (approved):
LGTM

Thank you for doing that ❤️

PR #4762, opened 2026-01-16
Generate profile IDs from trace metadata

simonswine:
Generate stable profile IDs from caller-supplied identity.

The feature is experimental and disabled by default. Enable it with `-validation.profile-id-deterministic=true`. Per-tenant overrides are supported.

ID selection policy:
- Preserve a valid user-supplied UUID regardless of whether deterministic generation is enabled, including profile_id supplied through OTLP. (today we ignore OTLP's profile_id)
- Invalid IDs are treated as absent and follow the fallback policy.
- When enabled and no user-supplied UUID is available, derive the ID from the original nonzero timestamp, profile type (__name__), and ingress labels.
- If the timestamp is zero or missing, derive the ID from the upstream caller's trace ID, profile type, and ingress labels instead.
- If neither a timestamp nor an upstream trace ID is available, generate   a random UUID v4.
- When disabled and no user-supplied UUID is available, use a random UUID v4; do not derive an ID from a timestamp or trace ID.

Previously this had a per-profile content hashing, which required a remarhsalling for OTLP and would have generated a different hash with a different amount of profiles included in a batch, so I am going with this.

Also exports a metric to see what case of 4 above was triggered.

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -322,6 +322,8 @@ func (d *Distributor) Push(ctx context.Context, grpcReq *connect.Request[pushv1.
 	if err != nil {
 		return nil, connect.NewError(connect.CodeUnauthenticated, err)
 	}
+	traceID, hasTraceID := tracing.ExtractTraceID(ctx)
+	deriveProfileID := d.limits.ProfileIDDeterministic(tenantID) && hasTraceID
 
 	defer func() {
 		if err == nil {
@@ -349,8 +351,11 @@ func (d *Distributor) Push(ctx context.Context, grpcReq *connect.Request[pushv1.
 		RawProfileType: distributormodel.RawProfileTypePPROF,
 	}
 	allErrors := multierror.New()
+	var profilePosition uint64
 	for _, grpcSeries := range grpcReq.Msg.Series {
 		for _, grpcSample := range grpcSeries.Samples {
+			position := profilePosition
+			profilePosition++
 			profile, err := pprof.RawFromBytesWithLimit(grpcSample.RawProfile, maxProfileSizeBytes)
 			if err != nil {
 				// check if decompression size has been exceeded
@@ -377,17 +382,13 @@ func (d *Distributor) Push(ctx context.Context, grpcReq *connect.Request[pushv1.
 				ID:         grpcSample.ID,
 			}
 
-			// Generate deterministic ID if not provided by client
-			if series.ID == "" && d.limits.ProfileIDDeterministic(tenantID) {
-				// Extract trace ID from context
-				traceID, _ := tracing.ExtractTraceID(ctx)
-
-				series.ID = profileid.GenerateFromRequest(
+			if series.ID == "" && deriveProfileID {
+				series.ID = profileid.GenerateFromTrace(
 					tenantID,
-					grpcSeries.Labels,
-					grpcSample.RawProfile,
-					profile.Profile.TimeNanos, // Original TimeNanos before modification
 					traceID,
+					grpcSeries.Labels,
+					profile.Profile.TimeNanos,
+					position,
 				).String()
 			}
 
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -47,6 +47,7 @@ import (
 	"github.com/grafana/pyroscope/v2/pkg/distributor/writepath"
 	phlaremodel "github.com/grafana/pyroscope/v2/pkg/model"
 	"github.com/grafana/pyroscope/v2/pkg/model/pprofsplit"
+	"github.com/grafana/pyroscope/v2/pkg/model/profileid"
 	"github.com/grafana/pyroscope/v2/pkg/model/relabel"
 	"github.com/grafana/pyroscope/v2/pkg/model/sampletype"
 	"github.com/grafana/pyroscope/v2/pkg/pprof"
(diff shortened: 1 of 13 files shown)


pkg/ingester/otlp/ingest_handler.go
```diff
@@ -276,13 +279,46 @@ func (h *ingestHandler) export(ctx context.Context, er *pprofileotlp.ExportProfi
...
+					for _, series := range req.Series {
+						if series.ID == "" {
+							// Marshal profile to get bytes for hashing
+							rawProfile, err := series.Profile.MarshalVT()
+							if err != nil {
+								// Log but continue with random ID
+								level.Warn(h.log).Log("msg", "failed to marshal profile for ID generation", "err", err)
+								series.ID = uuid.NewString()
+								continue
+							}
```
simonswine: This is not great, but I am also aware we do have very limited traffic via this path.

Maybe we could get hold of the hash of the full `*pprofileotlp.ExportProfilesServiceRequest` and append the sub series index or something?

pkg/ingester/pyroscope/ingest_adapter.go
```diff
@@ -196,6 +199,31 @@ func (p *pyroscopeIngesterAdapter) parseToPprof(
 			"orgID", tenantID)
 		return nil
 	}
+
+	// Generate IDs for series before calling PushBatch
+	tenantID, _ := tenant.ExtractTenantIDFromContext(ctx)
```
marcsanmi: Is it needed? It's already done [here](https://github.com/grafana/pyroscope/pull/4762/files#diff-4b425a973a854b3c016ee576530e7f015c20ab82acbcb3c931be1689aef9d343R196)

pkg/ingester/pyroscope/ingest_adapter.go
```diff
@@ -196,6 +199,31 @@ func (p *pyroscopeIngesterAdapter) parseToPprof(
...
+	if p.limits.ProfileIDDeterministic(tenantID) {
+		traceID, _ := tracing.ExtractTraceID(ctx)
+		for _, series := range plainReq.Series {
```
marcsanmi: I think the logic of generating IDs conditionally is repeated in 3 places (distributor, ingest_handler, ingest_adapter). Does it make sense to add a function like this in the generator.go (needs review, didn't check it in deep)?

```go
  func EnsureSeriesIDs(
      series []*distributormodel.Series,
      tenantID string,
      traceID string,
      deterministic bool,
  ) error {
      for _, s := range series {
          if s.ID != "" {
              continue
          }

          if deterministic && s.RawProfile != nil {
              s.ID = GenerateFromRequest(
                  tenantID,
                  s.Labels,
                  s.RawProfile,
                  s.Profile.Profile.TimeNanos,
                  traceID,
              ).String()
          }

          if s.ID == "" {
              s.ID = uuid.NewString()
          }
      }
      return nil
  }
  ```

File: pkg/ingester/otlp/ingest_handler.go (hunk 1 above)
Comment (marcsanmi) [reply pkg/ingester/otlp/ingest_handler.go]: We could accept the cost for now if we want to ship this quickly to unblock other initiatives. However, I'd create a task to address it next quarter since we should expect more traffic once Profiling signal is finally out.

**Review by marcsanmi — DISMISSED** [review]
LGTM!

File: pkg/distributor/distributor.go [hunk 4]
```diff
@@ -813,10 +834,19 @@ func (d *Distributor) sendRequestsToSegmentWriter(ctx context.Context, req *dist
...
 		profileID := uuid.New()
+		if s.ID != "" {
+			var parseErr error
+			profileID, parseErr = uuid.Parse(s.ID)
+			if parseErr != nil {
+				// Log warning but continue with random ID
+				level.Warn(d.logger).Log("msg", "invalid profile ID, using random", "id", s.ID, "err", parseErr)
+				profileID = uuid.New()
+			}
+		}
```
alsoba13: Did you mean this?

```suggestion
		profileID := profileid.GenerateRandom()
		if s.ID != "" {
			var parseErr error
			profileID, parseErr = uuid.Parse(s.ID)
			if parseErr != nil {
				// Log warning but continue with random ID
				level.Warn(d.logger).Log("msg", "invalid profile ID, using random", "id", s.ID, "err", parseErr)
				profileID = profileid.GenerateRandom()
			}
		}
```

otherwise the new introduced function `GenerateRandom()` seems unused and could be removed

simonswine: Removed GenerateRandom and its test in d75eb37be. The ingestion paths retain their existing random UUID fallbacks.

pkg/ingester/pyroscope/ingest_adapter.go
simonswine: Moved tenant extraction before the empty-series branch and reuse it for deterministic ID generation in 6f0316956.

pkg/distributor/distributor.go
```diff
@@ -836,7 +853,10 @@ func (d *Distributor) sendRequestsToIngester(ctx context.Context, req *distribut
 		if _, err = p.WriteTo(bw); err != nil {
 			return nil, err
 		}
-		series.ID = uuid.NewString()
+		// Only generate ID if not already set
+		if series.ID == "" {
+			series.ID = uuid.NewString()
+		}
```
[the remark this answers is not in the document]
simonswine: Fixed in 0d8e416a8. visitSampleSeries now copies the source profile ID to every split series, with an end-to-end assertion that the derived IDs reach the ingester.

pkg/ingester/otlp/ingest_handler.go
simonswine: Implemented a cheaper alternative in 6f0316956: IDs now derive from trace metadata, canonical ingress labels, timestamp, and profile position. This removes per-series pprof marshaling; #5563 is therefore closed as superseded.

alsoba13 (approved):
LGTM

PR #4776, opened 2026-01-21
profilecli: add --force/-f flag to overwrite output files

korniltsev-grafanista:
## Summary
- Add `-f`/`--force` flag to profilecli commands that write pprof output files
- When enabled, allows overwriting existing files instead of failing with "file exists" error
- Flag is disabled by default to preserve the current safe behavior

**Affected commands:**
- `profilecli query profile`
- `profilecli query go-pgo`
- `profilecli admin blocks query profile`

## Test plan
- [ ] Run `profilecli query profile --output pprof=./test.pprof` twice without `--force` - should fail on second run
- [ ] Run `profilecli query profile --output pprof=./test.pprof -f` twice - should succeed both times
- [ ] Verify `--help` shows the new flag for all three commands


marcsanmi (approved):
LGTM

PR #4777, opened 2026-01-21
feat(profilecli): add kube-proxy command for unified microservice access

simonswine:
This adds a reverse proxy command that unifies all Pyroscope microservices in Kubernetes into a single endpoint.

The command:
- Auto-discovers services using label selectors
- Provides path-based routing to correct services
- Uses kubectl proxy on Unix socket for secure K8s API access
- Listens on configurable TCP address (default 127.0.0.1:4242)
- Filters to user-facing services only (distributor, query-frontend,
  tenant-settings, ad-hoc-profiles)

cc @jake-kramer we discussed the best way for debugging a problem, not too sure that is. But it works for me. In theory we could also inject a tenant id using a flag.

### DIFF
--- a/cmd/profilecli/kube_proxy.go
+++ b/cmd/profilecli/kube_proxy.go
@@ -381,11 +381,22 @@ func (m *reverseProxyManager) proxyToService(w http.ResponseWriter, r *http.Requ
 		return
 	}
 
-	// Copy headers
+	// Copy headers, but skip Authorization to avoid interfering with kubectl's auth
 	for key, values := range r.Header {
+		// Skip Authorization header - kubectl proxy uses kubeconfig auth, not incoming auth
+		if key == "Authorization" {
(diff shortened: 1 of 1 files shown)


cmd/profilecli/kube_proxy.go
```diff
@@ -0,0 +1,554 @@
...
+				level.Info(logger).Log("msg", "kubectl proxy is ready")
+				return nil
+			}
+		}
+	}
+}
+
+func (m *reverseProxyManager) initRoutingRules() {
+	// Order matters - most specific first!
+	m.routingRules = []routingRule{
```
jake-kramer, line 280: Is there any straightforward way of programmatically generating this list?

simonswine: No, we only have that list in a private repo, that get's manually updated. (as far as I am aware).

simonswine:
@jake-kramer can you give it another try, is should now work in all namespaces without specifying custom selectors?

PR #4779, opened 2026-01-22
fix: Correct datatype for Exemplar value uint64->int64

simonswine:
pprof and our database parquet schema is using `int64` for values


marcsanmi (approved):
Thanks for taking care of it!

PR #4780, opened 2026-01-22
Add `.pre-commit-config.yaml` for pre-commit support

jake-kramer:
Adds pre-commit hooks for code formatting and code generation. Code generation is only run when necessary (for protobuf-related files).

To use:
- Install pre-commit tool (via Homebrew, pip, etc.)
- Run `pre-commit install` in root of `pyroscope`

Only staged changes will be subject to pre-commit hooks for `git commit`.

To skip pre-commit, use `git commit --no-verify`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,22 @@
+# Pre-commit hooks that delegate to Makefile targets
+# This ensures CI and local development use the same formatting/linting logic
+repos:
+  - repo: local
+    hooks:
+      - id: fmt
+        name: Format code (Go, protobuf, jsonnet)
+        entry: make fmt
+        language: system
+        pass_filenames: false
+        # Run on any file change - make fmt handles file filtering internally
+        always_run: true
+
+      - id: generate
+        name: Check generated protobuf files
+        entry: make generate check/unstaged-changes
+        language: system
+        pass_filenames: false
+        # Only run when protobuf-related files in api/ or pkg/ are staged
+        # This matches: api/**/*.proto, pkg/**/*.proto, api/buf*.yaml, pkg/buf*.yaml
+        # Root-level YAML files (renovate.json, .golangci.yml, etc.) are excluded
+        files: '^(api|pkg)/.*\.(proto|yaml)$'


korniltsev-grafanista:
`make fmt` is too slow for each commit, at least for me and my laptop

jake-kramer:
> `make fmt` is too slow for each commit, at least for me and my laptop

@korniltsev-grafanista I've been playing around with trying to speed it up. Looks like after warming the caches with one run, subsequent runs are a few seconds or less for me. Are you seeing the same? Can also disable with `git commit --no-verify` or uninstall with `pre-commit uninstall`

korniltsev-grafanista:
yeah, no big problem I just disable/uninstall it.

yeah it runs 1.5 seconds for me (and 3 seconds after reboot) and it definetely faster than not doing the check at all and waiting for CI to fail.

```
~/p/pyroscope (main) > time make fmt
git ls-files '*.go' | grep -v 'vendor/' | grep -v 'og/' | xargs gofmt -s -w
/home/korniltsev/p/pyroscope/.tmp/bin/golangci-lint run --fix
0 issues.
cd api/ && /home/korniltsev/p/pyroscope/.tmp/bin/buf format -w .
cd pkg && /home/korniltsev/p/pyroscope/.tmp/bin/buf format -w .
/home/korniltsev/p/pyroscope/.tmp/bin/tk fmt ./operations/pyroscope/jsonnet/
All discovered files are already formatted. No changes were made

________________________________________________________
Executed in    1.59 secs    fish           external
   usr time    4.53 secs  599.00 micros    4.53 secs
   sys time    0.99 secs    0.00 micros    0.99 secs
```

It's not much slower than  `go fmt` 

```
~/p/pyroscope (main) > time go fmt ./...

________________________________________________________
Executed in  956.21 millis    fish           external
   usr time    1.99 secs    581.00 micros    1.99 secs
   sys time    0.44 secs     91.00 micros    0.44 secs

```

I wonder if we can optimize this and only run the precommit hook on changed / commiting files

```
~/p/pyroscope (main) > time /home/korniltsev/p/pyroscope/.tmp/bin/golangci-lint run
 --fix pkg/symbolizer/...
0 issues.

________________________________________________________
Executed in  379.00 millis    fish           external
   usr time  840.00 millis  664.00 micros  839.34 millis
   sys time  225.00 millis   98.00 micros  224.90 millis
```

Or maybe have an option to run `go fmt` instead (obviously will catch less problems with less realiability)

```
~/p/pyroscope (main) > time go fmt ./pkg/symbolizer/...

________________________________________________________
Executed in   59.91 millis    fish           external
   usr time  198.25 millis    0.00 micros  198.25 millis
   sys time   42.55 millis  739.00 micros   41.81 millis
```

I just love to do a lot of small (often meaningless, just to back up the progress) commits. And 1.5 commit time would be a noticable slowdown which .

jake-kramer:
@korniltsev-grafanista This might be what you're looking for: `pre-commit uninstall && pre-commit install --hook-type pre-push`. I'm inclined to make this the default using [`default_install_hook_types`](https://pre-commit.com/#top_level-default_install_hook_types).. thoughts?

PR #4782, opened 2026-01-23
lidia: add gopclntab support for Go symbol resolution

korniltsev-grafanista:
## Summary
- Add gopclntab parsing to the lidia package for Go symbol resolution without relying on symtab
- Integrate lidia package into the pyroscope repository
- Add unit test that validates gopclntab-only symbol resolution using `/proc/self/exe`

## Test plan
- [x] Run `go test -v -run TestGoPclntabSelfExe` in the lidia package
- [x] Verify existing lidia tests still pass
- [ ] Integration testing with pyroscope symbolizer

### DIFF
--- a/lidia/gosym/parse.go
+++ b/lidia/gosym/parse.go
@@ -38,8 +38,6 @@ func ParseRuntimeTextFromPclntab18(pclntab []byte) uint64 {
 }
 
 var errEmptyText = errors.New("empty text")
-var errGoPCLNTabNotFound = errors.New(".gopclntab not found")
-var errGoTooOld = errors.New("go too old")
 
 func GoFunctions(f *elf.File) ([]Func, error) {
 	const headerSize = 64
@@ -50,7 +48,7 @@ func GoFunctions(f *elf.File) ([]Func, error) {
 	}
 	pclntab := f.Section(".gopclntab")
 	if pclntab == nil {
-		return nil, errGoPCLNTabNotFound
+		return nil, nil // not a go binary - return no functions, no error
 	}
 	pclntabData, err := pclntab.Data()
 	if err != nil {
@@ -75,7 +73,7 @@ func GoFunctions(f *elf.File) ([]Func, error) {
 	pcln := NewLineTable(pclntabData, textStart)
 
 	if !pcln.IsGo12() {
-		return nil, errGoTooOld
+		return nil, nil // too old - return no functions, no error
 	}
 	return pcln.Go12Funcs(), nil
 }
--- a/lidia/lidia.go
+++ b/lidia/lidia.go
@@ -196,7 +196,10 @@ func CreateLidiaFromELF(elfFile *elf.File, output io.WriteSeeker, opts ...Option
 		}
 	}
 	if rc.opt.parseGoPclntab {
-		functions, _ := gosym.GoFunctions(elfFile)
+		functions, err := gosym.GoFunctions(elfFile)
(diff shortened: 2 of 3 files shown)


korniltsev-grafanista:
note: it's not very efficient - it reads everything into memory twice, in a memory cache unfriendly  way. We can optimize it later(we had a faster version the ebpf module some time ago).

lidia/lidia.go
```diff
@@ -156,36 +158,53 @@ func CreateLidiaFromELF(elfFile *elf.File, output io.WriteSeeker, opts ...Option
...
+				File:      "",
+				CallFile:  "",
+				CallLine:  0,
+				Depth:     0,
+				LineTable: nil,
+			})
+		}
+	}
+	if rc.opt.parseGoPclntab {
+		functions, _ := gosym.GoFunctions(elfFile)
```
marcsanmi: Do we want to know if it fails?

lidia/gosym/parse.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"debug/elf"
+	"encoding/binary"
+	"errors"
+	"fmt"
+)
+
+func ParseRuntimeTextFromPclntab18(pclntab []byte) uint64 {
+	if len(pclntab) < 64 {
```
marcsanmi, line 11: Maybe adding `const minPclntabHeaderSize = 64` since it's also used below?

lidia/gosym/parse.go
```diff
@@ -0,0 +1,80 @@
...
+
+	textStart := ParseRuntimeTextFromPclntab18(pclntabHeader)
+
+	if textStart == 0 {
+		// for older versions text.Addr is enough
+		// https://github.com/golang/go/commit/b38ab0ac5f78ac03a38052018ff629c03e36b864
+		textStart = text.Addr
+	}
+	if textStart < text.Addr || textStart >= text.Addr+text.Size {
+		return nil, fmt.Errorf(" runtime.text out of .text bounds %d %d %d", textStart, text.Addr, text.Size)
```
marcsanmi:
```suggestion
		return nil, fmt.Errorf("runtime.text out of .text bounds %d %d %d", textStart, text.Addr, text.Size)
```

lidia/gosym/parse.go
```diff
@@ -0,0 +1,80 @@
...
+
+	return 0
+}
+
+var errEmptyText = errors.New("empty text")
+var errGoPCLNTabNotFound = errors.New(".gopclntab not found")
+var errGoTooOld = errors.New("go too old")
+
+func GoFunctions(f *elf.File) ([]Func, error) {
+	obj := f
```
marcsanmi: is it needed? Can we use just `f`?

lidia/options.go
```diff
@@ -31,3 +34,15 @@ func WithFiles() Option {
...
 	}
 }
+
+func WithParseGoPclntab(parse bool) Option {
+	return func(o *options) {
+		o.parseGoPclntab = parse
+	}
+}
+
+func WithSymTab(parse bool) Option {
```
marcsanmi:
```suggestion
func WithSymtab(parse bool) Option {
```

lidia/lidia.go
korniltsev-grafanista: I removed two errors (not a  go binary, too old go binary). Now returned error is treated as failure and returned up to the caller.

PR #4783, opened 2026-01-23
add an extraContainers field to pyroscope helm chart

luweglarz:
Hello everyone,

I would like to propose to add a extraContainers field so we can add sidecars to Pyroscope, it can be useful for authentification proxy for example.


simonswine (approved):
LGTM.

I think this is a valid use-case, btw since kubernetes v1.33, `initContainers` also supports to run as sidecar ([see](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/)).

I do think it might even make more sense to to use that (edit before first coffee: it is exposed  ~is also not exposed currently~).

PR #4790, opened 2026-01-26
Implement `SelectMergeTree` via the model.Tree

simonswine:
We currently use a completely separate query path for SelectMergeStacktraces vs. SelectMergePprof calls. This PR is a first step in unifying the two paths, by implementing a model.Tree based merge.

This refactors model.Tree to be generic over its node name type, supporting both string-based function names (FuntionName) and integer location references (LocationRefName). This enables the query-backend path to build flamegraph trees using location refs directly, deferring string resolution until the final merge step.

Most complexity is in the SymbolMerger in symdb that deduplicates and merges symbols (strings, mappings, functions, locations) across profile blocks using content hashing (xxhash), with linear probing to gracefully handle  hash collisions.

After merging this, the old path is still active by default, there is a per-tenant override, gating the tree-based SelectMergeProfile code path. Also adds test coverage for the query backend, frontend query path, and the symbol merger.

Note: The integration tests after this PR will only target the tree based variant.

---

> [!NOTE]
> **Medium Risk**
> Adds new protobuf fields and a new tree-based query path that reconstructs pprof profiles from tree+symbol tables, which can affect API compatibility and profile correctness. Risk is mitigated by being *tenant-gated/experimental*, but touches core query/frontend/model serialization paths.
> 
> **Overview**
> Adds a *tree+symbols* response shape to the query API: `TreeQuery` gains `full_symbols`, `TreeReport` can now return `symbols`, and a new `TreeSymbols` message carries mappings/locations/functions/string tables (plus hashes) for reconstructing profiles.
> 
> Refactors `pkg/model` trees to be generic over node-name type (function names vs location refs), updates all tree marshal/unmarshal/merge/flamegraph/diff code to use typed trees, and updates callers to pass the appropriate type parameters.
> 
> Updates `SelectMergeProfile` in the query-frontend to optionally use an *experimental, tenant-gated* tree-based backend query (`QueryTreeEnabled` / `querier.query-tree-enabled`), reconstructing a pprof profile from `TreeReport.Tree` + `TreeReport.Symbols` (with special handling for truncated/"other" nodes). Test coverage is expanded significantly around the new gating and reconstruction behavior, while the existing pprof-based path remains the default (and is forced for Go PGO queries).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 582d8623bfc9aad2c6a530bddecd0c0cfae8ca1b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/frontend/readpath/queryfrontend/query_select_merge_profile.go
+++ b/pkg/frontend/readpath/queryfrontend/query_select_merge_profile.go
@@ -159,6 +159,13 @@ func (q *QueryFrontend) selectMergeProfileTree(
 
 	var p profilev1.Profile
 
+	if report.Tree.Symbols == nil {
+		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("symbols not present in tree response"))
+	}
+	if len(report.Tree.Symbols.Mappings) < 1 || len(report.Tree.Symbols.Locations) < 1 || len(report.Tree.Symbols.Functions) < 1 {
+		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("incomplete symbols in tree response"))
+	}
+
 	p.StringTable = report.Tree.Symbols.Strings
 	p.Mapping = report.Tree.Symbols.Mappings[1:]
 	p.Location = report.Tree.Symbols.Locations[1:]
--- a/pkg/model/stacktraces.go
+++ b/pkg/model/stacktraces.go
@@ -314,6 +314,9 @@ func unsafeStringBytes(s string) []byte {
 }
 
 func (t *StacktraceTree) Tree(maxNodes int64, names []FunctionName) *FunctionNameTree {
+	if len(names) == 0 {
+		return new(FunctionNameTree)
+	}
 	lookup := func(i int32) FunctionName {
 		return names[i]
 	}


pkg/phlaredb/symdb/symbol_merger.go
```diff
@@ -0,0 +1,614 @@
+package symdb
```
simonswine, line 1: This is the most complex part of this PR, worth starting to look at it @aleks-p

pkg/phlaredb/symdb/symbol_merger.go
```diff
@@ -0,0 +1,614 @@
...
+
+func (h *hashedSlice[A]) len() int {
+	return len(h.sl)
+}
+
+func (h *hashedSlice[A]) grow(size int) {
+	h.sl = slices.Grow(h.sl, size)
+	h.hashes = slices.Grow(h.hashes, size)
+	// grow map if still empty
+	if len(h.m) == 0 {
```
aleks-p: nit, this is perhaps never hit, we always call `add` when creating new mergers

pkg/phlaredb/symdb/symbol_merger.go
```diff
@@ -0,0 +1,614 @@
...
+		merger: merger,
+		m:      make(map[uint64]int),
+		equal:  equal,
+	}
+}
+
+type equalityFn[A any] func(A, A) bool
+
+type hashedSlice[A any] struct {
+	merger *SymbolMerger
```
aleks-p: nit, it doesn't seem like we use this

pkg/phlaredb/symdb/symbol_merger.go
aleks-p: @simonswine looks good to me overall, I think we can move it forward once we address the few TODOs

simonswine:
I really should have enabled the integration tests a longer time ago, basically the symbolizer isn't able to work correctly with this feature. I will have to spent some more time on this.

simonswine:
After refactoring the symbolizer in #4887, keeping the symbolizer with this was super straightforward.

Other changes include: Refactoring after changes in main and I am dealing with hash collision properly.

pkg/phlaredb/symdb/resolver.go
```diff
@@ -236,13 +237,57 @@ func (r *Resolver) partition(partition uint64) *lazyPartition {
...
+			return err
+		}
+		resolved.FormatNodeNames(cb)
+
+		lock.Lock()
+		tree.Merge(resolved)
+		lock.Unlock()
+		return nil
+	})
+	tree.Total()
```
aleks-p, line 276: nit, this doesn't seem to be needed?

marcsanmi:
LGTM. As @aleks-p said, one minor note — not a blocker, but worth considering: both the [1:] slicing and the removed len(names) == 0 guard rely on implicit initialization invariants that aren't enforced at usage. Safe today, but a future refactor could break them silently. Defensive guards wouldn't make any harm.

marcsanmi:
@cursor push a289586886

PR #4793, opened 2026-01-27
chore: Refactor attribute table into its own package

simonswine:
This was something I have thought about in my code review of #4756. It should allow better code reuse between heatmap and time series.

- **Move attribute table**
- **Refactor merger to be able to be reused**

### DIFF
--- a/pkg/model/heatmap/merger.go
+++ b/pkg/model/heatmap/merger.go
@@ -53,7 +53,7 @@ func (m *Merger) MergeHeatmap(report *queryv1.HeatmapReport) {
 				existing = &atHeatmapSeries{
 					attributeRefs: remappedSeriesRefs,
 				}
-			m.series[key] = existing
+				m.series[key] = existing
 			}
 
 			// Ensure slice is big enough to hold all the points


pkg/model/attributetable/attributetable.go
```diff
@@ -0,0 +1,73 @@
...
+// It maps attribute keys to integer references, allowing compact representation of labels.
+type Table struct {
+	table   map[Key]int64
+	entries []Key
+}
+
+func New() *Table {
+	return &Table{
+		table:   make(map[Key]int64),
+		entries: make([]Key, 0),
```
marcsanmi: nit: I think setting it as `nil` wouldn't make any harm.

pkg/model/attributetable/merger.go
```diff
@@ -0,0 +1,119 @@
...
+
+// MergedSeries represents a merged series with generic point type.
+type MergedSeries[T any] struct {
+	AttributeRefs []int64
+	Points        []T
+}
+
+// Merger provides generic merging functionality for series with attribute tables.
+// T is the point type, K is the series key type.
+type Merger[T any, K comparable] struct {
```
marcsanmi: Since it has an internal mutex but also exposes `Lock()`/`Unlock()` publicly, and some methods lock internally (`IsEmpty()`) while others assume the caller holds the lock would it be worth adding a brief comment explaining this threading contract?

simonswine: You are totally right, refactored this to be clear and no locks exposed

simonswine:
This needed a fair bit of slop removal, can you take another look @marcsanmi 

I've merged #4794 into this and think its probably best if you now take over and implement #4756

marcsanmi (approved):

PR #4795, opened 2026-01-28
feat(profilecli): use SelectMergeStacktraces instead of SelectMergeProfile when --function-names-only

alsoba13:
Generating pprof from the tree at `SelectMergeStacktraces`, instead of using pprof at `SelectMergeProfile`, if `--function-names-only` param is given to `profilecli query profile`.

Tree format may be faster for some queries. Reconstructing pprof from there comes with data loss (best effort)

We lose:
* mappings
* line numbers
* context of in-line functions (we don't know whether these are inline or not)

### DIFF
--- a/cmd/profilecli/main.go
+++ b/cmd/profilecli/main.go
@@ -74,7 +74,7 @@ func main() {
 	queryProfileCmd := queryCmd.Command("profile", "Request merged profile.").Alias("merge")
 	queryProfileOutput := queryProfileCmd.Flag("output", "How to output the result, examples: console, raw, pprof=./my.pprof").Default("console").String()
 	queryProfileForce := queryProfileCmd.Flag("force", "Overwrite the output file if it already exists.").Short('f').Default("false").Bool()
-	queryProfileTree := queryProfileCmd.Flag("tree", "Reconstruct pprof from the native profile tree.").Default("false").Bool()
+	queryProfileFunctionNamesOnly := queryProfileCmd.Flag("function-names-only", "Faster call, without details about mappings, line number, and inlining").Default("false").Bool()
 	queryProfileParams := addQueryProfileParams(queryProfileCmd)
 	queryGoPGOCmd := queryCmd.Command("go-pgo", "Request profile for Go PGO.")
 	queryGoPGOOutput := queryGoPGOCmd.Flag("output", "How to output the result, examples: console, raw, pprof=./my.pprof").Default("pprof=./default.pgo").String()
@@ -144,7 +144,7 @@ func main() {
 			}
 		}
 	case queryProfileCmd.FullCommand():
-		if err := queryProfile(ctx, queryProfileParams, *queryProfileOutput, *queryProfileForce, *queryProfileTree); err != nil {
+		if err := queryProfile(ctx, queryProfileParams, *queryProfileOutput, *queryProfileForce, *queryProfileFunctionNamesOnly); err != nil {
 			os.Exit(checkError(err))
 		}
 	case queryGoPGOCmd.FullCommand():
--- a/cmd/profilecli/query.go
+++ b/cmd/profilecli/query.go
@@ -203,7 +203,7 @@ func queryProfileTree(ctx context.Context, params *queryProfileParams, from time
 		return nil, err
 	}
 
-	return pprof.FromTree(tree, ty, req.End), nil
+	return pprof.FromTree(tree, ty, req.End*1e6), nil
 }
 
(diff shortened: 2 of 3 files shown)


pkg/pprof/pprof_test.go
```diff
@@ -1441,6 +1442,40 @@ func Test_SampleExporter_Partial(t *testing.T) {
...
+	pprofFromTree := FromTree(treeFromPprof, &profileType, timeNanos)
+
+	// repeat the process
+	bytes, err = model.TreeFromBackendProfile(pprofFromTree, maxNodes)
+	require.NoError(t, err)
+	treeFromPprofFromTree, err := model.UnmarshalTree(bytes)
+	require.NoError(t, err)
+	pprofFromTreeFromTree := FromTree(treeFromPprofFromTree, &profileType, timeNanos)
+
+	// FromTree <-> model.TreeFromBackendProfile should be inverse:
```
simonswine: > // FromTree <-> model.TreeFromBackendProfile should be inverse:

Maybe I am not familiar enough with the term, I would call this behaviour immutable. Inverse would mean I am getting the mappings,lines,inline functions back.

cmd/profilecli/main.go
```diff
@@ -74,6 +74,7 @@ func main() {
 	queryProfileCmd := queryCmd.Command("profile", "Request merged profile.").Alias("merge")
 	queryProfileOutput := queryProfileCmd.Flag("output", "How to output the result, examples: console, raw, pprof=./my.pprof").Default("console").String()
 	queryProfileForce := queryProfileCmd.Flag("force", "Overwrite the output file if it already exists.").Short('f').Default("false").Bool()
+	queryProfileTree := queryProfileCmd.Flag("tree", "Reconstruct pprof from the native profile tree.").Default("false").Bool()
```
simonswine: To a user `--tree` means not that much, I would prefer `--function-names-only` (or something equally as descriptive)

I think you should also mention in the help the gist of what is in the PR description: "faster call, without details about mappings,linenumber,inlining"

cmd/profilecli/query.go
```diff
@@ -125,19 +154,56 @@ func queryProfile(ctx context.Context, params *queryProfileParams, outputFlag st
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to unmarshal tree")
+	}
+
+	ty, err := model.ParseProfileTypeSelector(params.ProfileType)
+	if err != nil {
+		return nil, err
+	}
+
+	return pprof.FromTree(tree, ty, req.End), nil
```
simonswine: This should likely be `req.Start` to match better what the profile endpoint would spit out.

Also you pass in milliseconds rather than nano
```
Time: 1970-01-01 01:29:29.504116728 +0100 BST
```

alsoba13: Good catch on the nano, I accidentally removed the conversion during a refactor.

I chose `req.End` because it was the one matching the pprof with my testing data:

```
$ go run . query merge --tree --from=2026-01-26T16:00:20.000000+01:00 --to=2026-01-26T16:01:00.000000+01:00 --query='{service_name="fire-dev-001/distributor"}' --output raw --force --max-nodes 1048576 | grep Time
level=info msg="query aggregated profile from profile store" url=https://profiles-dev-001.grafana-dev.net from=2026-01-26T16:00:20+01:00 to=2026-01-26T16:01:00+01:00 query="{service_name=\"fire-dev-001/distributor\"}" type=process_cpu:cpu:nanoseconds:cpu:nanoseconds
  TimeNanos:     1769439660000000000,
$ go run . query merge --from=2026-01-26T16:00:20.000000+01:00 --to=2026-01-26T16:01:00.000000+01:00 --query='{service_name="fire-dev-001/distributor"}' --output raw --force --max-nodes 1048576 | grep Time
level=info msg="query aggregated profile from profile store" url=https://profiles-dev-001.grafana-dev.net from=2026-01-26T16:00:20+01:00 to=2026-01-26T16:01:00+01:00 query="{service_name=\"fire-dev-001/distributor\"}" type=process_cpu:cpu:nanoseconds:cpu:nanoseconds
  TimeNanos:     1769439660000000000,
```

I've double checked with other queries and it's on the `req.End`. 

Edit: ok I found it: https://github.com/grafana/pyroscope/blob/9a2ae046642a1cc9a897df227043c327b518b309/pkg/phlaredb/block_querier.go#L1306

pkg/pprof/pprof_test.go
alsoba13: Thanks, better wording!

PR #4796, opened 2026-01-28
feat: Add configuration to skip label sanitization

jake-kramer:
Label sanitization would convert labels with dots to underscores. This change updates the validation logic to accept any utf-8 valid label name. The read path has been [previously](https://github.com/grafana/pyroscope/pull/4442) updated to filter out all non-legacy label names (ascii with underscores) unless the `allow-utf8-labelnames` client capability is set.

**By default, there is no change in logic/behavior**

If `disable_label_sanitization` is set to `true`, these are the breaking changes:
- Labels with dots will no longer be sanitized in to underscores.
- Any labels written that do not match the regex `^[a-zA-Z_][a-zA-Z0-9_]*$` will be written as-is
- Any labels written that do not match the regex `^[a-zA-Z_][a-zA-Z0-9_]*$` will be filtered out of the read path unless the `allow-utf8-labelnames` client capability is set.

Consider using `IngestionRelabelRules` for label sanitization moving forward.

### DIFF
--- a/cmd/pyroscope/help-all.txt.tmpl
+++ b/cmd/pyroscope/help-all.txt.tmpl
@@ -1157,6 +1157,8 @@ Usage of ./pyroscope:
     	[experimental] Set to true to enable profiling integration.
   -usage-stats.enabled
     	Enable anonymous usage statistics collection. For more details about usage statistics, refer to https://grafana.com/docs/pyroscope/latest/configure-server/anonymous-usage-statistics-reporting/ (default true)
+  -validation.disable-label-sanitization
+    	Disable label name sanitization (converting dots to underscores). When disabled, labels with dots are accepted as-is using UTF-8 validation.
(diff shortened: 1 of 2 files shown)


jake-kramer:
Won't be merged until relabeling rules are all set on the cloud side

jake-kramer:
> Not too sure what others are thinking, but I fear this feels like a proper breaking change to our OSS users and therefore, I think we should not do it between releases, and bundle the change into a Pyroscope v2.
> 
> For now I would introduce a flag or even better a per tenant override controlling this change and once we are working on we can do the breaking change and provide instructions, how to get the old behaviour back via relabeling rules.
> 
> Wdyt?

Addressed in [af4f74d](https://github.com/grafana/pyroscope/pull/4796/commits/af4f74df09400461692da1e9aef47d438319ab30).

PR #4797, opened 2026-01-28
feat: Add support for Javascript and Typescript source code integration

jake-kramer:


pkg/frontend/vcs/config/config.go
```diff
@@ -14,10 +14,12 @@ type Language string
...
-	LanguageUnknown = Language("")
-	LanguageGo      = Language("go")
-	LanguageJava    = Language("java")
-	LanguagePython  = Language("python")
+	LanguageUnknown    = Language("")
+	LanguageGo         = Language("go")
+	LanguageJava       = Language("java")
+	LanguagePython     = Language("python")
+	LanguageJavaScript = Language("javascript")
+	LanguageTypeScript = Language("typescript")
```
jake-kramer: rfc: Should `typescript` simply be implied under `javascript`? Or does it make sense to have both? I'm leaning toward consolidating under `javascript`.

jake-kramer:
This worked end-to-end with our NodeJS SDK for first-party code mapped with source code mappings in `.pyroscope.yaml`. It should also work for 3rd party libraries. I did not test with Otel eBPF.

pkg/frontend/vcs/config/config.go
simonswine: Yeah the lines are murky, I would maybe suggest this practice: The list of languages should match up with what we would have an SDK for.

That also mean that java would potentially include scala,kotlin,...

PR #4806, opened 2026-02-04
admin(v2): query diagnostics

aleks-p:
Adds an admin page to the Query Frontend component that allows us to collect and visualize query diagnostics. Diagnostics are written to the bucket and are shareable. Whether to collect diagnostics or not is decided on a request level, via a header (can be passed via `profilecli --collect-diagnostics` as well).

A big portion of the change is React code, written by Claude Code. I chose React because it made it easier to visualize the query execution graph. We can add more visualizations later on.

Things I left for a future iteration:
- linking block IDs to the block inspector page (#4480)
- downloading individual blocks or all blocks from a plan (for doing local tests)
- autocomplete in the query editor or alternative UIs for query composition
- capturing tracing identifiers

## Quick demo

A query section:

Screenshot 2026-02-04 at 12 08 10

Visualize the query plan:

Screenshot 2026-02-04 at 12 09 01

Visualize the query execution graph:

https://github.com/user-attachments/assets/a41f873f-f201-43db-9387-35700ac9084c

Stored diagnostics:

Screenshot 2026-02-04 at 12 23 36

### DIFF
--- a/cmd/profilecli/client.go
+++ b/cmd/profilecli/client.go
@@ -8,6 +8,8 @@ import (
 	"connectrpc.com/connect"
 	"github.com/prometheus/common/version"
 	"gopkg.in/alecthomas/kingpin.v2"
+
+	querydiagnostics "github.com/grafana/pyroscope/pkg/frontend/readpath/queryfrontend/diagnostics"
 )
 
 const (
@@ -80,7 +82,7 @@ func (a *authRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) 
 			req.Header.Set("Authorization", "Bearer "+c.BearerToken)
 		}
 		if c.CollectDiagnostics {
-			req.Header.Set("X-Pyroscope-Collect-Diagnostics", "true")
+			req.Header.Set(querydiagnostics.RequestHeader, "true")
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


cmd/profilecli/client.go
```diff
@@ -117,6 +121,10 @@ type commander interface {
 	Arg(name, help string) *kingpin.ArgClause
 }
 
+const (
+	diagnosticsIDHeader = "X-Pyroscope-Diagnostics-Id"
```
simonswine: nit: One day it would be nice having a central point for those header constants

aleks-p:
> LGTM. Thanks for working on this, this is super useful.
> 
> My only concern really is, that this bolts another admin page to the query-frontend, which endpoints might impact customer queries and also is more likely to be exposed to accidentally. (In order to use Pyroscope users need to expose distributor/query-frontend).
> 
> From what I can see there is no reason for the UI and related endpoint not to be moved to the admin component, which also might make it easier to integrate with e.g. the block explorer in the future.

Yeah, this work actually went through a few iterations - the first one was more tightly coupled with query-frontend code but the latest one isn't and it shouldn't be too hard to move it to the admin component. I'll give that a try!

PR #4812, opened 2026-02-06
DONOTMERGE feat: /ingest accept metric name

korniltsev-grafanista:


korniltsev-grafanista:
I use Push api instead

PR #4828, opened 2026-02-12
perf: replace regex with manual string scanning in DropGoTypeParameters

simonswine:
The `\[go\.shape\..*\]` regex caused expensive backtracking on every
string in the profile string table. Replace it with strings.Index /
strings.LastIndex, which replicates the greedy match semantics without
the regex engine overhead.

```
name                                          old time/op    new time/op    delta
_dropGoTypeParameters/with_type_params-12       2.70µs ± 0%    0.04µs ± 0%  -98.45%  (p=0.004)
_dropGoTypeParameters/without_type_params-12    30.7ns ± 0%     6.5ns ± 0%  -79.00%  (p=0.002)

name                                          old alloc/op   new alloc/op   delta
_dropGoTypeParameters/with_type_params-12         402B ± 0%       64B ± 0%  -84.07%  (p=0.002)
_dropGoTypeParameters/without_type_params-12     0.00B          0.00B          ~

name                                          old allocs/op  new allocs/op  delta
_dropGoTypeParameters/with_type_params-12         4.00 ± 0%      1.00 ± 0%  -75.00%  (p=0.002)
_dropGoTypeParameters/without_type_params-12      0.00           0.00          ~
```

### DIFF
--- a/pkg/pprof/fix_go_profile_test.go
+++ b/pkg/pprof/fix_go_profile_test.go
@@ -57,3 +57,84 @@ func Test_DropGoTypeParameters(t *testing.T) {
 	require.NoError(t, expected.Err())
 	require.False(t, expected.Scan())
 }
+
+func Test_dropGoTypeParameters(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		expected string
+	}{
+		{
+			name:     "no type parameters",
+			input:    "github.com/grafana/pyroscope/pkg/distributor.(*Distributor).Push",
+			expected: "github.com/grafana/pyroscope/pkg/distributor.(*Distributor).Push",
+		},
+		{
+			name:     "empty string",
+			input:    "",
+			expected: "",
+		},
+		{
+			name:     "simple type parameter",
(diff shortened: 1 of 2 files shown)


pkg/pprof/fix_go_profile_test.go
```diff
@@ -57,3 +57,84 @@ func Test_DropGoTypeParameters(t *testing.T) {
...
+			expected: "go.shape.int",
+		},
+		{
+			name:     "multiple separate type parameter sections",
+			input:    "pkg.Func[go.shape.int].Method[go.shape.string]",
+			expected: "pkg.Func.Method",
+		},
+		{
+			name:     "multiple type parameter sections with nested brackets",
+			input:    "pkg.Func[go.shape.struct { F []uint8 }].Method[go.shape.int]",
```
simonswine, line 114: This actually is also a bug fix, previously we wouldn't have treated a generic struct method like this correctly:

```
 type Func[T any] struct { }

 func (c *Func[T]) Method[U any](U) error {
 }
```

PR #4838, opened 2026-02-13
fix(renovate): Address invalid renovate.json syntax

jake-kramer:
Fixes https://github.com/grafana/pyroscope/issues/4835

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -115,5 +115,5 @@
         "executionMode": "update"
       }
     }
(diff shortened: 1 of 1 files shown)


jake-kramer:
@marcsanmi I just tested and apparently the validator doesn't check if it's valid JSON syntax?? I added this: https://github.com/grafana/pyroscope/pull/4839

Issue #4178, opened 2025-05-06
Migrate Pyroscope tracing to opentelemetry


simonswine:
#### Is your feature request related to a problem? Please describe.

Pyroscope tracing is based on jaeger client and opentracing, which are deprecated.

#### Describe the solution you'd like

- Migrate to opentelemetry.
- Being consistent between Mimir, Loki and Tempo (hopefully sharing the same code)
  - Tempo introduced opentelemetry tracer support (optional) in http://github.com/grafana/tempo/pull/842
  - Dskit has introduced support in https://github.com/grafana/dskit/pull/681
  -
#### Describe alternatives you've considered

_Nothing else considered._

callpraths:
This looks to be a dupe of #4127

Issue #4127, opened 2025-04-22
Migrate trace instrumentation from opentracing to open-telemetry


simonswine:
With dskit moving to otel/trace, we should also migrate our opentracing instrumentation to use otel/trace.

See https://github.com/grafana/dskit/pull/681

PR #4855, opened 2026-02-20
chore(tracing): support for open telemetry traces

aleks-p:
See https://github.com/grafana/pyroscope/issues/4127 and https://github.com/grafana/pyroscope/issues/4178.

Migrates all tracing instrumentation from OpenTracing to OpenTelemetry, using dskit's `tracing.StartSpanFromContext` bridge for span creation and dskit's `NewOTelOrJaegerFromEnv` for tracer initialization.

Includes an E2E verification tool (`tools/tracing/verify.sh`) that starts Pyroscope + Tempo, sends traffic, and asserts write-path and read-path traces appear in Tempo via TraceQL.

Existing Jaeger deployments continue to work without config changes, `dskit` auto-detects `JAEGER_*` env vars and configures Jaeger exporters via the OTel SDK. Trace export can be disabled entirely with `OTEL_TRACES_EXPORTER=none`.

TODO:
- [x] go over the diff
- [x] deploy in a dev environment and verify traces are flowing

---

> [!NOTE]
> **Medium Risk**
> Large dependency upgrades (observability stack, networking/memberlist, cloud SDKs) can introduce subtle runtime behavior changes despite most code changes here being documentation/help regeneration.
> 
> **Overview**
> Refreshes generated CLI help (`help.txt.tmpl`, `help-all.txt.tmpl`) and configuration reference docs to document new flags/options, including `-server.create-new-traces`, gRPC per-connection buffer sizing, expanded trace-header exclusion defaults, memberlist join parsing and new experimental memberlist knobs (fast-join min nodes, rejoin seed nodes, zone-aware routing, propagation delay tracker), and `server.cluster-validation.additional-labels`.
> 
> Updates tracing examples/docs to reference OpenTelemetry (`otel-profiling-go`) instead of OpenTracing, and bumps Go/tooling + a large set of dependencies (notably `dskit`, OpenTelemetry stack, Prometheus, AWS SDK, and memberlist fork) to versions consistent with the updated tracing/membership behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 00d32bff3ee22ebe76b83a5ef018a1076cadf7e2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


pkg/metastore/tracing/util.go
```diff
@@ -3,16 +3,16 @@ package tracing
...
-// Otherwise, it returns a noop span. To be used in places where we might not have access to the original context.
-func StartSpanFromContext(ctx context.Context, operationName string) (opentracing.Span, context.Context) {
-	if opentracing.SpanFromContext(ctx) != nil {
-		return opentracing.StartSpanFromContext(ctx, operationName)
+// StartSpanFromContext starts a new span from context only if the context
+// already carries a parent span. This avoids creating orphaned root spans
+// in places like metastore followers where no trace context is available.
+func StartSpanFromContext(ctx context.Context, operationName string) (*dskittracing.Span, context.Context) {
+	if !oteltrace.SpanFromContext(ctx).SpanContext().IsValid() {
+		return &dskittracing.Span{}, ctx
```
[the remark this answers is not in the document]
aleks-p: not a real issue, `SetTag`, `Finish` and other methods in dskit check if the inner opentracing or otel span is nil, which is the case for a `dskittracing.Span{}`

pkg/ingester/pyroscope/ingest_adapter.go
```diff
@@ -116,7 +116,7 @@ func (p *pyroscopeIngesterAdapter) Put(ctx context.Context, pi *storage.PutInput
 		Labels: make([]*typesv1.LabelPair, 0, 3+len(pi.LabelSet.Labels())),
 	}
 	series.Labels = append(series.Labels, &typesv1.LabelPair{
-		Name:  labels.MetricName,
+		Name:  string(prommodel.MetricNameLabel),
```
aleks-p, line 119: originally deprecated in https://github.com/prometheus/prometheus/pull/16228, more context in https://github.com/prometheus/prometheus/pull/17590

pkg/model/relabel/relabel_test.go
```diff
@@ -579,7 +579,7 @@ func TestRelabel(t *testing.T) {
 			if cfg.Replacement == "" {
 				cfg.Replacement = relabel.DefaultRelabelConfig.Replacement
 			}
-			require.NoError(t, cfg.Validate())
+			require.NoError(t, cfg.Validate(model.UTF8Validation))
```
aleks-p, line 582: `model.UTF8Validation` is the default,  https://github.com/prometheus/prometheus/pull/16928

pkg/validation/relabeling_test.go
```diff
@@ -195,8 +200,9 @@ func Test_defaultRelabelRules(t *testing.T) {
 			kept: true,
 		},
 	} {
-		result, kept := relabel.Process(tc.input, defaultRelabelRules...)
-		require.Equal(t, tc.expected, result)
+		lb := labels.NewBuilder(tc.input)
+		kept := relabel.ProcessBuilder(lb, defaultRelabelRules...)
+		require.Equal(t, tc.expected, lb.Labels())
```
aleks-p, line 205: https://github.com/prometheus/prometheus/pull/17531

pkg/settings/recording/recording_test.go
```diff
@@ -3,15 +3,14 @@ package recording
...
+	"math/rand/v2"
 	"os"
 	"testing"
-	"time"
 
 	"connectrpc.com/connect"
 	"github.com/go-kit/log"
 	"github.com/grafana/dskit/user"
 	"github.com/stretchr/testify/require"
-	"golang.org/x/exp/rand"
```
aleks-p, line 14: https://go-review.googlesource.com/c/exp/+/644475

pkg/validation/relabeling.go
```diff
@@ -85,7 +85,7 @@ func (p *RelabelRules) Set(s string) error {
 	}
 
 	for idx, rule := range v {
-		if err := rule.Validate(); err != nil {
+		if err := rule.Validate(model.UTF8Validation); err != nil {
```
aleks-p, line 88: model.UTF8Validation is the default, https://github.com/prometheus/prometheus/pull/16928

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/grafana/pyroscope
 
-go 1.24.9
+go 1.25.7
```
aleks-p, line 3: comes from dskit, https://github.com/grafana/dskit/blob/3167f499dfc047ca2fb20ce7d78eeed9fae6b954/go.mod#L3

pkg/util/http.go
```diff
@@ -238,9 +234,9 @@ func (w *reqBody) Read(p []byte) (int, error) {
 	if err == io.EOF {
 		w.duration = timeNow().Sub(w.start)
 		if w.sp != nil {
-			w.sp.LogFields(otlog.String("msg", "read body from request"))
+			w.sp.AddEvent("read body from request")
 			if w.read > 0 {
-				w.sp.SetTag("request_body_size", w.read)
+				w.sp.SetAttributes(attribute.String("request_body_size", w.read.String()))
```
marcsanmi: Is this changing from numeric to string? Could it break? Should we keep it as numeric?

pkg/compactor/bucket_compactor.go
```diff
@@ -773,10 +776,8 @@ func NewBucketCompactor(
 // Compact runs compaction over bucket.
 // If maxCompactionTime is positive then after this time no more new compactions are started.
 func (c *BucketCompactor) Compact(ctx context.Context, maxCompactionTime time.Duration) (rerr error) {
-	sp := opentracing.SpanFromContext(ctx)
-	if sp == nil {
-		sp, ctx = opentracing.StartSpanFromContext(ctx, "Compact")
-	}
+	sp, ctx := tracing.StartSpanFromContext(ctx, "Compact")
```
marcsanmi, line 779: The new code always creates a child "Compact" span. This changes trace structure (extra nesting level) and moves those attributes to a new child span. Is this intentional?

pkg/util/httpgrpcutil/carrier.go
```diff
@@ -6,43 +6,46 @@
...
+func (c *HttpgrpcHeadersCarrier) Keys() []string {
+	keys := make([]string, len(c.Headers))
+	for i, h := range c.Headers {
+		keys[i] = h.Key
 	}
+	return keys
+}
 
+// GetParentContextForRequest extracts parent trace context from HTTP request headers using OTel propagation.
+func GetParentContextForRequest(req *httpgrpc.HTTPRequest) context.Context {
```
bryanhuhta: I think we should pass a context into this function or rename this function to be a bit more precise. In its current form, it returns a brand new `context.Background()` which has no cancellation, no deadline, no extra keys except for the span.

This makes it easy to misuse by accident:

```go
// Extract trace context so queries appear in the right distributed trace.
reqCtx := httpgrpcutil.GetParentContextForRequest(req.HttpRequest)

// reqCtx looks like a real context. It has a valid span, it's non-nil,
// it was returned from a function called "GetParentContextForRequest".
// Nothing in the signature or name signals it's rooted at Background().
go func() {
    result, err := b.executeQuery(reqCtx, req)  // <-- never cancelled
    // ...
}()
```

Changing the signature to accept a context avoids this issue. The passed in context gets the span metadata attached to it and returned. All the cancellations, etc are preserved.

At the very least, if we still want to keep the signature, we should rename this to something like

```go
func NewContextWithParentSpan(req *httpgrpc.HTTPRequest) context.Context
```

pkg/phlaredb/query/iters.go
```diff
@@ -1084,11 +1084,10 @@ func (c *SyncIterator) seekPages(seekTo RowNumber, definitionLevel int) (done bo
...
 			c.metrics.pageReadsTotal.WithLabelValues(c.table, c.columnName).Add(1)
-			c.span.LogFields(
-				log.String("msg", "reading page (seekPages)"),
-				log.Int64("page_num_values", pg.NumValues()),
-				log.Int64("page_size", pg.Size()),
-			)
+			oteltrace.SpanFromContext(c.ctx).AddEvent("reading page (seekPages)", oteltrace.WithAttributes(
+				attribute.Int64("page_num_values", pg.NumValues()),
+				attribute.Int64("page_size", pg.Size()),
+			))
```
bryanhuhta: I'm unclear on the API, but it looks like `SyncIterator` has a field to cache the span:

https://github.com/grafana/pyroscope/blob/73c10ec4396e31e88c5a852a058f6d45e74947ee/pkg/phlaredb/query/iters.go#L847

I suspect this is to avoid context lookups in hot loops like what is happening here. Can we reuse the cached span? If not, can we cache this current span in a similar fashion to avoid context lookups?

This is mostly speculation based on the  fact that the previous span was cached, I have no hard numbers to prove there will be a performance penalty.

Theoretically we could use the [`IsRecording()`](https://pkg.go.dev/go.opentelemetry.io/otel/trace#Span) method on the span to further reduce the cost of annotating the span:

```go
if c.otelSpan.IsRecording() {
    // Add event to c.otelSpan
}
```

pkg/phlaredb/query/iters.go
```diff
@@ -1156,11 +1155,10 @@ func (c *SyncIterator) next() (RowNumber, *parquet.Value, error) {
...
 			c.metrics.pageReadsTotal.WithLabelValues(c.table, c.columnName).Add(1)
-			c.span.LogFields(
-				log.String("msg", "reading page (next)"),
-				log.Int64("page_num_values", pg.NumValues()),
-				log.Int64("page_size", pg.Size()),
-			)
+			oteltrace.SpanFromContext(c.ctx).AddEvent("reading page (next)", oteltrace.WithAttributes(
+				attribute.Int64("page_num_values", pg.NumValues()),
+				attribute.Int64("page_size", pg.Size()),
+			))
```
bryanhuhta: Same as above.

pkg/phlaredb/query/repeated.go
```diff
@@ -336,14 +337,13 @@ func (x *repeatedRowColumnIterator) readPage(rn int64) bool {
...
-		otlog.Int64("page_read_ms", pageReadDurationMs),
-		otlog.Int64("page_num_rows", pageNumRows),
-	)
+	oteltrace.SpanFromContext(x.ctx).AddEvent("Page read", oteltrace.WithAttributes(
+		attribute.Int64("min_rg_row", x.minRGRowNum),
+		attribute.Int64("max_rg_row", x.maxRGRowNum),
+		attribute.Int64("seek_row", x.minRGRowNum+rn),
+		attribute.Int64("page_read_ms", pageReadDurationMs),
+		attribute.Int64("page_num_rows", pageNumRows),
+	))
```
bryanhuhta: Here we have the same concern with extracting a span from a context in a hot loop. The `IsRecording()` method could applicable here too.

pkg/querybackend/query.go
```diff
@@ -200,8 +201,7 @@ func (b *blockContext) lookupDatasets(ds *metastorev1.Dataset) error {
 	md.Datasets = md.Datasets[:j]
 	b.obj.SetMetadata(md)
 
-	opentracing.SpanFromContext(b.ctx).
-		LogFields(otlog.String("msg", "dataset tsdb index lookup complete"))
+	otelTrace.SpanFromContext(b.ctx).SetAttributes(attribute.String("msg", "dataset tsdb index lookup complete"))
```
bryanhuhta: We probably want to add an event here instead:

```suggestion
	otelTrace.SpanFromContext(b.ctx).AddEvent("dataset tsdb index lookup complete")
```

pkg/querybackend/query.go
```diff
@@ -8,9 +8,10 @@ import (
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
+	"github.com/grafana/dskit/tracing"
 	"github.com/iancoleman/strcase"
-	"github.com/opentracing/opentracing-go"
-	otlog "github.com/opentracing/opentracing-go/log"
+	"go.opentelemetry.io/otel/attribute"
+	otelTrace "go.opentelemetry.io/otel/trace"
```
bryanhuhta:
```suggestion
	oteltrace "go.opentelemetry.io/otel/trace"
```

pkg/phlaredb/symdb/block_reader_parquet.go
```diff
@@ -57,11 +57,11 @@ func (t *parquetTable[M, P]) fetch(ctx context.Context) (err error) {
...
-			span.LogFields(
-				otlog.Uint32("row_group", h.RowGroup),
-				otlog.Uint32("index_row", h.Index),
-				otlog.Uint32("rows", h.Rows),
-			)
+			oteltrace.SpanFromContext(ctx).AddEvent("fetch row group", oteltrace.WithAttributes(
+				attribute.Int("row_group", int(h.RowGroup)),
+				attribute.Int("index_row", int(h.Index)),
+				attribute.Int("rows", int(h.Rows)),
+			))
```
bryanhuhta: You could lookup this span once and reuse it across loop iterations

pkg/util/http.go
bryanhuhta: I vote to keep it numeric, it'll make querying spans much easier.

pkg/pyroscope/tracing.go
```diff
@@ -0,0 +1,91 @@
...
+	closer, err := wwtracing.NewOTelOrJaegerFromEnv(serviceName, logger, opts...)
+	if err == nil {
+		return closer, nil
+	}
+	level.Warn(logger).Log("msg", "dskit tracing init failed, falling back to direct OTel SDK init", "err", err)
+	closer, fallbackErr := initTracingDirect(serviceName, logger, profilingEnabled)
+	if fallbackErr != nil {
+		// Return the original error as it is likely more informative.
+		return nil, err
+	}
```
bryanhuhta, line 42: Should we at least log the fallback error?

bryanhuhta (approved):
Took a peek at this. I think in general everything looks good. Maybe take a closer look at some of the hot path changes to make sure we don't introduce a performance regression.

Other than that, LGTM!

pkg/compactor/bucket_compactor.go
aleks-p: I checked and there is no good API to achieve what we had here before, without it becoming too complex. 

There is no `tracing.SpanFromContext` that returns a wrapped span (the existing method returns too many things and not the dskit's own wrapper). 

I think it is simpler to keep it as is, and deal with an extra nested span. This is also v1, which is on the path to deprecation.

PR #4866, opened 2026-02-27
feat(helm): support global.imageRegistry/global.image.registry

okhowang:
> [!NOTE]
> **Low Risk**
> Low risk: a small Helm templating change that only affects how container image names are constructed when optional registry values are set.
> 
> **Overview**
> Adds support for overriding the container image registry in the Pyroscope Helm chart via **either** a new `global.imageRegistry` value (applies across the chart) or `pyroscope.image.registry` (per-chart).
> 
> Updates the deployment/statefulset template to prefix images with the chosen registry, and documents/threads the new values through `values.yaml` and generated `jsonnet/values.json`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ef56d98c36a77e9278c90ece57e31352898ec48b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/operations/pyroscope/helm/pyroscope/README.md
+++ b/operations/pyroscope/helm/pyroscope/README.md
@@ -35,6 +35,7 @@
 | architecture.storage.migration.segmentWriterWeight | float | `1` | Specifies the fraction [0:1] that should be send to the v2 write path / segment-writer in combined mode. 0 means no traffics is sent to segment-writer. 1 means 100% of requests are sent to segment-writer. |
 | architecture.storage.v1 | bool | `true` | Enable v1 storage layer. |
 | architecture.storage.v2 | bool | `false` | Enable v2 storage layer. |
+| global.imageRegistry | string | `nil` | Overrides the Docker registry globally for all images |
 | ingress.annotations | object | `{}` |  |
 | ingress.className | string | `""` |  |
 | ingress.enabled | bool | `false` |  |
@@ -59,6 +60,7 @@
 | pyroscope.grpc.port | int | `9095` |  |
 | pyroscope.grpc.port_name | string | `"grpc"` |  |
 | pyroscope.image.pullPolicy | string | `"IfNotPresent"` |  |
+| pyroscope.image.registry | string | `""` |  |
 | pyroscope.image.repository | string | `"grafana/pyroscope"` |  |
 | pyroscope.image.tag | string | `""` |  |
 | pyroscope.imagePullSecrets | list | `[]` |  |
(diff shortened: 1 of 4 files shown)


marcsanmi:
I think this would be a breaking change for users who have a full registry in `image.repository` today, e.g.:

`myregistry.com/grafana/pyroscope:tag` → `docker.io/myregistry.com/grafana/pyroscope:tag`

Could you default `image.registry` to empty instead of `docker.io`, and only prepend the registry when explicitly set? Something like:

```yaml
{{- $registry := .Values.global.imageRegistry | default $values.image.registry -}}
image: "{{ if $registry }}{{ $registry }}/{{ end }}{{ $values.image.repository }}:{{ $values.image.tag | default .Chart.AppVersion }}"
```

This way existing users with repository: `myregistry.com/grafana/pyroscope` aren't broken, while the new `global.imageRegistry / image.registry` feature still works.

marcsanmi:
Thanks for addressing my earlier feedback @okhowang 

To address @simonswine's comment about sub-charts: could you rename `global.imageRegistry` → `global.image.registry`? The Grafana sub-charts ([alloy](https://github.com/grafana/alloy/blob/main/operations/helm/charts/alloy/values.yaml#L16), [grafana-agent](https://github.com/grafana/agent/blob/main/operations/helm/charts/grafana-agent/values.yaml#L13)) already read `global.image.registry`, so Helm will propagate it automatically.

Also, CI needs the generated files updated — please run `make helm/docs && make lint` and commit the results.

okhowang:
I read both `global.imageRegistry` and `global.image.registry` now.
because I notice that loki support them both. https://github.com/grafana/loki/pull/19246

marcsanmi:
Thanks for the updates. As you might have noticed, there isn't a consistent standard across Grafana charts for the global image registry key:

- `global.imageRegistry` — used by Grafana, Loki, Tempo (single-binary)
- `global.image.registry` — used by Alloy, Grafana Agent, Tempo-distributed

Loki ended up supporting both with `global.imageRegistry` taking precedence and `global.image.registry` as a deprecated fallback. This PR follows that same approach.

@simonswine — should we support both keys like Loki does (which is what this PR currently implements), or pick just
one? Worth noting that `global.image.registry` automatically propagates to our alloy/grafana-agent sub-charts since they already read that key, while `global.imageRegistry` does not.

marcsanmi:
For reference, the Bitnami common library only [uses](https://github.com/bitnami/charts/blob/main/bitnami/common/templates/_images.tpl#L13) `global.imageRegistry` and doesn't support `global.image.registry`.

Given there's no consistent standard across Grafana charts either, maybe I'd lean toward following Bitnami's convention and only supporting `global.imageRegistry` — simpler, no deprecated fallback to maintain. The trade-off is that it won't
automatically propagate to our sub-charts (alloy/grafana-agent use global.image.registry)

simonswine:
> Given there's no consistent standard across Grafana charts either, maybe I'd lean toward following Bitnami's convention and only supporting `global.imageRegistry`

Agree with that, I have also done a github code search for either and it is about 44k matches for `global.imageRegistry` vs. 4k `global.image.registry`.

marcsanmi:
@okhowang I think you need to run `make helm/check` to make the CI happy.

PR #4877, opened 2026-03-03
chore: bump Go to 1.25.7

marcsanmi:
## Summary

- Bumps Go from 1.24.x (EOL) to **1.25.7** across the entire codebase
- Updates `go` directive to `1.25.0`, `toolchain` to `go1.25.7` in all go.mod/go.work files
- Updates CI workflows, Dockerfiles, goreleaser, .pyroscope.yaml, and update_examples.Dockerfile
- Removes `tools/upgrade-go-version.sh` (superseded by `/update-go-version` skill from #4876)
- Small legitimate dependency adjustments in `api/go.mod` from `make go/mod` (otel 1.39→1.40, sys 0.39→0.40)

Generated with `/update-go-version 1.25.7` Claude skill.

**Depends on #4876** (skill PR) — rebase onto main after that merges.

## Test plan

- [ ] CI passes (build, test, lint)
- [ ] `go version` in CI matches 1.25.7
- [ ] Docker images build with golang:1.25.7
- [ ] goreleaser version check passes

---

> [!NOTE]
> **Medium Risk**
> Broad Go toolchain bump can surface build/test or runtime regressions across CI, release packaging, and Docker images. The `kuberesolver` waitgroup change also touches goroutine lifecycle/shutdown behavior, which could impact service discovery stability.
> 
> **Overview**
> Upgrades the repo to **Go 1.25.7**: CI workflows, release/weekly-release, goreleaser version guard, `.pyroscope.yaml` source mapping, `tools/update_examples.Dockerfile`, and all example Dockerfiles now target `1.25.7`, while `go.mod`/`go.work` directives move to `go 1.25.0` with `toolchain go1.25.7`.
> 
> Includes small generated module churn in `api/go.mod`/`go.sum` (notably OpenTelemetry and `x/sys` bumps), removes the now-obsolete `tools/upgrade-go-version.sh`, and adjusts `pkg/metastore/discovery/kuberesolver` so the watcher `WaitGroup` is incremented inside `watch()` to better align `Close()` with watcher goroutine lifetime.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8c5b8286fa95f6fa4d4147a6c3741d999dd03eed. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


pkg/metastore/discovery/kuberesolver/builder.go
```diff
@@ -93,6 +92,7 @@ func (k *KResolver) resolve() {
 }
 
 func (k *KResolver) watch() error {
+	k.wg.Add(1)
```
aleks-p, line 95: is this intentional and considered part of the Go upgrade?

aleks-p: nvm, I see the commit now

> fix: move WaitGroup.Add inside watch() to fix Go 1.25 vet check

marcsanmi: Kind of, the reason is explained in this commit 8c5b8286fa95f6fa4d4147a6c3741d999dd03eed. I think it started failing because of the vamp a new go vet check that was added. Initially I wanted to do it in a separate PR, but since the change was trivial I added it here... I can quickly create another PR, lmk 🙂

go.work
```diff
@@ -1,6 +1,6 @@
-go 1.24.6
```
simonswine, line 1: I think we should only update the `go` directive when we want a go feature from a newer version or we have a dependency that requires a feature from a newer go version.

From: https://go.dev/doc/modules/gomod-ref#go-notes

> The go directive sets the minimum version of Go required to use this module.

Not the go version we compile this with, that's `toolchain`

> Declares a suggested Go toolchain to use with this module.

tools/upgrade-go-version.sh
```diff
@@ -1,30 +0,0 @@
-#!/usr/bin/env bash
```
simonswine: I am not fully convinced this skill is doing an equally good job as this bash script

go.work
marcsanmi: Now the skill asks on every run whether you want to also update the `go` directive.

korniltsev-grafanista (commented):
I dont understand why we need a skill for this. Should not this be a script?

marcsanmi:
> I dont understand why we need a skill for this. Should not this be a script?

The script stays as-is and can be used standalone. The skill just wraps it and adds the go.mod/go.work toolchain updates + make go/mod + build verification — steps that were previously done manually and undocumented. The skill also handles things like diagnosing build failures interactively (e.g. the Go 1.25 vet check we hit in this PR). 

If you'd prefer these extra steps in the script or done manually, happy to close the PR 🙂

pkg/metastore/discovery/kuberesolver/builder.go
simonswine: That vet fail is probably a bug

PR #2508, opened 2023-10-10
Extend profile schema to support spans

kolesnikovae:
The PR extends the Profile parquet schema with `SpanID` field: this an optional array of uint64 for span identifiers associated with each sample. See #2458 for the context

PR #4884, opened 2026-03-06
feat: add trace ID field to profile storage schema

marcsanmi:
## Description
Adds a trace ID column to the Parquet profile storage schema, enabling cross-signal correlation between profiles and traces.

## Changes

- **Parquet schema**: Add optional `TraceID` column (`FixedLenByteArray(16)`) to samples, following the same pattern as `SpanID`
- **pprof extraction**: Extract `trace_id` labels from pprof samples into `[16]byte` arrays
- **OTLP ingestion**: Extract trace ID from OTLP `Link` objects and store as pprof labels
- **Ingestion path**: Wire trace ID extraction through `dedup_slice.go` into `Samples.TraceIDs`

## Compatibility

- **Backward compatible**: Old blocks without TraceID load without errors (`HasTraceID()` returns false)
- **Forward compatible**: Old code reading new blocks silently ignores the extra column
- Follows the exact same optional column pattern established by SpanID

## Scope

Storage and ingestion only. Query-layer changes (exemplars, heatmap, filtering) are follow-up work.

---

> [!NOTE]
> **Medium Risk**
> Schema and ingestion-path changes affect how profiles are encoded/decoded on disk; while the new column is optional and guarded for backward compatibility, any mismatch in parquet handling could break reads or increase storage/memory usage.
> 
> **Overview**
> Adds an optional per-sample `TraceID` column to the v1 Parquet profiles schema (fixed_len_byte_array(16)) and wires it through the in-memory `Samples` representation and row de/serialization, including schema compatibility checks for older blocks.
> 
> Extends ingestion to populate trace IDs: OTLP conversion now emits a `trace_id` label from link data, `pprof` extraction parses `trace_id` labels into `[16]byte`, and `symdb`/head ingestion carries these into stored samples.
> 
> Simplifies `Samples.Compact` to always drop zero/negative values (removing the unused dedupe flag) and updates callers, fixtures, and metrics/test expectations impacted by the schema size change.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cb082349145d2feb5a853b34b900c3e3fdffd2f0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


pkg/phlaredb/schemas/v1/profiles.go
```diff
@@ -422,6 +436,9 @@ func (s Samples) Range(n, m int) Samples {
 	if len(s.Spans) > 0 {
 		x.Spans = s.Spans[n:m]
 	}
+	if len(s.TraceIDs) > 0 {
+		x.TraceIDs = s.TraceIDs[n:m]
+	}
```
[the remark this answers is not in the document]
marcsanmi: The bug is real but pre-existing and not introduced by this PR.

`trimDuplicateSamples` also drops spans the same way — this was broken before TraceID was added. However, all production callers use `Compact(false)`, which skips `trimDuplicateSamples` entirely.

Food for thought: When two samples with the same stacktrace ID but different trace/span IDs are merged, which ID should win? I'd say, let's address this in another PR, if needed.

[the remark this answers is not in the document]
simonswine: Probably best to get rid of `dedupe` from `Compact` now all together.

pkg/phlaredb/schemas/v1/profiles.go
```diff
@@ -39,6 +39,7 @@ var (
 		phlareparquet.NewGroupField("Value", parquet.Encoded(parquet.Int(64), &parquet.DeltaBinaryPacked)),
 		phlareparquet.NewGroupField("Labels", pprofLabels),
 		phlareparquet.NewGroupField("SpanID", parquet.Optional(parquet.Encoded(parquet.Uint(64), &parquet.RLEDictionary))),
+		phlareparquet.NewGroupField("TraceID", parquet.Optional(parquet.Leaf(parquet.FixedLenByteArrayType(16)))),
```
simonswine: The cost of not using a dictionary might be quite substantial, that said it only impacts rows (through `optional`) that actually send TraceID. 

Not too sure if we ever collected actual evidence for the SpanID case, but I still assume the cardinality of trace IDs per block to be low and they repeating quite often per profile (so they should be close to what SpanIDs) are. So I think using a dictionary will be quite beneficial.

pkg/phlaredb/schemas/v1/schema_test.go
```diff
@@ -26,6 +26,17 @@ func TestSchemaMatch(t *testing.T) {
...
+	// parquet-go maps FixedLenByteArray to []byte in Go structs (same as Tempo).
+	// The struct produces "optional binary TraceID" while ProfilesSchema uses
+	// "optional fixed_len_byte_array(16) TraceID" — functionally equivalent,
+	// just a schema string mismatch that we normalize here.
+	profilesStructSchema = strings.Replace(
+		profilesStructSchema,
+		"optional binary TraceID;",
+		"optional fixed_len_byte_array(16) TraceID;",
+		1,
+	)
```
simonswine, line 36: If you use fixed length here you wouldn't need this exception:

```
--- a/pkg/phlaredb/schemas/v1/profiles.go
+++ b/pkg/phlaredb/schemas/v1/profiles.go
@@ -203,7 +203,7 @@ type Sample struct {
        Value        int64              `parquet:",delta"`
        Labels       []*profilev1.Label `parquet:",list"`
        SpanID       uint64             `parquet:",optional"`
-       TraceID      []byte             `parquet:",optional"`
+       TraceID      [16]byte           `parquet:",optional"`
 }
```

marcsanmi: I think `[16]byte` doesn't work for optional `FixedLenByteArray` in parquet-go.  I initially tried it but tests panicked during parquet reconstruction on profiles without trace context. When the value is null, [fixedLenByteArrayType.AssignValue](https://github.com/parquet-go/parquet-go/blob/v0.24.0/type.go#L896) receives a zero-length byte array, the guard dst.Len() == len(v) fails (16 != 0), and the fallback at https://github.com/parquet-go/parquet-go/blob/v0.24.0/type.go#L912-L913 panics trying to set `[]byte` into `[16]byte`. 

Tempo also uses `[]byte` for [TraceID](https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet4/schema.go#L249) maybe for same reason?

pkg/phlaredb/schemas/v1/profiles.go
```diff
@@ -501,6 +507,9 @@ func (s SamplesBySpanID) Swap(i, j int) {
 	if len(s.Spans) > 0 {
 		s.Spans[i], s.Spans[j] = s.Spans[j], s.Spans[i]
 	}
+	if len(s.TraceIDs) > 0 {
+		s.TraceIDs[i], s.TraceIDs[j] = s.TraceIDs[j], s.TraceIDs[i]
+	}
```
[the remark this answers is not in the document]
marcsanmi: Valid point — `Size()` is already missing Spans since it was added I think in #2508 (~2 years ago), so this is pre-existing. However, **impact is negligible**: both fields are nil for most profiles, and V2 doesn't use
it at all. Happy to open a follow-up PR if we think it's worth fixing.

pkg/phlaredb/schemas/v1/schema_test.go
simonswine: Hm that's frustrating. Do you have a branch where it panics, at least the integration tests seem to work fine locally?

I will also try to update parquet-go separately

marcsanmi: Here you have it https://github.com/grafana/pyroscope/tree/marc/repro-16byte-traceid-panic. Reproduce it with: 
```
go test ./pkg/phlaredb/... -run TestProfileStore_RowGroupSplitting -count=1
```

simonswine: Thanks, it seems like It no longer panics after we upgrade to latest parquet-go: https://github.com/grafana/pyroscope/pull/4925

That PR also still has a few regressions

simonswine: Main is currently on 0.26.4, that still panics, Once my upstream PR merges, we can update to the latest and use `[16]byte`.

For now I don't think we should block this, please go ahead with `[]byte`, the change won't be very disruptive.

simonswine (approved):

PR #4887, opened 2026-03-11
refactor(queryfrontend): move symbolization into per-query backendWrapper

simonswine:
Replace the centralized shouldSymbolize+modifiedQueries+processAndSymbolizeProfiles
approach in doQuery with a backendWrapper callback pattern. Each call site
(SelectMergeProfile, SelectMergeStacktraces, SelectMergeSpanProfile) now captures
whether symbolization is needed after QueryMetadata and applies it inline, rather than
having the query frontend rewrite queries and post-process responses globally.

This makes symbolization explicit at the call site and opens the door to
per-query-type symbolization strategies. Symbolization logic is extracted into
symbolizer.go alongside a TODO noting that it should eventually move into the
query backends as part of the query plan.

Known limitation: queries without a strict service_name label selector fall back
to tenant-wide TSDB index blocks which do not carry per-dataset unsymbolized=true
labels, so symbolization is silently skipped for those queries.

---

> [!NOTE]
> **Medium Risk**
> Moderate risk: refactors the core query execution path and changes when/how symbolization is applied, which could affect query types that rely on TREE/PPROF handling and span/stacktrace selectors.
> 
> **Overview**
> Refactors `QueryFrontend` to route all query execution through a new `doQuery` helper that accepts an optional `backendWrapper`, allowing call sites to wrap/alter the backend after `QueryMetadata` returns blocks.
> 
> Removes the previous centralized symbolization flow (global query rewriting + post-processing of responses) and instead applies symbolization per endpoint: `SelectMergeProfile` now symbolizes returned pprof inline, while `SelectMergeStacktraces`/`SelectMergeSpanProfile` conditionally wrap the backend with `backendTreeSymbolizer` to rewrite `QUERY_TREE`→`QUERY_PPROF`, symbolize, and convert back.
> 
> Adds `symbolizer.go` to house the symbolization helpers and introduces new tests covering symbolization behavior for merge-profile, stacktraces, and span-profile paths; non-symbolization query helpers were updated to pass `nil` wrappers.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit af7edde09597d32f999884026a404ceeecaa57e5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/frontend/readpath/queryfrontend/symbolizer.go
+++ b/pkg/frontend/readpath/queryfrontend/symbolizer.go
@@ -29,7 +29,7 @@ type backendTreeSymbolizer struct {
 }
 
 func (b *backendTreeSymbolizer) Invoke(ctx context.Context, req *queryv1.InvokeRequest) (resp *queryv1.InvokeResponse, err error) {
-	span, ctx := opentracing.StartSpanFromContext(ctx, "QueryFrontend.processAndSymbolizeProfiles")
+	span, ctx := opentracing.StartSpanFromContext(ctx, "backendTreeSymbolizer.Invoke")
 	defer func() {
 		if err != nil {
 			ext.LogError(span, err)
@@ -46,7 +46,12 @@ func (b *backendTreeSymbolizer) Invoke(ctx context.Context, req *queryv1.InvokeR
 		if q.QueryType == queryv1.QueryType_QUERY_TREE {
 			q.QueryType = queryv1.QueryType_QUERY_PPROF
 			q.Pprof = &queryv1.PprofQuery{
-				MaxNodes: q.Tree.GetMaxNodes(),
+				MaxNodes:           q.Tree.GetMaxNodes(),
+				ProfileIdSelector:  q.Tree.GetProfileIdSelector(),
+				StackTraceSelector: q.Tree.GetStackTraceSelector(),
+				// SpanSelector is not forwarded: PprofQuery has no span_selector field.
+				// To support span-filtered symbolization, span_selector must be added to
+				// the PprofQuery proto and the backend must handle it.
 			}
 			q.Tree = nil
 		}
@@ -91,7 +96,6 @@ func (b *backendTreeSymbolizer) Invoke(ctx context.Context, req *queryv1.InvokeR
 	}
 
 	return resp, nil
-
 }
 
 // hasUnsymbolizedProfiles checks if a block has unsymbolized profiles
--- a/pkg/frontend/readpath/queryfrontend/query_select_merge_span_profile_test.go
+++ b/pkg/frontend/readpath/queryfrontend/query_select_merge_span_profile_test.go
@@ -1,0 +1,202 @@
+package queryfrontend
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"connectrpc.com/connect"
+	"github.com/go-kit/log"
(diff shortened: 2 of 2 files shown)


pkg/frontend/readpath/queryfrontend/symbolizer.go
```diff
@@ -0,0 +1,146 @@
...
+
+	modifiedReq := req.CloneVT()
+
+	// check all queries for the ones using tree
+	for _, q := range modifiedReq.Query {
+		// If this is a TREE query, convert it to PPROF
+		if q.QueryType == queryv1.QueryType_QUERY_TREE {
+			q.QueryType = queryv1.QueryType_QUERY_PPROF
+			q.Pprof = &queryv1.PprofQuery{
+				MaxNodes: q.Tree.GetMaxNodes(),
```
marcsanmi: I know this was like this before, but I've just realised we are dropping `ProfileIdSelector`, can we add it? (same for other selectors e.g `SpanSelector`)

pkg/frontend/readpath/queryfrontend/symbolizer.go
```diff
@@ -0,0 +1,146 @@
...
+// so that symbolization can be distributed and executed closer to the data.
+
+// backendTreeSymbolizer allows to symbolize FunctionName Tree queries, by converting them into pprof queries, symbolize them and converting them back.
+type backendTreeSymbolizer struct {
+	upstream   QueryBackend
+	symbolizer Symbolizer
+}
+
+func (b *backendTreeSymbolizer) Invoke(ctx context.Context, req *queryv1.InvokeRequest) (resp *queryv1.InvokeResponse, err error) {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "QueryFrontend.processAndSymbolizeProfiles")
```
marcsanmi: This is legacy, maybe `backendTreeSymbolizer.Invoke"`?

pkg/frontend/readpath/queryfrontend/query_select_merge_span_profile.go
marcsanmi, line 1: Should we also add a test for this as in the others?

pkg/frontend/readpath/queryfrontend/symbolizer.go
```diff
@@ -0,0 +1,146 @@
...
+				return nil, fmt.Errorf("failed to build tree: %w", err)
+			}
+			r.Tree = &queryv1.TreeReport{Tree: treeBytes}
+			r.ReportType = queryv1.ReportType_REPORT_TREE
+			r.Pprof = nil
+		}
+	}
+
+	return resp, nil
+
```
marcsanmi:
```suggestion
```

pkg/frontend/readpath/queryfrontend/symbolizer.go
simonswine: SpanSelector is not supported in pprof and it won't work as expected.

pkg/frontend/readpath/queryfrontend/query_select_merge_span_profile.go
simonswine: SpanSelector is not supported in pprof and it won't work as expected, this can be fixed once we have the tree query path merged. Added comments and skipped test for this

PR #4888, opened 2026-03-12
fix(querybackend): filter profile IDs post-join instead of predicate push-down

marcsanmi:
## Summary

- When `ProfileIdSelector` is set (exemplar "View profile" click), the ID column predicate is no longer pushed down into the parquet column iterator. Instead, the ID column is fetched without a predicate and filtered post-join via `profileIDFilterIterator`.
- This avoids a known alignment bug in nested `BinaryJoinIterator` that triggers when one column iterator is highly selective (sparse), returning empty results for some profile IDs.

## Test plan

- [ ] Manual: click exemplar "View profile" for previously-failing profile IDs/types
- [ ] `go test ./pkg/querybackend/...`

### DIFF
--- a/pkg/phlaredb/query/iters_test.go
+++ b/pkg/phlaredb/query/iters_test.go
@@ -467,65 +467,6 @@ func TestBinaryJoinIterator(t *testing.T) {
 	}
 }
 
-// TestBinaryJoinIterator_SparseLeftIterator verifies that BinaryJoinIterator
-// finds matches when the left iterator is highly selective (few matching rows).
-// This reproduces the bug fixed by the converge() method.
-func TestBinaryJoinIterator_SparseLeftIterator(t *testing.T) {
-	type T struct {
-		ID        string
-		SeriesID  uint32
-		TimeNanos int64
-	}
-
-	targetID := "target__________" // 16 bytes
-	otherID := func(i int) string {
-		return fmt.Sprintf("other_%010d_", i) // 16 bytes
-	}
-
-	// Target ID at rows 50, 100, 150, 180. SeriesID cycles 0-7.
-	// Only row 150 matches all three predicates (ID=target, seriesID=6, time in range).
-	rowCount := 200
-	rows := make([]T, rowCount)
-	targetRows := map[int]bool{50: true, 100: true, 150: true, 180: true}
-	for i := range rows {
-		id := otherID(i)
-		if targetRows[i] {
-			id = targetID
-		}
-		rows[i] = T{
-			ID:        id,
-			SeriesID:  uint32(i % 8),
-			TimeNanos: int64(i) * 1000,
-		}
-	}
-
-	pf := createFileWith[T](t, rows, 1)
-
-	ctx := context.Background()
-	reg := prometheus.NewRegistry()
-	metrics := NewMetrics(reg)
-	ctx = AddMetricsToContext(ctx, metrics)
-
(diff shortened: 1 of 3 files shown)


pkg/querybackend/query_profile_entry.go
alsoba13, line 1: Not sure about this change. If the SelectSeries widens the start by one step, let's do the same for the Tree call instead of completely ignore time ranges?

alsoba13: Also, I don't think this contributes a lot to the empty flame graph bug. 
This should only happen in the edges (and we certainly already have some problems in there - image attached)

Image

pkg/phlaredb/query/iters_test.go
```diff
@@ -467,6 +467,65 @@ func TestBinaryJoinIterator(t *testing.T) {
 	}
 }
 
+// TestBinaryJoinIterator_SparseLeftIterator verifies that BinaryJoinIterator
+// finds matches when the left iterator is highly selective (few matching rows).
+// This reproduces the bug fixed by the converge() method.
+func TestBinaryJoinIterator_SparseLeftIterator(t *testing.T) {
```
simonswine: I do think it looks like there is a bug in nesting two BinaryJoins, but I don't think the fix is reasonable and it is super risky, I would need to spend a lot more time reviewing this. Also the area is not particularly well tested.

I rather have a work around for now, e..g if the nesting is a problem, how about you don't do the predicate push down and filter e.g. the time when going through the results in query_profile_entry.go.

And once we worked around we should look a lot more into detail how a reuse bug can happen on the BinaryJoinIterator and if nested binary joins work on both sides, this tests already covers that pretty well.

pkg/querybackend/query_profile_entry.go
marcsanmi: > Not sure about this change. If the SelectSeries widens the start by one step, let's do the same for the Tree call instead of completely ignore time ranges?

I initially tried this but we don't have step in `SelectMergeStacktracesRequest`. The alternatives would be (maybe there're more?):
- Use an arbitrary constant (e.g., 5min) — I tried this, but it's fragile
- Remove the time predicate when ProfileIdSelector is set — what I did. This is safe because profile IDs are globally unique (UUIDs), so the ID predicate is already maximally selective. The time predicate adds zero filtering value when you're looking for specific UUIDs.

This basically fixes some corner cases where flamegraph is also empty. Can you think of a better approach? Should we fix SelectSeries instead?

pkg/phlaredb/query/iters_test.go
simonswine: Btw if the AI comment is right on the converge, a fix might also be two swap the iterator legs:

```--- a/pkg/querybackend/query_profile_entry.go
+++ b/pkg/querybackend/query_profile_entry.go
@@ -143,8 +143,8 @@ func (c queryColumns) join(q *queryContext) parquetquery.Iterator {
                        continue
                }
                result = parquetquery.NewBinaryJoinIterator(0,
-                       result,
                        it,
+                       result,
                )
        }
        return result```
```

pkg/querybackend/query_profile_entry.go
marcsanmi: Let's leave this out of the scope of this PR

pkg/phlaredb/query/iters_test.go
marcsanmi: > Btw if the AI comment is right on the converge, a fix might also be two swap the iterator legs:

I tried this but didn't worked. Same corruption happens just on the other side

PR #4892, opened 2026-03-12
feat(metastore): add log store write timeout to prevent stuck raft leader

aleks-p:
## Summary

- Adds a `timeoutLogStore` wrapper that caps raft log store write latency, converting blocking I/O into errors after a configurable deadline (default 10s). This prevents a stuck disk from permanently stalling the raft leader while heartbeats continue on separate goroutines, blocking follower elections.
- Emits `raft_log_store_write_duration_seconds` (histogram) and `raft_log_store_write_timeouts_total` (counter) metrics for alerting on disk degradation.
- Adds an end-to-end integration test that exercises the full metastore stack: writes block metadata, injects a blocking disk fault on the leader, waits for a new leader election, then verifies writes and reads through the new leader.

See [hashicorp/raft#503](https://github.com/hashicorp/raft/issues/503) for the upstream issue

## Context

When a disk becomes slow (high I/O wait), the raft leader's `StoreLogs` call can block indefinitely, freezing its main goroutine. Because heartbeats run on separate goroutines, followers never time out and no election is triggered — making the entire cluster unavailable until manual intervention.

The `timeoutLogStore` wrapper converts this blocking scenario into fast errors, which raft already handles correctly by stepping down and allowing a new election.

Fixes https://github.com/grafana/pyroscope-squad/issues/605

---

> [!NOTE]
> **Medium Risk**
> Touches Raft persistence write path by adding goroutine-based timeouts, which can affect leader stability and error behavior under load or slow I/O. Includes new metrics and an end-to-end test, but incorrect timeout values or wrapper behavior could still cause unexpected step-downs or write failures.
> 
> **Overview**
> Adds a `timeoutLogStore` wrapper around the Raft `LogStore` to bound `StoreLog(s)` write latency and return an error after a configurable `log-store-timeout` (default 10s), preventing a leader from hanging indefinitely on slow/stuck disk I/O.
> 
> Exposes new Prometheus telemetry (`raft_log_store_write_duration_seconds` and `raft_log_store_write_timeouts_total`) and wires the wrapper into node initialization, plus a test-only `LogStoreMiddleware` hook to inject failures.
> 
> Introduces an integration test (`metastore_disk_failure_test.go`) that simulates a blocked leader log store, verifies a new leader election occurs, and confirms the cluster can resume reads/writes after recovery.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ac5f5fac90d3854a81206d22c61acd9c013a40aa. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/metastore/raftnode/node_metrics.go
+++ b/pkg/metastore/raftnode/node_metrics.go
@@ -23,10 +23,8 @@ func newMetrics(reg prometheus.Registerer) *metrics {
 			Help:                            "Duration of applying a command to the Raft log",
 			Buckets:                         prometheus.DefBuckets,
 			NativeHistogramBucketFactor:     1.1,
-			NativeHistogramZeroThreshold:    0,
 			NativeHistogramMaxBucketNumber:  50,
 			NativeHistogramMinResetDuration: time.Hour,
-			NativeHistogramMaxZeroThreshold: 0,
 		}),
 
 		read: prometheus.NewHistogram(prometheus.HistogramOpts{
@@ -34,10 +32,8 @@ func newMetrics(reg prometheus.Registerer) *metrics {
 			Help:                            "Duration of the Raft log read index wait",
 			Buckets:                         prometheus.DefBuckets,
 			NativeHistogramBucketFactor:     1.1,
-			NativeHistogramZeroThreshold:    0,
 			NativeHistogramMaxBucketNumber:  50,
 			NativeHistogramMinResetDuration: time.Hour,
-			NativeHistogramMaxZeroThreshold: 0,
 		}),
 
 		state: prometheus.NewGaugeVec(
@@ -53,10 +49,8 @@ func newMetrics(reg prometheus.Registerer) *metrics {
 			Help:                            "Duration of log store write operations",
 			Buckets:                         prometheus.DefBuckets,
 			NativeHistogramBucketFactor:     1.1,
-			NativeHistogramZeroThreshold:    0,
 			NativeHistogramMaxBucketNumber:  50,
 			NativeHistogramMinResetDuration: time.Hour,
-			NativeHistogramMaxZeroThreshold: 0,
 		}),
 
 		logStoreTimeout: prometheus.NewCounter(prometheus.CounterOpts{


aleks-p:
@cursor push 136b4e0839

pkg/metastore/raftnode/node_metrics.go
```diff
@@ -45,12 +47,30 @@ func newMetrics(reg prometheus.Registerer) *metrics {
 			},
 			[]string{"state"},
 		),
+
+		logStoreWrite: prometheus.NewHistogram(prometheus.HistogramOpts{
+			Name:                            "raft_log_store_write_duration_seconds",
+			Help:                            "Duration of log store write operations",
+			Buckets:                         prometheus.DefBuckets,
+			NativeHistogramBucketFactor:     1.1,
+			NativeHistogramZeroThreshold:    0,
```
simonswine: Nit: Not too sure what this does, but I can see it's copy pasted from the other histograms, likely the default any how. (same for `NativeHistogramMaxZeroThreshold`)

pkg/metastore/test/metastore_disk_failure_test.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/oklog/ulid/v2"
+	"github.com/stretchr/testify/require"
+
+	metastorev1 "github.com/grafana/pyroscope/api/gen/proto/go/metastore/v1"
+	"github.com/grafana/pyroscope/pkg/metastore"
+	"github.com/grafana/pyroscope/pkg/metastore/raftnode/raftnodepb"
+	"github.com/grafana/pyroscope/pkg/objstore/providers/memory"
+)
+
+// TestMetastoreDiskFailure_ClusterRecovery verifies that a full Pyroscope
```
simonswine, line 22: Feels quite slow until leader is elected, than another 2s for the timeout.

```
ok      github.com/grafana/pyroscope/pkg/metastore/test 9.162s
ok      github.com/grafana/pyroscope/pkg/metastore/test 7.875s
ok      github.com/grafana/pyroscope/pkg/metastore/test 11.405s
ok      github.com/grafana/pyroscope/pkg/metastore/test 26.278s
ok      github.com/grafana/pyroscope/pkg/metastore/test 7.574s
```

I don't think there is a easy way, maybe the raft election can be faster or preloaded or with some tweaks of the parameters.

aleks-p: We can bring it down substantially by tweaking the raft config (`HeartbeatTimeout`, `ElectionTimeout`, `LeaderLeaseTimeout`) for the test. The raft library does this as well:

https://github.com/hashicorp/raft/blob/10dcdbf1b7946ec5a4c25b38023e5a64b38a2f63/raft-compat/testcluster/cluster.go#L172-L174

Before
```
  - Run 1: 15.39s
  - Run 2: 5.64s
  - Run 3: 5.73s                                       
  - Run 4: 19.03s
  - Run 5: 15.34s
  - Run 6: 5.32s
  - Run 7: 9.91s
  - Run 8: 22.06s
```

After setting those I get `3–7s` consistently.

The downside is that we need to expose these config fields in our wrapper, just for config purposes. Probably worth adding those fields to the config as undocumented, what do you think?

pkg/metastore/raftnode/node_metrics.go
aleks-p: copy-pasted indeed, removed them from this and the other places in this file for consistency

pkg/metastore/test/metastore_disk_failure_test.go
aleks-p: Did a more comprehensive test (100 runs each way), I was wrong and the difference is smaller, we can probably skip this optimization for now

```
  Without Raft timing overrides (n=66):

  ┌────────┬────────┐
  │        │        │
  ├────────┼────────┤
  │ min    │ 4.85s  │
  ├────────┼────────┤
  │ mean   │ 9.02s  │
  ├────────┼────────┤
  │ median │ 6.43s  │
  ├────────┼────────┤
  │ p90    │ 16.84s │
  ├────────┼────────┤
  │ p95    │ 22.05s │
  ├────────┼────────┤
  │ max    │ 26.48s │
  ├────────┼────────┤
  │ stdev  │ 5.19s  │
  └────────┴────────┘

  With Raft timing overrides (n=86, 1 failure):

  ┌────────┬───────────────┐
  │        │               │
  ├────────┼───────────────┤
  │ min    │ 2.78s         │
  ├────────┼───────────────┤
  │ mean   │ 6.96s         │
  ├────────┼───────────────┤
  │ median │ 6.79s         │
  ├────────┼───────────────┤
  │ p90    │ 13.28s        │
  ├────────┼───────────────┤
  │ p95    │ 13.47s        │
  ├────────┼───────────────┤
  │ max    │ 32.27s (FAIL) │
  ├────────┼───────────────┤
  │ stdev  │ 4.55s         │
  └────────┴───────────────┘
```

simonswine: Thanks for investigating! At worst we need to run them separately one day when it all takes too long :)

PR #4648, opened 2025-11-28
feat: add debug info upload service

korniltsev-grafanista:
## Summary

Add a new `DebuginfoService` that allows profiling agents (e.g., the OTel eBPF profiler) to upload debug information (ELF executables) to Pyroscope's object storage. The symbolizer then uses these uploaded files as an additional source when resolving symbols, falling back to debuginfod if no upload is available.

This is an ALPHA state implementation for internal use and testing. Breaking changes will follow.

### What's included

- **New proto API** (`debuginfo.v1alpha1`): Bidirectional streaming `Upload` RPC using Connect. Clients send a `ShouldInitiateUploadRequest`, receive a response indicating whether to proceed, then stream `UploadChunk` messages. The API is marked `v1alpha1` to signal experimental status.
- **Debug info store** (`pkg/debuginfo/store.go`): Handles upload lifecycle — validates GNU build IDs, checks for duplicate/in-progress/stale uploads via `ObjectMetadata` persisted as JSON in object storage, streams the uploaded binary to the bucket.
- **Upload reader** (`pkg/debuginfo/reader/`): An `io.Reader` adapter that assembles streamed chunks into a contiguous reader for the object storage upload.
- **Symbolizer integration**: The symbolizer now checks the debug info upload bucket (keyed by tenant + GNU build ID) before falling back to debuginfod. The symbolizer bucket was changed from a prefixed bucket to the shared storage bucket, with explicit path prefixes (`symbolizer/`, `debug-info/`).
- **API wiring**: The service is registered on the distributor's HTTP handler with auth and a 5MB per-message limit.

### Key design decisions

- **Own proto, not Parca's**: Inspired by Parca's debug info upload but uses Pyroscope's own protobuf definitions with Connect-Go handlers instead of gRPC.
- **Metadata in object storage**: Upload state (`UPLOADING`/`UPLOADED`) is tracked via a JSON metadata file alongside the binary in the bucket, avoiding the need for an additional database.
- **Stale upload recovery**: If a previous upload started but never finished and exceeds `MaxUploadDuration + 2min`, it's considered stale and can be retried.

## Test plan

- [x] Unit tests for `Store` upload flow, metadata lifecycle, and edge cases (`pkg/debuginfo/store_test.go`)
- [x] Unit tests for `UploadReader` chunk assembly (`pkg/debuginfo/reader/reader_test.go`)
- [x] Symbolizer tests for the uploaded debug info fetch path (`pkg/symbolizer/symbolizer_test.go`)
[...]

PR #4895, opened 2026-03-13
lidia: tolerate elf.ErrNoSymbols when both symbol tables are missing

korniltsev-grafanista-yolo-vibecoder239:
## Summary

Tolerate `elf.ErrNoSymbols` in `CreateLidiaFromELF` when both `elfFile.Symbols()` and `elfFile.DynamicSymbols()` fail.

Previously, if both calls returned errors, the function would unconditionally return a hard error. This is incorrect for stripped ELF binaries that have no symbol tables at all — `elf.ErrNoSymbols` is expected in that case, not a fatal condition.

Now, the two errors are wrapped into a combined error using `fmt.Errorf` with two `%w` verbs, and `errors.Is` checks whether the combined error wraps `elf.ErrNoSymbols`. If it does, processing continues; otherwise, the error is returned as before.

## Context

Extracted from #4648 per [review comment](https://github.com/grafana/pyroscope/pull/4648/files#diff-2a29d4c7082fc45c52bbc42fac4141d5f392968e977802c9da2508facb8dd6c8R179) requesting this change be submitted separately.

## Changes

- `lidia/lidia.go`: Added `"errors"` import; changed the error path in `CreateLidiaFromELF` to check `errors.Is(err, elf.ErrNoSymbols)` before returning when both symbol table reads fail.

---

> [!NOTE]
> **Low Risk**
> Small, localized change to ELF symbol error handling; main risk is unintentionally suppressing unexpected symbol-read errors, mitigated by only tolerating `elf.ErrNoSymbols` on both calls.
> 
> **Overview**
> `CreateLidiaFromELF` now treats the case where both `elfFile.Symbols()` and `elfFile.DynamicSymbols()` return `elf.ErrNoSymbols` as non-fatal, allowing lidia generation to continue for stripped ELF binaries.
> 
> All other symbol-read failures still return an error; this change adds an `errors` import and gates the existing combined error return on `errors.Is(..., elf.ErrNoSymbols)` checks.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 221da0f0933a5b22312d443f690f1a88da9144ed. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lidia/lidia.go
+++ b/lidia/lidia.go
@@ -176,9 +176,8 @@ func CreateLidiaFromELF(elfFile *elf.File, output io.WriteSeeker, opts ...Option
 		if symErr != nil {
 			symbols, dynSymErr = elfFile.DynamicSymbols()
 			if dynSymErr != nil {
-				err := fmt.Errorf("failed to read symbols from ELF file: %w, %w", symErr, dynSymErr)
-				if !errors.Is(err, elf.ErrNoSymbols) {
-					return err
+				if !errors.Is(symErr, elf.ErrNoSymbols) || !errors.Is(dynSymErr, elf.ErrNoSymbols) {
+					return fmt.Errorf("failed to read symbols from ELF file: %w, %w", symErr, dynSymErr)
 				}
 			}
 		}


korniltsev-grafanista:
@codex review

PR #4995, opened 2026-04-07
feat(experimental): support async query execution

aleks-p:
@bryanhuhta  Reopening this PR to follow up on testing this implementation further.

I've tested the latest commit in dev and it seems to work as expected. This feature is gated under the flag:

```
-query-frontend.async-queries-enabled="true"
```

I've somewhat modified this POC to create a discrete Async query path, where clients must explicitly call the Async path to get async query support. This isn't a permanent decision and in the long term we likely would want to consider leveraging the capabilities api for clients to signal they support async queries. Then clients can continue to use the existing `QuerierService` API.


cmd/profilecli/query-top.go
```diff
@@ -34,7 +35,7 @@ func addQueryTopParams(queryCmd commander) *queryTopParams {
 	return params
 }
 
-func queryTop(ctx context.Context, params *queryTopParams) error {
+func queryTop(ctx context.Context, params *queryTopParams, async bool) error {
```
aleks-p: I think this came from the POC itself, but should we fold `async` into the shared `queryParams` (and use across sub-commands)?

cmd/profilecli/query.go
```diff
@@ -150,22 +162,31 @@ func queryProfile(ctx context.Context, params *queryProfileParams, outputFlag st
 	}
 
 	var profile *googlev1.Profile
-
-	if len(params.SpanSelector) > 0 {
+	if async {
+		if len(params.SpanSelector) > 0 {
+			return fmt.Errorf("--async does not support --span-selector")
```
aleks-p: Is this a real limitation? I remember specifically not covering the v1 query path in the POC, but we should in theory support any v2 operation. 

Somewhat related https://github.com/grafana/pyroscope/pull/5273, which actually implemented span-selector support in v2 for the missing queries.

cmd/profilecli/query_async.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	resp, err := runAsyncQuery(ctx, client, req)
+	if err != nil {
+		return nil, err
+	}
+	if len(resp.Reports) == 0 {
+		return nil, nil
+	}
+	if resp.Reports[0].SeriesLabels == nil {
+		return nil, fmt.Errorf("unexpected report type: expected series_labels")
```
aleks-p: nit, this sounds like we found a report type that is not the one we were expecting, we could say "missing report of type series_labels in response" or similar

pkg/frontend/async/coordinator.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	c.inFlight[tenantID]++
+	c.asyncQueriesCurrent.WithLabelValues(tenantID).Set(float64(c.inFlight[tenantID]))
+	return nil
+}
+
+// PromoteToAsync promotes an already-running query to async execution.
+// It takes ownership of the result channel: the coordinator will read from it,
+// store the result, and manage the heartbeat loop. Returns the request ID.
+func (c *Coordinator) PromoteToAsync(ctx context.Context, tenantID string, resultCh <-chan QueryResult) (string, error) {
```
aleks-p: Is this name and description still accurate? It might be from the POC concept of promoting long queries to async, which was stripped I think.

pkg/frontend/async/store.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+	if err := s.bucket.Upload(ctx, base+"result.pb", bytes.NewReader(data)); err != nil {
+		return fmt.Errorf("failed to upload result: %w", err)
+	}
+
+	meta := &Metadata{
+		RequestID: requestID,
+		TenantID:  tenantID,
+		Status:    StatusSuccess,
+		CreatedAt: time.Now().UTC(),
```
aleks-p: this seems like a bug (from the POC), not sure what the field is used for but we are effectively overriding the original CreatedAt (same in Fail below)

aleks-p (commented):
I see no major hurdles to move this forward in an experimental state. I think we are lacking test coverage beyond the integration test, but this stems from the POC itself. 

Some other random things I noticed (all from the POC itself)
- there is no upper bound for how much time we execute an async query, maybe we need something and set it high (we can also do this separately)
- it looks like we submit queries before we check the concurrency limit, the client will get an error but the query will be executed anyway

pkg/pyroscope/modules_experimental.go
```diff
@@ -147,15 +150,18 @@ func (f *Pyroscope) initQueryFrontendV12() (services.Service, error) {
...
 		f.reg,
 	)
 
 	f.API.RegisterFrontendForQuerierHandler(f.frontend)
-	f.API.RegisterQuerierServiceHandler(spanlogger.NewLogSpanParametersWrapper(handler, queryFrontendLogger))
-	f.API.RegisterPyroscopeHandlers(spanlogger.NewLogSpanParametersWrapper(handler, queryFrontendLogger))
+	f.API.RegisterQuerierServiceHandler(wrappedHandler)
+	f.API.RegisterPyroscopeHandlers(wrappedHandler)
 	f.API.RegisterVCSServiceHandler(vcsService)
+	f.registerQueryFrontendServiceHandler()
```
simonswine: I think we should not register it here, as far as I can tell in the code we don't support a mode split for query-backend/queriers and then the result will differ depending if people use async or not

pkg/frontend/async/handler.go
```diff
@@ -0,0 +1,123 @@
...
+	resultCh := make(chan QueryResult, 1)
+	go func() {
+		resp, err := h.queryFn(queryCtx, req)
+		resultCh <- QueryResult{Response: resp, Err: err}
+	}()
+
+	requestID, err := h.coordinator.PromoteToAsync(ctx, tenantID, resultCh)
+	if err != nil {
+		return nil, connect.NewError(connect.CodeResourceExhausted, err)
+	}
```
simonswine, line 83: I guess Aleks already mention this but the `PromoteToAsync` needs to happen before we actually dispatch the query in go routine.

simonswine (dismissed):
I think this is ready to be merged and tested beyond feature flag.

My main worry would be to enable the `queryv1` API on cloud-backend-gateways, as it  would be exposing quite a lot of surface currently hidden behind the query-frontend.

I think it would better (also for ease of adoption) to incorporate that into the querierv1 API.

cmd/profilecli/query-top.go
bryanhuhta: Probably something we should do long term. I have this issue https://github.com/grafana/pyroscope-squad/issues/835 which tracks clean up work for profilecli. Generally doing all the necessary work to make it a good reference for other clients (e.g. gcx). I think we should defer that work to there since this PR is already really beefy.

cmd/profilecli/query_async.go
bryanhuhta: No longer relevant now that I moved the async RPC to the QuerierService, but yes you're right 🙂

pkg/frontend/async/coordinator.go
bryanhuhta: Fixed in https://github.com/grafana/pyroscope/pull/4995/commits/22802ca9c5859dded46dd2a6152e9c32339c24ba. Changed to `Register`, as in "register an async query to be executed".

pkg/frontend/async/store.go
bryanhuhta: Fixed in https://github.com/grafana/pyroscope/pull/4995/commits/22802ca9c5859dded46dd2a6152e9c32339c24ba

pkg/frontend/async/handler.go
bryanhuhta: In https://github.com/grafana/pyroscope/pull/4995/commits/22802ca9c5859dded46dd2a6152e9c32339c24ba I changed to check if there's budget to dispatch another query before starting the goroutine (in `pkg/frontend/async/handler.go`.

```go
queryCtx := pyrotenant.InjectTenantID(context.Background(), tenantID)
resultCh := make(chan QueryResult, 1)

dispatch, err := h.dispatcherFor(queryCtx, req, resultCh)
if err != nil {
    return nil, err
}

// Reserve the concurrency slot before dispatching so a rejected submit
// never starts background work.
requestID, err := h.coordinator.Register(ctx, tenantID, resultCh)
if err != nil {
    return nil, connect.NewError(connect.CodeResourceExhausted, err)
}

go dispatch()
```

cmd/profilecli/query.go
bryanhuhta: No, this isn't a limitation. I don't think my agent (or me) was watching super closely when implementing this part.

Added in https://github.com/grafana/pyroscope/pull/4995/commits/3290cbce4c798af773679a1d585d3db1576155ac.

bryanhuhta:
> I think it would better (also for ease of adoption) to incorporate that into the querierv1 API.

Valid point. In https://github.com/grafana/pyroscope/pull/4995/commits/22802ca9c5859dded46dd2a6152e9c32339c24ba, I switched the API definition to put it on the QuerierService service.

```proto
// AsyncQuery submits or polls a long-running query. Experimental, gated by
// -query-frontend.async-queries-enabled.
rpc AsyncQuery(AsyncQueryRequest) returns (AsyncQueryResponse) {
  option (gnostic.openapi.v3.operation).tags = "scope/internal";
}

// AsyncQueryRequest is dual-purpose: leave request_id empty and set one of
// the query variants to submit a new query; set request_id to poll an
// in-flight query (query variants are ignored in that case).
message AsyncQueryRequest {
  string request_id = 1;
  oneof query {
    SelectMergeStacktracesRequest select_merge_stacktraces = 2;
    SelectMergeSpanProfileRequest select_merge_span_profile = 3;
  }
}

message AsyncQueryResponse {
  string request_id = 1;
  AsyncQueryStatus status = 2;
  // Populated on FAILURE.
  string error_message = 3;
  // Populated on SUCCESS. The variant matches the submission's query oneof.
  oneof result {
    SelectMergeStacktracesResponse select_merge_stacktraces = 4;
    SelectMergeSpanProfileResponse select_merge_span_profile = 5;
  }
}

enum AsyncQueryStatus {
  ASYNC_QUERY_STATUS_UNSPECIFIED = 0;
  ASYNC_QUERY_STATUS_IN_PROGRESS = 1;
  ASYNC_QUERY_STATUS_SUCCESS = 2;
  ASYNC_QUERY_STATUS_FAILURE = 3;
}

```

Not sure this is the final API contract we want long term, but it's a good starting point. We will need to wire up the `AsyncQuery` RPC with the cortex gateway, but this way we convert any existing sync RPC into an async RPC.

api/querier/v1/querier.proto
```diff
@@ -136,6 +141,36 @@ message SelectMergeStacktracesResponse {
   string dot = 3;
 }
 
+// AsyncQueryRequest is dual-purpose: leave request_id empty and set one of
+// the query variants to submit a new query; set request_id to poll an
+// in-flight query (query variants are ignored in that case).
+message AsyncQueryRequest {
+  string request_id = 1;
+  oneof query {
```
simonswine: Warning: I can't remember exactly what broke, but I think there are some intricacies when using `oneof` and the json marshalling/unmashalling.

api/querier/v1/querier.proto
```diff
@@ -67,6 +67,11 @@ service QuerierService {
   rpc AnalyzeQuery(AnalyzeQueryRequest) returns (AnalyzeQueryResponse) {
     option (gnostic.openapi.v3.operation).tags = "scope/internal";
   }
+  // AsyncQuery submits or polls a long-running query. Experimental, gated by
+  // -query-frontend.async-queries-enabled.
+  rpc AsyncQuery(AsyncQueryRequest) returns (AsyncQueryResponse) {
+    option (gnostic.openapi.v3.operation).tags = "scope/internal";
+  }
 }
```
simonswine: Note: this is not blocking, but I find the following more elegant, than a separate call:

- Add an optional `AsyncQueryRequest` field to `SelectMergeStacktracesRequest` and `SelectMergeSpanProfileRequest`
- Add an optional `AsyncQueryStatus` field to `SelectMergeStacktracesResponse` and `SelectMergeSpanProfileResponse`

This would make sure that currently attribution e.g. in metrics/logs is automatically counting it as the particular request. You would also not have to introduce any gateway change.

In order to signal support, you expect the client to set the message and `request_id`. Then you know they support the handling of the `AsyncQueryStatus` in the response.

The only one that is tricky is SelectMergeProfile as it replies with the upstream proto, but we can and we should integrate that into `SelectMergeStacktraces`.

cmd/profilecli/query_async.go
```diff
@@ -47,120 +55,57 @@ func runAsyncQuery(ctx context.Context, client queryv1connect.QueryFrontendServi
...
+		case querierv1.AsyncQueryStatus_ASYNC_QUERY_STATUS_IN_PROGRESS:
 			level.Info(logger).Log("msg", "waiting for async query", "request_id", poll.Msg.RequestId, "elapsed", time.Since(start).Truncate(time.Second))
 			continue
-		case queryv1.AsyncQueryStatus_ASYNC_QUERY_STATUS_SUCCESS:
+		case querierv1.AsyncQueryStatus_ASYNC_QUERY_STATUS_SUCCESS:
 			level.Info(logger).Log("msg", "async query completed", "request_id", poll.Msg.RequestId, "elapsed", time.Since(start).Truncate(time.Second))
-			return poll.Msg.Response, nil
-		case queryv1.AsyncQueryStatus_ASYNC_QUERY_STATUS_FAILURE:
+			result, ok := poll.Msg.Result.(*querierv1.AsyncQueryResponse_SelectMergeStacktraces)
+			if !ok || result.SelectMergeStacktraces == nil {
```
aleks-p: In the current read path, is an empty result an error or valid state?

simonswine (dismissed):
No hard no for me, I wonder if you want to reconsider the API choice in `querierv1`, as per my comment. 

Otherwise LGTM

api/querier/v1/querier.proto
bryanhuhta: Yeah, this makes sense. It could look something like this:

```proto
message SelectMergeStacktracesRequest {
  // old fields ...

  optional AsyncQueryRequest async = 9;
}

message SelectMergeStacktracesResponse {
  FlameGraph flamegraph = 1;
  // Pyroscope tree bytes.
  bytes tree = 2;
  // DOT graph representation.
  string dot = 3;

  optional AsyncQueryResponse async = 4;
}

message AsyncQueryRequest {
  // If set, this is a polling request.
  string request_id = 1;
  // If set, this is an async query.
  bool async = 2;
}

message AsyncQueryResponse {
  // Id of the async query.
  string request_id = 1;
  // Status of the query: unknown, in_progress, success, failed
  AsyncQueryStatus status = 2;
  // Populated on FAILURE.
  string error_message = 3;
}
```

This of course means we need to explicitly enable async queries for each RPC individually. We can focus on `/SelectMergeStacktraces` and `/SelectMergeSpanProfile` at the start. The implementation I [outlined earlier](https://github.com/grafana/pyroscope/pull/4995#issuecomment-4796500956) would get us async variants of all query RPCs. The drawback is the `oneof` marshal/unmarshal idiosyncrasies that you mentioned though.

aleks-p: I have mixed feelings about this.

I specifically went for a (somewhat) new RPC service in the POC because of `SelectMergeProfile` having a fixed response type that we can't control and to avoid having partial async support (going for all-or-none).

Sunsetting it altogether in favor of `SelectMergeStacktraces` makes sense, but I wonder if this "deprecation" and the fact that async support is a relatively major and structural API change warrants a **v2** (`querier.v2.QuerierService`), while keeping the v1 service as is.

> This would make sure that currently attribution e.g. in metrics/logs is automatically counting it as the particular request.

While that is true, wouldn't that also reduce visibility of async queries without changes to the metrics and logs?


I support either approach, but wanted to put forward v2 as an option. It could be more disruptive but at the same time could end up being cleaner. In a v2 world we could even merge `SelectMergeStacktracesRequest` with `SelectMergeSpanProfileRequest`, I always found this separation to be confusing.

cmd/profilecli/query_async.go
bryanhuhta: Empty results shouldn't be an error (they aren't in the current read path). The latest commits changed this erroneous behavior through refactoring.

bryanhuhta:
Final consensus for this PR after speaking with @simonswine and @aleks-p:

* We will use add fields in the `SelectMergeStacktraces` RPC to toggle async queries
* Any additional RPCs that would benefit from being async are required to get the relevant async fields added to them

A rough outline of what this looks like:

```proto
message SelectMergeStacktracesRequest {
  // old fields ...

  optional AsyncQueryRequest async = 9;
}

message SelectMergeStacktracesResponse {
  FlameGraph flamegraph = 1;
  // Pyroscope tree bytes.
  bytes tree = 2;
  // DOT graph representation.
  string dot = 3;

  optional AsyncQueryResponse async = 4;
}

message AsyncQueryRequest {
  // If set, this is a polling request.
  string request_id = 1;
  // Sets the kind of async query.
  AsyncQueryType type = 2;
}

message AsyncQueryResponse {
  // Id of the async query.
  string request_id = 1;
  // Status of the query: unknown, in_progress, success, failed
  AsyncQueryStatus status = 2;
  // Populated on FAILURE.
  string error_message = 3;
}

enum AsyncQueryType {
  // The query should not be async.
  ASYNC_QUERY_TYPE_DISABLED = 0;
  // The query must be async.
  ASYNC_QUERY_TYPE_FORCE = 1;
}

enum AsyncQueryStatus {
  ASYNC_QUERY_STATUS_UNKNOWN = 0;
  ASYNC_QUERY_STATUS_IN_PROGRESS = 1;
  ASYNC_QUERY_STATUS_SUCCESS = 2;
  ASYNC_QUERY_STATUS_FAILURE = 3;
}k
```

In the long term, we want to consider creating `querier.v2` to shrink and update the RPCs to query profiling data. There are several that are only minor variations of the same thing:

* `rpc SelectMergeStacktraces(SelectMergeStacktracesRequest) returns (SelectMergeStacktracesResponse)`
* `rpc SelectMergeSpanProfile(SelectMergeSpanProfileRequest) returns (SelectMergeSpanProfileResponse)`
* `rpc SelectMergeProfile(SelectMergeProfileRequest) returns (google.v1.Profile)`

That is out of scope of this work stream to implement, but important context nonetheless as it motivates some of the rationale for taking the approach of adding async fields directly to the `SelectMergeStacktraces` RPC fields.

cmd/profilecli/query.go
```diff
@@ -150,22 +150,30 @@ func queryProfile(ctx context.Context, params *queryProfileParams, outputFlag st
 	}
 
 	var profile *googlev1.Profile
-
-	if len(params.SpanSelector) > 0 {
+	if async {
+		if len(params.SpanSelector) > 0 {
+			return errors.New("--async is not supported with --span-selector (only SelectMergeStacktraces queries can run async)")
```
bryanhuhta, line 155: It was explicitly decided to only support `SelectMergeStacktraces` so this error is accurate.

bryanhuhta (dismissed):

bryanhuhta (approved):

PR #5041, opened 2026-04-09
fix(ui): Align time series line graph with actual data points

bryanhuhta:
Before this change the time series data points would render evenly across the time series graph window, regardless of whether the data point aligned with the x-axis time value.

This PR makes sure that data points are rendered above the correct time marker tick on the x-axis.

---

> [!NOTE]
> **Medium Risk**
> Changes the timeline API contract from `number[]` to timestamped `Point[]`, which touches query state and chart rendering and could break downstream consumers if any were missed. Adds a new test runner and CI job, which is low functional risk but may affect pipeline stability/time.
> 
> **Overview**
> Fixes the time series chart so points are positioned on the x-axis using their actual `timestamp` instead of being evenly distributed across the width, updating the area/line paths accordingly.
> 
> Updates `fetchTimeline`/`usePyroscopeQuery`/`TimeSeries` to pass full `{ value, timestamp }` points through the UI rather than mapping to just values.
> 
> Introduces `vitest` with a new `yarn test` script, moves time-series helper functions into `components/timeseries.ts` with unit tests, and adds a `test` job to the `frontend` GitHub Actions workflow.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 697753d798986723a0babfd72dd9cd9a8334d332. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -106,6 +106,27 @@ jobs:
           path: ui
       - name: Run lint
         run: yarn lint
+  test:
+    needs: install
+    runs-on: ${{ github.repository_owner == 'grafana' && 'ubuntu-x64' || 'ubuntu-latest' }}
+    defaults:
+      run:
+        working-directory: ui
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
+        with:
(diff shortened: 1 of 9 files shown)


aleks-p (approved):
LGTM

PR #5043, opened 2026-04-10
[dont merge] V2 by default, V1 opt-in

alsoba13:
[WIP]
DO NOT MERGE. This is part of the v2.0.0 release

# Description
This PR addresses two main aspects:
* the usage for pyroscope >=v2.0.0 in terms of default / flags
* helm chart migration/breaking changes

## v2 defaults and new flags
We are basically negating all the v1-v2 flags:
* The idea is to consider v2 the default. For that, we remove the config `flags.V2` which used to meant "we are either fully v2 or migrating to v2" to use `flags.V1` which reads as "We are running V1 ONLY (no v2, no migrating to v2)".
* Any user that wants to use >=v2.0.0 without migrating will need to set that flag by using the env var `PYROSCOPE_V1_LEGACY`. Otherwise, pyroscope will default to v2 storage architecture.
* The intermediate states such "migrating to v2", where v1 and v2 components coexisted, is now obtained by NOT setting `PYROSCOPE_V1_LEGACY` and also use the new flags `enable-querier` and `enable-querier-until`, which are the opposite to `enable-query-backend` and `enable-query-backend-from`.

Summary of the above described:
| pyroscope version     | 1.x.x                                                                                  | 2.x.x                                              |
|-----------------------|----------------------------------------------------------------------------------------|----------------------------------------------------|
| Run v1 storage only   | by default                                                                             | Set PYROSCOPE_V1_LEGACY                            |
| Run v1 and v2 storage | Set PYROSCOPE_V2_EXPERIMENT, -enable-query-backend=true and -enable-query-backend-from | Set -enable-querier=true and -enable-querier-until |
| Run v2 storage        | Set PYROSCOPE_V2_EXPERIMENT and -enable-query-backend=true                             | by default                                         |

## helm chart updates
Helm chart users that are moving from 1.x.x to 2.x.x while doing an `upgrade`, will be required to take action:
* either migrate from v1 to v2 (following the migration guide).
* or keep v1 by setting `architecture.storage.v1=true` explicitly, and migrate later (following the exact same guide).
* or acknowledge the breaking changes and YOLO to v2 (skipping migration, hence losing data)
NOTE: the logic above is just done for single binary yet. microservices WIP.

Finally, This is the PR changing the described logic. There's a clean-up PR coming, where we rename files, clean a bit more the code so v2 is the default happy path, and v1 stuff is the exception.


cmd/pyroscope/main_test.go
```diff
@@ -125,7 +125,7 @@ func TestHelp(t *testing.T) {
 
 			expected, err := os.ReadFile(tc.filename)
 			require.NoError(t, err)
-			assert.Equalf(t, string(expected), stdout, "%s %s output changed; try `make reference-help`", cmd, tc.arg)
+			assert.Equalf(t, string(expected), stdout, "%s %s output changed; try `make reference-help`", cmd, tc.arg) // TODO: make reference-help is broken
```
alsoba13, line 128: TODO: something's going on since we removed old ui/assets. I need to look deeper

operations/pyroscope/helm/pyroscope/templates/pre-upgrade-check.yaml
alsoba13, line 1: I guess this is not perfect, but is something.
First, this is only for single-binary (microservices coming next).
If someone is going to upgrade from an older version to v2 and accidentally setting storage v2, then this will fail and prompt them for action.

CONS:
Users at chart version 1.20.3, they can safely do `helm upgrade ... --reuse-values`, with the original default `architecture.storage.v1=true` and `architecture.storage.v2=false`. This will be safe in terms of migration (v1->v1) but they will likely stick in v1. We could set a Notice in NOTES so they get a migration suggestion

operations/pyroscope/helm/pyroscope/rendered/micro-services-legacy-v1.yaml
alsoba13, line 1: This is the old `micro-services.yaml`. Now on v2, that file is for this same specific but with v2 storage by default (users upgrading from v1 to v2 with `micro-services.yaml` should get the pre-upgrade-check as a safety net)

operations/pyroscope/helm/pyroscope/rendered/micro-services-v2.yaml
alsoba13, line 1: This is a copy of new `micro-services.yaml`, for people using `micro-services-v2.yaml` to continue to have it.

operations/pyroscope/helm/pyroscope/rendered/micro-services.yaml
alsoba13, line 1: The new micro-services.yaml uses v2 storage. Any user upgrading with this file should get pre-upgrade-check prompting for error if no v2 migration has been done yet (WIP)

operations/pyroscope/helm/pyroscope/rendered/single-binary-legacy-v1.yaml
alsoba13, line 1: Same as micro-services(.*).yaml but for single binary:
single-binary-legacy-v1.yaml‎ is the old single-binary.yaml‎
single-binary.yaml‎ uses v2 by default
single-binary-v2.yaml‎ is a copy of new single-binary.yaml‎ for backwards compatibility

pkg/frontend/readpath/query_backend_from.go
```diff
@@ -9,21 +9,23 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-// QueryBackendFrom is a flag/config type that accepts either an RFC3339
+// TODO RENAME FILE
```
alsoba13, line 12: TODO: rename this file

There are a lot of other things we need to refactor (get rid of modules_experimental.go or change the code branching so v2 is the default while v1 is the exception). We can do all of these refactors later as these are likely hard to review, and provide few value now.

alsoba13:
Failing tests:
* Integration tests: looking into it now.
* [helm-ci / call-lint-test-pyroscope / lint-test (pull_request)](https://github.com/grafana/pyroscope/actions/runs/24337613667/job/71058277330?pr=5043) -> wont work, pyroscope 2.0.0 is not there, hence helm can't pull it, I guess we will need to do 2 different PR for this and merge the second one once we release 2.0.0

pkg/frontend/readpath/read_path.go
```diff
@@ -5,16 +5,16 @@ import (
...
-	EnableQueryBackend     bool             `yaml:"enable_query_backend" json:"enable_query_backend" doc:"hidden"`
-	EnableQueryBackendFrom QueryBackendFrom `yaml:"enable_query_backend_from" json:"enable_query_backend_from" doc:"hidden"`
-	QueryTreeEnabled       bool             `yaml:"query_tree_enabled" json:"query_tree_enabled" doc:"hidden"`
+	EnableQuerier      bool        `yaml:"enable_querier" json:"enable_querier" doc:"hidden"`
+	EnableQuerierUntil QuerierFrom `yaml:"enable_querier_until" json:"enable_querier_until" doc:"hidden"` // TODO
+	QueryTreeEnabled   bool        `yaml:"query_tree_enabled" json:"query_tree_enabled" doc:"hidden"`
 }
 
 func (o *Config) RegisterFlags(f *flag.FlagSet) {
-	f.BoolVar(&o.EnableQueryBackend, "enable-query-backend", false,
```
simonswine, line 14: Not really sure why we are renaming this. Shouldn't it be enough to flip the default from `false` to `true`?

pkg/frontend/readpath/read_path.go
```diff
@@ -5,16 +5,16 @@ import (
...
-		"This parameter specifies whether the new query backend is enabled.")
-	f.Var(&o.EnableQueryBackendFrom, "enable-query-backend-from",
```
simonswine, line 16: Simiar like in L14, this renaming seems to create pain without gain

Makefile
```diff
@@ -433,23 +433,25 @@ helm/check: $(BIN)/kubeconform $(BIN)/helm
...
+	$(BIN)/helm template -n default --kube-version "1.23.0" pyroscope-dev ./operations/pyroscope/helm/pyroscope/ --values operations/pyroscope/helm/pyroscope/values-micro-services.yaml $(HELM_FLAGS_V1) \
 		| tee ./operations/pyroscope/helm/pyroscope/rendered/legacy-micro-services.yaml \
 		| $(BIN)/kubeconform --summary --strict --kubernetes-version 1.23.0
-	$(BIN)/helm template -n default --kube-version "1.23.0" pyroscope-dev ./operations/pyroscope/helm/pyroscope/ --values operations/pyroscope/helm/pyroscope/values-micro-services-hpa.yaml \
+	$(BIN)/helm template -n default --kube-version "1.23.0" pyroscope-dev ./operations/pyroscope/helm/pyroscope/ --values operations/pyroscope/helm/pyroscope/values-micro-services-hpa.yaml $(HELM_FLAGS_V1) \
 		| tee ./operations/pyroscope/helm/pyroscope/rendered/legacy-micro-services-hpa.yaml \
 		| $(BIN)/kubeconform --summary --strict --kubernetes-version 1.23.0
+	cp ./operations/pyroscope/helm/pyroscope/rendered/micro-services.yaml ./operations/pyroscope/helm/pyroscope/rendered/micro-services-v2.yaml
+	cp ./operations/pyroscope/helm/pyroscope/rendered/single-binary.yaml ./operations/pyroscope/helm/pyroscope/rendered/single-binary-v2.yaml
```
simonswine, line 454: Not quite sure why we need to rename all the output files, I think this is a cleanup task after we have release v2.0

Makefile
```diff
@@ -433,23 +433,25 @@ helm/check: $(BIN)/kubeconform $(BIN)/helm
...
-		| tee ./operations/pyroscope/helm/pyroscope/rendered/micro-services-v2.yaml \
+		| tee ./operations/pyroscope/helm/pyroscope/rendered/micro-services.yaml \
 		| $(BIN)/kubeconform --summary --strict --kubernetes-version 1.23.0
-	$(BIN)/helm template -n default --kube-version "1.23.0" pyroscope-dev ./operations/pyroscope/helm/pyroscope/ --values operations/pyroscope/helm/pyroscope/values-micro-services.yaml \
+	$(BIN)/helm template -n default --kube-version "1.23.0" pyroscope-dev ./operations/pyroscope/helm/pyroscope/ --values operations/pyroscope/helm/pyroscope/values-micro-services.yaml $(HELM_FLAGS_V1) \
 		| tee ./operations/pyroscope/helm/pyroscope/rendered/legacy-micro-services.yaml \
 		| $(BIN)/kubeconform --summary --strict --kubernetes-version 1.23.0
-	$(BIN)/helm template -n default --kube-version "1.23.0" pyroscope-dev ./operations/pyroscope/helm/pyroscope/ --values operations/pyroscope/helm/pyroscope/values-micro-services-hpa.yaml \
+	$(BIN)/helm template -n default --kube-version "1.23.0" pyroscope-dev ./operations/pyroscope/helm/pyroscope/ --values operations/pyroscope/helm/pyroscope/values-micro-services-hpa.yaml $(HELM_FLAGS_V1) \
```
simonswine, line 450: Ideally we tests the HPA against v2, but not critical for v2 release

cmd/pyroscope/main.go
```diff
@@ -55,7 +55,7 @@ func errorHandler() {
 func main() {
 	var flags mainFlags
 
-	flags.V2 = os.Getenv("PYROSCOPE_V2") != "" || os.Getenv("PYROSCOPE_V2_EXPERIMENT") != ""
+	flags.V1 = os.Getenv("PYROSCOPE_V1_LEGACY") != ""
```
simonswine, line 58: In an ideal world we would get rid of this environment variable: `PYROSCOPE_V1_LEGACY` and all the related branching. 

I am a bit unsure but what difference those it make when running with we should achieve exactly the same?

```
-enable-query-backend=false
-write-path=ingester
```

simonswine (commented):
Genearlly that goes in the right direction. I think the biggest ask would be to review the environment variable.

Also as you pointed out separately this PR should target #5038

pkg/frontend/readpath/read_path.go
alsoba13: In v2 pyroscope "query backend" is always enabled (unless the V1-only flag is enabled), and having a enable-query-backend flag feels redundant/error prone. In v2, What needs to be enabled/disabled is the querier (for the case users are doing v1->v2 migrations), that fundamentally the reason of this refactor.

Using `enable-query-backend=true` as a default is redundant in this case, hence we can rid of it, but we still need to tell the service to also run querier somehow, that's why I introduced the new flag.

I guess this is tricky to understand because the logic changes slightly (refer to table in the PR description): 
* PYROSCOPE_V1_LEGACY is for v1-only (while PYROSCOPE_V2_EXPERIMENT was for v2-only AND v1+v2 states). 

The alternative:
* We could use PYROSCOPE_V1_LEGACY for v1-only AND v1+v2 states, and keep the query backend flags.

The first option was flipping the flag logic while keeping the code almost exactly. The second one was negating the env var flag but with more changes in code. I can draft the second one and see how it looks, but still feels weird to have the query backend flag around instead of opt-in v1 stuff if needed.

Makefile
alsoba13: I want to achieve 2 things:
1. `single-binary.yaml` and `micro-services.yaml` defaults to new v2 architecture (for fresh installs of users following old guides).
2. backwards support for users already using `single-binary-v2.yaml` and `micro-services-v2.yaml` templates

But honestly, I'm not sure how people consume these templates, and I might be assuming something wrong about how there are actually used

cmd/pyroscope/main.go
alsoba13: Good point, I haven't paid enough attention to write-path flags

pkg/frontend/readpath/read_path.go
aleks-p: I think `enable-querier` could create more confusion for folks not entirely aware of our naming, because a "querier" is a very obvious thing that you want to enable. Calling it `enable-legacy-querier` could help, but considering many people are running v2 already we'd be introducing a breaking change which we can avoid.

I think keeping the flag as it was but flipping the defaults will be slightly smoother, and in v3 we can aim to remove the flags entirely (or have the app emit "warning" messages for one release).

operations/pyroscope/helm/pyroscope/templates/_helpers.tpl
```diff
@@ -167,3 +167,28 @@ behavior:
...
+{{-       (index $distributor.spec.template.spec.containers 0).image -}}
+{{-     end -}}
+{{-   end -}}
+{{- end -}}
+
+{{- define "pyroscope.installedStorageVersionPyroscopeV1" -}}
+{{-   $fullname := (include "pyroscope.fullname" .) }}
+{{-   $sts := (lookup "apps/v1" "StatefulSet" .Release.Namespace "pyroscope") }}
+{{-   $storageVersion := "v1" -}}
+{{-   range (index $sts.spec.template.spec.containers 0).env }}
```
aleks-p, line 188: probably need a guard for `sts` like in detectImage above

operations/pyroscope/helm/pyroscope/templates/_helpers.tpl
```diff
@@ -167,3 +167,28 @@ behavior:
...
+{{-     $distributor := (lookup "apps/v1" "Deployment" .Release.Namespace (printf "%s-distributor" $fullname)) }}
+{{-     if $distributor -}}
+{{-       (index $distributor.spec.template.spec.containers 0).image -}}
+{{-     end -}}
+{{-   end -}}
+{{- end -}}
+
+{{- define "pyroscope.installedStorageVersionPyroscopeV1" -}}
+{{-   $fullname := (include "pyroscope.fullname" .) }}
+{{-   $sts := (lookup "apps/v1" "StatefulSet" .Release.Namespace "pyroscope") }}
```
aleks-p, line 186: should this be `$fullname` instead of `"pyroscope"`?

operations/pyroscope/helm/pyroscope/templates/_helpers.tpl
```diff
@@ -167,3 +167,28 @@ behavior:
...
+{{-   $storageVersion := "v1" -}}
+{{-   range (index $sts.spec.template.spec.containers 0).env }}
+{{-     if and (eq .name "PYROSCOPE_V2_EXPERIMENT") (ne (.value | default "") "") }}
+{{-       $storageVersion = "v2" }}
+{{-     end }}
+{{-   end }}
+{{-   (index $storageVersion) -}}
```
aleks-p, line 193: this seems broken, do we need just `$storageVersion` without the index?

PR #5047, opened 2026-04-13
feat(querier): add DOT format support to SelectMergeStacktraces Connect API

marcsanmi:
### Summary
- Add `PROFILE_FORMAT_DOT` to the `ProfileFormat` enum
- Add `string dot` field to `SelectMergeStacktracesResponse`
- When `format = DOT`, delegate to `SelectMergeProfile` and convert pprof to DOT server-side
- Extract shared DOT conversion to `pkg/frontend/dot/convert.go` (replaces private `pprofToDotProfile` in `http.go`)
- Reject DOT format in v1 querier (v2 only)

### Why
The Grafana Assistant needs to query specific exemplar profiles in DOT format for LLM analysis. Today it uses the legacy `/pyroscope/render?format=dot` HTTP endpoint, which doesn't support `profile_id_selector`. Rather than extending the render endpoint, this adds DOT as a first-class format option on the Connect API, where `profile_id_selector` is already supported.

The assistant can now call `SelectMergeStacktraces` via Connect with `format: DOT` and `profile_id_selector: [uuid]` to get the exact same DOT output it previously got from render — with exemplar filtering.

### Implementation
The DOT path in `selectMergeStacktracesDot` delegates to `SelectMergeProfile`, then converts the result using the shared `dot.FromProfile()`.  The DOT rendering path validates `maxNodes` with tenant limits via `ValidateMaxNodes`, consistent with the tree/flamegraph paths.

### Not included: DOT support for `SelectMergeSpanProfile`
`SelectMergeSpanProfile` also has the `ProfileFormat` field, but adding DOT there requires `span_selector`
support in `PprofQuery`, which doesn't exist today. This can be added when span exemplar support lands in the assistant.

### Related
- A couple of PRs needed in `grafana/grafana` (pass profile_id in assistant context) and `grafana-assistant-app` (switch from render to Connect)

---

> [!NOTE]
> **Medium Risk**
> Introduces a new response format that adds server-side pprof parsing/rendering and a new response field, which could impact performance/memory and client compatibility if mishandled; v1 paths fail fast as `Unimplemented` to limit blast radius.
> 
> **Overview**
> Adds a new `PROFILE_FORMAT_DOT` option to the `ProfileFormat` enum and extends `SelectMergeStacktracesResponse` with a `dot` string payload, propagating the change through generated protobuf/OpenAPI artifacts and public API docs.
> 
> Implements DOT handling in the v2 read-path `QueryFrontend.SelectMergeStacktraces` by delegating to `SelectMergeProfile` (pprof) and converting server-side to DOT via a new shared helper `pkg/frontend/dot/convert.go`; the legacy HTTP render path is updated to reuse this helper. v1 frontends/querier explicitly reject DOT requests as `Unimplemented`, and a new unit test validates the DOT response shape/content.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c1aa26a77f3e45d298d4307c46d402c85ebc0060. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


simonswine (dismissed):
Not a particular big fan of supporting more functionality in  `/render`.

Do you mind writing up why this is needed. It clearly can't be GRPC or protobuf as connect supports http/json as well. Also the mentioned client already calls Series/LabelValues through the connect API.

I think the only way this would be acceptable, is if we add connect support for the dot format at the same time.

I don't think this is that hard: https://github.com/grafana/pyroscope/blob/15c7dba3bbaa5aa6e7d9b94255084416088988d5/api/querier/v1/querier.proto#L124-L127

marcsanmi:
Thanks for the suggestion. 

> Not a particular big fan of supporting more functionality in /render.

I agree — the goal should be to deprecate it, not extend it.

> I think the only way this would be acceptable, is if we add connect support for the dot format at the same time.

I like this approach. My original reasoning for going through render was to avoid changing the DOT output passed to the LLM, since that might affect analysis quality. The goal of this PR was narrowly scoped to fix https://github.com/grafana/pyroscope-squad/issues/745 and migrating to gRPC felt like a bigger, unrelated change.

However, adding DOT as a `ProfileFormat` option in the Connect API solves both problems: the assistant gets the same DOT output, and we don't need to extend render. I'll move forward with this.

pkg/frontend/readpath/queryfrontend/query_select_merge_stacktraces.go
```diff
@@ -35,6 +40,44 @@ func (q *QueryFrontend) SelectMergeStacktraces(
...
+	// Use the same validated max nodes for DOT rendering as SelectMergeProfile
+	// uses for the pprof fetch. This ensures tenant limits are respected.
+	tenantIDs, err := tenant.TenantIDs(ctx)
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, err)
+	}
+	maxNodes, err := validation.ValidateMaxNodes(q.limits, tenantIDs, c.Msg.GetMaxNodes())
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, err)
+	}
```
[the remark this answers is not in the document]
marcsanmi: `SelectMergeProfile` validates `maxNodes` internally before running the query, so no resources are wasted on invalid input. Our `ValidateMaxNodes` call after is only to obtain the resolved value for DOT rendering.

pkg/frontend/readpath/queryfrontend/query_select_merge_stacktraces.go
```diff
@@ -35,6 +40,44 @@ func (q *QueryFrontend) SelectMergeStacktraces(
 	return connect.NewResponse(&resp), nil
 }
 
+func (q *QueryFrontend) selectMergeStacktracesDot(
+	ctx context.Context,
+	c *connect.Request[querierv1.SelectMergeStacktracesRequest],
+) (*connect.Response[querierv1.SelectMergeStacktracesResponse], error) {
+	pprofResp, err := q.SelectMergeProfile(ctx, connect.NewRequest(&querierv1.SelectMergeProfileRequest{
```
simonswine, line 61: It is a bit frustrating that we need to call `SelectMergeProfiles` for this, because the information we are requiring is mostly in the SelectMergeStacktraces, it is only about the mapping. 

This might be blow the scope of this PR to implement and currently we also do call SelectMergeProfiles from render, but I would like you to create an issue to address this after this merges.

pkg/frontend/readpath/queryfrontend/query_select_merge_stacktraces.go
```diff
@@ -35,6 +40,44 @@ func (q *QueryFrontend) SelectMergeStacktraces(
...
+		ProfileTypeID:      c.Msg.ProfileTypeID,
+		LabelSelector:      c.Msg.LabelSelector,
+		Start:              c.Msg.Start,
+		End:                c.Msg.End,
+		MaxNodes:           c.Msg.MaxNodes,
```
simonswine: I think we need separate handling of max nodes like /render does currently, otherwise it will be incompatible with current calls:

https://github.com/grafana/pyroscope/blob/8dbabce8507ba0015e569540cb6bdf03caddaca1/pkg/querier/http.go#L150-L168

simonswine (commented):
Thanks for looking at this, a couple of comments

pkg/frontend/readpath/queryfrontend/query_select_merge_stacktraces.go
marcsanmi: Done in 2d2d0c209f48a0a8bfea9dcabdd9676af946a6ab.

simonswine (approved):

pkg/frontend/readpath/queryfrontend/query_select_merge_stacktraces.go
simonswine: It is actually also providing the full file path and line number information in the dot format. (I was not aware when I wrote that).

Not too sure what the best course of action is in theory the better fit call wise is in the SelectMergeProfiles instead, as we are actually processing the whole Profile rather than just Stacktraces.

I also don't want to block this feature from rolling out.

marcsanmi: We discussed this offline. We went with `SelectMergeStacktraces` because `SelectMergeProfile` returns `google.v1.Profile` directly — o we can't add a DOT field without changing the return type (breaking all existing callers) or creating yet another new endpoint. `SelectMergeStacktraces` already had the `format` field and its own extensible response message.

Longer term, we agreed all three RPCs (`SelectMergeStacktraces`, `SelectMergeProfile`, `SelectMergeSpanProfile`) could be unified into one (it could addressed during https://github.com/grafana/pyroscope/pull/4995). That would resolve the semantic mismatch naturally.

PR #5048, opened 2026-04-13
monitoring: migrate dashboards to native histograms

simonswine:
## Summary

Converts all classic Prometheus histogram queries to native histogram equivalents across all monitoring dashboards and Helm chart templates.

- Remove `_bucket` suffix and `by (le)` from `histogram_quantile` calls
- Replace `rate(metric_count{...})` with `histogram_count(rate(metric{...}))`
- Replace `rate(metric_sum{...})` with `histogram_sum(rate(metric{...}))`
- Replace `_sum/_count` average ratio with `histogram_avg(sum(rate(metric{...})))`
- Update `values.yaml` `tenantQuery` default

**Dashboards affected:** `operational`, `v2-metastore`, `v2-read-path`, `v2-write-path`

The raw JSON dashboards under `operations/monitoring/dashboards/` are generated via `make helm/check` — only the Helm templates were hand-edited.

## Test plan

- [ ] `make helm/check` passes (already verified locally)
- [ ] Confirm NHCB scraping is enabled before merging (requires `scrape_native_histograms: true` on the Prometheus scrape config for Pyroscope)

---

> [!NOTE]
> **Low Risk**
> Low risk: this only changes Makefile generation steps for rendered dashboards and does not affect runtime code paths, but it can change generated dashboard artifacts and CI outputs.
> 
> **Overview**
> `make helm/check` now generates monitoring dashboards/rules for **native histograms by default**, and additionally renders a **classic Prometheus histogram** dashboard set by templating `pyroscope-monitoring` with `--set dashboards.nativeHistograms=false` and exporting to `operations/monitoring/dashboards-classic-histogram/`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 82cbf686996f33d9e56418da9d3f5819ea5fe016. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


marcsanmi:
afaiu, after this change users who don't have `scrape_native_histograms: true` (or NHCB conversion) on their config will see empty panels after upgrading.

Should we document this in some way?

simonswine:
> afaiu, after this change users who don't have `scrape_native_histograms: true` (or NHCB conversion) on their config will see empty panels after upgrading.
> 
> Should we document this in some way?

I am not a fan of this aspect either, we could have native dashboards an non native dashboard with ifs in the helm. that's also very ugyl, wdyt?

simonswine:
> * operations/monitoring/helm/pyroscope-monitoring/charts/k8s-monitoring-3.8.8.tgz

https://github.com/grafana/k8s-monitoring-helm/blob/6da004416be81e67ddedd6199cd3d8fddeb37721/charts/k8s-monitoring-v1/values.yaml#L2584

Not too sure if we can ignore this the same way

simonswine:
Okay I have implemented a variant with and without native histograms also our helm chart now supports native histograms since the helm chart update: Bumped k8s-monitoring 3.7.2 → 3.8.8

@marcsanmi please take another look

operations/monitoring/helm/pyroscope-monitoring/values.yaml
```diff
@@ -143,7 +143,16 @@ dashboards:
...
+  # Set to true when Prometheus scrapes native histograms (scrape_native_histograms: true).
+  # When false, dashboards use classic _bucket/_sum/_count suffix queries instead.
+  nativeHistograms: true
+
   tenantQuery: |
+    sum by (tenant, slug, org_name, environment) (
+      histogram_sum(rate(pyroscope_distributor_received_decompressed_bytes{cluster=~"$cluster",namespace=~"$namespace"}[$__rate_interval]))
+    )
+
+  tenantQueryClassic: |
```
marcsanmi, line 155: I might be wrong but `_dashboard_v2-write-path.yaml` Tenants panel still uses `tenantQuery` without the nativeHistograms guard?

marcsanmi (dismissed):
Sorry, somehow I missed this PR. 

LGTM, just a minor thing to check.

marcsanmi (approved):

marcsanmi (approved):

PR #5049, opened 2026-04-13
Remove PYROSCOPE_V2_EXPERIMENTAL

alsoba13:
Note base branch is oss-v2-breaking-changes. Do not merge. 
Trying to focus only in code and operativity. 

## New Proposal:
Get rid of PYROSCOPE_V2_EXPERIMENT, and introducing a new flag instead: 
```-architecture.storage=(v1|v1-v2-dual|v2)```

* When using `v2` we only have v2 stuff (default value)
* When using `v1-v2-dual` we have both v1 and v2 modules.
* When using `v1` we only have v1 modules.

If using `v1-v2-dual`, we rely on the `write-path` variables for the write path. And the `enable-query-backend-from` for the read path (`enable-query-backend` is not needed anymore). 

For single binary, it works the same way, but I removed the old `all.enable-v1-read-path` and `all.enable-v1-write-path` (use `legacy` or `dual` instead). We lose flexibility here but this becomes a bit simpler.

---

> [!NOTE]
> **Medium Risk**
> Touches core configuration defaults and module dependency wiring, which can change what services start and how traffic is routed (write/read paths) across deployments. Risk is mitigated by added validation and updated integration tests, but misconfiguration could lead to missing components or unexpected routing.
> 
> **Overview**
> Replaces the `PYROSCOPE_V2(_EXPERIMENT)` switch with a first-class `-architecture.storage=(v1|v1-v2-dual|v2)` flag and refactors startup to select modules, feature flags, and admin behavior based on that mode (including dropping `all.enable-v1-*` toggles).
> 
> Changes defaults toward the v2 stack: `write-path` now defaults to `segment-writer`, `enable-query-backend` defaults to true with `enable-query-backend-from=auto`, and several server/storage client defaults are bumped (gRPC msg sizes/keepalive, S3/GCS idle connection settings). Adds config validation to enforce a compatible `write-path` for `v1` vs `v2` modes, and updates generated help/docs, IDE run configs, and integration/unit tests (including status code expectations and ensuring tests ingest data so matcher parsing is exercised).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2fe1bfd0b15b27d13f36800e0e05cd28ba238cde. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### CODE REVIEW

File: cmd/pyroscope/help-all.txt.tmpl [hunk 1]
```diff
@@ -15,6 +51,26 @@ Usage of ./pyroscope:
...
+  -compaction-worker.metrics-exporter.rules-source.client-address string
+    	The address to use for the recording rules client connection.
+  -compaction-worker.metrics-exporter.rules-source.static value
+    	[experimental] List of static recording rules of the type settingsv1.RecordingRule. Will only be use in the absence of a recording rules client. (default [])
+  -compaction-worker.request-timeout duration
+    	Job request timeout. (default 5s)
+  -compaction-worker.small-object-size-bytes int
+    	Size of the object that can be loaded in memory. (default 8388608)
+  -compaction-worker.temp-dir string
+    	Temporary directory for compaction jobs. (default "/tmp")
```
alsoba13, line 73: This file was crafted to pass the test, but in my local "/tmp" becomes "/var/...". I'll address later so we can pass tests in local too

pkg/pyroscope/pyroscope.go
```diff
@@ -456,118 +451,109 @@ func (f *Pyroscope) setupModuleManager() error {
...
-			PlacementAgent:        {Overrides, API, Storage},
-			PlacementManager:      {Overrides, API, Storage},
-			Symbolizer:            {Overrides, Storage},
-			QueryDiagnosticsStore: {Storage},
-			QueryDiagnosticsAdmin: {QueryDiagnosticsStore, API, MetastoreClient},
+		Server:                {GRPCGateway, HealthServer},
+		API:                   {Server},
+		Metastore:             {Overrides, API, MetastoreClient, Storage, PlacementManager},
+		MetastoreAdmin:        {API, MetastoreClient},
+		Distributor:           {Overrides, SegmentWriterClient, API, UsageReport, Storage, IngesterRing},
```
alsoba13, line 513: IngesterRing dependency still exists in Distributor. TODO: clean up later (non-blocking)

alsoba13: Apart of that, there may be more things loading that could be avoided. We should do a deeper clean up.
This is something that happened before: when running v2, we were actually also running ingesterRings. So we were loading unnecessary components. With this PR, the other way around will also happen: if we run the "legacy" code, we are likely including some v2 stuff.

cmd/pyroscope/help-all.txt.tmpl
simonswine: Maybe setting TMPDIR env variable as part of the generation might help avoid it

pkg/pyroscope/storage_mode.go
```diff
@@ -0,0 +1,37 @@
...
+	"errors"
+	"fmt"
+)
+
+type StorageMode string
+
+const (
+	Legacy  StorageMode = "legacy"
+	Dual    StorageMode = "dual"
+	Default StorageMode = "default"
```
simonswine: I would call this more explicit, in case we want/need a v2, also would align it with the naming of StorageLayer in helm comment (nit):


```suggestion
	V1 StorageLayer = "v1" // deprecated: will be removed in Pyroscope 3.0
 	V1V2Dual    StorageLayer = "v1-v2-dual" 
	V2 StorageLayer = "v2" //
```

simonswine (approved):
Thanks for iterating on this quickly, this is a lot cleaner ❤️

pkg/test/integration/http_status_codes_test.go
```diff
@@ -851,7 +851,7 @@ func TestStatusCodes(t *testing.T) {
 						"start":    time.Now().Add(-1 * time.Hour).UnixMilli(),
 						"end":      time.Now().UnixMilli(),
 					}),
-					WantV1StatusCode: http.StatusOK,
+					WantV1StatusCode: http.StatusInternalServerError,
```
alsoba13, line 856: Very weird that we had a 200 here and passed. In v2 we actually fixed this to be a 4xx error (https://github.com/grafana/pyroscope/pull/4685), but in v1 we should still have a 500.

I've tested myself with a old v1 cell and this is actually a 500.
```
curl ...
HTTP/2 500
...

{"code":"unknown","message":"{\"code\":\"internal\",\"message\":\"unknown: rpc error: code = InvalidArgument desc = failed to parse label selector: 1:1: parse error: unexpected character after '!': 'b'\"}"}
```

Not sure where the 200 comes from and why my changes made this surface this

pkg/test/integration/http_status_codes_test.go
```diff
@@ -2307,7 +2307,7 @@ func TestStatusCodes(t *testing.T) {
 						"labelSelector": "{}",
 						"start":         time.Now().Add(-1 * time.Hour).UnixMilli(),
 						"end":           time.Now().UnixMilli(),
-						"spanSelector":  []string{"span1", "span2"},
+						"spanSelector":  []string{"0123456789012345", "0123456789012346"},
```
alsoba13, line 2312: Same here, some of these tests where failing with 4xx (invalid span id lenght) when 2xx was expected (others where expecting 4xx already and passed anyway). Unsure why this surfaced now.

pkg/test/integration/http_status_codes_test.go
alsoba13: Btw, apparently this test is flaky also, I've run it once in my machine that was returning 200, but it usually returns 500

pkg/pyroscope/modules.go
```diff
@@ -480,7 +481,7 @@ func (f *Pyroscope) initServer() (services.Service, error) {
 	}
 
 	f.setupWorkerTimeout()
-	if f.isModuleActive(QueryScheduler) {
+	if f.isModuleActive(QueryScheduler) || f.isModuleActive(QueryFrontend) {
```
alsoba13, line 484: Now, we are not always enabling QueryScheduler, which means that this read timeout increase is not set. This was making ingestOTLP tests fail.

.idea/runConfigurations/v1.xml
```diff
@@ -3,6 +3,7 @@
     <module name="pyroscope" />
     <working_directory value="$PROJECT_DIR$" />
     <go_parameters value="-tags embedassets" />
+    <parameters value="-architecture.storage=legacy  -write-path=ingester" />
```
aleks-p: Getting this when running v1 

`invalid value "legacy" for flag -architecture.storage: invalid storage mode: legacy; valid options: v1, v1-v2-dual, v2`

pkg/pyroscope/pyroscope.go
aleks-p: Seems fairly straightforward to override dependencies if running purely v1, anything preventing us from doing that (aside from "it was like that the other way around")?

pkg/pyroscope/storage_layer.go
```diff
@@ -0,0 +1,37 @@
...
+
+func (m *StorageLayer) Set(text string) error {
+	x := StorageLayer(text)
+	for _, name := range storageModes {
+		if x == name {
+			*m = x
+			return nil
+		}
+	}
+	return fmt.Errorf("%w: %s; %s", ErrInvalidStorageMode, x, validStorageModeOptionsString)
```
aleks-p: this only covers the flag route, if someone sets `-config.file config.yaml` and has `architecture_storage: garbage` in the yaml, the error will come later and be less clear

adding something like this should be enough

```
func (m *StorageLayer) UnmarshalText(text []byte) error {
	return m.Set(string(text))
}
```

pkg/distributor/writepath/write_path.go
```diff
@@ -92,7 +92,7 @@ type Config struct {
 }
 
 func (o *Config) RegisterFlags(f *flag.FlagSet) {
-	o.WritePath = IngesterPath
+	o.WritePath = SegmentWriterPath
```
aleks-p, line 95: If someone runs v1 (`architecture-storage v1`) without changing this, Pyroscope will be up but the write path won't work (RPCs will fail). We should either
- detect this and fail (e.g., in Config.Validate() in pyroscope.go)
- remove the default and force users to set it
- auto-correct it somewhere to "ingester" for v1

pkg/pyroscope/storage_layer.go
```diff
@@ -0,0 +1,37 @@
...
+
+type StorageLayer string
+
+const (
+	V1       StorageLayer = "v1"
+	V1V2Dual StorageLayer = "v1-v2-dual"
+	V2       StorageLayer = "v2"
+)
+
+var ErrInvalidStorageMode = errors.New("invalid storage mode")
```
aleks-p: nit, should this be `layer` instead of `mode` (or rename the type and flags if we prefer mode)?

aleks-p (commented):
Makes sense to me, left a few minor remarks.

.idea/runConfigurations/v1.xml
alsoba13: fixed!

pkg/pyroscope/storage_layer.go
alsoba13: fixed, thanks!

pkg/pyroscope/storage_layer.go
alsoba13: addressed in b38ac801c9c70f1d621a5209402dd2172225ea7e

pkg/distributor/writepath/write_path.go
alsoba13: addressed in b38ac801c9c70f1d621a5209402dd2172225ea7e

pkg/pyroscope/pyroscope.go
alsoba13: Yeah I guess this is enough: cf865873defc265830dbeb9c2122bb3348b46c24

aleks-p (approved):

pkg/pyroscope/pyroscope.go
```diff
@@ -196,21 +197,25 @@ func (c *Config) RegisterFlagsWithContext(f *flag.FlagSet) {
 	)
 	f.BoolVar(&c.ShowBanner, "config.show_banner", true, "Prints the application banner at startup.")
 	f.DurationVar(&c.ShutdownDelay, "shutdown-delay", 0, "Wait time before shutting down after a termination signal.")
+	c.ArchitectureStorage = V2
```
simonswine: I think this should be the default otherwise user without any other config and single binary will loose access to v1 data at time of upgrading.

I also think we could remove the V2 mode, but not feeling to strongly about this

```suggestion
	c.ArchitectureStorage = V1V2Both
```

pkg/pyroscope/pyroscope.go
```diff
@@ -196,21 +197,25 @@ func (c *Config) RegisterFlagsWithContext(f *flag.FlagSet) {
...
+	f.Var(&c.ArchitectureStorage, "architecture.storage", "Storage architecture. Use 'v1', 'v1-v2-dual' or 'v2'.")
```
simonswine: I think this needs slightly better help text

simonswine (commented):
I have not been able to figure this status code mistmatch. But after changing default and help text am happy to merge this in my branch

pkg/test/integration/http_status_codes_test.go
simonswine: I can explain it, when there is no data ingested, the whole later part of the validations will never run and we will therefore return a 200 with empty data.

Or how claude put it:

```
Root Cause

  The chain is:

  1. Test sends !bad_syntax! as a matcher with a time range [now-1h, now]
  2. Querier's Series handler (querier.go:443) dispatches to seriesFromIngesters
  3. Inside the ingester: PhlareDB.Series → Queriers.Series → Series(ctx, req, blockGetter) at block_querier.go:1427
  4. blockGetter is ForTimeRange, which filters queriers using InRange (block_querier.go:608)
  5. The head has no data, so head.Bounds() returns (0, 0) (head.go:380)
  6. InRange check: start (now-1h in ms, e.g. 1745843951000) > max (0) → false → head querier is excluded
  7. No queriers in range → the loop in Series never runs → matcher is never parsed → returns empty LabelsSet → 200 {}

  The matcher syntax validation in Head.Series (head.go:500-503) is never reached:
  selectors, err := parseSelectors(req.Msg.Matchers)
  if err != nil {
      return nil, err  // <-- never gets here
```

simonswine (approved):

PR #5052, opened 2026-04-15
fix(scheduler): replace bufconn with TCP listener to fix flaky test

marcsanmi:
## Summary
- Replace `bufconn` (in-memory listener) with httptest's default TCP listener in scheduler tests — bufconn has a subtle incompatibility with h2c-hijacked HTTP/2 bidi streams that causes some EOF failures under load from time to time e.g. https://github.com/grafana/pyroscope/actions/runs/24395819884/job/71253229300
- Replace `time.Sleep(500ms)` with `connectedFrontendClients` metric polling in `TestCancelRequestInProgress`

## Context
`TestCancelRequestInProgress` was failing intermittently on CI with `rpc error: code = Unavailable desc = error reading from server: EOF` during the initial enqueue call. The flake was reproduced locally with `-race -count=1000` (~2 failures per 1000 runs) and traced to `bufconn` by swapping it for a real TCP listener (0 failures in 4500+ race detector runs).

---

> [!NOTE]
> **Low Risk**
> Test-only changes that adjust transport setup and timing assertions; low risk to production code, but may affect test behavior/assumptions around connection lifecycle.
> 
> **Overview**
> Deflakes `pkg/scheduler/scheduler_test.go` by dropping the `bufconn` in-memory listener for the scheduler’s h2c/HTTP2 test server and dialing the real `httptest` TCP listener address instead.
> 
> Replaces a fixed `time.Sleep` in `TestCancelRequestInProgress` with polling on the `connectedFrontendClients` metric to wait for frontend disconnect detection before completing the querier response cycle.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 59417fc35ad660f560feb208b7bf14ee52697949. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/scheduler/scheduler_test.go
+++ b/pkg/scheduler/scheduler_test.go
@@ -68,12 +68,7 @@ func setupScheduler(t *testing.T, args schedulerArgs) (*Scheduler, schedulerpb.S
 	mux := mux.NewRouter()
 	server.Config.Handler = h2c.NewHandler(mux, &http2.Server{})
 
-	// Use an in-memory network connection to avoid test flake from network access.
-	listener := bufconn.Listen(256 << 10)
-	server.Listener = listener
-
 	server.Start()
-	require.NoError(t, err)
 	schedulerpbconnect.RegisterSchedulerForFrontendHandler(mux, s, args.handlerOpts...)
 	schedulerpbconnect.RegisterSchedulerForQuerierHandler(mux, s, args.handlerOpts...)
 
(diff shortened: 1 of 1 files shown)


simonswine (approved):
Makes sense! Thanks for fixing this ❤️ 

Should we replace the other uses of bufconn as well?

marcsanmi:
> Should we replace the other uses of bufconn as well?

I don't see any other combined usage of `h2c + bufconn`, but I'll check carefully and andress it in a followup PR if I find anything else :)

PR #5078, opened 2026-04-17
docs: fix vertical stacking of contributors grid in README

marcsanmi:
## Summary

The contributor avatars in the README have been rendering one-per-row (vertically stacked) since the April 1 monthly contributors update landed (#4963). This PR restores the horizontal grid.

- **Current broken render on `main`:** https://github.com/grafana/pyroscope?tab=readme-ov-file#thanks-to-the-contributors-of-pyroscope
- **Fixed render on this branch:** https://github.com/grafana/pyroscope/blob/fix/readme-contributors-layout/README.md#thanks-to-the-contributors-of-pyroscope

## Root cause

GitHub's markdown renderer was updated (first reports: [community discussion #165090](https://github.com/orgs/community/discussions/165090), July 2025) to inject aspect-ratio placeholder CSS on `` tags that have both `width` and `height` attributes:

```html
<!-- What we write -->
<a href="..."><img src="..." width="80" height="80"></a>

<!-- What GitHub renders -->
<a href="..."><img src="..." width="80" height="80"
  style="max-width: 100%; height: auto; max-height: 80px; aspect-ratio: 80/80;
         background-color: var(--bgColor-muted); border-radius: 6px;
         display: block"
  class="js-gh-image-fallback"></a>
```

The injected `display: block` propagates up through the enclosing inline `<a>`, making each anchor block-level. Combined with GFM turning newlines between `<a>` lines into `<br>`, every avatar ends up on its own row.

PR #4963 itself was harmless — only reordered/added names — but it was the first README edit since GitHub's sanitizer change rolled out to the cache for this repo, so it was the commit that made the regression visible. Nothing in our repo caused this; our format simply happens to trigger the new renderer behavior.

## Fix

Two small changes, applied both to the live README and to the generator so the monthly automation produces the same shape (no churn PR next month):

1. **Wrap the block in `<p>...</p>`.** This makes the section a CommonMark HTML block, so GFM stops converting each source-line newline into `<br>`.
2. **Drop `height="80"` from each `` (keep `width="80"`).** GitHub only injects the `display: block` style when *both* attributes are present. With width-only, the injection is just `max-width: 100%;` — no block-ification.

Avatars are square (GitHub serves them that way), so `width=80` alone still renders at 80×80 — the browser derives the height from the intrinsic aspect ratio. Visually identical to the pre-break grid.

This is the same workaround confirmed by multiple users in the GitHub community thread above.

## Files changed

- `tools/update-contributors/main.go` — emit `<p>` / `</p>` wrappers; drop `height="80"` from generated `` tags.
- `README.md` — mirror the same shape in the currently-committed content, so the fix renders immediately and the next monthly cron produces a zero-diff regeneration.

## Test plan

- [x] Verified rendered output via `POST /markdown` against GitHub's API — no `<br>`, no `display: block`, just `max-width: 100%;`.
- [ ] Visually check the rendered README on this branch (link above).
- [ ] After merge, confirm `github.com/grafana/pyroscope` shows avatars in a grid again.
- [ ] Next month's automated contributors PR produces only name-level changes (no structural diff).

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,68 +159,70 @@ To start contributing, check out our [Contributing Guide](docs/internal/contribu
 ### Thanks to the contributors of Pyroscope!
 
 [//]: contributor-faces
-<a href="https://github.com/simonswine"><img src="https://avatars.githubusercontent.com/u/223048?v=4" title="simonswine" width="80" height="80"></a>
-<a href="https://github.com/petethepig"><img src="https://avatars.githubusercontent.com/u/662636?v=4" title="petethepig" width="80" height="80"></a>
-<a href="https://github.com/cyriltovena"><img src="https://avatars.githubusercontent.com/u/1053421?v=4" title="cyriltovena" width="80" height="80"></a>
-<a href="https://github.com/eh-am"><img src="https://avatars.githubusercontent.com/u/6951209?v=4" title="eh-am" width="80" height="80"></a>
-<a href="https://github.com/kolesnikovae"><img src="https://avatars.githubusercontent.com/u/12090599?v=4" title="kolesnikovae" width="80" height="80"></a>
-<a href="https://github.com/korniltsev"><img src="https://avatars.githubusercontent.com/u/331773?v=4" title="korniltsev" width="80" height="80"></a>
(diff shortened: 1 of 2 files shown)


aleks-p (approved):
Looks good at https://github.com/grafana/pyroscope/blob/c66aa7c4e71307c7cd87111b3f4fbc29fee53e66/README.md

PR #5067, opened 2026-04-17
helm: add pyroscope.rbac.create toggle

luweglarz:
Adds a pyroscope.rbac.create toggle to allow disabling creation for RBAC when managing RBAC externally or in restricted environments.

---

> [!NOTE]
> **Low Risk**
> Low risk Helm values change that only gates creation of the v2 discovery `Role`/`RoleBinding`; misconfiguration (disabling it without providing equivalent RBAC) could break service discovery in v2 mode.
> 
> **Overview**
> Adds a new Helm value, `pyroscope.rbac.create` (default `true`), to allow disabling generation of Pyroscope RBAC resources when RBAC is managed externally.
> 
> The v2 discovery RBAC template (`rbac-discovery.yaml`) is now rendered only when both `architecture.storage.v2` is enabled and `pyroscope.rbac.create` is `true`, and the new option is documented in `README.md` and propagated into `values.yaml` and `jsonnet/values.json`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aab7e80ef3f716725b5c0a24ce9aaa92f2d6df15. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #5079, opened 2026-04-17
docs(v2): improve v1->v2 migration docs

aleks-p:
`helm upgrade --reuse-values` will fail if the new chart has features that rely on values bundled with the chart (e.g. failed for me because of how #5067 and #4983 are set up).

Hence we switch to `--reset-then-reuse-values` which will overlay the installation values on top of the chart-bundled defaults.

Also updating the target chart version to 2.0.0 which is not yet released. We might want to backport this, but it is probably fine if we don't.

---

> [!NOTE]
> **Low Risk**
> Docs-only changes with no runtime impact, though pinning to an unreleased chart version could confuse users if they follow the commands verbatim.
> 
> **Overview**
> Adjusts the v1→v2 storage migration guide to recommend `helm upgrade` with `--reset-then-reuse-values` (instead of `--reuse-values`) so chart-default values are retained while overlaying existing install values.
> 
> Pins the single-binary migration commands (Phase 1 dual ingest and Phase 3 cutover) to `grafana/pyroscope --version 2.0.0`, and removes the prerequisite callout about persistence being enabled.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 37f2e1244457128da6aa15628be802d44b6e58f0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/sources/reference-pyroscope-v2-architecture/migrate-from-v1/index.md
+++ b/docs/sources/reference-pyroscope-v2-architecture/migrate-from-v1/index.md
@@ -31,8 +31,6 @@ Before starting the migration, make sure you have:
 
   You should see `v1: true` and `v2: false`. If you see `v2: true`, your installation is already using v2 or is mid-migration.
 
-- **Persistence enabled** (`pyroscope.persistence.enabled=true`).
-
 - **Object storage configured.** v2 writes directly to object storage — it doesn't use local disk for block storage. If you haven't configured object storage yet, add it to your Helm values. For example, for S3:
 
   ```yaml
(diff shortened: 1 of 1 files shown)


docs/sources/reference-pyroscope-v2-architecture/migrate-from-v1/index.md
```diff
@@ -31,8 +31,6 @@ Before starting the migration, make sure you have:
 
   You should see `v1: true` and `v2: false`. If you see `v2: true`, your installation is already using v2 or is mid-migration.
 
-- **Persistence enabled** (`pyroscope.persistence.enabled=true`).
```
aleks-p, line 34: I think this causes more confusion than anything. If a user tries to set it globally all components will request a volume, some will not get one and the installation will hang.

simonswine (approved):

marcsanmi (approved):

PR #5094, opened 2026-04-22
compactionworker: default job concurrency to 1

rabunkosar-dd:
Summary
- default compaction-worker.job-concurrency to 1
- treat values below 1 as 1
- update help text and add a focused unit test

Why
compaction-worker currently defaults job-concurrency to 0, which resolves to runtime.GOMAXPROCS(-1).

For a memory-bound worker, CPU count is not a safe default for concurrency. On large nodes this can cause the worker to advertise a very large job capacity and run too many compactions in parallel by default.

Using 1 as the default makes the worker conservative and predictable, while still allowing operators to explicitly raise concurrency when they have enough memory headroom.

Changes
- change the flag default for -compaction-worker.job-concurrency from 0 to 1
- add effectiveJobConcurrency() so values below 1 resolve to 1
- remove the GOMAXPROCS fallback from worker initialization
- update CLI help text
- add a unit test for the fallback behavior

Validation
- go test ./pkg/compactionworker

---

> [!NOTE]
> **Medium Risk**
> Changes the default `compaction-worker.job-concurrency` behavior, reducing parallelism by default and clamping misconfigured values, which can affect compaction throughput and job scheduling capacity in production deployments.
> 
> **Overview**
> Updates `compaction-worker.job-concurrency` to default to **1** (instead of deriving from CPU count) and introduces `effectiveJobConcurrency()` to clamp configured values `<1` up to 1.
> 
> This removes the previous `GOMAXPROCS` fallback during worker initialization, updates the CLI help text, and adds a focused unit test covering the new clamping behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b38eeffeba7783d68cae0835124760e95b2ed6d9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/pyroscope/help-all.txt.tmpl
+++ b/cmd/pyroscope/help-all.txt.tmpl
@@ -54,7 +54,7 @@ Usage of ./pyroscope:
   -compaction-worker.cleanup-max-duration duration
     	Maximum duration of the cleanup operations. (default 15s)
   -compaction-worker.job-concurrency int
-    	Number of concurrent jobs compaction worker will run. Defaults to the number of CPU cores.
+    	Number of concurrent jobs compaction worker will run. Values less than 1 are treated as 1.
   -compaction-worker.job-poll-interval duration
     	Interval between job requests (default 5s)
   -compaction-worker.metrics-exporter.enabled
--- a/cmd/pyroscope/help-all.txt.tmpl
+++ b/cmd/pyroscope/help-all.txt.tmpl
@@ -54,7 +54,7 @@ Usage of ./pyroscope:
   -compaction-worker.cleanup-max-duration duration
     	Maximum duration of the cleanup operations. (default 15s)
   -compaction-worker.job-concurrency int
-    	Number of concurrent jobs compaction worker will run. Values less than 1 are treated as 1.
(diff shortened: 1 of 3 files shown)


aleks-p:
Thanks for your contribution! The reasoning is sound - CPU count isn't a great proxy for a memory-bound worker.
  
That said, I don't think we should change the shipped default here. The conservative behavior you are after is already available without a code change: setting `-compaction-worker.job-concurrency=1` (or `job_capacity: 1`) gives you exactly one thread today. So if you're hitting memory pressure on large nodes, that's the immediate fix.

Flipping the default from `GOMAXPROCS` to 1 is a different thing: it silently slows compaction on upgrade for every existing deployment that relies on the default, which risks compaction falling behind (object buildup, query degradation). That's a deliberate tradeoff we'd want to discuss as a team rather than fold into a single PR - and if we did decide to change it, we'd need to communicate it (Helm upgrade note, release note).

Since the behavior you need is already reachable via the existing flag, I'm going to close this. If you think the out-of-the-box default is genuinely unsafe for most users, please open an issue and we can weigh the tradeoff properly there.

PR #5112, opened 2026-05-05
feat(segmentwriter): build dataset index in segments

### LINKED ISSUE #4193: V2: Ensure ingested data becomes deterministically even if there is no profile flow
As part of extending the integration tests to cover V2 experimental storage, we noticed, after stopping ingestion all together, we are not able to deterministically query profiles.

As a workaround we keep pushing some extra profiles for another tenant before we querying them back.

https://github.com/grafana/pyroscope/pull/4185#discussion_r2082396157

aleks-p:
Currently, the per-tenant dataset TSDB index is only built at compaction time. Queries that do not include a `service_name` label rely on the dataset index to identify which datasets within a block match the query. As a result, freshly ingested data is effectively invisible to such queries until compaction completes.

This change extends the segment writer to build a per-tenant dataset index for each tenant present in a segment. The index is appended as a `Format1` pseudo-dataset to the block metadata, annotated with the `__tenant_dataset__ = dataset_tsdb_index` label, mirroring the layout produced by the compactor.

The implementation:

- Preserves the series labels and fingerprints captured by the in-memory profile index so that we can build the dataset index without re-reading the TSDB.
- Promotes the dataset index writer used by compaction to a public block.DatasetIndexWriter and adds an AddSeries entry point (compaction continues to use the row-by-row path).
- Groups contiguous tenant heads in a segment, builds an index per tenant and writes a pseudo-dataset entry on each tenant boundary.

The integration test workaround that pushed extra profiles to force compaction so a tenant-wide index would become available is no longer necessary and is removed (closes the TODO referencing pyroscope#4193).

The `pushProfiles` helper is also adjusted to stamp profiles at the test's reference time instead of the (overwritten) future timestamps, so queries with default time ranges can find the data without depending on wall-clock progress.

TODO:
- [x] test in a dev environment
- [x] simplify tests for federated (multi-tenant) queries

Performance impact (after 447b0d1):

```
  ┌──────────────────────┬────────────────┬───────────────┬────────┐                               
  │     Profile type     │ New-code total │ Service total │ Share  │                               
  ├──────────────────────┼────────────────┼───────────────┼────────┤                               
  │ CPU (process_cpu)    │ 4.19 s         │ 109.51 s      │ 3.83%  │                               
  ├──────────────────────┼────────────────┼───────────────┼────────┤                             
  │ Memory alloc_space   │ 793 MB         │ 53.86 GB      │ 1.47%  │                               
  ├──────────────────────┼────────────────┼───────────────┼────────┤                               
  │ Memory alloc_objects │ 2.82 M         │ 211 M         │ 1.34%  │                               
  ├──────────────────────┼────────────────┼───────────────┼────────┤                               
  │ Memory inuse_space   │ 538 KB         │ 236 MB        │ 0.23%  │                               
  ├──────────────────────┼────────────────┼───────────────┼────────┤                               
  │ Mutex delay          │ 5.5 µs         │ 221 ms        │ <0.01% │                               
  ├──────────────────────┼────────────────┼───────────────┼────────┤                               
  │ Block delay          │ not present    │ —             │ 0%     │                               
  └──────────────────────┴────────────────┴───────────────┴────────┘                               
```

Fixes #4193

---

> [!NOTE]
> **Medium Risk**
> Changes block layout/metadata by adding per-tenant `DatasetFormat1` pseudo-datasets at segment flush time and updates query resolution logic to union results across multiple indices, which can affect dataset selection for queries. Risk is moderated by added unit/integration coverage but touches critical query/compaction paths.
> 
> **Overview**
> Freshly-flushed segment (L0) blocks now include a **per-tenant `DatasetFormat1` `dataset_tsdb_index` pseudo-dataset** appended at each tenant boundary, so queries without `service_name` can discover datasets without waiting for compaction.
> 
> This introduces a pooled `block.DatasetIndexWriter` (used by both compaction and segment flush), preserves series labels/fingerprints from `memdb` flush to feed the index, and adds segment-writer metrics/debug counters for emitted tenant indexes.
> 
> The query backend now supports **resolving multiple Format1 indices in a single block** (multi-tenant segments) by opening each index, unioning matched dataset IDs, and rewriting block metadata accordingly; tests/integration were updated to skip pseudo-datasets, add federated coverage, and remove the prior “force compaction” workaround.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eef33f5caa2820b0e2909374e1c6a0c39e9084a6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/block/dataset_index.go
+++ b/pkg/block/dataset_index.go
@@ -2,7 +2,9 @@ package block
 
 import (
 	"context"
+	"io"
 	"sort"
+	"sync"
 
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/storage"
@@ -11,6 +13,12 @@ import (
 	"github.com/grafana/pyroscope/v2/pkg/phlaredb/tsdb/index"
 	memindex "github.com/grafana/pyroscope/v2/pkg/segmentwriter/memdb/index"
 )
+
+// datasetIndexEncBufferSize is the initial capacity hint for the
+// underlying TSDB index writer's scratch encoding buffers. Dataset
+// indexes have one entry per dataset within a tenant block, so a small
+// hint is plenty; the buffers grow if needed.
+const datasetIndexEncBufferSize = 1 << 14
 
 // DatasetIndexWriter builds the per-tenant dataset TSDB index.
 //
@@ -23,19 +31,43 @@ import (
 // It is used both at compaction time (where series labels are discovered
 // row-by-row via WriteRow) and at segment flush time (where series are
 // already known and added in bulk via AddSeries).
+//
+// Writers are pooled. Obtain one with NewDatasetIndexWriter and return
+// it with Close once WriteTo has been called.
 type DatasetIndexWriter struct {
 	series   []seriesLabels
 	chunks   []index.ChunkMeta
 	previous model.Fingerprint
 	symbols  map[string]struct{}
 	idx      uint32
-	buf      []byte
 }
 
+var datasetIndexWriterPool = sync.Pool{
+	New: func() any {
+		return &DatasetIndexWriter{symbols: make(map[string]struct{})}
+	},
+}
+
+// NewDatasetIndexWriter returns a DatasetIndexWriter ready for use,
+// reusing one from a pool when possible. Callers must call Close to
+// return it once they are done.
 func NewDatasetIndexWriter() *DatasetIndexWriter {
-	return &DatasetIndexWriter{
-		symbols: make(map[string]struct{}),
-	}
+	return datasetIndexWriterPool.Get().(*DatasetIndexWriter)
+}
+
+// Close returns the writer to the pool. After Close, the writer must
+// not be used.
+func (rw *DatasetIndexWriter) Close() {
+	rw.reset()
+	datasetIndexWriterPool.Put(rw)
+}
+
+func (rw *DatasetIndexWriter) reset() {
+	rw.series = rw.series[:0]
+	rw.chunks = rw.chunks[:0]
+	clear(rw.symbols)
+	rw.previous = 0
+	rw.idx = 0
 }
 
 // SetIndex sets the dataset index assigned to subsequent series added
@@ -77,40 +109,47 @@ func (rw *DatasetIndexWriter) addSeries(labels phlaremodel.Labels, fp model.Fing
 // Empty reports whether the writer has no series.
 func (rw *DatasetIndexWriter) Empty() bool { return len(rw.series) == 0 }
 
-// Buf returns the serialised index bytes. Only valid after Flush.
-func (rw *DatasetIndexWriter) Buf() []byte { return rw.buf }
-
-func (rw *DatasetIndexWriter) Flush() error {
-	// TODO(kolesnikovae):
-	//  * Estimate size.
-	//  * Use buffer pool.
-	w, err := memindex.NewWriter(context.Background(), 1<<20)
-	if err != nil {
-		return err
+// WriteTo encodes the accumulated series into a TSDB index and writes
+// the encoded bytes directly to dst, returning the number of bytes
+// written. After WriteTo, the accumulated state is reset; the writer
+// can be reused for another tenant or returned to the pool via Close.
+func (rw *DatasetIndexWriter) WriteTo(dst io.Writer) (int64, error) {
+	if len(rw.series) == 0 {
+		return 0, nil
 	}
 
-	// Sort symbols
+	w, err := memindex.NewWriter(context.Background(), datasetIndexEncBufferSize)
+	if err != nil {
+		return 0, err
+	}
+
 	symbols := make([]string, 0, len(rw.symbols))
 	for s := range rw.symbols {
 		symbols = append(symbols, s)
 	}
 	sort.Strings(symbols)
 
-	// Add symbols
 	for _, symbol := range symbols {
 		if err = w.AddSymbol(symbol); err != nil {
-			return err
+			return 0, err
 		}
 	}
 
-	// Add Series
 	for i, series := range rw.series {
 		if err = w.AddSeries(storage.SeriesRef(i), series.labels, series.fingerprint, rw.chunks[i]); err != nil {
-			return err
+			return 0, err
 		}
 	}
 
-	err = w.Close()
-	rw.buf = w.ReleaseIndex()
-	return err
+	if err = w.Close(); err != nil {
+		return 0, err
+	}
+
+	bw := w.ReleaseIndexBuffer()
+	defer memindex.PutBufferWriterToPool(bw)
+	rw.reset()
+
+	buf, _, _ := bw.Buffer()
+	n, err := dst.Write(buf)
+	return int64(n), err
 }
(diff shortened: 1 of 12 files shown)


pkg/querybackend/query_label_values.go
```diff
@@ -30,6 +31,15 @@ func queryLabelValues(q *queryContext, query *queryv1.Query) (*queryv1.Report, e
...
 		values, err = q.ds.Index().LabelValues(query.LabelValues.LabelName)
+		// Reader.LabelValues returns strings that alias the underlying
+		// index buffer ("It is not safe to use the return value beyond
+		// the lifetime of the byte slice passed into the Reader"). The
+		// dataset's buffer is recycled to a pool on close, which may
+		// happen before the staged report is marshaled, so we clone the
+		// values here to decouple them from the buffer lifetime.
+		for i, v := range values {
+			values[i] = strings.Clone(v)
+		}
```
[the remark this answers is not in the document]
aleks-p: This is a false positive, the buffer pool is not used in this path

pkg/segmentwriter/memdb/profile_index.go
```diff
@@ -91,19 +96,16 @@ func (pi *profilesIndex) Add(ps *schemav1.InMemoryProfile, lbs phlaremodel.Label
 	pi.metrics.profilesCreated.WithLabelValues(profileName).Inc()
 }
 
-func (pi *profilesIndex) Flush(ctx context.Context) ([]byte, []schemav1.InMemoryProfile, error) {
+func (pi *profilesIndex) Flush(ctx context.Context) ([]byte, []schemav1.InMemoryProfile, []flushedSeries, error) {
```
korniltsev-grafanista: nit: can we return `[]*profileSeries` here? or do we add a new separate slice to let the `[]*profileSeries` be GCed?

pkg/querybackend/query_label_values.go
```diff
@@ -30,6 +31,15 @@ func queryLabelValues(q *queryContext, query *queryv1.Query) (*queryv1.Report, e
 	var err error
 	if len(q.req.matchers) == 0 {
 		values, err = q.ds.Index().LabelValues(query.LabelValues.LabelName)
+		// Reader.LabelValues returns strings that alias the underlying
```
korniltsev-grafanista, line 34: nit: Is this an unrelated drive-by fix? should we fix other callers?
1. pkg/operations/v2/blocks/dataset_handlers.go:255
2. pkg/querybackend/query_label_values.go:32 
3. pkg/phlaredb/block_querier.go:375
Should this be submitted as a separate PR maybe? or remove yolostring from the `LabelValues`

pkg/segmentwriter/segment.go
```diff
@@ -273,21 +273,40 @@ func (s *segment) flushBlock(stream flushStream) ([]byte, *metastorev1.BlockMeta
...
-		// TODO(kolesnikovae): Build dataset index for the tenant.
-		//   Note that the heads are flushed concurrently, so we cannot build
-		//   during head flush. I'd prefer to delegate it to the head itself:
-		//      WriteDatasetIndex(w *memindex.Writer, idx uint32)
-		//   Tenant datasets follow sequentially; when all tenant datasets
-		//   are flushed, we can build the index and create a metadata
-		//   entry for it.
+		if dsIndex == nil || f.dataset.key.tenant != curTenant {
+			if dsIndex != nil {
+				if err := writeTenantDatasetIndex(w, meta, stringTable, curTenant, curTenantIdx, dsIndex); err != nil {
```
korniltsev-grafanista: nit: consider adding some visibility into how much memory new tenant dataset indexes occupy, maybe add a new metric or a log kv entry to the "writing segment block done" logline.

pkg/segmentwriter/segment.go
```diff
@@ -273,21 +273,40 @@ func (s *segment) flushBlock(stream flushStream) ([]byte, *metastorev1.BlockMeta
...
+					return nil, nil, err
+				}
+			}
+			curTenant = f.dataset.key.tenant
+			curTenantIdx = len(meta.Datasets)
+			dsIndex = block.NewDatasetIndexWriter()
```
korniltsev-grafanista, line 303: nit: it would be nice to have a unit test inspecting the new tenant indixes from a flushed segment.

korniltsev-grafanista (dismissed):
LGTM

only couple notes, feel free to ignore all of them
nit: there may be query failures during rollout if segment-writer starts producing new segments with new tenatn indexes and query backend is not rolled out yet and checks for `len(DataSets) != 1`

pkg/querybackend/query_label_values.go
aleks-p: Thanks for flagging. This surfaced in the tests now that we query segments in service-less queries, so we need to address it in some way.

> pkg/operations/v2/blocks/dataset_handlers.go:255

This is a handler for an admin page, can definitely apply the fix there but it is low risk as we access datasets one by one there.

> pkg/querybackend/query_label_values.go:32

This is the code we are discussing here :)

> pkg/phlaredb/block_querier.go:375

There is no buffer pool there to reuse from what I can see, so the chance of yoloString surfacing data in the same way is very low.

simonswine: I agree with @aleks-p assessment of the other paths and have moved this fix into a separate PR: https://github.com/grafana/pyroscope/pull/5116

pkg/block/dataset_index.go
```diff
@@ -0,0 +1,155 @@
...
+		rw.symbols[l.Value] = struct{}{}
+	}
+	rw.series = append(rw.series, seriesLabels{
+		labels:      cloned,
+		fingerprint: fp,
+	})
+	rw.chunks = append(rw.chunks, index.ChunkMeta{
+		SeriesIndex: rw.idx,
+	})
+}
```
simonswine: This interface feels quite ugly, I think it would be cleaner by removing `WriteRow`, `SetIndex` and only have `AddSeries` with an idx parameter. 

Compaction then can implement the wrapper that adds the previous fingerprint stuff

pkg/segmentwriter/segment.go
aleks-p: did both in 35ae9c9

pkg/segmentwriter/segment.go
aleks-p: done in 35ae9c9

pkg/block/dataset_index.go
aleks-p: done in eef33f5caa2820b0e2909374e1c6a0c39e9084a6

pkg/segmentwriter/memdb/profile_index.go
aleks-p: We can, it was done to make what we return more narrow, but it doesn't impact the lifecycle in any way. Changed to what you suggested in eef33f5caa2820b0e2909374e1c6a0c39e9084a6

aleks-p:
> nit: there may be query failures during rollout if segment-writer starts producing new segments with new tenatn indexes and query backend is not rolled out yet and checks for `len(DataSets) != 1`

@korniltsev-grafanista thanks for flagging this. I checked and this would only impact federated (multi-tenant) queries. For single-tenant queries we will still have 1 dataset index per block.

From what I see, we started supporting federated queries in v2 and I am not aware of them being broadly used, so my gut feeling is we can roll this out as is. If you or @simonswine have a strong opinion about this, the only way to address that would be to control this through a flag.

simonswine (approved):

PR #5137, opened 2026-05-08
fix(read-path): do not allow sub-ms step parameter

aleks-p:
> [!NOTE]
> **Low Risk**
> Low risk input-validation change that tightens API parameter constraints to prevent divide-by-zero/unbounded loops; behavior changes only for requests with `step < 1ms`.
> 
> **Overview**
> Prevents `SelectSeries`/`SelectHeatmap` queries from accepting *sub-millisecond* `step` values by rejecting any `step < 0.001` with `InvalidArgument` (protecting downstream millisecond arithmetic from truncating to 0 and causing divide-by-zero/unbounded loops).
> 
> Adds frontend unit tests to ensure both endpoints consistently reject `step` values below 1ms (including `0`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1d00a37c1af0c62adaa2e1c69ae738610e7110a3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/frontend/readpath/queryfrontend/query_heatmap.go
+++ b/pkg/frontend/readpath/queryfrontend/query_heatmap.go
@@ -25,9 +25,11 @@ func (q *QueryFrontend) SelectHeatmap(
 	ctx context.Context,
 	c *connect.Request[querierv1.SelectHeatmapRequest],
 ) (*connect.Response[querierv1.SelectHeatmapResponse], error) {
-	// Validate step
-	if c.Msg.Step <= 0 {
-		return nil, connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("step must be greater than 0, got %f", c.Msg.Step))
+	// Sub-millisecond step values truncate to 0 in the backend's millisecond
+	// arithmetic and would cause divide-by-zero / unbounded loop downstream;
+	// reject anything below 1ms.
+	if c.Msg.Step < 0.001 {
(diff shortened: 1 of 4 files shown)


pkg/frontend/readpath/queryfrontend/query_heatmap.go
```diff
@@ -25,9 +25,11 @@ func (q *QueryFrontend) SelectHeatmap(
 	ctx context.Context,
 	c *connect.Request[querierv1.SelectHeatmapRequest],
 ) (*connect.Response[querierv1.SelectHeatmapResponse], error) {
-	// Validate step
-	if c.Msg.Step <= 0 {
-		return nil, connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("step must be greater than 0, got %f", c.Msg.Step))
+	// Sub-millisecond step values truncate to 0 in the backend's millisecond
+	// arithmetic and would cause divide-by-zero / unbounded loop downstream;
+	// reject anything below 1ms.
+	if c.Msg.Step < 0.001 {
```
aleks-p, line 31: Current UIs go down to 10-20ms for very short time ranges but in reality I feel like 1s should be the minimum. Since we can't effectively change the UIs, we could alternatively clamp to 1s instead of rejecting. On top of this we should probably check the overall time range and limit the number of points we want to return.

korniltsev-grafanista (approved):

alsoba13 (approved):

PR #5139, opened 2026-05-09
Fix TLS support in distributor-to-ingester connections

### LINKED ISSUE #4585: Distributor connects to ingester via plan http despite configured TLS
#### Describe the bug

You can see the code here:

https://github.com/grafana/pyroscope/blob/39e0e7c9f5ee9c61b352ce0cafaf033f79be2205/pkg/clientpool/ingester_client_pool.go#L65

It is using plain HTTP despite me having TLS configured throughout.

In the logs:

```
Oct 31 03:59:30 143dm6 pyroscope[4186086]: http: TLS handshake error from [2400:...2844]:48136: tls: first record does not look like a TLS handshake
```

```
Oct 31 03:59:30 143dm6 pyroscope[4186086]: ts=2025-10-31T03:59:30.77202765Z caller=distributor.go:396 component=distributor tenant=anonymous user=anonymous level=warn msg="profile rejected" service_name=grafana.alloy.ebpf profile_type=process_cpu matched_usage_groups=[] detected_language=go profile_time=2025-10-31T03:59:27.807Z ingestion_delay=2.963s decompressed_size=21244 sample_count=149 err="unavailable: write tcp [2400:...2844]:48136->[2400:...2844]:4040: write: connection reset by peer"
```

```
Oct 31 03:59:30 143dm6 pyroscope[4186086]: ts=2025-10-31T03:59:30.207856516Z caller=distributor.go:396 component=distributor tenant=anonymous user=anonymous level=warn msg="profile rejected" service_name=grafana.alloy.ebpf profile_type=process_cpu matched_usage_groups=[] detected_language=unknown profile_time=2025-10-31T03:59:27.812Z ingestion_delay=2.394s decompressed_size=420 sample_count=1 err="unavailable: unexpected EOF"
```

I can see plaintext queries in tcpdump:

Image

#### To Reproduce

1. Configure TLS for all components in a single node deployment.
2. Send some data to the node via HTTPS.
3. Observe complaints about ingestion.

#### Expected behavior

I expect TLS to work.

#### Environment

- Infrastructure: bare metal, single node
- Deployment tool: manual

kimjune01:
Fixes #4585.

The ingester client pool hardcodes `insecure.NewCredentials()` for gRPC transport and `http://` for the Connect HTTP client. When TLS is configured between distributor and ingester, the distributor still connects over plaintext.

This adds a `TLSEnabled` field to `PoolConfig` (flag: `--tls-enabled`, default `false`). When enabled, the pool factory uses `credentials.NewTLS` for gRPC and `https://` for the Connect client URL scheme. The default behavior is unchanged.

---

> [!NOTE]
> **Medium Risk**
> Changes how distributor ingester clients dial gRPC and Connect endpoints when `--tls-enabled` is set, which can affect inter-service connectivity if misconfigured. Default behavior remains plaintext and should be unchanged for existing deployments.
> 
> **Overview**
> Adds an optional `tls_enabled`/`--tls-enabled` switch to the ingester client pool configuration.
> 
> When enabled, ingester pool clients use TLS credentials for gRPC dialing and switch the Connect client base URL from `http://` to `https://`, instead of always forcing insecure transport.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 93de26e22f7247c366cd510abde19045c9116fa9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/clientpool/ingester_client_pool.go
+++ b/pkg/clientpool/ingester_client_pool.go
@@ -2,6 +2,7 @@ package clientpool
 
 import (
 	"context"
+	"crypto/tls"
 	"flag"
 	"io"
 	"time"
@@ -12,6 +13,7 @@ import (
 	ring_client "github.com/grafana/dskit/ring/client"
 	"github.com/prometheus/client_golang/prometheus"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/health/grpc_health_v1"
 
@@ -24,6 +26,7 @@ type PoolConfig struct {
 	ClientCleanupPeriod  time.Duration `yaml:"client_cleanup_period"`
 	HealthCheckIngesters bool          `yaml:"health_check_ingesters"`
 	RemoteTimeout        time.Duration `yaml:"remote_timeout"`
+	TLSEnabled           bool          `yaml:"tls_enabled"`
 }
 
 // RegisterFlagsWithPrefix adds the flags required to config this to the given FlagSet.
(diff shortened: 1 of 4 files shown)


simonswine (commented):
Thanks for the contribution. We are also having a slightly changed architecture since v2.0, where there is no more ingester component. Still happy for this to be merged, but won't work on it. 

Here the main points to address:

### Inconsistent TLS between gRPC and Connect transports

The gRPC connection uses the explicit (empty) `tls.Config{}`, but the Connect HTTP client (`httpClient`) is left untouched and uses the default OS trust store. If a custom CA is ever added to the gRPC side, the two transports will silently diverge. They should share the same TLS configuration.

---

### No tests

There are no tests for the new `TLSEnabled` path. A small table-driven test in `ingester_client_pool_test.go` verifying the correct dial options are chosen based on config would be a good baseline.

pkg/clientpool/ingester_client_pool.go
```diff
@@ -24,18 +26,20 @@ type PoolConfig struct {
 	ClientCleanupPeriod  time.Duration `yaml:"client_cleanup_period"`
 	HealthCheckIngesters bool          `yaml:"health_check_ingesters"`
 	RemoteTimeout        time.Duration `yaml:"remote_timeout"`
+	TLSEnabled           bool          `yaml:"tls_enabled"`
```
simonswine, line 29: This only works if the ingester's server certificate is signed by a CA in the OS trust store (e.g. a public CA). Most Kubernetes deployments use private/internal CAs (cert-manager, etc.), so this will fail with a certificate verification error in practice.

There's no way to specify:
- A custom CA cert (`tls_ca_path`)
- Client certificates for mTLS (`tls_cert_path` / `tls_key_path`)
- Skip-verify for testing (`tls_insecure_skip_verify`)

The dskit ecosystem has a standard pattern for exactly this — `github.com/grafana/dskit/crypto/tls.ClientConfig`. Other Grafana components use it, and it exposes all of the above fields via YAML and flags automatically. I'd recommend replacing the `bool` with that struct:

```go
import dskittls "github.com/grafana/dskit/crypto/tls"

type PoolConfig struct {
    // ...
    GRPCClientTLS dskittls.ClientConfig `yaml:"grpc_tls_config"`
}
```

Then derive `*tls.Config` from it via `cfg.GRPCClientTLS.GetTLSConfig()` for both the gRPC dial and the HTTP transport.

simonswine (changes requested):

PR #5143, opened 2026-05-11
fix(speedscope): replace panic with error return for unknown unit values

simonswine:
## Summary

- `pkg/og/convert/speedscope/units.go`: `defaultSampleRate()` and `precisionMultiplier()` called `panic()` on unrecognised `unit` values that come directly from attacker-controlled JSON input
- Replaced both `panic()` calls with `fmt.Errorf()` returns and propagated the error up through all three call sites in `parser.go`
- Simplified `precisionMultiplier()` by collapsing five identical cases into one multi-value `case`
- Added four new tests covering both functions at the unit level and both profile types (`sampled`, `evented`) end-to-end through `Parse()`

## Impact

An write path caller could previously crash a handler goroutine by `POST /ingest?format=speedscope` with any non-standard `unit` string (e.g. `"INVALID"`). The panic propagated past the ingestion stack and was only recovered by Go's `net/http` top-level handler, producing a 500 response.

## Test plan

- [ ] `go test ./pkg/og/convert/speedscope/...` passes (7/7)
- [ ] `make lint` passes clean

---

> [!NOTE]
> **Medium Risk**
> Changes ingestion-time parsing behavior for attacker-controlled Speedscope JSON units from process panics to handled errors; low code churn but impacts error paths and profile rejection behavior.
> 
> **Overview**
> Prevents Speedscope ingestion from crashing on unrecognized `unit` values by changing `unit.defaultSampleRate()` and `unit.precisionMultiplier()` to return errors instead of `panic`.
> 
> Propagates these errors through `parser.go` so `Parse()` fails cleanly for both `sampled` and `evented` profiles when the unit is invalid, and adds tests asserting the new error behavior (including end-to-end parse cases that previously would have panicked).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f0dc3e807cf06a8a6e03a0c690a42d4833d8224b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/og/convert/speedscope/parser.go
+++ b/pkg/og/convert/speedscope/parser.go
@@ -125,7 +125,11 @@ func parseOne(prof *profile, putInput storage.PutInput, frames []frame, multi bo
 	// TODO(petethepig): We need a way to tell if it's a default or a value set by user
 	//   See https://github.com/pyroscope-io/pyroscope/issues/1598
 	if putInput.SampleRate == 100 {
-		putInput.SampleRate = uint32(prof.Unit.defaultSampleRate())
+		rate, err := prof.Unit.defaultSampleRate()
+		if err != nil {
+			return nil, fmt.Errorf("invalid profile unit: %w", err)
+		}
+		putInput.SampleRate = rate
 	}
 
 	var err error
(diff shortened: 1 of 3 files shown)


aleks-p (approved):
LGTM

PR #5145, opened 2026-05-12
fix(tree): add bounds checks to DeserializeNoDict

simonswine:
## Summary

- Varint-encoded values from the input stream were used directly in `make([]byte, nameLen)` and a children-count loop without any upper bound, which could cause a runtime panic or unbounded memory growth on malformed input
- Guards are sourced from existing per-tenant limits (`MaxProfileSymbolValueLength` for node name length, `MaxProfileStacktraceSamples` for children-per-node) — no new configuration surface introduced; `0` disables the check, consistent with all other limits
- Also fixes a silently-dropped error on the `nameLen` varint read

---

> [!NOTE]
> **Medium Risk**
> Touches the ingestion interface and tree-format parsing path used for profile ingestion; incorrect limit wiring or tenant ID extraction could reject valid payloads or change behavior, but changes are narrow and covered by new tests.
> 
> **Overview**
> Adds bounds checks to `tree.DeserializeNoDict` to prevent panics/unbounded memory growth on malformed tree-format payloads, returning explicit errors when node name lengths or child counts exceed configured maxima.
> 
> Threads per-tenant limits through the ingestion pipeline by extending `ingestion.RawProfile.Parse` and `Limits` interfaces, wiring limits from the pyroscope ingester adapter into tree parsing (`convert.ParseTreeNoDict`) so the new guards use `MaxProfileSymbolValueLength` and `MaxProfileStacktraceSamples`.
> 
> Updates parsers and tests (pprof/jfr/speedscope/tree profile) to the new `Parse(..., limits)` signature and adds regression tests covering oversized varints/children counts and limit enforcement behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 91ea150e55de97f74239fc3001bbde4cadafbe58. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/og/convert/parser.go
+++ b/pkg/og/convert/parser.go
@@ -3,14 +3,26 @@ package convert
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"io"
 	"strconv"
 
+	"github.com/grafana/pyroscope/v2/pkg/og/ingestion"
 	"github.com/grafana/pyroscope/v2/pkg/og/storage/tree"
+	"github.com/grafana/pyroscope/v2/pkg/tenant"
 )
 
-func ParseTreeNoDict(r io.Reader, cb func(name []byte, val int)) error {
-	t, err := tree.DeserializeNoDict(r)
+func ParseTreeNoDict(ctx context.Context, r io.Reader, cb func(name []byte, val int), limits ingestion.Limits) error {
+	var maxNameLen, maxChildren int
+	if limits != nil {
+		tenantID, err := tenant.ExtractTenantIDFromContext(ctx)
+		if err != nil {
+			return err
+		}
+		maxNameLen = limits.MaxProfileSymbolValueLength(tenantID)
+		maxChildren = limits.MaxProfileStacktraceSamples(tenantID)
+	}
+	t, err := tree.DeserializeNoDict(r, maxNameLen, maxChildren)
 	if err != nil {
 		return err
 	}
--- a/pkg/og/convert/profile/profile_test.go
+++ b/pkg/og/convert/profile/profile_test.go
@@ -1,15 +1,27 @@
 package profile
 
 import (
+	"bytes"
 	"context"
 	"testing"
 
 	"github.com/stretchr/testify/require"
 
+	"github.com/grafana/dskit/user"
+
 	"github.com/grafana/pyroscope/api/model/labelset"
 	"github.com/grafana/pyroscope/v2/pkg/og/ingestion"
 	"github.com/grafana/pyroscope/v2/pkg/og/storage"
 )
+
+type mockLimits struct {
+	maxSymbolLen int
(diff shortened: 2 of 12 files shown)


pkg/og/convert/profile/profile_test.go
```diff
@@ -88,3 +100,46 @@ var _ = Describe("metrics exporter", func() {
...
+		payload = append(payload, 0x01)
+		p := RawProfile{Format: ingestion.FormatTree, RawData: payload}
+		err := p.Parse(ctx, ingester, newMockExporter(false), md, mockLimits{maxSymbolLen: 65535, maxSamples: 16000})
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("exceeds maximum"))
+	})
+
+	It("does not enforce limits when nil is passed", func() {
+		// With nil limits the oversized varint reaches make([]byte, n) and panics
+		// without the serialize.go guard — the guard still fires because maxNameLen=0
```
marcsanmi: Should it be?
```suggestion
		// without the serialize.go guard — the guard doesn't fire because maxNameLen=0
```

pkg/og/convert/parser.go
```diff
@@ -3,14 +3,26 @@ package convert
...
-	t, err := tree.DeserializeNoDict(r)
+func ParseTreeNoDict(ctx context.Context, r io.Reader, cb func(name []byte, val int), limits ingestion.Limits) error {
+	var maxNameLen, maxChildren int
+	if limits != nil {
+		tenantID, err := tenant.ExtractTenantIDFromContext(ctx)
+		if err != nil {
+			return err
+		}
+		maxNameLen = limits.MaxProfileSymbolValueLength(tenantID)
+		maxChildren = limits.MaxProfileStacktraceSamples(tenantID)
```
marcsanmi, line 23: nit: `MaxProfileStacktraceSamples` is [documented](https://github.com/grafana/pyroscope/blob/d55b18e5775adc062962ee1848929ad30930ff02/pkg/validation/validate.go#L340) as the cap on total sample count per profile, but here it's "reused" as a per-node fanout cap, a different dimension. Should we update the flag help text to show the dual semantics.? Adding a new flag..?

simonswine: Each child will require a sample in the profile, so naturally the MaxProfileStacktraceSamples will be an upper bound to how many nodes can be children of a node.

Maybe I am misunderstanding your point?

marcsanmi: You're right. I was treating them as independent, but children-per-node is always ≤ total samples, so `MaxProfileStacktraceSamples` is a conservative bound.

marcsanmi (approved):
LGTM

PR #5153, opened 2026-05-13
refactor(queryplan): Refactor `queryplan.Build` and add table-driven tests

bryanhuhta:
## Summary

Rewrite the merge-tree construction loop in `Build` to use the same chunk-by-stride pattern as the leaf-creation loop: each parent's `Children` is now a single three-index slice (`nodes[i:end:end]`) instead of being built by a per-element head-pop and append loop.

Convert `Test_Plan` to a table-driven `Test_Build` and split the single golden file into per-case files under `testdata/`. Golden paths are derived from `t.Name()`, so adding a new case requires only a table row and one run with `-update`.

Expand edge-case coverage: empty input, `maxReads < 1`, `maxMerges < 2`, single block, exactly-one-leaf, leaf-with-leftover, perfectly-filled two-level plan, and the three-level case with a degenerate trailing merge. The previous 25-block scenario is preserved as `twenty_five_blocks`.

## Test plan
- [x] `go test ./pkg/querybackend/queryplan/` passes
- [x] `go test ./pkg/querybackend/queryplan/ -update` produces no diff in `testdata/` (goldens already match Build's output)
- [x] Spot-check the new golden files against a hand trace of the algorithm for each scenario

---

> [!NOTE]
> **Medium Risk**
> Touches query-plan construction logic (which can affect read/merge fanout and execution behavior), but is largely a mechanical refactor with expanded golden coverage to catch regressions.
> 
> **Overview**
> Refactors `queryplan.Build` to build leaf and merge nodes using preallocated contiguous node storage and chunk-by-stride slicing (`nodes[start:end:end]`), removing the previous pop/append merge loop and `math.Ceil` usage.
> 
> Reworks tests into a table-driven `Test_Build` with an `-update` flag and per-case golden files under `testdata/`, adding coverage for empty/invalid params and multiple tree shapes (including the existing 25-block scenario).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1e4108147ba4854f7a8fc6260183826ca49de05a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/querybackend/queryplan/query_plan.go
```diff
@@ -35,38 +33,39 @@ func Build(
 	}
 
 	// create leaf nodes and spread the blocks in a uniform way
-	var leafNodes []*queryv1.QueryNode
+	var nodes []*queryv1.QueryNode
```
simonswine: You can right size this too.

Nit: `[]*T` are kind of annoying but forced by proto generator. In theory that would get trid of the small allocations from append:

```
func allocateContiguousGeneric[T any](size int) []*T {
	// 1. Allocate all structs in one contiguous memory block
	values := make([]T, size)

	// 2. Allocate the slice of pointers
	pointers := make([]*T, size)

	// 3. Link pointers to the contiguous block
	for i := range values {
		pointers[i] = &values[i]
	}

	// Note: values stays alive because pointers holds references to its elements
	return pointers
}
```

simonswine (dismissed):

pkg/querybackend/queryplan/query_plan.go
bryanhuhta: This is a good point. Addressed in https://github.com/grafana/pyroscope/pull/5153/commits/6fea3e211b0d319d3510273b4e91c5665e93f0bb

simonswine (approved):

PR #5159, opened 2026-05-15
perf(tsdb/index): allocation-free postings offset table scan

simonswine:
## Summary

`newReader` spent ~96% of its allocations inside `ReadOffsetTable`: every entry materialised a `[]string` plus two string clones for name and value, even for the ~31/32 entries that symbolFactor sampling immediately discards.

This PR introduces `readPostingsOffsetTable`, a sibling of `ReadOffsetTable` that yields `(name, value []byte)` slices aliasing the mmap buffer instead of allocating strings. The `newReader` callback:

- uses `yoloString` for map lookups (zero allocation),
- defers `string(name)` / `string(value)` conversion until an entry is actually retained, and
- tracks the "last value" candidate as raw byte-slice aliases, cloning only at flush time.

## Impact

This path represents 66% of the allocation counts in query-backends (and a very small amount of compaction workers)

image

## Benchmark

`BenchmarkNewReader` with 10 label names × 200 values each:

```
             │      sec/op       │      sec/op      vs base  │
NewReader-12      305.1µ ± 1%       124.2µ ± 1%    -59.30%

             │       B/op        │       B/op       vs base  │
NewReader-12      270.7Ki ± 0%      169.8Ki ± 0%   -37.30%

             │     allocs/op     │   allocs/op    vs base    │
NewReader-12       6075.0 ± 0%       244.0 ± 0%   -95.98%
```

Allocation count drops from 6 075 → 244 (−96%), wall time from 305 µs → 124 µs (−59%).

## Notes

- `readPostingsOffsetTable` is unexported and used only by `newReader` and `PostingsRanges`.
- This builds on #5158 (FormatV1 reader support removed), which clears away the V1 branch that previously shared `ReadOffsetTable`.
- Inspired by the newer version of Prometheus code base: https://github.com/prometheus/prometheus/blob/43e5fc6a248c1a73d30dff86428f74d11aedbedc/tsdb/index/index.go#L1297-L1302

## Test plan

- [ ] `go test ./pkg/phlaredb/tsdb/index/... -run . -bench BenchmarkNewReader` passes and shows regression-free numbers
- [ ] Existing index reader tests (`TestNewReader`, table-driven cases) pass unchanged

---

> [!NOTE]
> **Medium Risk**
> Touches TSDB index reading/parsing and introduces more buffer-aliasing (`[]byte`/`yoloString`) to avoid allocations, which could cause subtle lifetime/aliasing bugs if misused despite being localized and covered by tests/bench.
> 
> **Overview**
> Reduces `NewReader` startup allocations/time by replacing the generic `ReadOffsetTable` postings scan with an allocation-free `readPostingsOffsetTable` that returns `(name, value []byte)` slices aliasing the index buffer, and only materializes `string` values for sampled/required entries (including correct handling of the per-label “last value” entry).
> 
> Updates `PostingsRanges` to use the new postings-table iterator, adjusts the label-indices test to decode the table directly, and adds `BenchmarkNewReader` to track open-index performance/regressions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e07c1e058ce94d818a9a6de5e15a89e4c7f135bc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/phlaredb/tsdb/index/index.go
+++ b/pkg/phlaredb/tsdb/index/index.go
@@ -589,7 +589,7 @@ func (w *Writer) finishSymbols() error {
 
 	hashPos := w.f.pos
 	// Leave space for the hash. We can only calculate it
-	// now that the number of symbols is known, so mmap and do it from there.
+	// now that the number of symbols is known, so seek back and write it.
 	if err := w.write([]byte("hash")); err != nil {
 		return err
 	}
@@ -1249,8 +1249,8 @@ func newReader(b ByteSlice, c io.Closer) (*Reader, error) {
 		return nil, errors.Wrap(err, "read symbols")
 	}
 
-	// lastNameB/lastValueB alias the mmap buffer and are only converted to
-	// owned strings if the entry turns out to be the final one for its label.
+	// lastNameB/lastValueB alias the underlying buffer and are only converted
+	// to owned strings if the entry turns out to be the final one for its label.
 	var lastNameB, lastValueB []byte
 	lastOff := 0
 	valueCount := 0
--- a/pkg/phlaredb/tsdb/index/index.go
+++ b/pkg/phlaredb/tsdb/index/index.go
@@ -1249,30 +1249,35 @@ func newReader(b ByteSlice, c io.Closer) (*Reader, error) {
 		return nil, errors.Wrap(err, "read symbols")
 	}
 
-	var lastKey []string
+	// lastNameB/lastValueB alias the mmap buffer and are only converted to
+	// owned strings if the entry turns out to be the final one for its label.
+	var lastNameB, lastValueB []byte
 	lastOff := 0
 	valueCount := 0
 	// For the postings offset table we keep every label name but only every nth
 	// label value (plus the first and last one), to save memory.
-	if err := ReadOffsetTable(r.b, r.toc.PostingsTable, func(key []string, _ uint64, off int) error {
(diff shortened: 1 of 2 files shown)


pkg/phlaredb/tsdb/index/index.go
```diff
@@ -1249,39 +1249,45 @@ func newReader(b ByteSlice, c io.Closer) (*Reader, error) {
 		return nil, errors.Wrap(err, "read symbols")
 	}
 
-	var lastKey []string
+	// lastNameB/lastValueB alias the mmap buffer and are only converted to
```
aleks-p: nit, maybe we can remove the mention of mmap here to avoid confusion, it looks like the data is always heap allocated (and it shouldn't matter in general)

aleks-p (dismissed):
Great improvement!

pkg/phlaredb/tsdb/index/index.go
simonswine: Likely coming from the prometheus context 👍

simonswine: Fixed in [e07c1e0](https://github.com/grafana/pyroscope/pull/5159/commits/e07c1e058ce94d818a9a6de5e15a89e4c7f135bc)

aleks-p (approved):

PR #5166, opened 2026-05-19
ci(update-contributors): Use GATB for update-contributors workflow

alsoba13:
Migrating old vault workflows to use the Github App Token Broker. 

Starting with update-contributors, because it's an easy one to test.


marcsanmi (approved):
LGTM

PR #5178, opened 2026-05-22
fix(ci): update helm-charts action version

alsoba13:
I missed to update the pinned version commit in the `helm-release.yml` workflow in #5171  which made the run [fail](https://github.com/grafana/pyroscope/actions/runs/26288284384/workflow).

Using the last commit in the helm-charts repo with the last improvements and fixes included.


simonswine (approved):
LGTM

PR #5180, opened 2026-05-22
feat(querybackend): per-tenant per-query bytes-fetched metrics

simonswine:
Tracks actual bytes read from object storage and metastore per query, exposing them as a `pyroscope_query_frontend_fetched_bytes_total` Prometheus counter and logging them (humanized, with latency) on every query log line. Counting is done at the `ReadCloser` level so short reads are not over-counted, MERGE nodes correctly sum bytes from all children, and the tenant label uses `tenant.JoinTenantIDs` (|-separated) to match the standard dskit multi-tenant org ID format.

### DIFF
--- a/pkg/frontend/readpath/queryfrontend/query_frontend.go
+++ b/pkg/frontend/readpath/queryfrontend/query_frontend.go
@@ -14,6 +14,7 @@ import (
 	"github.com/go-kit/log/level"
 	"github.com/grafana/dskit/tenant"
 	"github.com/grafana/dskit/tracing"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/prometheus/model/labels"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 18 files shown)


pkg/frontend/readpath/queryfrontend/query_frontend.go
```diff
@@ -56,6 +58,30 @@ type QueryFrontend struct {
...
+	m := &queryFrontendMetrics{
+		fetchedBytesTotal: prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Namespace: "pyroscope",
+				Subsystem: "query_frontend",
+				Name:      "fetched_bytes_total",
+				Help:      "Total bytes fetched per tenant per source (object_storage, metastore).",
+			},
+			[]string{"tenant", "kind"},
+		),
```
bryanhuhta, line 79: Do we also want to publish a metric that tracks estimated query size based off metastore metadata?

bryanhuhta (approved):
LGTM I ran this in dev for a number of days and found it useful. Could be helpful to have an estimated_bytes metric, but I'm happy to merge this as-is.

pkg/frontend/readpath/queryfrontend/query_frontend.go
simonswine: I guess we might loose the individual view on it, be probably want a histogram of "estimation accuracy". Will think about this more and follow it up with a PR.

simonswine: Done in https://github.com/grafana/pyroscope/pull/5252. Adds a `pyroscope_query_frontend_estimation_accuracy_ratio` histogram (ratio of pre-execution metadata size estimate to actual `BytesFetched`) with classic buckets centred around 1.0 and native histogram support. The ratio also surfaces on the existing "query finished" log line as `estimated_object_bytes` and `estimation_ratio`.

PR #5184, opened 2026-05-24
fix: repair Go standard library package generator

immanuwell:
Tiny fix, there was a problem with generator

Repro before this change:
```sh
go generate ./pkg/frontend/vcs/source/golang
```

It failed on missing `github.com/github/go-pipe/pipe` and the old `pkg/querier/golang` import.

This switches the generator to the current package path and `go/format`, then refreshes `packages_gen.go` for the current Go baseline. No extra moving parts.

Checked:
```sh
go generate ./pkg/frontend/vcs/source/golang
go test ./pkg/frontend/vcs/...
git diff --check
```


simonswine (approved):
LGTM

Thanks for updating this

PR #5186, opened 2026-05-24
fix(vcs): resolve Kotlin files from Java mappings

### LINKED ISSUE #4878: Support Kotlin (.kt) file extension in Java source code integration
## Problem

The current Java source code integration (`find_java.go`) hardcodes the `.java` file extension when resolving function names to source file paths:

```go
const ExtJava = ".java"
```

The `convertJavaFunctionNameToPath` function always appends `.java`:
- `com/example/MyClass.method` → `com/example/MyClass.java`

This means **Kotlin applications get 404 errors for every source code lookup**, since Kotlin source files use the `.kt` extension.

## Context

Kotlin is a JVM language officially supported by Google for Android and widely used in server-side applications (Spring Boot, Ktor, etc.). Kotlin compiles to JVM bytecode, so the async-profiler / JFR profiling agent captures Kotlin classes the same way as Java classes. However, when Pyroscope tries to fetch the source code from GitHub, it looks for `MyClass.java` instead of `MyClass.kt`, resulting in a 404.

## Reproduction

1. Set up a Kotlin project with `.pyroscope.yaml`:
   ```yaml
   version: v1
   source_code:
     mappings:
       - function_name:
           - prefix: com/example/myapp
         language: java
         source:
           github:
             owner: my-org
             repo: my-kotlin-app
             ref: main
             path: src/main/kotlin
   ```
2. Profile the application and click on any function in the flame graph
3. Pyroscope requests `src/main/kotlin/com/example/myapp/MyClass.java` from GitHub
4. GitHub returns 404 because the actual file is `MyClass.kt`

## Proposed Solution

When `language: java`, try multiple JVM file extensions before returning "not found":

1. Try `.java` first (current behavior)
2. If 404, try `.kt` (Kotlin)
3. Optionally try `.scala`, `.groovy` for other JVM languages

Alternatively, add a new `language: kotlin` option that uses `.kt` as the primary extension.

### Suggested change in `find_java.go`:

```go
var jvmExtensions = []string{".java", ".kt"}

func (ff *FileFinder) findJavaFile(ctx context.Context, mapping []MappingConfig) (*connect.Response[vcsv1.GetFileResponse], error) {
    span, ctx := opentracing.StartSpanFromContext(ctx, "findJavaFile")
    defer span.Finish()

    path := convertJavaFunctionNameToPath(ff.file)
    // path has .java extension, try .kt as fallback
    basePath := strings.TrimSuffix(path, ExtJava)
    
    for _, ext := range jvmExtensions {
        candidate := basePath + ext
        for _, m := range mapping {
            resp, err := ff.fetchMappingFile(ctx, candidate, m)
            if err == nil {
                return resp, nil
            }
            // continue on 404, return on other errors
        }
    }
    return nil, CodeNotFound
}
```

## Impact

This affects all JVM projects using non-Java languages:
- **Kotlin** — widely used in Android and server-side (Spring Boot)
- **Scala** — used in data engineering (Spark, Akka)
- **Groovy** — used in Gradle build scripts and some web apps

## Environment

- Grafana Cloud with Pyroscope
- Kotlin 1.9.x + Spring Boot 3.x application
- Profiled via Grafana Alloy with `pyroscope.java` component
- `.pyroscope.yaml` with 33 mappings, all returning 404 due to `.java` extension

officialasishkumar:
## Summary

Closes #4878

Java source mappings now keep the existing `.java` lookup first and then fall back to `.kt` when the Java source file is not found. This lets JVM applications written in Kotlin use the existing `language: java` mapping flow without changing the config schema.

The PR also adds an integration-style `FileFinder` test for a Kotlin local mapping and documents the fallback behavior in the source code integration guide.

## Test plan

- [x] `go test ./pkg/frontend/vcs/source`

### DIFF
--- a/docs/sources/view-and-analyze-profile-data/line-by-line.md
+++ b/docs/sources/view-and-analyze-profile-data/line-by-line.md
@@ -416,7 +416,7 @@ The system extracts Go version information from paths like `/usr/local/go/go1.24
 
 Java applications **require** explicit mappings in `.pyroscope.yaml`:
 
-- **Function name conversion**: Java function names like `org/example/App$Inner.method` are automatically converted to `org/example/App.java`
+- **Function name conversion**: Java function names like `org/example/App$Inner.method` are automatically converted to `org/example/App.java`. If the Java source file is not found, Pyroscope also tries the Kotlin `.kt` extension for JVM applications written in Kotlin.
 - **No fallback**: Unlike Go and Python, Java files cannot be resolved without configuration
 - **Inner classes**: Automatically handled (inner class markers are stripped)
 
--- a/pkg/frontend/vcs/source/find_java.go
+++ b/pkg/frontend/vcs/source/find_java.go
@@ -15,10 +15,17 @@ import (
 )
 
 const (
-	ExtJava = ".java"
+	ExtJava   = ".java"
+	ExtKotlin = ".kt"
 )
 
+var javaSourceFileExtensions = []string{ExtJava, ExtKotlin}
+
 func convertJavaFunctionNameToPath(functionName string) string {
+	return convertJavaFunctionNameToPathWithExt(functionName, ExtJava)
+}
+
+func convertJavaFunctionNameToPathWithExt(functionName, ext string) string {
 	pathSegments := strings.Split(functionName, "/")
 	last := len(pathSegments) - 1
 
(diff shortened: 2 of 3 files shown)


simonswine (approved):
This all looks good, before we merge it though we need a @officialasishkumar cla assistant green light.

jake-kramer:
Ping @officialasishkumar can you please sign the CLA?

PR #5189, opened 2026-05-26
fix(metastore/index): avoid stale shard cache reads in GetBlocks and QueryMetadata

simonswine:
## Summary

This fixes a metastore shard cache bug where a shard loaded from an older read transaction could remain in the shared cache and later be reused with newer block metadata from a different transaction snapshot.

That mismatch could cause:
- `QueryMetadata` to panic with `index out of range`
- `GetBlocks` to export metadata using a stale string table

## What changed

- add a persisted per-shard `Version` to `ShardIndex`
- bump and store the shard version on every shard write
- make shard cache reads version-aware
- reload cached read-only shards when the caller's transaction snapshot sees a newer shard version
- use the version-aware read path in both `QueryMetadata` and `GetBlocks`
- make direct shard version lookup return an error on malformed shard index data
- add a regression test covering the stale-read cache poisoning sequence

## Why

A stale read transaction could populate the shared shard cache with an older string table. Later, a fresh transaction could read newer block bytes for the same shard while reusing that cached shard snapshot. The version check prevents mixing block metadata from one snapshot with a string table from another.

## Testing

- `go test ./pkg/metastore/index/... -count=1`

## Notes

- The shard index format remains backward compatible:
  - new code can read old shard indexes without a version
  - old code ignores the trailing version bytes in the new format
- shards still on the old format report version `0` until rewritten by the new code

### DIFF
--- a/pkg/metastore/index/store/shard.go
+++ b/pkg/metastore/index/store/shard.go
@@ -78,19 +78,15 @@ func (s *Shard) Store(tx *bbolt.Tx, md *metastorev1.BlockMeta) error {
 		return err
 	}
 
-	var updateIndex bool
+	s.ShardIndex.Version++
 	if s.ShardIndex.MinTime == 0 || s.ShardIndex.MinTime > md.MinTime {
 		s.ShardIndex.MinTime = md.MinTime
-		updateIndex = true
 	}
 	if s.ShardIndex.MaxTime < md.MaxTime {
 		s.ShardIndex.MaxTime = md.MaxTime
-		updateIndex = true
 	}
-	if updateIndex {
-		if err = shardBucket.Put(tenantShardIndexKeyNameBytes, s.ShardIndex.MarshalBinary()); err != nil {
-			return err
-		}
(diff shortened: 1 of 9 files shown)


pkg/metastore/index/store/shard.go
```diff
@@ -78,19 +78,15 @@ func (s *Shard) Store(tx *bbolt.Tx, md *metastorev1.BlockMeta) error {
...
 		s.ShardIndex.MaxTime = md.MaxTime
-		updateIndex = true
 	}
-	if updateIndex {
-		if err = shardBucket.Put(tenantShardIndexKeyNameBytes, s.ShardIndex.MarshalBinary()); err != nil {
-			return err
-		}
+	if err = shardBucket.Put(tenantShardIndexKeyNameBytes, s.ShardIndex.MarshalBinary()); err != nil {
+		return err
 	}
```
simonswine, line 90: This obviously means that instead of only writing the index whenever the boundary changes, we are doing it for every write.

My gut feeling is this is okey as we do a store anyhow and value is much bigger than the `tenantShardIndex`.

I don't think there is an easy way around it though.

aleks-p (approved):
Makes sense to me.

PR #5190, opened 2026-05-26
ui: reduce dependencies

aleks-p:
Trims the list of dependencies for the ui project. Done primarily by vendoring `@grafana/flamegraph` in (while stripping things we don't need), replacing `@grafana/ui` and `@grafana/data` with inline components, dropping `react-use` and `@emotion/css` and replacing `vitest` with `node:test`. 

The flamegraph component could be further simplified in a future iteration.

I added e2e tests before doing the change, to make sure we capture the core functionalities without breakage. There could still be broken things though, the test suite is not all encompassing.

Before:

```
yarn info --recursive --name-only | sort -u | wc -l
     656
```

After:

```
yarn info --recursive --name-only | sort -u | wc -l
     278
```

I also couldn't resist and removed the blue tint in the UI, let me know if you have a strong opinion about this :) 

Original UI:

UI after #4948:

UI with this PR:


ui/e2e/flamegraph.spec.ts
```diff
@@ -0,0 +1,418 @@
+import { test, expect, type Page, type Locator } from '@playwright/test';
```
aleks-p, line 1: These tests don't run in the CI right now. If we want that, we'll need to rework them to run in Docker and regenerate snapshots.

ui/package.json
```diff
@@ -12,19 +12,24 @@
     "format": "prettier --check .",
     "format:fix": "prettier --write .",
     "preview": "vite preview",
-    "test": "vitest run"
```
bryanhuhta, line 15: Are you removing vitest to cut down on more deps?

When I opened this branch and ran `yarn test`, Node failed to run the tests properly:

```
$ yarn test
node:internal/modules/esm/get_format:219
  throw new ERR_UNKNOWN_FILE_EXTENSION(ext, filepath);
        ^

TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for /Users/bryan/grafana/pyroscope/ui/src/components/timeseries-utils.test.ts
    at Object.getFileProtocolModuleFormat [as file:] (node:internal/modules/esm/get_format:219:9)
    at defaultGetFormat (node:internal/modules/esm/get_format:245:36)
    at defaultLoad (node:internal/modules/esm/load:120:22)
    at async ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:514:32)
    at async ModuleJob._link (node:internal/modules/esm/module_job:115:19) {
  code: 'ERR_UNKNOWN_FILE_EXTENSION'
}

Node.js v22.14.0
✖ src/components/timeseries-utils.test.ts (33.365208ms)
ℹ tests 1
ℹ suites 0
ℹ pass 0
ℹ fail 1
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 39.929792

✖ failing tests:

test at src/components/timeseries-utils.test.ts:1:1
✖ src/components/timeseries-utils.test.ts (33.365208ms)
  'test failed'
```

bryanhuhta (dismissed):
Nice work. I took it for a spin locally and everything seems to work well. The unit tests might need to be spruced up. I have no opinion on using Node or Vitest, but the Node approach is broken, I think.

I know the e2e tests aren't part of CI (and in general I don't like e2e UI test 😅 ), but this PR has images generated from Darwin Chrome which won't really work with those generated in a (probably Linux-based) CI runner.

bryanhuhta (approved):

ui/package.json
aleks-p: > Are you removing vitest to cut down on more deps?

Yes, it adds [~20 depedencies](https://www.npmjs.com/package/vitest?activeTab=dependencies) but we only have one small test. From what I've read, `node:test` has been improving over the years and should be good enough for our needs.

> When I opened this branch and ran yarn test, Node failed to run the tests properly:

Enforced Node 24 with 008a4ab, that should help

aleks-p:
> I know the e2e tests aren't part of CI (and in general I don't like e2e UI test 😅 ), but this PR has images generated from Darwin Chrome which won't really work with those generated in a (probably Linux-based) CI runner.

They help greatly to reduce uncertainty (increase confidence) when doing large refactors like this one. They don't help as much when doing small things, especially if they get flaky. I'd keep them around for the next person who needs to touch this so they don't have to reinvent basic E2E tests.

PR #5191, opened 2026-05-26
feat: add optional admin HTTP server on separate port

simonswine:
Adds an optional secondary HTTP server (-admin-server.mode=disabled|additional|exclusive) that exposes operational endpoints (metrics, pprof, rings, config, admin/ops) on a dedicated port (default localhost:4042), so they can be firewalled independently from the public API port.

This is step one and a couple of extra things needs adressing:

- Include this in the helm chart so we can roll this out
- Decide how to handle query diagnostics, they currently sit weirdly in between admin and normal query API

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -14,6 +14,7 @@ import (
 	"connectrpc.com/connect"
 	"github.com/felixge/fgprof"
 	"github.com/go-kit/log"
+	"github.com/gorilla/mux"
 	"github.com/grafana/dskit/kv/memberlist"
 	"github.com/grafana/dskit/middleware"
 	"github.com/grafana/dskit/server"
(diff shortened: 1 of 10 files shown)


pkg/api/api.go
```diff
@@ -260,27 +332,27 @@ func (a *API) RegisterIngester(svc *ingester.Ingester) {
 }
 
 func (a *API) RegisterReadyHandler(handler http.Handler) {
-	a.RegisterRoute("/ready", handler, WithMethod("GET"))
```
marcsanmi, line 263: Was moving /ready onto the admin server intentional, or a side effect of routing everything non-data-plane through registerAdminRoute?

In exclusive mode it's removed from the main port and only served on the admin server (defaults to localhost:4042), so the helm readiness probe on the main service port would 404 and pods would never go Ready.

simonswine: Yes that was intentional as readiness error messages might return errors not to be seen for the public, there needs to be helm changes for metrics/profile collection anyhow in exclusive mode, so that is totally fine.

marcsanmi (approved):
LGTM

PR #5195, opened 2026-05-27
fix: clean up zip-slip/tar-slip in archive extraction

simonswine:
## Summary

Adds zip-slip/tar-slip path traversal prevention to `clearPath` in `pkg/embedded/grafana/assets.go`.

Previously, archive entries with paths like `../../evil.txt` could escape the destination directory when extracted. The fix resolves both the entry path and the destination to absolute paths and rejects any entry whose resolved path does not sit under the destination.

## Changes

- `clearPath` now returns `(string, error)` and rejects paths that escape `destPath`
- `extractZip` and `extractTarGz` propagate the new error
- Tests added covering `clearPath` directly (normal paths, strip behaviour, traversal variants) and integration-level extraction tests for both zip and tar.gz formats including filesystem assertions that no file escapes the destination

## Testing

```
go test ./pkg/embedded/grafana/...
```

### DIFF
--- a/pkg/embedded/grafana/assets.go
+++ b/pkg/embedded/grafana/assets.go
@@ -151,7 +151,7 @@ func (release *releaseArtifact) download(ctx context.Context, logger log.Logger,
 	return targetPath, nil
 }
 
-func clearPath(name string, destPath string, stripComponents int) string {
+func clearPath(name string, destPath string, stripComponents int) (string, error) {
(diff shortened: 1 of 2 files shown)


korniltsev-grafanista (approved):
lgtm

considering current week theme I'd probably consider removing the whole embeded grafana feature

PR #5196, opened 2026-05-27
fix: don't panic on unknown QueryNode type

simonswine:
Replace unconditional panic in QueryBackend.Invoke for unknown
node types with a proper connect.CodeInvalidArgument error, preventing
crash via a crafted query plan.


marcsanmi (approved):
LGTM

PR #5203, opened 2026-05-28
fix(helm): add apiVersion and kind to volumeClaimTemplates

slimes28:
When deploying chart with ArgoCD, using server-side apply, those fields are added automatically, and the sts resource stays out of sync due to that
image


simonswine (approved):
LGTM!

slimes28:
@simonswine Do I also need to bump Chart.yaml version, or are there some weekly releases?

PR #5206, opened 2026-05-28
fix: close files properly in extractZip and extractTarGz

immanuwell:
## Summary

extractZip and extractTarGz in pkg/embedded/grafana/assets.go leak file descriptors every time grafana assets are extracted.

In extractZip, both fileInArchive (the zip entry reader) and outFile are opened but never get closed on any code path. In extractTarGz, outFile has an explicit Close() on the happy path but if io.Copy fails the fd is just... gone.

The fix pulls the open/copy/close logic into small helper fns so defer handles cleanup properly on all paths.

## How to reproduce

Run pyroscope with `--target all,embedded-grafana` on a fresh setup and lsof/strace the pid while grafana assets are being extracted. You'll see unclosed fds piling up for each file in the archive. A typical grafana release has a few hundred files, so this ain't trivial.

You can also just read the code: extractZip opens fileInArchive and outFile in a loop with no defer and no explicit Close anywhere.

Related: #5195 touched this same file and code path.

### DIFF
--- a/pkg/embedded/grafana/assets.go
+++ b/pkg/embedded/grafana/assets.go
@@ -201,22 +201,44 @@ func extractZip(zipStream io.ReaderAt, size int64, destPath string, stripCompone
 			}
 		}
 
-		fileInArchive, err := f.Open()
-		if err != nil {
-			return fmt.Errorf("ExtractZip: Open() failed: %s", err.Error())
-		}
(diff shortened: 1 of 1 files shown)


simonswine (approved):
Thanks! Makes total sense

PR #5209, opened 2026-05-29
fix: nil dereference in embedded grafana download + drop deprecated ioutil

immanuwell:
Two small fixes in unrelated files, both pretty straightforward.

`pkg/embedded/grafana/assets.go` - nil deref in download

`req.Header.Set(...)` was called before checking the error from `http.NewRequestWithContext`. If the URL fails to parse, `req` is nil and the call panics instead of returning a clean error.

Reproduce:
```go
req, err := http.NewRequestWithContext(ctx, "GET", "://bad-url", nil)
req.Header.Set("User-Agent", "test") // panics here, req is nil
if err != nil { ... }
```

Every other `NewRequestWithContext` callsite in the codebase checks `err` first - this one slipped thru. Fix swaps the two lines.

`pkg/util/copy.go` - deprecated ioutil

`ioutil.ReadDir` has been deprecated since Go 1.16, replaced by `os.ReadDir`. Updated the call and switched `entry.Mode()` to `entry.Type()` since `os.DirEntry` exposes type bits via `Type()`.

### DIFF
--- a/pkg/embedded/grafana/assets.go
+++ b/pkg/embedded/grafana/assets.go
@@ -77,10 +77,10 @@ func (release *releaseArtifact) download(ctx context.Context, logger log.Logger,
 
 	level.Info(logger).Log("msg", "download new release", "url", release.URL)
 	req, err := http.NewRequestWithContext(ctx, "GET", release.URL, nil)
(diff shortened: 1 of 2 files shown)


aleks-p (approved):
Thank you @immanuwell

PR #5217, opened 2026-05-30
feat(debuginfo): add list and delete APIs with profilecli support

luweglarz:
Debug info upload existed, but there was no easy way to:
  - see what has already been uploaded
  - remove stale or incorrect entries
  
  So this PR aim to be able to List and Delete uploaded debuginfo, by adding `ListDebuginfo` and `DeleteDebuginfo` endpoints, with an extension of the profilecli to target them with the following commands:
- `debuginfo list`
- `debuginfo delete`

# Example usage:

### 1) List

```bash
go run ./cmd/profilecli debuginfo list --url http://localhost:14040 --tenant-id pr-demo-tenant
```

```text
build_id=05dad2c279f7651722809fa75adba6bf9ab1c209 name=ls type=TYPE_EXECUTABLE_FULL state=STATE_UPLOADED size=139.0 KiB uploaded_at=2026-05-30T11:37:01Z
build_id=2f77b36371c214e11670c7d9d92727e9a49f626b name=bash type=TYPE_EXECUTABLE_FULL state=STATE_UPLOADED size=1.4 MiB uploaded_at=2026-05-29T13:00:49Z
```

### 2) Delete

```bash
go run ./cmd/profilecli debuginfo delete 05dad2c279f7651722809fa75adba6bf9ab1c209 --url http://localhost:14040 --tenant-id pr-demo-tenant
```

```text
deleted debuginfo build_id=05dad2c279f7651722809fa75adba6bf9ab1c209
```

### 3) List again

```bash
go run ./cmd/profilecli debuginfo list --url http://localhost:14040 --tenant-id pr-demo-tenant
```

```text
build_id=2f77b36371c214e11670c7d9d92727e9a49f626b name=bash type=TYPE_EXECUTABLE_FULL state=STATE_UPLOADED size=1.4 MiB uploaded_at=2026-05-29T13:00:49Z
```


korniltsev-grafanista:
Hi. Thanks for contribution. I will ask my colleague opinions on this.

While we discuss this, may I ask if this is something that you operate and need or just trying to improve the situation for other operators?. I am asking because this feature is an alpha stage and was mostly merged for grafana to performance-test the symbolizer in our dev cluster. We don't have immediate plans to finish/improve it(we will I just dont know when). And so we don't know the final shape of the feature, and all of this can be changed/removed. 

Iam just trying to understand that someone already need this to operate or not

korniltsev-grafanista:
Also CI is red, please fix

luweglarz:
Hi thanks for your answer, as I want to experiment this symbolization feature I wanted to anticipate potential use case and improvements, operationally it is not urgent.

May I ask what are the plans for this feature? it is one I am really interested in and looking forward to experiment it soon :)

korniltsev-grafanista:
We don't know the future of this feature yet. Hopefully some time soon there will be an OTEL protocol for symbols upload and we remove all of our custom inventions. Until then we may productionise this feature if there are enough demand from GL customers. Other than that idk. Definetely no planned for this Q.

korniltsev-grafanista (dismissed):

korniltsev-grafanista (dismissed):

pkg/debuginfo/store.go
```diff
@@ -113,6 +113,72 @@ func (s *Store) checkShouldInitiateUpload(
...
+		return nil, connect.NewError(connect.CodeInternal, err)
+	}
+
+	id, err := ValidateGnuBuildID(req.Msg.GetGnuBuildId())
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("invalid gnu_build_id: %w", err))
+	}
+
+	deleted := false
+	for _, objectPath := range []string{ObjectPath(tenantID, id), MetadataObjectPath(tenantID, id)} {
```
marcsanmi: Could we delete the metadata before the binary? If the binary delete succeeds but the metadata delete then fails, the metadata stays in `STATE_UPLOADED` pointing at a missing binary, and `ShouldInitiateUpload` returns `ReasonDebuginfoAlreadyExists` blocking re-upload until someone deletes again. Metadata-first makes a partial failure self-healing via the normal upload path.

pkg/debuginfo/store.go
```diff
@@ -113,6 +113,72 @@ func (s *Store) checkShouldInitiateUpload(
...
+		case err == nil:
+			deleted = true
+		case s.bucket.IsObjNotFoundErr(err):
+			continue
+		default:
+			return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("failed to delete debuginfo object %q: %w", objectPath, err))
+		}
+	}
+
+	if !deleted {
```
marcsanmi: CodeNotFound here is backend-dependent. S3 and filesystem return nil when the key is missing, so deleted is always true and this path is unreachable; only GCS/Azure/in-memory return a not-found error. I I think test passes only because it uses the in-memory bucket. Maybe either pre-check with Attributes/Exists to return NotFound reliably, or drop the NotFound contract and treat delete as idempotent?

pkg/debuginfo/store.go
luweglarz: Thanks for the catch! I also considered adding an additional guard on `ShouldInitiateUpload` to check if the binary is present or not but deleting the metadata before the binary like you propose is simpler

pkg/debuginfo/store.go
luweglarz: Thanks, I agree I dropped the NotFound-on-missing contract and made delete idempotent

korniltsev-grafanista (approved):

PR #5218, opened 2026-05-31
fix: include remote write URL in error messages

### LINKED ISSUE #1961: [remoteWrite] add more info (like url) when showing errors
```
time="2023-05-25T15:55:10.103354" level=error msg="error happened while ingesting data" file=" remotewrite/queue.go:96" error="response not ok: status code: '404'. body: '404 page not found\n'" key="foo.cpu{}" remote_target=bar
```

This message would be more helpful if it had the url which is trying to be hit.

RoySerbi:
Fixes #1961

When remote write fails, the error message only shows the status code but not the URL being hit. This makes debugging difficult when multiple remote targets are configured.

Added the remote write URL to the error log message so operators can immediately identify which target is failing.

Before:
```
level=error msg="unable to store prompb.WriteRequest" err="response not ok: status code: '404'"
```

After:
```
level=error msg="unable to store prompb.WriteRequest" url="http://localhost:9009/api/v1/write" err="response not ok: status code: '404'"
```

### DIFF
--- a/pkg/metrics/exporter.go
+++ b/pkg/metrics/exporter.go
@@ -29,6 +29,7 @@ type StaticExporter struct {
 	wg     sync.WaitGroup
 
 	logger log.Logger
+	url    string
 
 	metrics *clientMetrics
 }
@@ -47,6 +48,7 @@ func NewExporter(remoteWriteAddress string, logger log.Logger, reg prometheus.Re
 		client:  client,
 		wg:      sync.WaitGroup{},
 		logger:  logger,
+		url:     remoteWriteAddress,
 		metrics: metrics,
 	}, nil
 }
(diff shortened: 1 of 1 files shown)


pkg/metrics/exporter.go
```diff
@@ -65,7 +67,7 @@ func (e *StaticExporter) Send(tenantId string, data []prompb.TimeSeries) error {
 		ctx := tenant.InjectTenantID(context.Background(), tenantId)
 		_, err := e.client.Store(ctx, snappy.Encode(nil, buf.Bytes()), 0)
 		if err != nil {
-			level.Error(e.logger).Log("msg", "unable to store prompb.WriteRequest", "err", err)
+			level.Error(e.logger).Log("msg", "unable to store prompb.WriteRequest", "url", e.url, "err", err)
```
korniltsev-grafanista, line 70:
```suggestion
			level.Error(e.logger).Log("msg", "unable to store prompb.WriteRequest", "url", e.url.Redacted(), "err", err)
```

korniltsev-grafanista: can we use https://pkg.go.dev/net/url#URL.Redacted ?

korniltsev-grafanista:
Note: the linked issue is incorrect and outdated and has already been solved

RoySerbi:
Thanks @korniltsev-grafanista — closing since the underlying issue is already resolved.

korniltsev-grafanista:
I did not mean to shut down the change request, just wanted to point out this is slightly different code.
It would not hurt adding an url to the logline as you did.

PR #5222, opened 2026-06-01
chore: Remove Go dependencies

bryanhuhta:
Removes unnecessary or replaceable Go dependencies.

Note that https://github.com/grafana/pyroscope/pull/5223 contains the changes necessary to remove the `github.com/pkg/errors` package, but that change touches ~100 files. In the interest of making this PR easier to review, I've split them out.

- `gonum.org/v1/plot` only used in a permanently-skipped test, deleted the dead test file
- `github.com/aybabtme/rgbterm` unmaintained since 2017, replaced with `fatih/color` which was already a dependency
- `github.com/briandowns/spinner` single cosmetic animation in profilecli, removed
- `github.com/k0kubun/pp/v3` + `github.com/mattn/go-isatty` replaced with `encoding/json` in profilecli output
- `github.com/gogo/status` all call sites already used `google.golang.org/grpc/status`
- `github.com/json-iterator/go` only used in non-hot telemetry reporting, replaced with `encoding/json`
- `github.com/hashicorp/go-multierror` single usage replaced with stdlib `errors.Join`
- `github.com/iancoleman/strcase` single usage replaced with a small helper in `pkg/util`
- `sigs.k8s.io/yaml` replaced with `go.yaml.in/yaml/v3` which was already a direct dependency
- `gopkg.in/yaml.v3` migrated all imports to `go.yaml.in/yaml/v3`, removing the stale 2022 fork replace directive
- `github.com/sirupsen/logrus` legacy logger only in `pkg/og/`


korniltsev-grafanista (approved):

go.work.sum
```diff
@@ -800,6 +800,8 @@ github.com/hashicorp/raft-boltdb/v2 v2.2.2 h1:rlkPtOllgIcKLxVT4nutqlTH2NRFn+tO1w
 github.com/hashicorp/raft-boltdb/v2 v2.2.2/go.mod h1:N8YgaZgNJLpZC+h+by7vDu5rzsRgONThTEeUS3zWbfY=
 github.com/huandu/xstrings v1.3.2 h1:L18LIDzqlW6xN2rEkpdV8+oL/IXWJ1APd+vsdYy4Wdw=
 github.com/huandu/xstrings v1.3.2/go.mod h1:y5/lhBue+AyNmUVz9RLU9xbLR0o4KIIExikq4ovT0aE=
+github.com/iancoleman/strcase v0.3.0 h1:nTXanmYxhfFAMjZL34Ov6gkzEsSJZ5DbhxWjvSASxEI=
```
korniltsev-grafanista, line 803: I wonder why does this add an entry here if we removed them?

Should we remove go.work alltogether as a followup?

bryanhuhta: The distinction is that `go.mod.sum` tracks the import graph while `go.work.sum` tracks the module graph. The import graph is all the modules that have been imported (this is why `strcase` doesn't appear in and `go.mod.sum`) but the module graph is a collection of all direct and indirect dependencies.

This chain of dependencies imports `strcase`:

- pkg/metrics imports prometheus/storage/remote imports envoyproxy/protoc-gen-validate@v1.3.3 (via discovery/xds)
- protoc-gen-validate's go.mod declares github.com/iancoleman/strcase as one of its requirements

The package that is actually imported (protoc-gen-validate/validate) doesn't import strcase, so strcase is never compiled into the build.

I'm in favor of removing `go.work.sum` as a follow up. I've never been completely convinced we need to checked in to begin with and it is very sensitive to changes.

PR #5225, opened 2026-06-02
examples(dotnet): switch to GH releases, improve update script

aleks-p:
Two improvements to the .NET examples

- switch away from the grafana/pyroscope-dotnet image for sourcing the profiler binarier, in favour of GitHub releases - to be more consistent with other examples
- fix the tracing example (it was using an incompatible version of the OTEL managed helper)

Also adapts the update examples script so that future updates will respect the binary sources and will also take into account examples/tracing/dotnet

### DIFF
--- a/tools/update_examples.go
+++ b/tools/update_examples.go
@@ -122,21 +122,29 @@ func updateDotnet() {
 	last := tags[len(tags)-1]
 	fmt.Println(last)
 
-	reDockerGlibc := regexp.MustCompile(`grafana/pyroscope-dotnet:\d+\.\d+\.\d+-glibc`)
-	replDockerGlibc := fmt.Sprintf("grafana/pyroscope-dotnet:%s-glibc", last.version())
-	replaceInplace(reDockerGlibc, "examples/language-sdk-instrumentation/dotnet/fast-slow/Dockerfile", replDockerGlibc)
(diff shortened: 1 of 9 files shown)


korniltsev-grafanista (dismissed):

tools/update_examples.go
```diff
@@ -345,6 +353,26 @@ func extractDotnetVersion() func(tag Tag) *version {
 	}
 }
 
+// extractDotnetOtelVersion matches the Pyroscope.OpenTelemetry helper release
+// tags (e.g. "v0.4.0-opentelemetry"), which are versioned independently of the
+// native profiler tags.
+func extractDotnetOtelVersion() func(tag Tag) *version {
+	return func(tag Tag) *version {
+		re := regexp.MustCompile(`^v(\d+)\.(\d+)\.(\d+)-opentelemetry$`)
```
korniltsev-grafanista: With release-please I had to change the tagging scheme as release-please does not support suffixes, only prefixes. It would likely be `opentelemetry-*` etc.

aleks-p: thanks, made an update to handle the transition - once we have all components released under the new names we can simplify the logic slightly

korniltsev-grafanista (approved):

PR #5233, opened 2026-06-03
fix: helm query backend SRV dns

apenen:
## Description
This PR resolves an issue in the Pyroscope Helm chart microservices (v2) mode where the `query-frontend` and other query-related components experience 30-second timeouts (`context deadline exceeded`) when
     trying to dispatch `SelectSeries` requests to the backends.

## Root Cause
In the default target address for `-query-backend.address` is generated by appending an explicit port suffix (`:{{ .Values.pyroscope.grpc.port }}`) or 9095 to an SRV-prefixed domain name:

`"-query-backend.address=dns:///_grpc._tcp.pyroscope-dev-headless.$(NAMESPACE_FQDN):9095"`
`"-query-backend.address=dns:///_grpc._tcp.{{ include "pyroscope.fullname" . }}{{ if $hasQueryBackend }}-query-backend{{ end }}-headless.$(NAMESPACE_FQDN):{{ .Values.pyroscope.grpc.port }}"`

According to the **gRPC Name Resolution specification**:
1. When using the `dns:///` scheme, if a target address **explicitly contains a port suffix** (such as `:9095`), the gRPC DNS resolver **bypasses SRV record lookups** entirely.
2. It then falls back to performing standard **A/AAAA (host) record lookups** directly on the entire target string—including the `_grpc._tcp.` prefix.
3. Because standard Kubernetes DNS (CoreDNS) does **not** create A/AAAA records for SRV names (which only have SRV records), the DNS lookup fails with `NXDOMAIN`.
4. Consequently, the `query-frontend` resolves **0 backend endpoints** and hangs waiting for backends to appear, eventually timing out after 30 seconds.

*Note: The `-metastore.address` is not affected by this bug because it uses the `kubernetes:///` scheme (from `kuberesolver`), which handles explicit ports correctly.*

## Solution
Remove the explicit port suffix from the SRV-prefixed DNS target address for the `query-backend` in the templates.

By removing the explicit port, the standard Go gRPC DNS client is allowed to perform the SRV record query properly, natively extracting both the backend pod hostnames and their respective ports from the returned SRV record.

### DIFF
--- a/operations/pyroscope/helm/pyroscope/rendered/legacy-micro-services-hpa.yaml
+++ b/operations/pyroscope/helm/pyroscope/rendered/legacy-micro-services-hpa.yaml
@@ -2220,7 +2220,7 @@ spec:
             - "-runtime-config.file=/etc/pyroscope/overrides/overrides.yaml"
             - "-log.level=debug"
             - "-store-gateway.sharding-ring.replication-factor=3"
-            - "-query-backend.address=dns:///_grpc._tcp.pyroscope-dev-headless.$(NAMESPACE_FQDN):9095"
+            - "-query-backend.address=dns:///_grpc._tcp.pyroscope-dev-headless.$(NAMESPACE_FQDN)"
             - "-metastore.address=kubernetes:///pyroscope-dev-headless.$(NAMESPACE_FQDN):9095"
             - "-architecture.storage=v2"
           env:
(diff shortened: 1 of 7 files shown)


apenen:
When the query-backend is run as a StatefulSet, SRV DNS resolution works seamlessly because of the headless service. However, if it is deployed as a Deployment,
  SRV DNS resolution is not supported. For those cases, users can easily override the address using extraArgs to use standard resolution with a port suffix.

PR #5241, opened 2026-06-04
ci(release): open a PR for Homebrew formula updates instead of pushing to main

marcsanmi:
## What
Changes the release workflow's "Update homebrew formulas" step to **open a PR** against `grafana/homebrew-pyroscope` instead of pushing directly to its `main`.

## Why
`grafana/homebrew-pyroscope`'s `main` is protected by an **org-level ruleset ("Public repo required PRs")** that requires reviewed PRs. The migrated release pipeline still did `git push origin main`, which is rejected with `GH013`. This is exactly what failed during the v2.0.3 release. See run [26893483450](https://github.com/grafana/pyroscope/actions/runs/26893483450/job/79390975828).

## How
The step now creates a branch `update-formulas-<tag>`, runs `make generate-formulas`, and opens a PR via `gh pr create`. A maintainer reviews and merges it to publish via `brew`. If there are no formula changes, it no-ops; if the PR already exists, the branch is just updated.

Also documents this in `docs/internal/RELEASE.md`, noting the step doesn't block the rest of the release.

## Notes
- Tradeoff: the formula PR still needs **1 human approval** , so Homebrew publishing is one manual merge per release rather than fully automated. The release itself is unaffected.


.github/workflows/release.yml
```diff
@@ -72,7 +72,25 @@ jobs:
...
+          # so we open a PR for a maintainer to merge instead of pushing to main.
+          gh pr create --repo grafana/homebrew-pyroscope \
+            --base main --head "${branch}" \
+            --title "chore: update formulas to ${PYROSCOPE_TAG}" \
+            --body "Automated formula update for Pyroscope ${PYROSCOPE_TAG}, generated by the release workflow. Review and merge to publish the new version via Homebrew." \
+            || echo "A pull request for ${branch} already exists; it was updated with the latest commit."
         env:
+          # App token scoped to grafana/{homebrew-pyroscope,pyroscope}, used for
+          # the git push, the formula generator, and `gh pr create`.
           HOMEBREW_GITHUB_TOKEN: ${{ steps.get-github-app-token.outputs.token }}
```
korniltsev-grafanista: can we keep only one env variable? they look the same

.github/workflows/release.yml
```diff
@@ -72,7 +72,25 @@ jobs:
           git config --global user.name "Pyroscope Bot <dmitry+bot@pyroscope.io>"
           git clone https://github.com/grafana/homebrew-pyroscope ../homebrew-pyroscope
           cd ../homebrew-pyroscope
-          make generate-formulas && git add Formula && git commit -m "chore: update formulas" && git push origin main
+          branch="update-formulas-${PYROSCOPE_TAG}"
+          git checkout -b "${branch}"
+          make generate-formulas
+          if git diff --quiet -- Formula; then
```
korniltsev-grafanista: why wold there be no changes? I would consider either failing the step, or removing the check at all

.github/workflows/release.yml
```diff
@@ -72,7 +72,25 @@ jobs:
...
+          git add Formula
+          git commit -m "chore: update formulas to ${PYROSCOPE_TAG}"
+          git push --force-with-lease origin "${branch}"
+          # The homebrew-pyroscope default branch requires reviewed PRs (org ruleset),
+          # so we open a PR for a maintainer to merge instead of pushing to main.
+          gh pr create --repo grafana/homebrew-pyroscope \
+            --base main --head "${branch}" \
+            --title "chore: update formulas to ${PYROSCOPE_TAG}" \
+            --body "Automated formula update for Pyroscope ${PYROSCOPE_TAG}, generated by the release workflow. Review and merge to publish the new version via Homebrew." \
+            || echo "A pull request for ${branch} already exists; it was updated with the latest commit."
```
korniltsev-grafanista: same, I would prefer to see a failure instead of a silent log lien in logs

korniltsev-grafanista (dismissed):
left couple of nits, consider making it branchless and fail fast.

lgtm in general, feel free to ignore ^^

aleks-p (approved):

PR #5245, opened 2026-06-06
fix: allow fixed delays in delayed bucket client

immanuwell:
`DelayedBucketClient` blows up when `minDelay == maxDelay`.

That config is legit for tests. `0,0` means no delay, `1ms,1ms` means fixed delay. Right now it hits `rand.Int63n(0)` and panics, kinda rough.

This patch uses a plain sleep when both bounds are equal, and adds a regression test.

Repro:
1. `b := objstore.NewDelayedBucketClient(memory.NewInMemBucket(), 0, 0)`
2. `b.Exists(context.Background(), "x")`
3. before: panic `invalid argument to Int63n`
4. after: returns `ok=false err=<nil>`

Test:
`go test ./pkg/objstore/... -count=1`

### DIFF
--- a/pkg/objstore/delayed_bucket_client.go
+++ b/pkg/objstore/delayed_bucket_client.go
@@ -112,5 +112,10 @@ func (m *DelayedBucketClient) Provider() objstore.ObjProvider {
 }
 
 func (m *DelayedBucketClient) delay() {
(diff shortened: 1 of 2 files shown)


simonswine (approved):
LGTM

PR #5247, opened 2026-06-09
feat: limit number of recording rules per tenant

alsoba13:
## What this does

Adds a new per-tenant limit, `max_recording_rules` (default `25`), that caps how many recording rules a tenant can create and store. When a tenant hits the limit, an upsert that would create a new rule fails with a `ResourceExhausted` error. The default can be raised (or lowered) per tenant via runtime-config `overrides`.

```yaml
overrides:
  my-tenant:
    max_recording_rules: 50
```

Set to `0` to disable the limit.

## How it works

- **New limit** `MaxRecordingRules` added to `pkg/validation` with the flag `-recording-rules.max-rules-per-tenant` (default `25`) and an `Overrides.MaxRecordingRules(tenantID)` accessor, following the existing per-tenant limit pattern.
- **Enforced atomically inside the store's write transaction.** A naive "list, count, then upsert" in the handler would race under concurrent creates and could exceed the cap. Instead, `GenericStore.Upsert` takes a `maxElements` guard that's checked while the write lock is held.
- **Only applies to new rules.** Updates to an existing rule never increase the count, so they're never blocked — even at the limit.
- **Provisioned rules don't count.** Operator-defined rules coming from config (`recording_rules` overrides) live outside the store and aren't tenant-creatable, so they're excluded from the cap. The limit governs only what a tenant creates itself.
- On the API path, `ErrMaxElementsExceeded` is mapped to `connect.CodeResourceExhausted` with the message `recording rules limit (N) exceeded for this tenant`.

## Tests

- `store_test.go`: cap enforcement at the generic-store level (insert to limit, reject beyond, updates allowed at limit, `0` disables).
- `recording_test.go`: handler-level `TestRecordingRules_MaxRecordingRules` — create up to limit → reject → update-existing still allowed → delete-then-create allowed.

## Generated files

Regenerated config reference docs (`docs/.../reference-configuration-parameters/index.md`) and `cmd/pyroscope/pyroscope.yaml`.


simonswine (dismissed):
LGTM. 10 feels a bit low, not too sure though. 

What would be good to check if the UI surfaces the error in a good way

alsoba13:
> LGTM. 10 feels a bit low, not too sure though.
> 
> What would be good to check if the UI surfaces the error in a good way

I bumped the default limit up to 25

on the UI, I had to fix that: https://github.com/grafana/profiles-drilldown/pull/1014

simonswine (approved):

PR #5248, opened 2026-06-09
chore: Ensure examples have consistent services

simonswine:
## What

This PR removes the duplication across our `examples/` directory by introducing
a single source of truth for the shared services (Grafana, Tempo & Pyroscope).
Each example now provisions these services from common template files under
`examples/templates/`, so configuration stays consistent across every example.

It also fixes the Tempo configuration so it works with the Traces Drilldown app,
and aligns the Pyroscope/Tempo datasource provisioning across all examples.

> Replaces https://github.com/grafana/pyroscope/pull/4329

## How

- Added `examples/templates/` containing the canonical `docker-compose` service
  definitions and provisioning files for `grafana`, `tempo` and `pyroscope`.
- Added `examples/templates/sync_test.go`, a test-driven sync/check tool driven
  by the `EXAMPLE_CHECK_MODE` env var:
  - `check` — verifies every registered example matches its templates
    (services checked structurally, other config files byte-for-byte).
  - `update` — rewrites the example files from the templates.
  - The `examples` slice in the test is the authoritative registry of which
    templates apply to each example. `serviceVolumes` / `fileAppends` allow an
    example to legitimately extend a templated service or config file while
    still enforcing the shared prefix.
- Added Makefile targets:
  - `make examples/sync-templates` → regenerate examples from templates.
  - `make examples/check-templates` → verify examples are in sync.
- Added a `check-example-templates` CI job that runs `make examples/check-templates`
  so drift is caught automatically.
- Regenerated all example `docker-compose.yml` and provisioning files from the
  templates, and updated Tempo configs to work with Traces Drilldown.

## Why replace #4329

The previous approach (#4329) generated examples via a standalone Go "script".
This PR instead leans on the standard Go test tooling (`go test` in `check` /
`update` modes), which removes the separate build step, wires drift detection
directly into CI, and keeps the template registry alongside the assertions.


examples/templates/sync_test.go
```diff
@@ -0,0 +1,525 @@
...
+    url: http://prometheus:9090
+    basicAuth: true #username: admin, password: admin
+    basicAuthUser: admin
+    jsonData:
+      manageAlerts: true
+      prometheusType: Prometheus #Cortex | Mimir | Prometheus | Thanos
+      prometheusVersion: 2.40.0
+    secureJsonData:
+      basicAuthPassword: admin #https://grafana.com/docs/grafana/latest/administration/provisioning/#using-environment-variables
+`,
```
aleks-p: this feels like something that might be difficult to maintain, not sure what the alternative would be, can it be moved to a prometheus.yml file not managed by the template system?

aleks-p (dismissed):
Thanks for picking this up again!

Looks great overall to me, I left one remark - but not critical.

Also, the example runner I added in #5221 probably picks up the template directory, we might need to adapt it slightly. Lastly, we probably also want to run those tests in any PR touching examples (like this one), that could be done in a separate PR (we need to ensure we don't allow arbitrary code execution).

examples/templates/sync_test.go
simonswine: Agreed, I have created a prometheus template responsible for it. (It doesn't contain prometheus's config as it contains local services which are quite specific)

examples/_templates/pyroscope/docker-compose.yml
```diff
@@ -0,0 +1,5 @@
+services:
+  pyroscope:
+    image: grafana/pyroscope:latest
```
aleks-p, line 3: should we pin this or will that complicate things as it will be hard to keep up to date?

examples/_templates/pyroscope/grafana-provisioning/datasources/pyroscope.yml
```diff
@@ -0,0 +1,14 @@
+---
```
aleks-p, line 1: should this file be under pyroscope/grafana-provisioning/datasources to be in line with the rest, or do we keep it here since it is always included?

examples/tracing/tempo/grafana/provisioning/plugins/explore-profiles.yml
aleks-p, line 1: I found 10+ more explore-profiles.yml files left, can we remove them?

aleks-p (commented):
Looking great, I added a few more questions and the only other thing left is skipping the _template directory from make examples/test as it reports failures

```
--- FAIL: TestExamples/examples/_templates/grafana (4.02s)
--- FAIL: TestExamples/examples/_templates/tempo (1.38s)
--- FAIL: TestExamples/examples/_templates/prometheus (1.95s)
...
```

examples/templates/sync_test.go
simonswine: Done, thanks.

examples/_templates/pyroscope/docker-compose.yml
simonswine: Keeping `:latest`, matches grafana/tempo templates.

examples/_templates/pyroscope/grafana-provisioning/datasources/pyroscope.yml
simonswine: Moved to pyroscope template, see d11afbf.

examples/tracing/tempo/grafana/provisioning/plugins/explore-profiles.yml
simonswine: Removed all, see e97323a.

simonswine:
Also fixed `_templates/` being picked up by the example runner and the CI selection script (5fdaa40). Follow-up PR for running the sync check on all example-touching PRs.

aleks-p (approved):

PR #5251, opened 2026-06-10
fix: enforce tenant header extraction in auth interceptor

simonswine:
## Summary

The server-side auth interceptor ignored errors from `ExtractTenantIDFromHeaders` in unary and streaming handlers. If the tenant header was missing or invalid, the request could continue with an inherited tenant already present in the context.

A remote client cannot directly inject context values, so this is primarily a fail-early hardening fix that prevents handlers from observing stale or pre-populated tenant context when header-based tenant auth fails.

## Changes

- Return `CodeUnauthenticated` immediately when tenant extraction from request headers fails in `WrapUnary`.
- Apply the same fail-early behavior to `WrapStreamingHandler`.
- Validate tenant headers with `TenantIDs` so legitimate pipe-separated multi-tenant headers are accepted.
- Canonicalize resolved tenant IDs before reinjecting them into context, including duplicate-only headers like `tenant-a|tenant-a`.

## Tests

- Rejects unary requests with an inherited tenant when `X-Scope-OrgID` is missing.
- Rejects streaming requests with an inherited tenant when `X-Scope-OrgID` is missing.
- Allows multi-tenant headers and canonicalizes them to sorted, deduplicated form.
- Canonicalizes duplicate-only tenant headers.

### DIFF
--- a/pkg/tenant/interceptor.go
+++ b/pkg/tenant/interceptor.go
@@ -75,7 +75,10 @@ func (i *authInterceptor) WrapStreamingHandler(next connect.StreamingHandlerFunc
 		if !i.enabled {
 			return next(InjectTenantID(ctx, DefaultTenantID), conn)
 		}
-		_, ctx, _ = ExtractTenantIDFromHeaders(ctx, conn.RequestHeader())
+		_, ctx, err := ExtractTenantIDFromHeaders(ctx, conn.RequestHeader())
+		if err != nil {
+			return connect.NewError(connect.CodeUnauthenticated, err)
+		}
 		if err := next(ctx, conn); err != nil {
 			if errors.Is(err, ErrNoTenantID) {
 				return connect.NewError(connect.CodeUnauthenticated, err)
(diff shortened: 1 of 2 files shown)


pkg/tenant/interceptor.go
```diff
@@ -85,20 +88,36 @@ func (i *authInterceptor) WrapStreamingHandler(next connect.StreamingHandlerFunc
...
 
-	tenantID, err := defaultResolver.TenantID(ctx)
+	tenantIDs, err := defaultResolver.TenantIDs(ctx)
 	if err != nil {
 		return "", ctx, err
 	}
 
-	return tenantID, ctx, nil
+	// Single tenant — already injected above, nothing more to do.
+	if len(tenantIDs) == 1 {
```
aleks-p: nit, I think this breaks the normalization claim for e.g. `tenant-a|tenant-a`, but that is fine

aleks-p (dismissed):
LGTM

A few questions / remarks
- does the same issue apply for WrapStreamingHandler?
- the summary perhaps overstates the exploit, a remote client can't "pre-inject" anything in the context (the fix still makes sense, but it is more of a fail-early / future-proofing than exploit prevention imo)

pkg/tenant/interceptor.go
simonswine: Yes true! See my changes in 37c4a4b

simonswine:
> * does the same issue apply for WrapStreamingHandler?

It does, now also fails early 

> * the summary perhaps overstates the exploit, a remote client can't "pre-inject" anything in the context (the fix still makes sense, but it is more of a fail-early / future-proofing than exploit prevention imo)

Yes it is defence in depth, comes from my security review workding and creeped into the context

aleks-p (approved):
LGTM, thanks for addressing the feedback!

PR #5252, opened 2026-06-10
feat(queryfrontend): add estimation accuracy histogram and log ratio

simonswine:
Follows up on the review comment in #5180.

Adds a `pyroscope_query_frontend_estimation_accuracy_ratio` histogram that records, per query, the ratio of the pre-execution metadata size estimate (`weight.Total()`) to actual object-storage bytes fetched (`BytesFetched`):

- **< 1.0** — fetched less than estimated (common; weight is an upper bound from section offsets)
- **1.0** — estimate matched exactly
- **> 1.0** — fetched more than estimated (e.g. Format1 index-lookup blocks whose profile/symbol sizes are unknown pre-execution)

Configured with both classic explicit buckets centred around 1.0 `{0.1, 0.25, 0.5, 0.75, 0.9, 1.0, 1.1, 1.25, 1.5, 2, 5, 10}` and native histogram parameters (`factor=1.1`, `maxBuckets=100`) so scrapers that support native histograms get higher resolution automatically.

The ratio is also surfaced on the existing **"query finished"** log line as two new fields:
- `estimated_object_bytes` — humanized pre-execution estimate
- `estimation_ratio` — 2 d.p. ratio (omitted when no bytes were fetched)

### DIFF
--- a/pkg/frontend/readpath/queryfrontend/query_frontend.go
+++ b/pkg/frontend/readpath/queryfrontend/query_frontend.go
@@ -63,7 +63,8 @@ type QueryFrontend struct {
 }
 
 type queryFrontendMetrics struct {
-	fetchedBytesTotal *prometheus.CounterVec
+	fetchedBytesTotal       *prometheus.CounterVec
+	estimationAccuracyRatio prometheus.Histogram
 }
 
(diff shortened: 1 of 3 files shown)


bryanhuhta (approved):
Nice, this looks good!

PR #5253, opened 2026-06-11
fix: reject non-positive ingest sample rates

immanuwell:
tiny footgun, but real.

`/ingest` accepts `sampleRate` from the query string. Before this patch, `sampleRate=-1` was parsed as an `int` and then cast to `uint32`, so it wrapped to `4294967295`. For CPU and wall profiles that can drive the computed period down to `0`, which is pretty cursed.

fix:
- treat `sampleRate <= 0` as invalid
- fall back to the documented default `100`
- add tests for `-1`, `0`, `abc`, and a valid positive value

repro:
1. hit `/ingest?...&sampleRate=-1`
2. before this patch, the handler stored `4294967295` as the sample rate
3. after this patch, it falls back to `100`

verified with `go test ./pkg/ingester/pyroscope -count=1`

### DIFF
--- a/pkg/ingester/pyroscope/ingest_handler.go
+++ b/pkg/ingester/pyroscope/ingest_handler.go
@@ -135,7 +135,10 @@ func (h ingestHandler) parseInputMetadataFromRequest(_ context.Context, r *http.
 
 	if sr := q.Get("sampleRate"); sr != "" {
 		sampleRate, err := strconv.Atoi(sr)
-		if err != nil {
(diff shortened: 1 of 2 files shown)


simonswine (approved):
This make perfect sense. Thanks for adding it

LGTM

korniltsev-grafanista:
Were there any problems with sample rate 0? I noticed a ton of errors in my logs https://github.com/grafana/pyroscope/pull/5276

PR #5254, opened 2026-06-11
feat: move symbolization to the query backend

marcsanmi:
Moves symbolization from the query frontend to the query backend, performed per dataset before merge. Fixes three problems of the frontend approach:

* A single unsymbolized dataset rewrote the whole query to the pprof path (TREE to PPROF to symbolize to TREE), so every dataset in the query paid the pprof materialization, wire and merge cost.
* Queries without a strict `service_name` selector are served via the tenant wide datasets index, whose metadata carries no per dataset labels, so symbolization was silently skipped. The backend resolves the index into the real datasets with their labels, closing the gap (covered by a new integration test).
* Span filtered tree queries could not be symbolized at all, since the rewrite dropped the span selector.

## How it works

```mermaid
flowchart LR
    QF[query frontend] -->|TREE query, Options.Symbolize| QB[query backend, per dataset]
    QB --> L{labeled __unsymbolized__?}
    L -->|no| N[native tree]
    L -->|yes| S[pprof in memory, symbolize, rebuild tree]
    N --> M[tree merge]
    S --> M
```

The frontend sets `InvokeOptions.Symbolize` when `symbolizer.enabled` holds for all request tenants. The backend re checks enablement against each dataset's owning tenant (cache keys and limits are tenant scoped, relevant for multi tenant blocks) and takes the pprof detour only for datasets labeled `__unsymbolized__` at ingest. The wire format stays a tree in all cases.

A request scoped cache (`symbolizer.WithRequestCache`) makes each binary's symbol data fetched once per request instead of once per dataset. It holds no memory between requests, retention is capped, and oversized entries are served fetch through.

The full symbols tree path (`SelectMergeProfile` with `query-tree-enabled`) keeps its single post merge symbolization in the frontend, now gated on tenant enablement alone, since the backend does not cover that path.

⚠️ Query backends now need the symbolizer configuration (`symbolizer.enabled` overrides, debuginfod URL, symbol size limit) and network access to the debuginfod server, previously required only on frontends. Deployment config change goes separately.

New metrics: `pyroscope_query_backend_symbolized_datasets_total{status}` and `cache_type="request"` on the existing symbolizer cache counters.


simonswine (commented):
From a quick look the two main points that are problematic are:

While the query frontend before would query once per build id and then have it cached. The query-backend will be running at a higher replica count and therefore will amplificate the traffic to the symbol store and the debuginfod servers. I don't think this is acceptable. I think we need to have a shared cache (do we have symbols for this buildid and if the location in the bucket if we have one).
I do think a single replica shared cache like Redis/Valkey is perfect for this use case. In case it is down we fall back to the per process cache.

I also think the query-backends should only load the symbols when they are hitting a location that they need and can symbolized. So rather than only check for the label we should only attemt to get symbols on the first unsymoblized location.

PR #5256, opened 2026-06-12
chore: Update dashboards to their internal version

simonswine:
## Summary

- Removes unused envoy-related panels from all monitoring dashboards
- Removes fields that typically cause drift between internal and published dashboard versions
- Adds a shared `_dashboards.tpl` helper to reduce duplication in Helm templates

## Test plan

- [ ] Verify dashboards render correctly in Grafana after applying the Helm chart changes
- [ ] Confirm no envoy panels appear in the updated dashboards


marcsanmi (approved):
LGTM

PR #5261, opened 2026-06-16
chore(deps): bump dskit to emit trace_id pprof label

marcsanmi:
Bumps dskit to pick up grafana/dskit#1006, so Pyroscope's own self-profiling emits a `trace_id` pprof label for traces↔profiles correlation by trace ID. The dskit bump also requires adapting to its new `dns.NewProvider` signature (3 call sites) and bumping the `grafana/memberlist` replace to match dskit's.


korniltsev-grafanista (approved):

pkg/metastore/discovery/parse.go
```diff
@@ -14,12 +14,11 @@ import (
 
 func NewDiscovery(l log.Logger, address string, reg prometheus.Registerer) (Discovery, error) {
 	if strings.HasPrefix(address, "dnssrvnoa+") {
-		p := dns.NewProvider(l,
+		p := dns.NewProvider(dns.MiekgdnsResolverType, 2, l,
```
korniltsev-grafanista, line 17: ~Any particular reason why MiekgdnsResolverType here and GolangResolverType in another place?~
nevermind

PR #5262, opened 2026-06-17
helm: add service_name relabel rule to alloy scrape config

### LINKED ISSUE #2051: Require presence of `service_name` in scrape configs
Similar to https://github.com/grafana/phlare/issues/711

Since the new App Selector was implemented(https://github.com/grafana/phlare/pull/712 and https://github.com/grafana/phlare/pull/725), it requires a field called `service_name` to be able to distinguish what's an "Application".

However, scraped data does not necessarily contain that label, making them hidden from the UI.
eg https://github.com/grafana/phlare/blob/60981cf248b377de6935d77f17793f0d3c9f7e11/tools/docker-compose/phlare.yaml#L1-L5

We could either fail to scrape that data altogether, or default to an `unknown` `service_name`, or default to `job_name`.

SLoeuillet:
## What

Adds a relabel rule to the Helm chart's default Alloy scrape config that maps the `profiles.grafana.com/service_name` pod annotation to the `service_name` label:

```alloy
rule {
    action        = "replace"
    source_labels = ["__meta_kubernetes_pod_annotation_profiles_grafana_com_service_name"]
    target_label  = "service_name"
}
```

It sits alongside the existing `service_repository` / `service_git_ref` / `service_root_path` rules in `templates/configmap-alloy.yaml`.

## Why

The chart already lets users set `service_repository`, `service_git_ref` and `service_root_path` via `profiles.grafana.com/*` annotations, but **not `service_name`** — the single most important label, used by the App Selector to identify an application (#2051).

Without it, Alloy falls back to inferring `service_name` from target metadata (namespace + `app.kubernetes.io/*`). When that inference is generic — e.g. a Helm component literally named `default` — pods surface under a useless/colliding `service_name` (`<namespace>/default`) with **no annotation-based way to override it**, unlike every other `service_*` label. This rule closes that gap so an operator can simply set `profiles.grafana.com/service_name: my-app`.

## Behaviour / compatibility

- **No-op unless a pod sets the annotation.** An explicit `service_name` label takes precedence over Alloy's inference, so existing deployments are completely unaffected until they opt in.
- Consistent with the existing `service_*` annotation rules — same block, same placement.

## Changes

- `templates/configmap-alloy.yaml` — the new rule.
- `rendered/*.yaml` — regenerated (the rule is in the static config block, so it appears identically in all six fixtures).
- `Chart.yaml` / `README.md` — chart `version` 2.0.3 → 2.0.4 (appVersion unchanged).

Closes #2051

### DIFF
--- a/operations/pyroscope/helm/pyroscope/templates/configmap-alloy.yaml
+++ b/operations/pyroscope/helm/pyroscope/templates/configmap-alloy.yaml
@@ -65,6 +65,21 @@ data:
     		target_label  = "container"
     	}
 
+    	// Set service_name by choosing the first non-empty value from the following ordered list:
+    	// - pod.annotation[profiles.grafana.com/service_name]
+    	// - pod.annotation[resource.opentelemetry.io/service.name]
+    	rule {
+    		action        = "replace"
+    		source_labels = [
+    			"__meta_kubernetes_pod_annotation_profiles_grafana_com_service_name",
+    			"__meta_kubernetes_pod_annotation_resource_opentelemetry_io_service_name",
+    		]
+    		separator    = ";"
+    		regex        = "^(?:;*)?([^;]+).*$"
+    		replacement  = "$1"
+    		target_label = "service_name"
+    	}
+
     	rule {
     		action        = "replace"
     		source_labels = ["__meta_kubernetes_pod_annotation_profiles_grafana_com_service_repository"]
--- a/operations/pyroscope/helm/pyroscope/Chart.yaml
+++ b/operations/pyroscope/helm/pyroscope/Chart.yaml
@@ -3,7 +3,7 @@ name: pyroscope
 description: horizontally-scalable, highly-available, multi-tenant continuous profiling aggregation system
 home: https://grafana.com/oss/pyroscope/
 type: application
-version: 2.1.1
+version: 2.1.2
 appVersion: 2.1.1
 dependencies:
   - name: grafana-agent
--- a/operations/pyroscope/helm/pyroscope/README.md
+++ b/operations/pyroscope/helm/pyroscope/README.md
@@ -1,6 +1,6 @@
 # pyroscope
 
-![Version: 2.1.1](https://img.shields.io/badge/Version-2.1.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.1.1](https://img.shields.io/badge/AppVersion-2.1.1-informational?style=flat-square)
+![Version: 2.1.2](https://img.shields.io/badge/Version-2.1.2-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.1.1](https://img.shields.io/badge/AppVersion-2.1.1-informational?style=flat-square)
 
 horizontally-scalable, highly-available, multi-tenant continuous profiling aggregation system
 
--- a/operations/pyroscope/helm/pyroscope/rendered/legacy-micro-services-hpa.yaml
+++ b/operations/pyroscope/helm/pyroscope/rendered/legacy-micro-services-hpa.yaml
@@ -6,7 +6,7 @@ metadata:
   name: pyroscope-dev-compactor
   namespace: default
   labels:
-    helm.sh/chart: pyroscope-2.1.1
+    helm.sh/chart: pyroscope-2.1.2
     app.kubernetes.io/name: pyroscope
     app.kubernetes.io/instance: pyroscope-dev
     app.kubernetes.io/version: "2.1.1"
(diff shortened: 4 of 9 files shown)


operations/pyroscope/helm/pyroscope/templates/configmap-alloy.yaml
```diff
@@ -65,6 +65,12 @@ data:
     		target_label  = "container"
     	}
 
+    	rule {
+    		action        = "replace"
+    		source_labels = ["__meta_kubernetes_pod_annotation_profiles_grafana_com_service_name"]
+    		target_label  = "service_name"
+    	}
+
```
simonswine, line 82: I think this is a real gap and I would like to improve the situation in this helm chart. 

In general this helm chart's Alloy config is very minimal for more comprehensive cases we usually suggest using k8s-monitoring-helm chart.

So ideally I want you to align with its logic, which you can find here: https://github.com/grafana/k8s-monitoring-helm/blob/ca0302f0f2203669d8457800e6c638f70daada37/charts/k8s-monitoring/charts/feature-profiling/templates/_ebpf.tpl#L84-L101

Although I would want to avoid breaking changes for existing users, 

https://github.com/grafana/k8s-monitoring-helm/blob/ca0302f0f2203669d8457800e6c638f70daada37/charts/k8s-monitoring/charts/feature-profiling/templates/_ebpf.tpl#L84-L101

My suggestion would be this:

```
  // Set service_name by choosing the first value found from the following ordered list:
  // - pod.annotation[profiles.grafana.com/service_name]
  // - pod.annotation[resource.opentelemetry.io/service.name]
  rule {
    action = "replace"
    source_labels = [
      "__meta_kubernetes_pod_annotation_profiles_grafana_com_service_name"
      "__meta_kubernetes_pod_annotation_resource_opentelemetry_io_service_name",
    ]
    separator = ";"
    regex = "^(?:;*)?([^;]+).*$"
    replacement = "$1"
    target_label = "service_name"
  }
```

simonswine (commented):
Let me know how you feel about my suggestion.

operations/pyroscope/helm/pyroscope/templates/configmap-alloy.yaml
SLoeuillet: Thanks @simonswine, that makes sense — aligned the rule with the k8s-monitoring-helm `feature-profiling` logic. It now sets `service_name` from the first non-empty value of the ordered annotation list:

- `profiles.grafana.com/service_name`
- `resource.opentelemetry.io/service.name`

using the same `separator`/`regex`/`replacement` first-match pattern you referenced. I kept it to those two annotations (rather than also folding in the `app.kubernetes.io/*` label + container fallbacks that k8s-monitoring uses) so Alloy's existing inference remains the last resort and this stays a no-op for current users — it only kicks in when one of the two annotations is set.

```alloy
// Set service_name by choosing the first non-empty value from the following ordered list:
// - pod.annotation[profiles.grafana.com/service_name]
// - pod.annotation[resource.opentelemetry.io/service.name]
rule {
	action        = "replace"
	source_labels = [
		"__meta_kubernetes_pod_annotation_profiles_grafana_com_service_name",
		"__meta_kubernetes_pod_annotation_resource_opentelemetry_io_service_name",
	]
	separator     = ";"
	regex         = "^(?:;*)?([^;]+).*$"
	replacement   = "$1"
	target_label  = "service_name"
}
```

(Added the missing comma between the two source labels from your snippet.) Rebased onto latest `main`, regenerated all six rendered fixtures, and bumped the chart to 2.1.1. PTAL.

SLoeuillet:
@simonswine is that OK for you this way ?

SLoeuillet:
The failing `test` jobs were `Test_FormatAgentRiverConfig` (`operations/pyroscope/alloy_test.go`) — not a logic issue, just Alloy-syntax formatting. The multi-line `source_labels = [ … ]` array breaks the `=` alignment group, so the `separator`/`regex`/`replacement`/`target_label` lines must align to `target_label` (12) rather than to `source_labels` (13) above the array — I had one extra space of padding on each.

Fixed the alignment in `templates/configmap-alloy.yaml`, regenerated the six `rendered/*.yaml` fixtures, and confirmed locally:

```
go test ./operations/pyroscope/    # ok
```

Amended into the single commit and force-pushed (`e23891b`). The new workflow runs are sitting in `action_required` — mind approving the CI re-run when you get a chance? 🙏

SLoeuillet:
Hi @simonswine — gentle nudge on this one, no rush. I've applied your suggestion from the earlier review, CI is green, and it should be mergeable. Whenever you have a moment to take another look I'd appreciate it. This also closes #2051. Thanks! 🙏

simonswine (approved):
LGTM

Sorry for keeping you waiting for so long

simonswine:
Sorry I had to force push, but our policy now requires all commits to be GPG signed before we can merge. 

More on this here: https://community.grafana.com/t/action-required-signed-commits-mandatory-for-all-grafana-repositories/163404

PR #5263, opened 2026-06-17
fix(deps): bump tar, js-yaml, @babel/core in /ui [security]

marcsanmi:
## What

Refreshes `/ui`'s lockfile (via `yarn up -R`) to pull the last remaining transitive dev/build dependencies up to their patched versions. **Lockfile-only** — `package.json` is unchanged, since its existing ranges already permit these versions.

| Package | From | To | CVE |
|---|---|---|---|
| `tar` | 7.5.13 | 7.5.16 | CVE-2026-53655 |
| `js-yaml` | 4.1.1 | 4.2.0 | CVE-2026-53550 |
| `@babel/core` | 7.29.0 | 7.29.7 | CVE-2026-49356 |

## Why

These are the only vulnerabilities still flagged on `main` (2 medium + 1 low, all inside SLO). They are **build-time only** deps (vite/eslint toolchain) — they do not ship in the runtime image — but the vulnerability scanner reads the committed `ui/yarn.lock`, so bumping the lockfile clears them. This brings `main` to a fully clean posture, which is then inherited by the upcoming **v2.0.4** patch release (the clean `ui/` is backported wholesale to `release/v2.0`).

## Verification

- `yarn lint` — clean
- `yarn test` — 25/25 pass
- `yarn build` — builds with vite 8.0.16
- `make frontend/build` (dockerized) — passes

### DIFF
--- a/ui/yarn.lock
+++ b/ui/yarn.lock
@@ -5,100 +5,100 @@ __metadata:
   version: 8
   cacheKey: 10c0
 
-"@babel/code-frame@npm:^7.28.6, @babel/code-frame@npm:^7.29.0":
-  version: 7.29.0
-  resolution: "@babel/code-frame@npm:7.29.0"
+"@babel/code-frame@npm:^7.29.7":
+  version: 7.29.7
+  resolution: "@babel/code-frame@npm:7.29.7"
   dependencies:
-    "@babel/helper-validator-identifier": "npm:^7.28.5"
+    "@babel/helper-validator-identifier": "npm:^7.29.7"
     js-tokens: "npm:^4.0.0"
     picocolors: "npm:^1.1.1"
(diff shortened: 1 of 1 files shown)


korniltsev-grafanista (dismissed):

ui/package.json
```diff
@@ -47,6 +47,9 @@
   },
   "resolutions": {
     "brace-expansion@npm:^1.1.7": "npm:1.1.13",
-    "brace-expansion@npm:^5.0.2": "npm:5.0.6"
+    "brace-expansion@npm:^5.0.2": "npm:5.0.6",
+    "tar": "npm:^7.5.16",
+    "js-yaml": "npm:^4.2.0",
+    "@babel/core": "npm:^7.29.6"
```
korniltsev-grafanista: nit: we don't have to list them here, we can just do `yarn up tar ...` and only update the lock file.

Not sure wha'ts the preferred way though

korniltsev-grafanista (approved):

PR #5273, opened 2026-06-18
fix(querybackend): forward and apply span selector in pprof and tree query paths

simonswine:
`SelectMergeSpanProfile` returned all samples regardless of span IDs in the V2 read path due to two bugs:

1. `backendTreeSymbolizer` rewrites `QUERY_TREE` → `QUERY_PPROF` for symbolization, but `PprofQuery` had no `span_selector` field — selector was silently dropped.
2. `queryPprof` always called `AddSamplesFromParquetRow` with no span filtering even when a selector was present.

Adds `span_selector` to `PprofQuery` proto, forwards it through the symbolizer, and applies filtering in the pprof backend. Also guards both backends against blocks without a SpanID column.


pkg/querybackend/query_pprof.go
```diff
@@ -77,9 +92,22 @@ func queryPprof(q *queryContext, query *queryv1.Query) (*queryv1.Report, error)
 	resolver := symdb.NewResolver(q.ctx, q.ds.Symbols(), resolverOptions...)
 	defer resolver.Release()
 
-	for profiles.Next() {
-		p := profiles.At()
-		resolver.AddSamplesFromParquetRow(p.Row.Partition, p.Values[0], p.Values[1])
+	if len(spanSelector) > 0 {
```
marcsanmi, line 95: Nit: Could we add a unit test for span-selector case? Likely in `block_reader_test.go`, similar to `Test_ProfileIDSelector`.

marcsanmi (dismissed):
LGTM

pkg/querybackend/query_pprof.go
```diff
@@ -55,14 +55,29 @@ func queryPprof(q *queryContext, query *queryv1.Query) (*queryv1.Report, error)
...
+	indices := []int{
 		columns.StacktraceID.ColumnIndex,
-		columns.Value.ColumnIndex)
+		columns.Value.ColumnIndex,
+	}
+	if len(spanSelector) > 0 {
+		if !columns.HasSpanID() {
+			// Block has no SpanID column: no samples can match the span selector.
+			return &queryv1.Report{Pprof: &queryv1.PprofReport{Query: query.Pprof.CloneVT()}}, nil
+		}
```
aleks-p, line 76: mostly curious, under what circumstances would this happen?

simonswine: A very old block, so something like from Phlare v0.x

pkg/querybackend/query_pprof.go
simonswine: Done this here: be47b818120b1fafbe91a90a840311938fe6187d

marcsanmi (approved):

pkg/querybackend/query_pprof.go
aleks-p: impossible with v2 then :)

PR #5274, opened 2026-06-19
fix(ui): do not parse label selector - send as is to server.

korniltsev-grafanista:
Before this change the UI does not allow to send arbitrary labels - the app parsed `service_name` and `profile_type` label from the query and only sent them.

In this change, I remove the `profile_type` from the label selector - it is stored separately and we do not append and parse back the query.
Furthermore we do not parse labelSelector at all - we send it as is to the server, letting the server validate the query. We still parse the label selector for service name detection, but it is no longer required and that means we can query even without `service_name`.
As a result we don't need the callbacks from the `userPyroscopeQuery` anymore and they are removed too.


marcsanmi (approved):

korniltsev-grafanista:
I think this change broke the refresh with run button, converting back to draft

korniltsev-grafanista:
closing in favour of https://github.com/grafana/pyroscope/pull/5172

PR #5284, opened 2026-06-23
feat(query): support trace_id_selector in the v2 read path

marcsanmi:
## What
Adds sample-level filtering by `trace_id` to the v2 read path. A `trace_id_selector` field is added to `TreeQuery`/`PprofQuery` and to the `SelectMergeProfile`/`SelectMergeStacktraces` APIs; the query backend keeps only samples whose `TraceID` column matches.

## Why
SDKs are starting to attach a 128-bit `trace_id` to profiling samples. The value is already persisted in the parquet `TraceID` column, but, unlike `span_id` and `profile_id`, nothing on the read path could query it. This wires up the missing path so a single distributed trace can be isolated within a merged profile.

## Notes for reviewers
- **Why a selector field, not a dedicated `SelectMergeTraceProfile` RPC** (as `span_id` has): `span_id` and `trace_id` are the same kind of sample-level filter — span only has its own RPC for historical reasons, not because it is a different kind of query. A field reuses the existing merge RPCs (like `profile_id_selector` / `stack_trace_selector`), so it works across every output format (pprof, tree, flamegraph, dot) whereas `SelectMergeSpanProfile` returns only tree/flamegraph; and at the engine level span/profile_id/trace are all just `TreeQuery`/`PprofQuery` fields, so a separate RPC would add API surface with no backend benefit.
- Filtering follows the sample-level `span_id` path, but trace ids are matched as **raw 16-byte fixed arrays, no endian swap**, unlike span's little-endian uint64. Verified byte-for-byte against the write path (`traceIDFromLabels`).
- **v2-only:** the v1 querier and the legacy time-splitting frontend (including the Diff path) return `CodeUnimplemented` rather than silently returning unfiltered results. This is intentionally stricter than `profile_id`, which still silently drops on those paths.
- `span` and `trace` selectors can't be combined (no public RPC sets both; rejected to avoid silently honoring one).


simonswine:
> Why a selector field, not a dedicated SelectMergeTraceProfile RPC (as span_id has): trace_id only narrows which samples contribute, it doesn't change the result shape, so it fits the existing merge RPCs the way profile_id/stack_trace_selector do.

I don't think this bit is true, trace_id/span_id works exactly on the same level. But the second part is the reason why I agree with your choice of it being a attribute

pkg/frontend/frontend_select_merge_profile.go
```diff
@@ -23,6 +24,13 @@ func (f *Frontend) SelectMergeProfile(
 	c *connect.Request[querierv1.SelectMergeProfileRequest],
 ) (*connect.Response[profilev1.Profile], error) {
 	ctx = connectgrpc.WithProcedure(ctx, querierv1connect.QuerierServiceSelectMergeProfileProcedure)
+	// trace_id_selector is a sample-level filter applied only by the v2 query
+	// backend. This legacy frontend rebuilds (and would drop) the selector when
+	// splitting by time, which would silently return unfiltered results, so we
+	// reject it here rather than answer the wrong question.
+	if len(c.Msg.TraceIdSelector) > 0 {
+		return nil, connect.NewError(connect.CodeUnimplemented, errors.New("trace_id_selector is only supported with the v2 query backend"))
+	}
```
simonswine: The comment is quite long, I think this can be condensed into 1-2 lines

The filter also needs to be placed into the Diff calls left/right parameters of v1 otherwise they are silently ignored.

pkg/frontend/frontend_select_merge_stacktraces.go
```diff
@@ -25,6 +25,13 @@ func (f *Frontend) SelectMergeStacktraces(
 	if c.Msg.Format == querierv1.ProfileFormat_PROFILE_FORMAT_DOT {
 		return nil, connect.NewError(connect.CodeUnimplemented, errors.New("dot format is only supported with the v2 query backend"))
 	}
+	// trace_id_selector is a sample-level filter applied only by the v2 query
```
simonswine: See comment above

simonswine (commented):
See my notes inside, otherwise looking good.

Adding test coverage for the full block path would be nice, kind of what has been done in 506c6f3f00.

korniltsev-grafanista (approved):

PR #5287, opened 2026-06-24
docs: Update docs to sign release tags

simonswine:
If we now require signed commits, we probably should also require signed release tags for pushing. Wdyt?

image

Updated the instructions accordingly. Let me know how you feel.

### DIFF
--- a/docs/internal/RELEASE.md
+++ b/docs/internal/RELEASE.md
@@ -13,7 +13,7 @@
    >
    > ⚠️  Incorrect: `release/v1.3.0`  
 3. Update `renovate.json` so `baseBranchPatterns` includes the new release branch and only the latest two release branches.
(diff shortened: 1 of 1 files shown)


aleks-p (approved):
I agree this makes sense.

I don't think it can be physically enforced, so someone will eventually forget to do it. We could either:
- add more automation around creating tags (like we do in other repos)
- add a validation step
- _commit_ to enable tag signing by [default](https://git-scm.com/docs/git-config#Documentation/git-config.txt-taggpgSign)

The PR still makes sense as is though.

simonswine:
I am not fully sure I like release-please to be that automation, the SDK experiences has been mixed I would say. Maybe we give a bit and think about.

If we forget at least we can still sign it at that time. (CI might re-run though)

PR #5289, opened 2026-06-25
querybackend: scan profile entries columnar

simonswine:
Something I came up with working on performance for #5290. I think this has wider benefit.

This replaces the profile entry iterator's row-oriented parquet joins with columnar predicate scans that build accepted-row bitmaps and materialize entries from scalar columns using native parquet typed readers when available. It also adds a real-block benchmark covering the default, grouped-label, and all-labels-with-profile-IDs iterator modes.

Not quite ready for merge, TODOs:
- [ ] Remove make benchmark against bucket/file more generic
- [ ] Asses in the real world against bucket

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	connectrpc.com/connect v1.19.2
 	connectrpc.com/grpchealth v1.4.0
 	github.com/PuerkitoBio/goquery v1.12.0
+	github.com/RoaringBitmap/roaring/v2 v2.18.2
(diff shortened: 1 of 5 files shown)


jake-kramer:
Cool idea, learned about RoaringBitmap. I wonder how its performance would compare vs using raw bitmap given I expect (?) the bitmaps to be relatively sparse (not too familiar with typical query patterns vs block density but my intuition says sparse?)

PR #5291, opened 2026-06-26
fix(vcs): match prefix only at path segment boundaries

immanuwell:
The `Match` method in `MappingConfig` used plain `strings.HasPrefix`, so a prefix like `java` would match `javax/swing/JFrame.java` - wrong VCS source gets picked for those files.

The fix adds `matchPrefix` that only accepts a match when the prefix ends at a `/` boundary (or is an exact match, or itself ends with `/`).

Reproduce:

```yaml
mappings:
  - path:
      - prefix: "java"
    source:
      github:
        repository: ...
```

Any file whose path starts with `javax` or `javafx` will be incorrectly matched by that rule. After this fix it won't be.

Related: issue #4878 (Kotlin `.kt` extension support - same VCS config area).

### DIFF
--- a/pkg/frontend/vcs/config/config.go
+++ b/pkg/frontend/vcs/config/config.go
@@ -193,21 +193,33 @@ func (c *PyroscopeConfig) FindMapping(file FileSpec) *MappingConfig {
 	return bestMatch
 }
 
+func matchPrefix(value, prefix string) int {
+	if !strings.HasPrefix(value, prefix) {
+		return -1
+	}
+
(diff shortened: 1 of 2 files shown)


jake-kramer:
Thanks for your contribution @immanuwell. I'm not sure that this is a bug to be fixed. The existing machinery allows you to specify any prefix you want, including with a trailing `/`. The example you provide in the description is working as expected, and I'd advise making a more specific prefix to capture exactly what you want.

PR #5293, opened 2026-06-26
feat(ingester): validate units and aggregationType in ingest handler

faisalaffan:
### What

Add input validation for `units` and `aggregationType` query parameters
in the Pyroscope ingest HTTP handler. Previously, these values were accepted
silently as arbitrary strings — invalid values now produce a warning log
and fall back to safe defaults.

### Why

The `parseInputMetadataFromRequest` method had two explicit TODOs from maintainers:

- `// TODO(petethepig): add validation for these?` (units)
- `// TODO(petethepig): add validation for these?` (aggregationType)

Without validation, a client sending `units=bogus` or `aggregationType=banana`
would create metadata with invalid values that propagate through the ingestion
pipeline undetected.

### Changes

**`pkg/og/storage/metadata/metadata.go`**
- Added `IsValidUnit(unit string) bool` — validates against the 6 known unit types
- Added `IsValidAggregationType(aggType string) bool` — validates against `sum` / `average`

**`pkg/ingester/pyroscope/ingest_handler.go`**
- `parseInputMetadataFromRequest` now validates `units` and `aggregationType`
- Invalid values log a warning with the original value and fall back to defaults:
  - Units → `samples`
  - AggregationType → `sum`
- Follows the same pattern already used for `sampleRate` validation

### Testing

- **37 unit tests** for `IsValidUnit` and `IsValidAggregationType` (100% coverage):
  - All valid values
  - Case sensitivity
  - Whitespace (leading, trailing)
  - Newline injection
  - Unicode characters
  - SQL injection patterns
  - Very long strings

- **35 integration tests** for `parseInputMetadataFromRequest` (88.6% coverage):
  - Valid and invalid units
  - Valid and invalid aggregation types
  - Both params invalid simultaneously
  - Empty and missing params (default behavior)
  - URL-encoded edge cases

### DIFF
--- a/pkg/ingester/pyroscope/ingest_handler.go
+++ b/pkg/ingester/pyroscope/ingest_handler.go
@@ -159,15 +159,29 @@ func (h ingestHandler) parseInputMetadataFromRequest(_ context.Context, r *http.
 	}
 
 	if u := q.Get("units"); u != "" {
-		// TODO(petethepig): add validation for these?
-		input.Metadata.Units = metadata.Units(u)
+		if !metadata.IsValidUnit(u) {
+			_ = h.log.Log(
+				"err", fmt.Errorf("invalid units: %q", u),
+				"msg", fmt.Sprintf("invalid units: %q, falling back to %q", u, metadata.SamplesUnits),
+			)
+			input.Metadata.Units = metadata.SamplesUnits
+		} else {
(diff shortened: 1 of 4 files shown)


korniltsev-grafanista:
these fields are unused

PR #5301, opened 2026-07-01
fix(otlp): coalesce per-profile pushes to fix ingest timeout (#5299)

### LINKED ISSUE #5299: eBPF/OTLP ingest (/v1development/profiles) times out: distributor fans out one segment-writer push per sample, latency scales with sample count
### Summary

On a v2 (segment-writer) cell, OTLP eBPF profiling ingestion via `/v1development/profiles` intermittently fails with `500`/`504` because the distributor fans a single OTLP request out into **one segment-writer `Push` per sample**. Requests carry 60–1300 samples each, so request latency scales with sample count and routinely exceeds the gateway's 15s upstream timeout. Classic SDK ingest (`push_v1_pusherservice_push`, ~1 series/request) is unaffected.

### Symptoms

- Alert: ~4–8% 5xx on `v1development_profiles`, oscillating 1–27% over 12h+.
- `distributor` returns `500`; `cortex-gw-internal` returns `504`.
- Distributor log on failure:
  ```
  level=warn uri=/v1development/profiles status=500 duration=15.000s
  msg="http request failed"
  response_body="failed to make a GRPC request: push series with index 0 and id … failed: canceled: rpc error: code = Canceled desc = context canceled"
  ```
  `duration` is consistently exactly `15.000s` → the gateway's upstream timeout cancels the in-flight push.

### Evidence / root cause

| Signal | Value |
|---|---|
| distributor `v1development` request rate | 0.96 req/s |
| segment-writer `Push` rate | 99.9 push/s → **~104× fan-out per request** |
| avg samples / request | 98.4 (p50 **62.8**, p99 **1298**) |
| segment-writer `Push` latency | p50 **0.3s**, p99 0.95s |
| distributor `v1development` latency | p50 **8.2s**, p90 **20.5s**, p99 24.5s |
| distributor CPU per request (span exemplars) | **~20–100ms** (i.e. almost entirely off-CPU wait) |
| normal `push_v1_pusherservice_push` latency | p50 **0.384s** (≈ the 350ms artificial delay, ~1 push) |

Distributor goroutine profile shows the requests parked off-CPU in the fan-out, not in CPU work:

```
distributor.(*Distributor).Push → PushBatch → sync.(*WaitGroup).Wait → semacquire        (~6%)
ingester/otlp.(*ingestHandler).export → PushBatch → sync.(*WaitGroup).Wait → semacquire   (~2.4%)
grpc.(*clientStream).withRetry → RecvMsg → waitOnHeader → gopark                          (~9.5%)
connect handler → tenant.authInterceptor → delayhandler.(*delayInterceptor).WrapUnary → chanrecv  (~4%, the 350ms artificial delay, applied once per inbound request)
```

**Mechanism:** `ingester/otlp.export` → `distributor.PushBatch` issues ~one segment-writer `Push` per sample. Each push only acks after its segment flushes (~0.3s p50). With bounded fan-out concurrency, an N-sample request takes roughly `N/concurrency × ~0.3s`; the observed p50 (8.2s @ ~63 samples) implies effective concurrency ≈ 3. Upper-tail requests (hundreds–1300 samples) take 20s+, exceeding the 15s gateway timeout → cancellation → 500/504. The `350ms` artificial delay is applied once per inbound request (confirmed by the 0.384s normal-path p50) and is a minor contributor here.

### Suggested fixes (in priority order)

1. **Batch segment-writer writes** — coalesce the per-sample pushes from one OTLP request into far fewer (ideally one) segment-writer `Push` calls, so request latency stops scaling with sample count. *(real fix)*
2. **Raise fan-out concurrency** for `PushBatch` on the OTLP path.
3. **Bound/split samples-per-request** for the eBPF/OTLP ingest path before fan-out.
4. Reconsider applying the per-request `ingestion-artificial-delay` to the high-fan-out OTLP path.

Timeout tuning (gateway 15s) only masks the underlying O(samples) fan-out.

jake-kramer:
Fixes #5299

## Problem
The OTLP profiling ingest endpoint (`/v1development/profiles`, used by eBPF) times out
under load. `ingestHandler.export()` called `PushBatch` **once per OTLP `Profile` message,
sequentially**, awaiting each before the next. Each `PushBatch` ultimately does a
segment-writer push that only acks after the segment flushes (~0.3s p50), so request
latency scaled with the number of messages an eBPF exporter packs into one request.

**Observed on a production deployment (current, unfixed):**
- OTLP p99 latency **~24.6s**, regularly exceeding the 15s gateway timeout.
- ~10% of OTLP requests erroring (`504` gateway timeouts + `500`).
- A slow request fans out to **~80–100 sequential segment-writer pushes** (24.6s ÷ ~0.3s
  per push; consistent with the reported ~104× fan-out).

## Fix
Restructure `export()` to accumulate the converted pprof profiles across the **entire**
`ResourceProfiles → ScopeProfiles → Profiles` walk into a **single** `PushRequest`,
**merging profiles that share a label set** via `pprof.ProfileMerge`, and call `PushBatch`
**once**. This turns N sequential per-message pushes into one concurrent fan-out (fixes the
latency regardless of mergeability) and collapses repeated series into one push (fewer
segment-writer pushes when messages repeat a service).

Two commits:
1. **`fix(pprof)`** — `equalValueType` treated two `nil` `PeriodType`/`SampleType` values as
   *incompatible*, so `ProfileMerge` refused to merge profiles that both omit a
   `PeriodType` (the common OTLP/eBPF shape) — it would even fail the *first* self-merge.
   `PeriodType` is optional/informational per the pprof spec, so two absent value types
   agree. Now `nil == nil → compatible`; present-vs-absent stays incompatible. This also
   fixes latent merge failures on the read/aggregation paths that share `ProfileMerge`.
2. **`fix(otlp)`** — the `export()` restructure above. Profiles that genuinely cannot be
   merged (incompatible sample/period types) are emitted as their own series — never
   dropped, never failing the request — using an `order` slice + `byKey` map so termination
   is O(N) regardless of how many incompatible profiles collide on a label set.

## Evidence (internal dev cluster canary, controlled single-service load)
Identical load both runs (`-messages 10 -rps 5 -concurrency 8`, one distributor pod):

| Metric | main (unfixed) | fix | change |
|---|---|---|---|
| OTLP ingest p99 (server) | 7.5 s | 0.81 s | ~9× faster |
| Sustained throughput @ conc 8 | ~1.7 req/s | ~4.3 req/s | ~2.5× |
| Segment-writer pushes / request | ~10 (= message count) | ~1–2 (merged) | ~10× fewer |
| Requests completed in 3 min | 299 | 722 | ~2.4× more |
| Server-side errors | 0 | 0 | — |

At 1300 messages/request, unfixed `main` cannot service the request at all (times out);
the fix returns in <1s.

## Testing
- New unit tests: batching-into-one-PushBatch, cross-message merge correctness (summed
  values, combined headers), nil-PeriodType merge, distinct-service separation,
  incompatible-type fallback (no drop / no 500), 3-incompatible no-hang regression,
  HTTP 400/500 status mapping, summed size accounting.
- `pkg/pprof/merge_test.go`: nil==nil compatible, nil-vs-non-nil still incompatible.
- `go test ./pkg/pprof/... ./pkg/ingester/otlp/... ./pkg/distributor/...` pass, incl.
  `-race`. Blast-radius checked: querier/phlaredb also green.
- No `make generate` needed (no proto/config/flag changes).

### DIFF
--- a/pkg/usagestats/stats.go
+++ b/pkg/usagestats/stats.go
@@ -418,9 +418,8 @@ func (s *MultiStatistics) Value() map[string]interface{} {
 }
 
 func (s *MultiStatistics) Record(v float64, key string) {
-	keyStats := s.getOrCreateStatistics(key)
-	keyStats.Record(v)
-	s.values["__total__"].Record(v)
+	s.getOrCreateStatistics(key).Record(v)
+	s.getOrCreateStatistics("__total__").Record(v)
 }
 
 func (s *MultiStatistics) getOrCreateStatistics(key string) *Statistics {
@@ -588,9 +587,8 @@ func (c *MultiCounter) reset() {
 }
 
 func (c *MultiCounter) Inc(i int64, keyValue string) {
-	v := c.getOrCreateCounter(keyValue)
-	v.Inc(i)
-	c.values["__total__"].Inc(i)
+	c.getOrCreateCounter(keyValue).Inc(i)
+	c.getOrCreateCounter("__total__").Inc(i)
 }
 
 func (c *MultiCounter) getOrCreateCounter(keyValue string) *Counter {
--- a/pkg/ingester/otlp/ingest_handler.go
+++ b/pkg/ingester/otlp/ingest_handler.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"sort"
 	"strconv"
 	"strings"
 
@@ -26,6 +27,7 @@ import (
 	pprofileotlp "go.opentelemetry.io/proto/otlp/collector/profiles/v1development"
 	v1 "go.opentelemetry.io/proto/otlp/common/v1"
 
+	googlev1 "github.com/grafana/pyroscope/api/gen/proto/go/google/v1"
 	typesv1 "github.com/grafana/pyroscope/api/gen/proto/go/types/v1"
 	distributormodel "github.com/grafana/pyroscope/v2/pkg/distributor/model"
 	"github.com/grafana/pyroscope/v2/pkg/model"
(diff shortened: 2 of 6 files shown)


pkg/usagestats/stats.go
```diff
@@ -418,9 +418,8 @@ func (s *MultiStatistics) Value() map[string]interface{} {
 }
 
 func (s *MultiStatistics) Record(v float64, key string) {
-	keyStats := s.getOrCreateStatistics(key)
-	keyStats.Record(v)
-	s.values["__total__"].Record(v)
+	s.getOrCreateStatistics(key).Record(v)
```
korniltsev-grafanista, line 421: can we submit usagestats change separately?

korniltsev-grafanista:
Do we really need to merge pprofs here? Would just sending a single PushRequest with multiple profiles instead of multiple requests per profile do the same trick?

pkg/usagestats/stats.go
jake-kramer: There's a dependency here because this bug breaks the new logic but I can split out in to separate PR

jake-kramer:
> Do we really need to merge pprofs here? Would just sending a single PushRequest with multiple profiles instead of multiple requests per profile do the same trick?

Yea that makes more sense to me, and should be simpler code

jake-kramer:
Closing in favor of https://github.com/grafana/pyroscope/pull/5304

PR #5303, opened 2026-07-02
fix(usagestats): guard MultiCounter/MultiStatistics total against data race

jake-kramer:
## What

`MultiCounter.Inc` and `MultiStatistics.Record` read `c.values["__total__"]` without holding the lock, while `getOrCreateCounter` / `getOrCreateStatistics` can concurrently insert new keys into the same map under the lock. That is a concurrent map read and map write: the race detector flags it, and the runtime can `fatal: concurrent map read and map write` in production.

The fix routes the `__total__` access through the existing lock-safe `getOrCreate` accessors, and adds `TestMultiCounter_ConcurrentInc`.

## Why now / why standalone

This was surfaced by `-race` integration CI while working on the OTLP ingest fix for #5299: batching profile ingestion makes the distributor's per-series fan-out actually concurrent, which makes this race hot. Both candidate fixes for #5299 (#5301 and a forthcoming alternate PR) currently carry this exact commit; splitting it out lets it merge independently, and both will drop their copy on rebase.

Refs #5299.

### DIFF
--- a/pkg/usagestats/stats.go
+++ b/pkg/usagestats/stats.go
@@ -418,9 +418,8 @@ func (s *MultiStatistics) Value() map[string]interface{} {
 }
 
 func (s *MultiStatistics) Record(v float64, key string) {
-	keyStats := s.getOrCreateStatistics(key)
-	keyStats.Record(v)
-	s.values["__total__"].Record(v)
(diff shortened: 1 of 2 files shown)


korniltsev-grafanista (approved):
lgtm

nit: maybe add a test for the second type too

PR #5307, opened 2026-07-03
fix(metastore): reconcile abandoned raft log writes

simonswine:
A log-store write may time out while continuing in the background. If it eventually succeeds, raft retries the same index and raft-wal rejects it as non-monotonic, leaving the follower permanently behind. This is a follow up to #4802

Serialize log mutations, reconcile late entries before retrying, and extend suffix truncation over stale physical tails. This allows followers to recover after temporary disk stalls.

Adds unit and cluster-level regression coverage for delayed writes.

### Limitations

A timed-out underlying operation cannot be cancelled through `raft.LogStore`. If it never returns, later mutations continue to time out waiting for the token; the safe recovery is to mark the node unhealthy and restart or replace it rather than issue concurrent WAL mutations.

### What changed since PR creation `950b6ab1551b3b811e78ceda43e91b595dba0283`

- Replaced the mutex plus truncation-epoch design with a capacity-one write token acquired before dispatch. Timed-out waiters now return without creating queued write goroutines, so only the one already-running operation can be abandoned.
- Removed lower-term rejection. Because no later mutation can pass the token while an abandoned write is active, any physical overlap beyond Raft's acknowledged tail can be reconciled by truncating the conflicting suffix and appending the requested entries.
- Reworked the unit tests around the token behavior, physical-tail reconciliation, timeout handling, and suffix versus prefix deletion. The cluster regression test remains focused on a delayed write that eventually succeeds.

### DIFF
--- a/pkg/metastore/raftnode/logstore_test.go
+++ b/pkg/metastore/raftnode/logstore_test.go
@@ -1,0 +1,596 @@
+package raftnode
+
+import (
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/hashicorp/raft"
+	raftwal "github.com/hashicorp/raft-wal"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stretchr/testify/require"
+)
+
+// gatedLogStore wraps a raft.LogStore and lets tests hold StoreLogs/StoreLog
+// calls open on a channel, so a write can be started, "abandoned" by a
+// timeout wrapper, and completed later out of band. This simulates a slow
+// disk: the write is real and eventually lands, but the caller (raft, via
+// timeoutLogStore) may have already given up waiting for it.
+type gatedLogStore struct {
+	store raft.LogStore
+
+	mu   sync.Mutex
+	gate chan struct{}
+
+	// holdDuration, if set, is slept once a call has passed the gate but
+	// before it reaches the real underlying store, simulating a slow disk
+	// write that stays "in flight" for a known, deterministic window. This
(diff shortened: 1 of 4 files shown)


pkg/metastore/raftnode/logstore_test.go
```diff
@@ -0,0 +1,337 @@
+package raftnode
```
aleks-p, line 1: the new tests were a bit flaky locally, when running with -race, I got

```
Received unexpected error:
      log store write timed out after 75ms
```

aleks-p (approved):
This looks safe enough to me to merge as it improves recovery after a stall.

I did find another scenario that this will not address:

1. The follower believes it has entries through index 100
2. A new leader sends a conflicting replacement for indexes 20–30
3. Raft successfully calls DeleteRange(20, 100)
4. The following StoreLogs(20…30) stalls and returns the new timeout error
5. That abandoned append later succeeds, so the physical WAL now ends at 30
6. However, Raft’s in-memory last index remains 100

This seems to be a [known issue](https://github.com/hashicorp/raft/blob/c0dc6a0b2c7e889f31e5ab2f7ed90ceb159acffe/raft.go#L1541) in Raft so this PR doesn't make anything worse. The only option other than fixing it in upstream might be to exit on a timeout (#5536).

pkg/metastore/raftnode/logstore_test.go
simonswine: Sorry I thought I had pushed that, here the follow up: https://github.com/grafana/pyroscope/pull/5643

PR #5308, opened 2026-07-03
docs: Make clear what flags are v1 exclusive

simonswine:
## Summary
- Warn when user on v2 architecture and using v1 only flags
- Fixed a bug where we rendered v2 manifests for the v1 job.
- Updated help text of flags that are unused in v2

### DIFF
--- a/pkg/cfg/cfg.go
+++ b/pkg/cfg/cfg.go
@@ -21,7 +21,7 @@ type Cloneable interface {
 	Clone() flagext.Registerer
 }
 
-type FlagSetRecorder interface {
+type SetFlagRecorder interface {
 	RecordSetFlag(name string)
 }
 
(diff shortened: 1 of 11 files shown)


pkg/cfg/cfg.go
```diff
@@ -21,6 +21,10 @@ type Cloneable interface {
 	Clone() flagext.Registerer
 }
 
+type FlagSetRecorder interface {
```
marcsanmi: Nit:
```suggestion
type SetFlagRecorder interface {
```

marcsanmi (dismissed):
LGTM

pkg/cfg/cfg.go
simonswine: Done in fafc14c

alsoba13 (approved):
LGTM

PR #5310, opened 2026-07-06
perf(distributor): compute profile size once per push

marcsanmi:
## What

`pushSeries` computed the identical pre-normalization profile size **twice per push**:

1. inside `calculateRequestSize` → `req.Profile.SizeVT()` (for `req.TotalBytesUncompressed`)
2. again ~60 lines later as `decompressedSize := p.SizeVT()` (for the per-profile metrics and `ValidateProfile`'s max-size check)

`SizeVT()` is a full recursive walk over the profile (every Sample, Location, Function, Mapping, and the whole string table). One walk costs **~350µs of pure CPU** on a 261KB uncompressed heap profile (`pkg/pprof/testdata/heap`, 0 allocs) — proportionally less for smaller profiles — so the redundant walk is a real per-push cost on the write path's front door.

This computes the size once and shares it.

Note when reading the diff: `calculateRequestSize` was **split**, not stripped — the `SizeVT()` call moved into the new one-line `calculateRequestSize` body, and the label loop became `labelsSize`. The split exists so the push site can combine the already-computed profile size with just the label bytes, while `calculateRequestSize` keeps doing a fresh walk for its remaining caller (the post-`Normalize()` metric, where the profile has genuinely changed and must be re-measured).

## Why it's safe

- The two values were identical by construction: nothing between the two sites mutates the profile. `checkIngestLimit`/`checkUsageGroupsIngestLimit`/`MarkSampledRequest` only append to `req.Annotations` (never part of either size), `rateLimit` only reads the request totals, and `GetProfileLanguage` reads the string table and caches on `req.Language`.
- Git history shows the duplication was incidental, not intentional: the two call sites were added independently, two years apart (#2659 in Nov 2023, #4367 in Aug 2025), for different accounting purposes.
- The **third** `SizeVT` call (inside `profileSizeBytes`) is deliberately left alone: it runs after `ValidateProfile`, which can truncate deep stack traces and over-long string-table entries, so its value can legitimately differ from the pre-validation size.
- `SizeVT` is generated vtproto code — a pure size computation with no caching or side effects — so removing a call cannot change behavior; only the values fed downstream matter, and they are unchanged.

## Verification

`Test_Limits` acts as an independent oracle for the exact byte counts: it pushes profiles through the full `Push → pushSeries` path and asserts the `DiscardedBytes` counter equals a value the test recomputes from scratch (its own label-length loop + its own `SizeVT()` call in `uncompressedProfileSize`, without going through `calculateRequestSize`/`labelsSize`). A one-byte change in `TotalBytesUncompressed` would fail it. It passes, along with the rest of the `pkg/distributor` suite.


pkg/distributor/distributor.go
```diff
@@ -535,7 +535,9 @@ func (d *Distributor) pushSeries(ctx context.Context, req *distributormodel.Prof
 	}
 
 	req.TotalProfiles = 1
-	req.TotalBytesUncompressed = calculateRequestSize(req)
+	// SizeVT is expensive; computed once and reused below (the profile is not mutated until normalization).
```
aleks-p: Nit: This comment is slightly misleading, validation (which runs before normalization) can also mutate the profile. 

Perhaps "not mutated before this value is last used" or similar?

aleks-p (dismissed):
LGTM, I left one nit.

marcsanmi:
We can close this since this was suggested and introduced in this other PR review https://github.com/grafana/pyroscope/pull/5354#discussion_r3595636959

PR #5311, opened 2026-07-06
perf(timeseries): reduce allocations in series build and merge

marcsanmi:
## What

Two allocation fixes in `pkg/model/timeseries`, which both read paths funnel through — v1 (`querier`/`phlaredb` `mergeByLabels`) and v2 (`querybackend` `executeTimeSeriesQuery` calls `Builder.Add` once per profile row; `RangeSeries`/`Merger` run per query).

### 1. `Builder.Add`: one heap allocation per profile row removed

`seriesKey := string(s.labelBuf)` materialized the key string on **every row**, even though nearly all rows hit an already-existing series — assigning the conversion to a variable defeats Go's zero-alloc `m[string(b)]` map-lookup optimization. Lookups now use the inline conversion; the key is only materialized on the rare new-series insert, where a persistent copy is genuinely required.

Verified with `-gcflags=-m`:

```
before:  timeseries.go:46: string(s.labelBuf) escapes to heap        // every row
after:   string(s.labelBuf) does not escape                          // both lookups
         string(s.labelBuf) escapes to heap                          // inserts only (required)
```

### 2. Finish the `sort.Slice` → `slices.SortFunc` migration

`sort.Slice` costs a reflect-based swapper + an escaping closure per call; `mergePoints` pays it once per merged series. `mergeAnnotations` in the same file already uses `slices.SortFunc` — this converts the remaining six call sites, making the package consistent. Same shape as #5159.

## Ordering safety

Both `sort.Slice` and `slices.SortFunc` are unstable, so no stability contract can regress. Checked every swapped sort for whether equal elements can even exist:

- series sorts (`TimeSeries`, `normalize`, `rangeSeriesWithLimit`): keys come from maps keyed by label hash/bytes → unique → **byte-identical output**
- `mergeExemplars` by ProfileId: map-keyed → unique → **byte-identical**
- `mergePoints` by timestamp / `selectTopNExemplarsProto` by value: ties possible, but their relative order was already unspecified before (unstable sort); sum mode merges timestamp ties away entirely

## Benchmarks

Apple M3 Pro, `benchstat` n=8. Adds `BenchmarkBuilderAdd` and `BenchmarkMergeSeries`:

| benchmark | metric | before | after | delta |
|---|---|---|---|---|
| BuilderAdd | sec/op | 142.0n | 109.6n | **-22.8%** |
| BuilderAdd | allocs/op | 2 | 1 | **-50%** |
| MergeSeries (geomean) | allocs/op | 333.7 | 125.3 | **-62.5%** |
| MergeSeries (geomean) | sec/op | 638.7µ | 602.5µ | -5.7% |
| MergeSeries series=1000 | sec/op | 814.7µ | 709.6µ | **-12.9%** |
| TimeSeriesQuery/NoExemplars (existing, end-to-end) | allocs/op | 33.53k | 30.52k | **-9.0%** |

End-to-end wall time is unchanged at test-fixture scale (dominated by block I/O) — the win here is per-row allocation/GC pressure on the read path plus the measured micro-level speedups.

Follow-up candidate (not included, unbenchmarked): `pkg/model/timeseriescompact` has 3 remaining `sort.Slice` sites of the same shape.

### DIFF
--- a/pkg/model/timeseries/timeseries_bench_test.go
+++ b/pkg/model/timeseries/timeseries_bench_test.go
@@ -1,0 +1,34 @@
+package timeseries
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/prometheus/common/model"
+
+	phlaremodel "github.com/grafana/pyroscope/v2/pkg/model"
+	schemav1 "github.com/grafana/pyroscope/v2/pkg/phlaredb/schemas/v1"
+)
+
+func BenchmarkBuilderAdd(b *testing.B) {
+	const numSeries = 20
+	labelSets := make([]phlaremodel.Labels, numSeries)
+	for i := range labelSets {
+		labelSets[i] = phlaremodel.LabelsFromStrings(
+			"__name__", "process_cpu",
+			"service_name", fmt.Sprintf("service-%d", i),
+			"pod", fmt.Sprintf("pod-%d", i),
+			"namespace", "default",
+		)
+	}
+
+	b.ReportAllocs()
+	builder := NewBuilder("service_name")
+	for i := 0; i < b.N; i++ {
+		lbs := labelSets[i%numSeries]
+		builder.Add(model.Fingerprint(i%numSeries), lbs, int64(i)*15000, 1.0, schemav1.Annotations{}, "")
+	}
+	if len(builder.Build()) == 0 {
+		b.Fatal("no series built")
+	}
+}
--- a/pkg/model/timeseries/timeseries_bench_test.go
+++ b/pkg/model/timeseries/timeseries_bench_test.go
@@ -22,13 +22,16 @@ func BenchmarkBuilderAdd(b *testing.B) {
 		)
 	}
 
+	const pointsPerSeries = 100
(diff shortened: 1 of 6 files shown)


aleks-p (commented):
There is one more instance of `sort.Slice` in exemplar_builder.go that we might want to convert, to keep things consistent.

pkg/model/timeseries/timeseries_bench_test.go
```diff
@@ -0,0 +1,34 @@
...
+		labelSets[i] = phlaremodel.LabelsFromStrings(
+			"__name__", "process_cpu",
+			"service_name", fmt.Sprintf("service-%d", i),
+			"pod", fmt.Sprintf("pod-%d", i),
+			"namespace", "default",
+		)
+	}
+
+	b.ReportAllocs()
+	builder := NewBuilder("service_name")
```
aleks-p: Is it deliberate to reuse the builder across b.N iterations? I imagine it makes each subsequent iteration more expensive.

aleks-p (approved):

PR #5312, opened 2026-07-06
perf(segmentwriter): observe ingest bytes without re-serializing the profile

marcsanmi:
## What

`segment.ingest` ended every push with `Observe(float64(p.SizeVT()))`. `SizeVT()` is a full recursive walk over the profile — every Sample, Location, Function, Mapping and the whole string table — **~350µs on a 261KB uncompressed heap profile** (`pkg/pprof/testdata/heap`), paid on every push purely to feed a histogram.

The uncompressed size is already known for free: `pprof.RawFromBytes` records it at parse time and exposes it as `Profile.RawSize()`. The observation moves to `SegmentWriterService.Push`, which already observes `segmentFlushWaitDuration`.

## Metric rename

This changes the recorded value from "bytes as re-serialized after the sample-series split" to "bytes as received", so per review the metric is renamed to say so:

`pyroscope_segment_writer_segment_ingest_bytes` → `pyroscope_segment_writer_received_bytes`

Also drops the `segment_writer_segment_` stutter and adds the `Help` string the metric never had. Same type, labels and buckets; one observation per push, as before.

Consumers updated: the 5 queries in `operations/monitoring/` (edited in the helm template and regenerated — the JSON dashboards are generated artifacts). No alerts or recording rules reference this metric. Cloud mixin dashboard: grafana/deployment_tools#708217, to merge after this rolls out.

## Notes

- `shardKey(req.Shard)` is hoisted into a local instead of being computed twice, and the shard label uses `strconv.FormatUint` rather than `fmt.Sprintf`, which allocates on the hot path.
- `pkg/segmentwriter` tests pass except `TestBusyIngestLoop`, which hangs identically on clean `main` (verified at `7a039f572`) — pre-existing, not a flake.


aleks-p (dismissed):
Looks good and based on real profiling data it should have measurable impact

Image

The only thing I would consider is reusing `shardKey(req.Shard)` (we do it twice now I think)

simonswine (commented):
I am not fully sure if we should do this, it obviously changes the metric from being "ingested", to the profiles received. In the wider picture I don't think those two values differ significantly, but I would rather rename the metric to make this clear. 

```
			prometheus.HistogramOpts{
				Namespace:                       "pyroscope",
				Subsystem:                       "segment_writer",
				Name:                            "segment_ingest_bytes",
				Buckets:                         prometheus.ExponentialBucketsRange(10*1024, 15*1024*1024, 20),
				NativeHistogramBucketFactor:     1.1,
				NativeHistogramMaxBucketNumber:  32,
				NativeHistogramMinResetDuration: time.Minute * 15,
			},
```

aleks-p:
Right, I thought at first this moves the code but actually doing it before the split changes the semantics.

marcsanmi:
Renamed to `pyroscope_segment_writer_received_bytes` with a Help string; also`shardKey(req.Shard)` is hoisted now.

Dashboards updated; happy to close it if you feel this is too much and leave things as they were.

pkg/segmentwriter/segment_metrics.go
```diff
@@ -37,11 +37,13 @@ func newSegmentMetrics(reg prometheus.Registerer) *segmentMetrics {
...
 	m := &segmentMetrics{
-		segmentIngestBytes: prometheus.NewHistogramVec(
+		receivedBytes: prometheus.NewHistogramVec(
 			prometheus.HistogramOpts{
-				Namespace:                       "pyroscope",
-				Subsystem:                       "segment_writer",
-				Name:                            "segment_ingest_bytes",
+				Namespace: "pyroscope",
+				Subsystem: "segment_writer",
+				Name:      "received_bytes",
```
simonswine: Nit: In the distributor we have `pyroscope_distributor_received_decompressed_bytes`, maybe this is more consistent:

```suggestion
				Name:      "received_decompressed_bytes",
```

pkg/segmentwriter/service.go
```diff
@@ -261,8 +262,12 @@ func (i *SegmentWriterService) Push(ctx context.Context, req *segmentwriterv1.Pu
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
+	shard := shardKey(req.Shard)
+	i.segmentWriter.metrics.receivedBytes.
+		WithLabelValues(strconv.FormatUint(uint64(shard), 10), req.TenantId).
```
simonswine: Nit: Maybe add this as `func (k shardKey) String() string`, could also replace: 

https://github.com/grafana/pyroscope/blob/07173f517a8a4de582045dbb4b70ce4634ce11b8/pkg/segmentwriter/segment.go#L187

simonswine (dismissed):
Two nits, otherwise LGTM

simonswine (approved):
Thanks for following up! LGTM

PR #5314, opened 2026-07-06
feat(symdb): cross-query decoded-symbol cache for the store-gateway

1linkovdim:
## Summary

Adds an opt-in, process-level, byte-bounded LRU cache of decoded symbol tables (locations / mappings / functions / strings) in symdb, keyed by (block ULID, partition). On a cache hit the store-gateway skips re-decoding a block's symbols from object storage.

## Motivation

Store-gateway blocks are opened once and reused across queries (singleBlockQuerier.Open is idempotent; blocks are closed only on drop, ~every 15m sync). But their symbol tables are decoded from parquet on every query and freed immediately after (refctr). For MetaVersion3 blocks — served via symdb.Open, where the decoded symbols are the data the query actually uses — this repeats a large, pure-overhead decode.

Measured on a real production MetaVersion3 block: a single fleet-wide (`{}`) SelectMergeByStacktraces decodes ~44 GiB across hundreds of partitions, with ~80% of query CPU in parquet-go page decode — re-done for every query even though the block is long-lived.

## Design

- Partition-level hook, transparent to the symdb internal format (V2 parquet or V3 raw) — it operates on the decoded slices, below initParquetTables/initTables.
- Ownership & threading: one cache per store-gateway BucketStores, passed to each block via NewSingleBlockQuerierFromMeta(..., WithSymbolCache(c)) → symdb.Open(..., WithSymbolCache(c)).
- Load-once: concurrent fetchers of the same partition share a single decode via a per-partition barrier (so N in-flight queries → one decode).
- Bounding & safety: eviction is LRU RemoveOldest to a byte budget, so the resident set is hard-bounded by maxBytes. Safety is GC-based — a running query holds a reference to the decoded slices, so an entry evicted mid-query is never freed out from under it (worst case: a later query re-decodes). Hot entries are protected by LRU recency. Total process memory is a soft bound (byte cost is estimated; entries evicted while borrowed linger until the query ends), with GOMEMLIMIT as the global backstop.
- Block-drop purge: Reader.Close purges the block's entries to reclaim budget — cold entries otherwise age out via LRU.
- Off by default (blocks-storage.bucket-store.symbol-cache-max-bytes=0): the reader takes a zero-overhead fast path (a single fetch transaction, identical to before). Metrics:
pyroscope_store_gateway_symbol_cache_{hits,misses,evictions}_total, _size_by

## Results (warm queries)
```
Real MetaVersion3 block (4 hour long L3 compaction block) (benchstat, n=6):
  metric      no cache  with cache  change
  ---------  ---------  ----------  ------
  sec/op     2616.9 ms    425.0 ms  -83.8%
  B/op       41.65 GiB   1.573 GiB  -96.2%
  allocs/op    23.96 M     4.488 M  -81.3%
```

```
Generated 256-partition block (benchstat, n=6):
  metric      no cache  with cache  change
  ---------  ---------  ----------  ------
  sec/op      866.6 ms    338.5 ms  -60.9%
  B/op       12.59 GiB   141.8 MiB  -98.9%
  allocs/op    7.609 M     1.058 M  -86.1%
```

## Tests

symbol_cache_test.go: consecutive-query reuse (1 miss + N−1 hits, same backing array reused, purged only on Close); miss ownership transfer (table wrappers released, decoded arrays
survive and stay valid); concurrent load-once (asserts a single decode under; block purge. All symdb / phlaredb / storegateway suites pass; symdb isclean under -race.

## Compatibility

Off by default and purely additive — no behavior change unless a byte budget is configured. Threading uses variadic options, so existing callers are unaffected.


1linkovdim:
I've now had a chance to run a build including these changes behind the flag on a production store-gateway deployment (8 instances, replication factor 3, MetaVersion3 blocks, multi-tenant), and it holds up well against the offline benchstat numbers in the PR description.

Setup: store_gateway.bucket_store.symbol_cache_max_bytes: 8589934592 (8 GiB). Backing store is an S3 bucket fronted by a local disk read-cache, so S3 latency is largely out of the picture — the remaining query cost is dominated by parquet symbol decode, which is exactly what this cache targets.

Method: canonical query held constant — SelectMergeStacktraces render, selector {}, max-nodes=16384, fixed 5h historical window, several iterations (iter 1 cold, rest warm). Compared flag-off vs flag-on on the same window, across two block formats.

Warm render latency, same-window A/B (flag off → on):
```
Block format                           Cache off  Cache on  Speedup
-------------------------------------  ---------  --------  -------
symbols split by fingerprint           ~7.1s      1.77s     ~4x
symbols split by stacktrace partition  ~3.9s      0.74s     ~5x
```
The ~4–5× warm speedup is independent of how the compactor lays out symbols. (The second format keeps each shard's symbols disjoint, so in our case there's ~2.4× less symbol data to begin with — the cache then removes the repeated decode of what remains. The two optimizations are complementary.)

Cache behavior over the run (summed across the 8 instances):
- 73–76% hit rate — misses are first-touch decodes of each (block, partition); every warm reuse is a hit. Hit rate matches the expected warm-up curve as partitions become resident.
- Resident footprint ~0.5–0.9 GiB per instance (~5 GiB fleet-wide) — well under the 8 GiB budget per instance.
- Zero evictions — the working set fit comfortably; nothing was re-decoded under memory pressure.
- No measurable movement against the soft memory limit; the default-off nil path remains zero-overhead when unset.

This lines up with the PR's offline result (−84% warm latency / −96% allocations on a real block) and with the measured sizing (~0.27 MiB resident per partition, ~1.7× the on-disk symbol parquet).

jake-kramer:
@1linkovdim Thank you for your contributions, these look like very impactful patches. Our v1 code paths are now under maintenance mode and we are encouraging users to migrate to v2; we are hesitant to merge PRs that are not fixing critical bugs or security vulnerabilities

The performance characteristics of v2 are fairly different than v1 and I imagine your usage patterns lend to better performance in a v2 deployment. Here are some high-level details about Pyroscope v2: https://grafana.com/blog/pyroscope-2-0-release. We would love to hear from you about your experiences using Pyroscope and if there are any major barriers to a v2 migration. You can join us in our [Community Slack](https://slack.grafana.com/) (#pyroscope) to get in touch.

PR #5318, opened 2026-07-08
feat(query): add SymbolRefTable and symbol_refs tree fields to the query API

jake-kramer:
## Part of a series: symbol-aware tree references (2 of 8)

Full context and series overview: #5317 (PR 1 of the series). Compact map:

2. **proto: `SymbolRefTable` + tree query/report fields ← this PR**
3. `model/symbolref`: intern table + merge rebasing
4. `model/symbolref`: job grouping + tree rebuild
5. symdb + querybackend: produce and aggregate symbol-ref trees

Tracking issue: grafana/pyroscope-squad#487.

## What

The wire schema for tree reports that carry unresolved symbol references — a compact side table
alongside the existing tree bytes:

```proto
message SymbolRefTable {
  // Resolved refs: 0 <= ref < len(names); names[ref] is the frame name.
  repeated string names = 1;
  // Unresolved refs: i = ref - len(names);
  //   build_ids[unresolved_build_id[i]], unresolved_address[i].
  repeated string build_ids = 2;            // distinct, each exactly once
  repeated string binary_names = 3;         // parallel to build_ids; fallback display
  repeated uint32 unresolved_build_id = 4;  // per unresolved entry
  repeated uint64 unresolved_address = 5;   // parallel to unresolved_build_id
}
```

A tree node reference is a single integer: references below `len(names)` are resolved frame
names; references at or above it index the unresolved `(build ID, address)` entries.
`TreeQuery.symbol_refs` requests the new report shape (mutually exclusive with `full_symbols`);
`TreeReport.symbol_refs` carries the table.

Schema and generated code only — nothing reads or writes these fields until later PRs in the
series, so this is inert on its own.

## Why this shape

- **Build IDs are interned** (each once, with per-entry indexes) instead of repeating a 40-64
  character string per reference. Later PRs defer tree truncation while unresolved references
  exist, so side tables travel through every merge uncapped — per-entry weight matters.
- **`binary_names`** (one per build ID) preserves the existing `binary!0xaddr` fallback rendering
  for addresses that fail to resolve.
- **Unresolved entries are sorted by `(build ID, address)`** at marshal time, giving deterministic
  merge output and single-scan grouping into per-build-ID resolution jobs.

## Testing

`make generate` with zero spillover outside `api/`; `go build` clean in the root and `api/`
modules. Field numbers verified against current `main` (`TreeQuery.symbol_refs = 7`,
`TreeReport.symbol_refs = 4`). No behavioral surface to test yet.

### DIFF
--- a/api/query/v1/query.proto
+++ b/api/query/v1/query.proto
@@ -258,9 +258,11 @@ message TreeSymbols {
 //                      unresolved_address[u]              address within it
 //
 // build_ids and binary_names are parallel and 1:1 (binary_names[k] is the
-// "binary!0xaddr" fallback display for build_ids[k]); each build ID appears
-// once. unresolved_build_id and unresolved_address are parallel, one pair per
-// unresolved entry, sorted by (build ID, address).
+// "binary!0xaddr" fallback display for build_ids[k]); each (build ID,
+// binary name) pair appears once — the same build ID may repeat under
+// different binary names, each retained exactly as stored.
+// unresolved_build_id and unresolved_address are parallel, one entry per
+// unresolved location, sorted by (build ID, binary name, address).
 message SymbolRefTable {
   repeated string names = 1;
   repeated string build_ids = 2;
--- a/api/query/v1/query.proto
+++ b/api/query/v1/query.proto
@@ -214,15 +214,26 @@ message TimeSeriesReport {
   repeated types.v1.Series time_series = 2;
 }
 
+// SymbolMode selects how a TreeQuery reports frame symbols.
+enum SymbolMode {
+  SYMBOL_MODE_UNSPECIFIED = 0;
+  SYMBOL_MODE_NAME = 1; // frame names only
+  SYMBOL_MODE_FULL = 2; // full pprof symbol tables alongside the tree
+  SYMBOL_MODE_REFS = 3; // compact symbol-ref side table for deferred resolution
+}
+
 message TreeQuery {
   int64 max_nodes = 1;
   repeated string span_selector = 2;
   optional types.v1.StackTraceSelector stack_trace_selector = 3;
   repeated string profile_id_selector = 4;
+  // Deprecated: use symbol_mode = SYMBOL_MODE_FULL. Retained for wire
+  // compatibility; will be removed in a couple of releases.
   bool full_symbols = 5;
   repeated string trace_id_selector = 6;
-  // symbol_refs and full_symbols are mutually exclusive.
-  bool symbol_refs = 7;
+  // symbol_mode selects the symbol output. When unset, full_symbols is honored
+  // for back-compat.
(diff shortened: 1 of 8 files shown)


jake-kramer:

- #5337
    - #5317
        - #5318 ◀
            - #5321
                - #5322
                    - #5332
                        - #5333
                            - #5334
                                - #5335
                                    - #5336

api/query/v1/query.proto
```diff
@@ -221,6 +221,8 @@ message TreeQuery {
   repeated string profile_id_selector = 4;
   bool full_symbols = 5;
   repeated string trace_id_selector = 6;
+  // symbol_refs and full_symbols are mutually exclusive.
+  bool symbol_refs = 7;
```
simonswine: I would prefer to mark `full_symbols` deprecated (and remove it in a couple of releases) and add a new enum symbol_mode:

```
SYMBOL_MODE_UNKNOWN
SYMBOL_MODE_NAME
SYMBOL_MODE_FULL
SYMBOL_MODE_REFS
```

 I think that provides most flexibility going forward, rather than having 2 (or more) mutually exclusive bools

api/query/v1/query.proto
```diff
@@ -234,10 +236,22 @@ message TreeSymbols {
...
 
+message SymbolRefTable {
+  // Resolved refs: 0 <= ref < len(names); names[ref] is the frame name.
+  repeated string names = 1;
+  // Unresolved refs: i = ref - len(names);
+  //   build_ids[unresolved_build_id[i]], unresolved_address[i].
+  repeated string build_ids = 2; // distinct, each exactly once
+  repeated string binary_names = 3; // parallel to build_ids; fallback display
+  repeated uint32 unresolved_build_id = 4; // per unresolved entry
+  repeated uint64 unresolved_address = 5; // parallel to unresolved_build_id
```
simonswine: Comments could a lot clearer. I think it makes sense to keep this flat and columnar (rather than a repeated struct), but I provide better examples and clearer language that build_ids and binary_names indexes are matching and so are the unresolved_*

simonswine (commented):
I think it makes sense, just would like to see slightly better explanations around it and an ENUM for the symbol output.

api/query/v1/query.proto
jake-kramer: Agreed, resolved in https://github.com/grafana/pyroscope/pull/5318/commits/bcdd64314c59e92eafa3a086b44d36aa61f005e1

api/query/v1/query.proto
jake-kramer: Resolved in https://github.com/grafana/pyroscope/pull/5318/commits/bcdd64314c59e92eafa3a086b44d36aa61f005e1

simonswine (dismissed):

simonswine (approved):

PR #5320, opened 2026-07-08
ci(backport): migrate to grafana-github-actions-go for signed commits

aleks-p:
Switch to `grafana-github-actions-go/backport` (see https://github.com/grafana/grafana-github-actions-go/pull/187), to unblock the backport workflow - currently broken because of the signed commit requirement.

Implementation borrowed (copied) from [Mimir](https://github.com/grafana/mimir/blob/3f0316590ef299b8bce578cc1b7f6bb592753ed0/.github/workflows/backport.yaml)

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,44 +1,52 @@
 name: Backport PR Creator
+run-name: "Backport for ${{ github.event.pull_request.title }} (#${{ github.event.pull_request.number }})"
+
 on:
   pull_request:
     types:
       - closed
       - labeled
 
+permissions:
+  contents: read
+  id-token: write
+
 jobs:
   main:
     # We don't run the backporting for PRs from forks because those can't access "pyroscope-development-app" secrets in vault.
     # We don't use GitHub actions app (secrets.GITHUB_TOKEN) because PRs created by the bot don't trigger CI.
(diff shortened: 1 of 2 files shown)


jake-kramer (commented):
I believe https://github.com/grafana/pyroscope/blob/6d07e117a36a52783d26b21651ad2c0a6fffd95a/docs/internal/RELEASE.md#backport should be updated (I don't think the expected labels are relevant anymore).

Is it possible to test this all in a fork?

aleks-p:
Thanks @jake-kramer 

I've updated the docs in 6d204e2 - the expected-labels gate was enforcement in the old action; the new one only looks at `backport release/vX.Y` labels.

> Is it possible to test this all in a fork?

To some extent yes but getting it to work in a fork won't guarantee it will work here as we are using Vault via OIDC, and org-wide rulesets for signed commits. The fact that the workflow works in [Mimir](https://github.com/grafana/mimir/actions/workflows/backport.yaml) and [Tempo](https://github.com/grafana/tempo/actions/workflows/backport.yml) should be a good enough signal imo, but I'd be happy to do a test in a fork if you think it will answer something those other repos don't.

jake-kramer (approved):
LGTM, ok to give it a try in main repo

PR #5321, opened 2026-07-08
feat(model): add symbolref package for symbol-aware tree references

jake-kramer:
## Part of a series: symbol-aware tree references (3 of 8)

3. **`model/symbolref`: intern table + merge rebasing ← this PR**
4. `model/symbolref`: job grouping + tree rebuild
5. symdb + querybackend: produce and aggregate symbol-ref trees

Tracking issue: grafana/pyroscope-squad#487. Based on #5318's branch; the diff here is only the
new package.

## What

`pkg/model/symbolref`: the intern table that pairs a `model.LocationRefNameTree` with the
`queryv1.SymbolRefTable` side message from #5318, so tree nodes can reference either resolved
frame names or unresolved `{build ID, address}` locations in one integer reference space.

- `Table` — content-addressed interning (`InternName`, `InternUnresolved`), safe for concurrent
  producers; structured as a lock-free core wrapped by a uniform locking boundary, so lock
  discipline is auditable at a glance and no function carries an unenforceable "must hold lock"
  precondition.
- `Table.Add` — merges another table's wire form and returns a remap function compatible with
  `TreeMerger.MergeTreeBytes(..., WithTreeMergeFormatNodeNames(remap))`, the same rebasing pattern
  the existing FullSymbols path uses. The `-1` truncation sentinel passes through unchanged.
- `ResultBuilder` — marshal-time compaction (`KeepRef`) and wire encoding (`Build`): only refs
  reachable in the marshaled tree are emitted; unresolved entries are sorted by
  `(build ID, address)` so merge output is deterministic and downstream grouping into
  per-build-ID jobs is a single scan. Built from a point-in-time snapshot taken under the table's
  lock.

Job grouping and the final tree rebuild (expanding resolved frames back into a plain tree) land
in the next PR; nothing consumes this package yet.

## Design notes

- **Reference 0 is reserved** (`InternName`'s first real call returns 1): any tree that
  round-trips the tree unmarshal path surfaces a synthetic virtual-root frame carrying the zero
  ref — existing consumers already special-case it — so a real name must never land on ref 0.
- **Merge determinism is semantic, not byte-identical**: per-stack values and the content-sorted
  unresolved wire fields are independent of partial arrival order; the name-table order is not.
  Final rendered output is unaffected (the eventual rebuild re-sorts structurally).

## Testing

Table-driven unit tests covering intern idempotency, remap round-trips, merge determinism under
permuted arrival order (all 3! orderings), wire-ordering, sentinel passthrough, empty/only-resolved
tables, and plain-tree absorption — 95.9% statement coverage under `-race`. Benchmarks for
intern+merge throughput at three cardinalities (100 / 10k / 1M addresses) and marshal-time
compaction of a 100k-node table, with throughput/size reported via custom metrics.

### DIFF
--- a/pkg/model/symbolref/table.go
+++ b/pkg/model/symbolref/table.go
@@ -240,19 +240,18 @@ func (c *tableCore) newResultBuilder() *ResultBuilder {
 		binaryNames:      c.binaryNames,
 		unresolvedBI:     c.unresolvedBI,
 		unresolvedAd:     c.unresolvedAd,
-		nameLookup:       make(map[int32]int32),
 		sortedUnresolved: sorted,
 		rank:             rank,
 	}
 }
 
-// ResultBuilder compacts and orders a Table's contents into a
-// queryv1.SymbolRefTable at marshal time. It is built from a point-in-time
-// snapshot of its Table's state, taken under the Table's lock (see
-// tableCore.newResultBuilder); ResultBuilder itself holds no reference back
-// to the Table and takes no lock of its own, so it is not safe for
-// concurrent use — it is meant to be driven single-threaded, once, after
-// every Add/Intern* call on its Table has completed.
+// ResultBuilder encodes a Table's contents into a queryv1.SymbolRefTable at
+// marshal time. It is built from a point-in-time snapshot of its Table's
+// state, taken under the Table's lock (see tableCore.newResultBuilder);
+// ResultBuilder itself holds no reference back to the Table and takes no
+// lock of its own, so it is not safe for concurrent use — it is meant to be
+// driven single-threaded, once, after every Add/Intern* call on its Table
+// has completed.
 type ResultBuilder struct {
 	namesLen int // snapshot of len(names) at construction; the resolved/unresolved dividing line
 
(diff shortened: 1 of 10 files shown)


jake-kramer:

- #5337
    - #5317
        - #5318
            - #5321 ◀
                - #5322
                    - #5332
                        - #5333
                            - #5334
                                - #5335
                                    - #5336

pkg/model/symbolref/table.go
```diff
@@ -0,0 +1,366 @@
...
+		biOut, ok := buildIDOut[bi]
+		if !ok {
+			biOut = int32(len(pb.BuildIds))
+			buildIDOut[bi] = biOut
+			pb.BuildIds = append(pb.BuildIds, rb.buildIDSl[bi])
+			pb.BinaryNames = append(pb.BinaryNames, rb.binaryNames[bi])
+		}
+		pb.UnresolvedBuildId[out] = uint32(biOut)
+		pb.UnresolvedAddress[out] = rb.unresolvedAd[idx]
+	}
```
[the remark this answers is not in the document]
jake-kramer: Resolved by https://github.com/grafana/pyroscope/pull/5321/commits/6244e3a4dfc384c52e1983225e527041596492f5 which emits full name snapshot

pkg/model/symbolref/table.go
```diff
@@ -0,0 +1,373 @@
...
+	if pb == nil {
+		pb = &queryv1.SymbolRefTable{}
+	}
+
+	pb.Names = make([]string, rb.namesLen)
+	copy(pb.Names, rb.namesSl)
+
+	n := len(rb.sortedUnresolved)
+	pb.BuildIds = make([]string, 0, n)
+	pb.BinaryNames = make([]string, 0, n)
```
simonswine: I imagine `n` might be a lot bigger than what I would expect in a typical query. Would probably have a constant like `8`/`16` or start with empty ones.

pkg/model/symbolref/table.go
```diff
@@ -0,0 +1,373 @@
...
+// hashedSlice is a content-addressed intern table for strings: repeated add
+// calls with equal values return the same index. Modeled on the hashedSlice
+// pattern in pkg/phlaredb/symdb/symbol_merger.go, specialized to strings
+// since both names and build IDs interned here are strings.
+type hashedSlice struct {
+	m  map[uint64]int32
+	sl []string
+}
+
+func newHashedSlice() *hashedSlice {
```
simonswine: Nit: There is a very similar generic `hashedSlice `implementation in pkg/phlaredb/symdb/symbol_merger.go. Maybe we can share it with little effort?

pkg/model/symbolref/table.go
```diff
@@ -0,0 +1,373 @@
...
+// internBuildID interns buildID (deduplicated) and records binaryName the
+// first time a given buildID is seen; later calls for the same buildID
+// keep the first binaryName.
+func (c *tableCore) internBuildID(buildID, binaryName string) int32 {
+	before := c.buildID.len()
+	bi := c.buildID.add(buildID)
+	if int(bi) == before {
+		c.binaryNames = append(c.binaryNames, binaryName)
+	}
+	return bi
```
simonswine: I don't think this will result in deterministic behaviour as the first arriving wins. This might see flamegraph for the same data change, which we should avoid.

I also think it might be possible to validly have multiple build IDs with different binary names:

Let's say I take a `pyroscope` binary and rename it into `pyroscope-server` and run both images' profiles in a query, they would have the same `build_Id` and their names would be jumping around from one to the other depending on query executing timings.

IMHO the correct option is to retain the binary name as it is stored and not make build ID values distinct on their own.

simonswine (commented):
A few comments but generally pretty close to LGTM

simonswine (dismissed):

simonswine (approved):

PR #5322, opened 2026-07-08
feat(read-path): produce and aggregate symbol-ref trees (model, symdb, query backend)

jake-kramer:
## Part of a series: symbol-aware tree references (4 of 6)

1. symbolizer: expose address-level `Resolve` API (#5317) — merged
2. proto: `SymbolRefTable` + `SymbolMode` tree query fields (#5318) — merged
3. `model/symbolref`: intern table + merge rebasing (#5321) — merged
4. **backend half: `model/symbolref` grouping + rebuild, symdb resolver, querybackend aggregation ← this PR**
5. frontend: orchestration behind a default-off per-tenant flag — nothing activates before this (#5334)
6. integration tests + docs (#5335)

Tracking issue: grafana/pyroscope-squad#487.

> **Note:** this PR consolidates what were previously three PRs (#5322, #5332, #5333) after
> feedback that the stack was too fine-grained to review. Commits remain separated per
> component for commit-by-commit review; earlier review discussion lives on the closed PRs.

## What

Everything the read path needs to *produce* symbol-ref trees, bottom to top. Nothing is
reachable until the frontend PR (#5334) wires it up behind the per-tenant flag.

**`pkg/model/symbolref` consumer half** (`unresolved.go`, `rebuild.go`):

- **`UnresolvedBinaries`** groups a table's unresolved references by build ID — one
  `UnresolvedBinary{BuildID, BinaryName, Addresses}` per distinct binary, addresses sorted and
  deduplicated, binary name carried for fallback rendering. Input is validated (paired-slice
  lengths, build-ID index range) before grouping; a single forward scan when the input follows
  the wire contract's ordering, with a defensive sort otherwise.
- **`Rebuild`** expands a merged reference tree back into a plain `FunctionName` tree:
  references resolve through a caller-supplied callback (memoized per distinct location within
  a call); multi-frame inline chains expand into that many tree levels (root-first); unresolved
  locations render the legacy `binary!0xaddr` fallback. Truncation happens exactly once, after
  every stack is expanded.

**symdb resolver output** (`Resolver.SymbolRefTree`):

A new output shape alongside `Tree`/`LocationRefNameTree`/`Pprof`. Locations with line info
contribute resolved frame names; line-less locations contribute unresolved
`{build ID, address}` references, carrying the mapping's binary name. A line-less location
without a build ID cannot be symbolized and renders the legacy `binary!0xaddr` fallback
inline. Trees build untruncated — truncation is deferred until after symbolization, since a
truncated-away unresolved node might merge with a kept one once resolved.

**querybackend production and aggregation**:

- **Per-dataset gating.** When a tree query requests `SYMBOL_MODE_REFS`, a dataset labeled
  `__unsymbolized__` at ingest builds a symbol-ref tree; an unlabeled, already-symbolized
  dataset keeps today's exact `FunctionName` path. The label is read from dataset metadata at
  query-execution time, so queryless queries routed through the tenant-wide index work too.
- **Content-driven collapse.** A labeled dataset (or a fully merged result) whose selected
  samples produced no unresolved locations converts back to a plain, truncatable tree — only
  data that actually needs symbolization pays the deferred-truncation cost.
- **Mixed-shape aggregation.** The aggregator merges partials of both shapes into one result:
  symbol-ref partials rebase their references through a shared table; plain partials are
  absorbed by interning their names, with the truncation "other" sentinel mapped explicitly.

No pprof detour is taken on any of these paths, so every tree-query selector (span, trace,
profile-ID, stack-trace) flows through unchanged.

## Bounding memory: the unresolved-locations limit

Deferred truncation must not mean unbounded memory. The query carries
`max_unresolved_locations` (stamped by the frontend from the per-tenant
`symbolizer.max-unresolved-locations` override, default 1,000,000; 0 disables), and the
backend enforces it **fail-fast at both levels**: each dataset's tree build stops at the cap
(`WithResolverSymbolRefCap`), and the aggregator re-checks the merged table after every
partial it absorbs. A query over the limit fails with `too many unresolved locations in
query` rather than degrading — a partial fallback rendering would keep growing the result
anyway and make output depend on merge order.

## Notes

- Call-site (`StackTraceSelector`) selection matches locally resolved names only, mirroring
  the `FunctionName` path: a stack whose match depends on a not-yet-symbolized frame is not
  selected. Called out in code; lifting it requires filtering after symbolization.
- The ingest rewrite assigns a partition's first mapping index 0 — the same value used for
  "no mapping" — so kernel/JIT frames and a line-less location on the first mapping are not
  always distinguishable. Documented and pinned by a test.

## Testing

`go test -race` green across `pkg/model/symbolref`, `pkg/phlaredb/symdb`, `pkg/querybackend`.
Coverage includes: grouping/rebuild round-trips with inline chains and malformed-input
validation; mixed lined/line-less fixtures; `MappingId==0` rendering; multi-partition
reference consistency; cap-overflow fail-fast at build and at merge; the dataset gate; the
aggregator across all-plain / all-ref / mixed / "other"-preservation; and a native-dataset
plain-path test on the real block fixture. `golangci-lint` clean; whole-repo build clean.

### DIFF
--- a/pkg/model/symbolref/rebuild.go
+++ b/pkg/model/symbolref/rebuild.go
@@ -1,0 +1,141 @@
+package symbolref
+
+import (
+	"fmt"
+	"slices"
+
+	queryv1 "github.com/grafana/pyroscope/api/gen/proto/go/query/v1"
+	"github.com/grafana/pyroscope/v2/pkg/model"
+)
+
+// Frame is one resolved stack frame; deliberately minimal, with no lidia
+// dependency.
+type Frame struct {
+	Name string
+}
+
+// resolveKey memoizes Rebuild's resolve calls by (build ID index, address)
+// within a single call, so a repeated address is resolved once.
+type resolveKey struct {
+	buildID uint32
+	addr    uint64
+}
+
+// Rebuild expands every model.LocationRefName ref in treeBytes (a
+// model.LocationRefNameTree marshal paired with pb) into resolved
+// model.FunctionNames, rebuilds a plain tree from the expanded stacks, and
+// truncates exactly once, via a single tree.Bytes(maxNodes, nil) call after
+// every stack has been expanded and reinserted.
+//
+// resolve returning nil or an empty slice for a given (buildID, addr) means
+// "could not resolve"; Rebuild then synthesizes exactly one fallback frame
+// named fmt.Sprintf("%s!0x%x", binaryName, addr), substituting "unknown" for
+// an empty binaryName — matching the pprof detour's legacy fallback
+// rendering byte for byte. A resolved chain's frame order is copied through
+// from resolve unchanged, expanding one address into that many tree levels.
+// Because every stack is reinserted from scratch, two refs that expand to
+// the same name at the same position merge structurally, with no separate
+// dedup pass.
+//
+// Returns an error only for malformed input: an unmarshalable treeBytes, a
+// structurally inconsistent pb, or a tree ref outside pb's valid range.
+func Rebuild(
+	treeBytes []byte,
+	pb *queryv1.SymbolRefTable,
+	resolve func(buildID string, addr uint64) []Frame,
+	maxNodes int64,
+) ([]byte, error) {
+	if err := validateSymbolRefTable(pb); err != nil {
+		return nil, err
+	}
+
+	src, err := model.UnmarshalTree[model.LocationRefName, model.LocationRefNameI](treeBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	names := pb.GetNames()
+	buildIDs := pb.GetBuildIds()
+	binaryNames := pb.GetBinaryNames()
+	unresolvedBuildID := pb.GetUnresolvedBuildId()
+	unresolvedAddress := pb.GetUnresolvedAddress()
+	numUnresolved := len(unresolvedAddress)
+
+	cache := make(map[resolveKey][]model.FunctionName)
+
+	expand := func(ref model.LocationRefName) ([]model.FunctionName, error) {
+		switch {
+		case ref == model.OtherLocationRef:
+			return []model.FunctionName{model.OtherFunctionName}, nil
+		case ref < 0:
+			return nil, fmt.Errorf("symbolref: ref %d out of range", ref)
+		case int(ref) < len(names):
+			return []model.FunctionName{model.FunctionName(names[ref])}, nil
+		}
+
+		i := int(ref) - len(names)
+		if i >= numUnresolved {
+			return nil, fmt.Errorf("symbolref: ref %d out of range (len(names)=%d, len(unresolved)=%d)", ref, len(names), numUnresolved)
+		}
+
+		bi := unresolvedBuildID[i]
+		addr := unresolvedAddress[i]
+		key := resolveKey{buildID: bi, addr: addr}
+		if frames, ok := cache[key]; ok {
+			return frames, nil
+		}
+
+		frames := expandOne(resolve(buildIDs[bi], addr), buildIDBinaryName(binaryNames, bi), addr)
+		cache[key] = frames
+		return frames, nil
+	}
+
+	dst := new(model.FunctionNameTree)
+	expanded := make([]model.FunctionName, 0, 64)
+	var rebuildErr error
+	src.IterateStacks(func(_ model.LocationRefName, self int64, stack []model.LocationRefName) {
+		if rebuildErr != nil {
+			return
+		}
+		slices.Reverse(stack)
+		expanded = expanded[:0]
+		for _, ref := range stack {
+			frames, err := expand(ref)
+			if err != nil {
+				rebuildErr = err
+				return
+			}
+			expanded = append(expanded, frames...)
+		}
+		dst.InsertStack(self, expanded...)
+	})
+	if rebuildErr != nil {
+		return nil, rebuildErr
+	}
+
+	return dst.Bytes(maxNodes, nil), nil
+}
+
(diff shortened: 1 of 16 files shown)


jake-kramer:

- #5337
    - #5317
        - #5318
            - #5321
                - #5322 ◀
                    - #5332
                        - #5333
                            - #5334
                                - #5335
                                    - #5336

pkg/model/symbolref/unresolved.go
```diff
@@ -0,0 +1,105 @@
...
+
+// UnresolvedBinaries groups a table's unresolved references by build ID,
+// one entry per distinct binary referenced. ResultBuilder.Build already
+// sorts unresolved entries by (buildID, address), making contiguous-run
+// grouping a single forward pass; a table from a different producer may not
+// be grouped that way, so UnresolvedBinaries falls back to sorting first
+// when it detects that.
+func UnresolvedBinaries(pb *queryv1.SymbolRefTable) []UnresolvedBinary {
+	buildIdx := pb.GetUnresolvedBuildId()
+	addrs := pb.GetUnresolvedAddress()
```
simonswine, line 29: Nit: Also validation here might be good

pkg/model/symbolref/rebuild.go
```diff
@@ -0,0 +1,149 @@
...
+	src, err := model.UnmarshalTree[model.LocationRefName, model.LocationRefNameI](treeBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	names := pb.GetNames()
+	buildIDs := pb.GetBuildIds()
+	binaryNames := pb.GetBinaryNames()
+	unresolvedBuildID := pb.GetUnresolvedBuildId()
+	unresolvedAddress := pb.GetUnresolvedAddress()
```
simonswine, line 61: Nit: Validate here that the length of the paired slices match and also if the buildID pointer is in range.

pkg/model/symbolref/rebuild.go
```diff
@@ -0,0 +1,149 @@
...
+	numUnresolved := len(unresolvedAddress)
+
+	cache := make(map[resolveKey][]model.FunctionName)
```
simonswine: I think this map is worth to be presizing:

```suggestion
	cache := make(map[resolveKey][]model.FunctionName, numUnresolved)
```

pkg/model/symbolref/rebuild.go
```diff
@@ -0,0 +1,143 @@
...
+
+// expandOne converts a resolve() result into the frame chain for one
+// unresolved location, synthesizing the fallback frame when resolution
+// failed.
+func expandOne(resolved []Frame, binaryName string, addr uint64) []model.FunctionName {
+	if len(resolved) == 0 {
+		if binaryName == "" {
+			binaryName = "unknown"
+		}
+		return []model.FunctionName{model.FunctionName(fmt.Sprintf("%s!0x%x", binaryName, addr))}
```
simonswine: I think the fallback naming should be sharing code.

pkg/phlaredb/symdb/resolver_symbolref_tree.go
```diff
@@ -0,0 +1,263 @@
...
+	return slot
+}
+
+func (b *symbolRefTreeBuilder) lookup(slot int32) model.LocationRefName {
+	return b.slots[slot]
+}
+
+// fallbackSymbolName matches Symbolizer.createFallbackSymbol
+// (pkg/symbolizer/symbolizer.go) byte for byte.
+func fallbackSymbolName(binaryName string, addr uint64) string {
```
simonswine: Should be sharing code with other fallback areas of pyroscope

- pkg/symbolizer/symbolizer.go:579
- pkg/model/symbolref/rebuild.go (noted below)

pkg/querybackend/query_tree.go
```diff
@@ -250,14 +375,115 @@ func (a *treeAggregator) build() *queryv1.Report {
...
+}
+
+// locationRefTreeToFunctionNameTree converts tree into a FunctionNameTree,
+// resolving every ref through rb. The caller must ensure rb's table has no
+// unresolved entries reachable from tree: any such ref renders as the
+// empty string, since there is no name to fall back to. rb may already
+// have observed refs from an earlier KeepRef/Build pass (e.g. a prior
+// marshal of the same tree): KeepRef is idempotent, so this only adds to
+// rb's state, never invalidates it.
+func locationRefTreeToFunctionNameTree(tree *model.LocationRefNameTree, rb *symbolref.ResultBuilder) *model.FunctionNameTree {
```
simonswine, line 445: I am a bit confused why this function has to do the 3 passes of the data.

I think it should be possible to do it in one pass. At the point of call we should be sure that it is all symbolised (`!HasUnresolved()`)

Maybe I am missing something?

pkg/phlaredb/symdb/resolver_symbolref_tree.go
```diff
@@ -0,0 +1,263 @@
...
+	var buildID, binaryName string
+	if int(loc.MappingId) < len(b.symbols.Mappings) {
+		mapping := b.symbols.Mappings[loc.MappingId]
+		buildID = b.symbols.Strings[mapping.BuildId]
+		binaryName = filepath.Base(b.symbols.Strings[mapping.Filename])
+	}
+	if buildID == "" {
+		return append(dst, b.toSlot(b.table.InternName(fallbackSymbolName(binaryName, loc.Address))))
+	}
+	if !b.capper.allow(buildID, loc.Address) {
```
simonswine: The cap key omits the binary name, so it counts differently from the table it's meant to bound. Seems like an artifact from refactoring earlier.

simonswine (approved):

PR #5332, opened 2026-07-08
feat(symdb): add symbol-ref tree resolver output

jake-kramer:
## Part of a series: symbol-aware tree references (5a of 8)

4. `model/symbolref`: unresolved-binary grouping + tree rebuild (#5322)
5. **symdb: symbol-ref tree resolver output ← this PR** · querybackend: query mode + aggregation (next)

Tracking issue: grafana/pyroscope-squad#487. Based on #5322's branch; the diff here is only the new symdb output path.

## What

Adds `Resolver.SymbolRefTree`, a new output shape alongside the resolver's existing `Tree`/`LocationRefNameTree`/`Pprof` methods. It builds a location-reference tree paired with a `symbolref` table: locations with line info contribute resolved frame names; line-less locations contribute unresolved `{build ID, address}` references, carrying the mapping's binary name so unresolved frames can render the existing `binary!0xaddr` fallback.

Trees build **untruncated** — the method takes no node limit, because truncation is deferred until after symbolization (a truncated-away unresolved node might merge with another into a kept node once resolved). To bound memory under that deferral, an optional cap limits distinct unresolved references; on overflow the remainder renders fallback names inline, which restores truncatability without failing the query or dropping samples.

Nothing calls `SymbolRefTree` yet — the query backend consumes it in the next PR.

## Notes

- Line-less locations without a mapping render as bare hex addresses, matching the existing pprof-path rendering.
- The ingest rewrite assigns a partition's first mapping index 0 — the same value used for "no mapping" — so kernel/JIT frames and a line-less location on the first mapping are not always distinguishable. The behavior is documented and pinned by a test rather than silently assumed.

## Testing

`go test ./pkg/phlaredb/symdb/... -race` green; new tests cover mixed lined/line-less fixtures, the `MappingId==0` rendering, multi-partition reference consistency, and cap-overflow semantics (no error, exact values, fallback rendering, truncation re-armed). `golangci-lint` clean; whole-repo build clean.

### DIFF
--- a/pkg/phlaredb/symdb/resolver_symbolref_tree.go
+++ b/pkg/phlaredb/symdb/resolver_symbolref_tree.go
@@ -1,0 +1,244 @@
+package symdb
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"strconv"
+	"sync"
+
+	"github.com/grafana/pyroscope/v2/pkg/model"
+	"github.com/grafana/pyroscope/v2/pkg/model/symbolref"
+	schemav1 "github.com/grafana/pyroscope/v2/pkg/phlaredb/schemas/v1"
+)
+
+// SymbolRefTree resolves the tree for the samples in appender into a
+// model.LocationRefNameTree whose node names are refs into table.
+//
+// A location with lines is interned per line into table via InternName (one
+// intern per line, including inlined frames), exactly as addFunctionNames
+// does for the FunctionName tree.
+//
+// A line-less location needs symbolization: it is interned via
+// InternUnresolved, keyed on the (build ID, address) of its mapping. The
+// mapping is read by direct index into symbols.Mappings, exactly as
+// copyMappings (resolver_pprof.go) does — in this representation index 0
+// is a real mapping, not a "no mapping" sentinel, so it must not be
+// special-cased. A location whose mapping carries no build ID cannot be
+// symbolized (a genuine no-mapping kernel/JIT frame, or a binary with no
+// GNU build ID): it renders as the bare hex address via InternName,
+// matching the legacy path, which leaves such locations unsymbolized.
+//
+// capper bounds the number of distinct unresolved entries table accepts
+// for this call; locations past the cap render an inline fallback name via
+// InternName instead of InternUnresolved (see unresolvedCap).
+//
+// The tree is always built without truncation (maxNodes 0): a tree with
+// unresolved entries must not be truncated before those are resolved.
+func (r *Symbols) SymbolRefTree(
+	ctx context.Context,
+	appender *SampleAppender,
+	selection *SelectedStackTraces,
+	table *symbolref.Table,
+	capper *unresolvedCap,
+) (*model.LocationRefNameTree, error) {
+	if !selection.HasValidCallSite() {
+		return &model.LocationRefNameTree{}, nil
+	}
+	samples := appender.Samples()
+	b := newSymbolRefTreeBuilder(r, samples, selection, table, capper)
+	if err := r.Stacktraces.ResolveStacktraceLocations(ctx, b, samples.StacktraceIDs); err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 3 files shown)


jake-kramer:

- #5337
    - #5317
        - #5318
            - #5321
                - #5322
                    - #5332 ◀
                        - #5333
                            - #5334
                                - #5335
                                    - #5336

pkg/phlaredb/symdb/resolver_symbolref_tree.go
```diff
@@ -0,0 +1,244 @@
...
+	defer c.mu.Unlock()
+	if _, ok := c.seen[key]; ok {
+		return true
+	}
+	if len(c.seen) >= c.max {
+		c.exceeded++
+		return false
+	}
+	c.seen[key] = struct{}{}
+	return true
```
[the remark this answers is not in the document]
jake-kramer: Resolved by https://github.com/grafana/pyroscope/pull/5332/commits/4b15fda0bf08d46a17b6755d4c3b45bf68b6fe56 which updates documentation to clarify intent of `exceeded`

pkg/phlaredb/symdb/resolver_symbolref_tree.go
```diff
@@ -0,0 +1,250 @@
...
+		buildID = b.symbols.Strings[mapping.BuildId]
+		binaryName = filepath.Base(b.symbols.Strings[mapping.Filename])
+	}
+	if buildID == "" {
+		return append(dst, b.toSlot(b.table.InternName(fallbackSymbolName(binaryName, loc.Address))))
+	}
+	if b.capper.allow(buildID, loc.Address) {
+		return append(dst, b.toSlot(b.table.InternUnresolved(buildID, binaryName, loc.Address)))
+	}
+	return append(dst, b.toSlot(b.table.InternName(fallbackSymbolName(binaryName, loc.Address))))
```
simonswine: I don't think the cap is really helping with the excessive memory consumption from excessive addresses. It will also make them not very comparable.

I think if we have a cap we should fail hard, and have it as an override to raise on a per tenant basis.

Falling back to string for some parts of the query and not for others will just create confusion and complexity

pkg/phlaredb/symdb/resolver_symbolref_tree.go
```diff
@@ -0,0 +1,250 @@
...
+		slotOf:    make(map[model.LocationRefName]int32),
+	}
+	if selection != nil && len(selection.callSite) > 0 {
+		b.funcNamesMatcher = b.funcNamesMatchSelection
+	}
+	return b
+}
+
+func (b *symbolRefTreeBuilder) InsertStacktrace(_ uint32, locations []int32) {
+	if b.funcNamesMatcher != nil {
```
simonswine, line 168: In order to support funcNamesMatcher we have to propagate all stacktrace that contain an unsymbolized location and only filter after the query has been symbolized.

I think it is fair enough to miss unsymbolized profiles, but we should clearly call it out in a comment and with a follow issue

simonswine (commented):
I think we should reevaluate the cap here

pkg/phlaredb/symdb/resolver_symbolref_tree.go
jake-kramer: Hard fails now with `ErrTooManyUnresolvedLocations`

pkg/phlaredb/symdb/resolver_symbolref_tree.go
jake-kramer: Comment added, issue created https://github.com/grafana/pyroscope-squad/issues/857

pkg/phlaredb/symdb/resolver_symbolref_tree.go
jake-kramer: Also have a per-tenant limit: `symbolizer.max-unresolved-locations `

jake-kramer:
Folded into #5322 together with #5333, consolidating the backend half of the series into one PR for easier review. The commits are preserved there per component; the cap-rework discussion on this PR remains the record — the fail-fast semantics it led to are described in #5322's updated description.

PR #5333, opened 2026-07-08
feat(querybackend): produce and aggregate symbol-ref tree reports

jake-kramer:
## Part of a series: symbol-aware tree references (5b of 8)

4. `model/symbolref`: unresolved-binary grouping + tree rebuild (#5322)
5. symdb: symbol-ref tree resolver output (#5332) · **querybackend: query mode + aggregation ← this PR**

Tracking issue: grafana/pyroscope-squad#487. Based on #5332's branch; the diff here is only the query-execution and aggregation changes.

## What

Teaches the query backend to produce and merge symbol-ref tree reports.

- **Per-dataset gating.** When a tree query requests symbol references (mutually exclusive with full symbols — validated), a dataset labeled `__unsymbolized__` at ingest builds a symbol-ref tree (via `Resolver.SymbolRefTree` from #5332); an unlabeled, already-symbolized dataset keeps today's exact `FunctionName` path. The label is read from dataset metadata available at query-execution time, so this works for queryless queries routed through the tenant-wide index — the gap that motivated the issue.
- **Content-driven collapse.** A labeled dataset whose selected samples produced no unresolved locations (rollout skew, or a selector that only matched resolved stacks) converts back to a plain, truncatable tree — only data that actually needs symbolization pays the deferred-truncation cost.
- **Mixed-shape aggregation.** The tree aggregator merges partials of both shapes into one result: symbol-ref partials rebase their references through a shared table; plain partials (unlabeled datasets, or older backends mid-rollout) are absorbed by interning their names, with the truncation "other" sentinel mapped explicitly across the representation boundary. The merged output stays untruncated while any unresolved reference remains, and collapses to a plain tree otherwise.

No pprof detour is taken on any of these paths, so every tree-query selector (span, trace, profile-ID, stack-trace) flows through unchanged.

## Cap

Distinct unresolved references per dataset are capped (1,000,000 by default, a limit chosen empirically from the worst-case rebuild cost); on overflow, further locations render inline fallback names so the tree stays bounded. A counter tracks how often the cap is exceeded — the one operationally meaningful signal here, since exceeding it means symbolization silently degraded.

## Testing

`go test ./pkg/querybackend/... -race` green. New tests: the dataset gate (5 cases), the aggregator across all-plain / all-ref / mixed / "other"-preservation, mutual-exclusion validation, and a native-dataset plain-path integration test on the real block fixture. `golangci-lint` clean; whole-repo build clean.

An end-to-end test driving the labeled-dataset paths through a real block needs fixtures with `__unsymbolized__` datasets, which the integration suite (final PR of the series) introduces for the queryless-query case; the gate, aggregator, and helpers are unit-tested directly here, and the resolver semantics are covered in #5332.

### DIFF
--- a/pkg/querybackend/query_tree.go
+++ b/pkg/querybackend/query_tree.go
@@ -2,6 +2,7 @@ package querybackend
 
 import (
 	"fmt"
+	"slices"
 	"strings"
 	"sync"
 
@@ -9,9 +10,12 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 
+	metastorev1 "github.com/grafana/pyroscope/api/gen/proto/go/metastore/v1"
 	queryv1 "github.com/grafana/pyroscope/api/gen/proto/go/query/v1"
 	"github.com/grafana/pyroscope/v2/pkg/block"
+	"github.com/grafana/pyroscope/v2/pkg/block/metadata"
 	"github.com/grafana/pyroscope/v2/pkg/model"
+	"github.com/grafana/pyroscope/v2/pkg/model/symbolref"
 	parquetquery "github.com/grafana/pyroscope/v2/pkg/phlaredb/query"
 	v1 "github.com/grafana/pyroscope/v2/pkg/phlaredb/schemas/v1"
 	"github.com/grafana/pyroscope/v2/pkg/phlaredb/symdb"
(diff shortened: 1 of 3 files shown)


jake-kramer:

- #5337
    - #5317
        - #5318
            - #5321
                - #5322
                    - #5332
                        - #5333 ◀
                            - #5334
                                - #5335
                                    - #5336

pkg/querybackend/query_tree.go
```diff
@@ -100,6 +117,9 @@ func queryTree(q *queryContext, query *queryv1.Query) (*queryv1.Report, error) {
 	if query.Tree.StackTraceSelector != nil {
 		resolverOptions = append(resolverOptions, symdb.WithResolverStackTraceSelector(query.Tree.StackTraceSelector))
 	}
+	if query.Tree.SymbolRefs {
+		resolverOptions = append(resolverOptions, symdb.WithResolverSymbolRefCap(defaultSymbolRefUnresolvedCap))
```
simonswine: As discussed in #5332 the cap should be reworked

pkg/querybackend/query_tree.go
```diff
@@ -215,14 +336,124 @@ func (a *treeAggregator) build() *queryv1.Report {
...
+// empty string, since there is no name to fall back to. rb may already
+// have observed refs from an earlier KeepRef/Build pass (e.g. a prior
+// marshal of the same tree): KeepRef is idempotent, so this only adds to
+// rb's state, never invalidates it.
+func locationRefTreeToFunctionNameTree(tree *model.LocationRefNameTree, rb *symbolref.ResultBuilder) *model.FunctionNameTree {
+	var stacks []stackToInsert
+	tree.IterateStacks(func(_ model.LocationRefName, self int64, stack []model.LocationRefName) {
+		cp := append([]model.LocationRefName(nil), stack...)
+		slices.Reverse(cp)
+		if len(cp) > 0 && cp[0] == 0 {
```
simonswine: This sounds like a bug somewhere else the AI tries to patch up for

jake-kramer: Resolved here: https://github.com/grafana/pyroscope/commit/9a459464f

jake-kramer:
Folded into #5322 together with #5332 — the backend half of the series now reviews as one PR, commits preserved per component.

PR #5335, opened 2026-07-09
test(symbolizer): add symbol-ref tree integration tests and docs

jake-kramer:
## Part of a series: symbol-aware tree references (7 of 8)

Full context: #5317 (PR 1). The final functional PR: end-to-end integration tests + docs.

1. symbolizer: expose address-level `Resolve` API (#5317)
2. proto: `SymbolRefTable` + tree fields (#5318)
3. `model/symbolref`: intern table + merge rebasing (#5321)
4. `model/symbolref`: grouping + rebuild (#5322)
5. symdb resolver output (#5332) · querybackend mode + aggregation (#5333)
6. frontend orchestration behind a default-off per-tenant flag (#5334)
7. **integration tests + docs ← this PR**
8. symdb: compaction preserves line-less locations (#5337) — stacked at the bottom of the series, merges first

Based on #5334.

## What

Integration coverage with `symbolizer.symbol-ref-trees-enabled=true`, each mapped to a fix:
- **Queryless** `{}` symbolization via the tenant-wide index (the scenario that motivated grafana/pyroscope-squad#487).
- **Span-filtered** tree query symbolizes correctly.
- **Mixed** symbolized + unsymbolized datasets in one query.
- **Unknown build ID** -> `binary!0xaddr` fallback, end to end.

Fixtures push unsymbolized profiles through the real ingest path (authentic `__unsymbolized__` label), using the common single-mapping shape. Docs: the new flag, resolve timeout, queryless-now-works, and the fallback rendering.

## Testing

The four symbol-ref integration tests pass; the existing symbolization test still passes; build/vet/gofmt clean.

PR #5334, opened 2026-07-08
feat(frontend): resolve symbol-ref trees after the final merge

jake-kramer:
## Part of a series: symbol-aware tree references (6 of 8)

Full context and series overview: #5317 (PR 1). Compact map:

1. symbolizer: expose address-level `Resolve` API (#5317)
2. proto: `SymbolRefTable` + tree query/report fields (#5318)
3. `model/symbolref`: intern table + merge rebasing (#5321)
4. `model/symbolref`: unresolved-binary grouping + tree rebuild (#5322)
5. symdb: symbol-ref tree resolver output (#5332) · querybackend: query mode + aggregation (#5333)
6. **frontend: orchestration behind a default-off per-tenant flag ← this PR**
7. integration tests + docs (#5335)
8. symdb: compaction preserves line-less locations (#5337) — stacked at the bottom of the series, merges first

Tracking issue: grafana/pyroscope-squad#487. Based on #5333's branch.

## What

The activation PR. Behind a new per-tenant flag `symbolizer.symbol-ref-trees-enabled` (default **off**), tree and span-profile queries request symbol-ref tree reports instead of being rewritten through the pprof path when symbolization is needed. With the flag off, request construction and the legacy symbolization path are byte-identical to today — that's the rollback story.

When a merged report comes back with unresolved references, the frontend:
- groups them by binary (`symbolref.UnresolvedBinaries`),
- resolves each through the symbolizer's address-level API under bounded concurrency (the existing `max-debuginfod-concurrency`, shared with the pprof path) and a per-binary timeout (`symbolizer.resolve-timeout`, default 20s),
- rebuilds the final tree (`symbolref.Rebuild`), rendering the existing `binary!0xaddr` fallback for anything unresolved.

**Error semantics:** a per-binary timeout degrades that one binary to fallbacks and lets the query succeed; cancellation of the *request* context fails the query, matching how `SymbolizePprof` errors are already handled at these call sites.

## Observability

A single unsampled counter, `symbol_ref_locations_total{result=resolved|miss|timeout}`, drives the fallback-rate signal (an alert can't be built on sampled traces). The per-request detail — resolution timing, binary count, unresolved-reference count — lives on a `resolveSymbolRefs` tracing span instead of dedicated metrics.

## Testing

[...]

PR #5336, opened 2026-07-09
test(symbolizer): comparative benchmark, symbol-ref vs pprof-detour

jake-kramer:
## Part of a series: symbol-aware tree references (benchmark, on PR 7)

A comparative benchmark: new symbol-ref tree path vs the legacy pprof detour, on identical unsymbolized input with warm, debuginfod-free resolution (isolating the tree/pprof machinery). Both paths verified to produce byte-identical trees. Based on #5334's series.

## Result (honest)

| Size | pprof detour | symbol-ref |
|---|---|---|
| 100 | ~102µs / 1319 allocs | ~137µs / 941 allocs |
| 5000 | ~29ms / 55k allocs | ~28ms / 35k allocs |
| 50000 | ~2.8s / 551k allocs | ~2.7s / 351k allocs |

- **Allocations/memory: consistent win** for symbol-ref (29-36% fewer allocs, 7-12% less memory, all sizes).
- **Wall-clock: a crossover, not a win** — slower at small N (fixed overhead), comparable-within-noise at medium/large.

**Takeaway:** the new path's justification is correctness + GC efficiency, not raw latency.

## Limitation

Single-dataset comparison; does not capture the structural mixed-query advantage (a query-planning effect above this layer). Large-size absolute numbers are a synthetic worst case.

PR #5337, opened 2026-07-09
fix(symdb): preserve line-less locations when rewriting partitions without functions

jake-kramer:
## Part of a series: symbol-aware tree references (8 of 8)

2. proto: `SymbolRefTable` + tree fields (#5318)
4. `model/symbolref`: grouping + rebuild (#5322)
5. symdb resolver output (#5332) · querybackend mode + aggregation (#5333)
6. frontend orchestration behind a default-off per-tenant flag (#5334)
8. **symdb: compaction preserves line-less locations ← this PR — stacked at the bottom of the series, merges first**

Tracking issue: grafana/pyroscope-squad#487. Unlike the rest of the series this fixes pre-existing
shared code, so it is stacked first: the feature must not be enabled anywhere before it lands.

## What

`symdb.Rewriter.populateUnresolved` replaces every location of a partition whose functions table is
empty with a zero value: the guard added in #4051 drops the location's line records and then
`continue`s, skipping `unresolvedLocs.setValue(location)`. The location is never stored into the
lookup table, so the zero value `InMemoryLocation{}` is what gets written to the rewritten
partition — all locations of such a partition collapse into `{Address: 0, MappingId: 0}` (and dedup
into a single entry).

A partition without functions is not malformed: it is precisely a native profile that has not been
symbolized, where every location is line-less by definition. Both compactors (v1
`pkg/phlaredb/compact.go`, v2 `pkg/block/compaction.go`) rewrite symbols through this path, so the
first compaction permanently destroys the addresses of fully-unsymbolized datasets — source blocks
are deleted after compaction, making the loss unrecoverable. Read-time symbolization of compacted
data then renders `binary!0x0` fallback frames.

Partitions with at least one function are unaffected, which is why even partially-symbolized
profiles survive compaction — and why this went unnoticed: most agents symbolize locally before
pushing, so zero-function partitions are rare until server-side symbolization is in play.

## Fix

Keep dropping the line records (with no functions in the partition they cannot reference anything),
but store the location so its address and mapping reference survive the rewrite.

## Testing

- New `Test_Rewriter_location_addresses`: a zero-functions partition (fails without the fix — all
  locations collapse to one zero-valued entry) and a control partition with functions.
- The symbolization end-to-end test now runs past the first L0 compaction and asserts the
  symbolized frames survive, order-insensitively (compaction may renumber locations). Filler pushes
  drive the planner, since age-based batch flushing is only evaluated when another block arrives at
  the level. Without the rewriter fix it fails in seconds with `libfoo.so!0x0` fallback frames.
- Full `pkg/phlaredb/symdb` suite (incl. `-race`), v1 compaction tests (`pkg/phlaredb`), and v2
  `Test_CompactBlocks` (golden unchanged) pass.

### DIFF
--- a/pkg/phlaredb/symdb/rewriter.go
+++ b/pkg/phlaredb/symdb/rewriter.go
@@ -141,11 +141,14 @@ func (p *partitionRewriter) populateUnresolved(stacktraceIDs []uint32) error {
 		location := p.src.Locations[unresolvedLocs.At()]
 		location.MappingId = p.mappings.tryLookup(location.MappingId)
 		if len(p.src.Functions) == 0 {
+			// Line records cannot reference any function and are dropped,
+			// but the location itself must be kept: not-yet-symbolized
+			// native partitions consist entirely of line-less locations.
 			location.Line = nil
-			continue
-		}
-		for j, line := range location.Line {
-			location.Line[j].FunctionId = p.functions.tryLookup(line.FunctionId)
+		} else {
+			for j, line := range location.Line {
+				location.Line[j].FunctionId = p.functions.tryLookup(line.FunctionId)
+			}
 		}
 		unresolvedLocs.setValue(location)
 	}
--- a/pkg/phlaredb/symdb/rewriter_test.go
+++ b/pkg/phlaredb/symdb/rewriter_test.go
@@ -1,9 +1,14 @@
 package symdb
 
 import (
+	"context"
+	"slices"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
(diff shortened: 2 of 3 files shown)


jake-kramer:

- #5337 ◀
    - #5317
        - #5318
            - #5321
                - #5322
                    - #5332
                        - #5333
                            - #5334
                                - #5335
                                    - #5336

simonswine (approved):

simonswine (approved):
I think this is a bug introduced as part of https://github.com/grafana/pyroscope/pull/4051.

The `continue` skips locations that are supposed to remain: 
`unresolvedLocs.setValue(location)`

PR #5338, opened 2026-07-10
Update golang version to 1.25.12

aleks-p:
Routine patch bump, like #5230 for 1.25.11.


simonswine (approved):
Thank you for looking at this!

PR #5350, opened 2026-07-13
PoC: keep totals and labels from sampled profiles (delete stacktraces)

alsoba13:


alsoba13:
This was only a PoC. A better implementation is under development. Closing

PR #5352, opened 2026-07-13
chore(helm): improve chart description, re-triggering the 2.1.1 release

aleks-p:
The v2.1.1 chart release [failed at the index.yaml push](https://github.com/grafana/pyroscope/actions/runs/29113496513/job/86431235372) (GH013, fixed by #5351), so the chart tgz and GitHub release exist but `pyroscope-2.1.1` is missing from https://grafana.github.io/helm-charts and cannot be installed via `helm repo`.

Re-running the workflow does not help: `ct list-changed` compares against the latest tag reachable from main. Opting for this minor improvement instead of bumping to 2.1.2.


jake-kramer (approved):
😍

PR #5353, opened 2026-07-14
fix(query): return trace IDs in span heatmaps

simonswine:
## What this PR does

- reads trace IDs from span heatmap sample rows and carries them through heatmap construction
- returns trace IDs in public span exemplars and profilecli output
- includes trace ID in point identity so equal span IDs from different traces are not merged
- adds generated protobuf and OpenAPI artifacts

### Example outout

```
profilecli query exemplars span \
                --profile-type=process_cpu:cpu:nanoseconds:cpu:nanoseconds \
                --query='{service_name="pyroscope-query-backend"}' \
                --from="now-30m" \
                --to="now"  --max-label-columns=0
+----------------------------------+------------------+---------------------------+---------------------+
|             Trace ID             |     Span ID      |         Timestamp         | Value (nanoseconds) |
+----------------------------------+------------------+---------------------------+---------------------+
| eadd35b6c29d493756d6191a96015937 | ab2fe93b328e9abd | 2026-07-15T09:23:03+01:00 |                70ms |
| 2426b8b2ed5d565455724d4acec79cab | 639766feed5f0d1b | 2026-07-15T09:23:24+01:00 |                70ms |
| 4243c9a8ab31751000b57117c0cd3e3a | 0c4df958381c4ced | 2026-07-15T09:18:48+01:00 |                60ms |
+----------------------------------+------------------+---------------------------+---------------------+
```

## Tests

- `go test ./pkg/model/heatmap ./pkg/model/attributetable ./pkg/querybackend ./pkg/frontend/readpath/queryfrontend ./cmd/profilecli`
- `go test ./api/gen/proto/go/query/v1 ./api/gen/proto/go/types/v1`

## Notes

- During a mixed-version rollout, an old query-backend (without this change) can return a heatmap point without a trace ID while a new one returns the same point with a trace ID populated. The merger treats these as distinct points instead of summing them, producing a transient duplicate for that span until the rollout completes. This is expected and self-resolves once all query-backends are upgraded.

### DIFF
--- a/api/connect-openapi/gen/ingester/v1/ingester.openapi.yaml
+++ b/api/connect-openapi/gen/ingester/v1/ingester.openapi.yaml
@@ -1089,6 +1089,12 @@ components:
           description: |-
             Labels specific to this exemplar (e.g., pod name, node name). When an exemplar
              label overlaps with the series label, it is not included here.
+        traceId:
+          type: string
+          examples:
+            - 4bf92f3577b34da6a3ce929d0e0e4736
+          title: trace_id
+          description: Trace ID associated with the span, encoded as 32 lowercase hexadecimal characters.
       title: Exemplar
       additionalProperties: false
       description: |-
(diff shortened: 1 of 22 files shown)


alsoba13 (approved):
LGTM

PR #5354, opened 2026-07-14
feat: keep totals and labels of sampled-out profiles

alsoba13:
## What

When adaptive sampling rejects a profile, we used to drop it entirely, which left gaps in the time series and made the affected series disappear.

This keeps sampled-out profiles as **totals-only**: their values (totals) and labels are retained, the stacktraces are removed.

## Write path (distributor)

When a profile is sampled out and keeping is enabled for the tenant, the distributor reduces it to totals instead of dropping it:

- Samples are collapsed to **one sample per distinct label set**, each holding the summed values, so totals stay accurate.
- Stacktraces are removed: sample location references are cleared, the location/function/mapping tables are dropped, and the string table is rebuilt to keep only the strings still referenced (sample/period types, labels, …).
- Sample labels are preserved, so span- and trace-attributed totals remain distinguishable.
- Each sampled-out series is marked with an internal `__sampled__="true"` label.

## Read path (query-backend)

- **Stack-based queries** (flamegraph/tree, pprof, heatmap) always exclude these series with a `__sampled__!="true"` matcher — they have no stacktraces to contribute.
- **Time-series / totals queries** include them **only when enabled for the querying tenant**; otherwise they are excluded so time series match the pre-sampling behavior. Exemplars are suppressed for these series, since there is no stacktrace to drill into.

## Configuration — two per-tenant overrides (both default off)

This replaces the earlier single distributor flag. Both are **per-tenant limits**, so they can be enabled globally via `limits:` or per tenant via runtime overrides:

- **`keep_stripped_profiles`** — write path (distributor). CLI: `-distributor.sampling.keep-stripped-profiles`.
  - off → sampled-out profiles are dropped as before; on → they are retained as totals and marked `__sampled__`.
- **`include_stripped_profiles`** — read path (query-backend). CLI: `-query-backend.include-stripped-profiles`.
  - off → `__sampled__` series are excluded from time series; on → included.
  - Resolved **per querying tenant** (for a multi-tenant query, included only when enabled for every tenant).

## Metrics (bytes ingested vs dropped)

A retained sampled-out profile is stripped to totals up front, then accounted for like a normal profile:

- **`discarded_bytes_total{reason="dropped_by_sampling_rules"}`** gets the removed portion: `full profile size − kept totals size` (stacktraces, symbols, and the duplicate samples collapsed away).
- The **ingested** byte metrics (`distributor_received_decompressed_bytes`, usage-group received bytes) count only the kept totals.
- Fully dropped profiles (write path disabled) still count the whole profile as discarded.
- `received_compressed_bytes` is left as the raw wire size.

## Rollout

Two independent knobs, both default off:

1. **Deploy the release** — behavior-neutral, nothing is written or shown.
2. **Enable `keep_stripped_profiles`** (write) to start retaining sampled-out profiles as totals, and **`include_stripped_profiles`** (read) to surface those totals in time series. Because the two are independent, the read side can be enabled/validated separately from the write side.

Rolling back is flipping the relevant override off: disabling the write override stops producing `__sampled__` data; disabling the read override hides already-written `__sampled__` series from time series, so historical totals are not left inflated. Stack queries and exemplar suppression behave correctly regardless.

## Testing

- `go build`, and `pkg/distributor`, `pkg/querybackend`, `pkg/validation` unit tests pass; `cmd/pyroscope` `TestHelp` golden files regenerated.
- Verified end-to-end locally (sampled-out profiles keep their totals in the time series when enabled; flamegraphs exclude them).


pkg/querybackend/query_heatmap.go
alsoba13, line 1: For now, I just excluded sampled profiles from heatmap. Ideally we should use that to generate richer heatmaps, but I think we need to tweak few things (for example, allow HeatmapPoints's ProfileId to be empty, potentially changing the wip UI too)

pkg/distributor/distributor.go
```diff
@@ -578,16 +581,27 @@ func (d *Distributor) pushSeries(ctx context.Context, req *distributormodel.Prof
 	}
 
 	willSample, samplingSource := d.shouldSample(tenantID, groups.Names())
+	sampledOut := false
+	var fullBytes int64
```
aleks-p: I think we could remove both of these. If we can't, we probably need to rename the first (e.g. `stripped`) and remove the second and calculate the original profile size somewhere higher up, e.g. like in #5310 since it is generally expensive to do it multiple times.

If we calculate the size further up, we might be able to update it after we modify the profile and keep the rest of the code unchanged (see my next comment).

pkg/distributor/distributor.go
```diff
@@ -578,16 +581,27 @@ func (d *Distributor) pushSeries(ctx context.Context, req *distributormodel.Prof
...
+		if !d.cfg.KeepStrippedProfiles {
+			validation.DiscardedBytes.WithLabelValues(string(validation.SkippedBySamplingRules), tenantID).Add(float64(req.TotalBytesUncompressed))
+			groups.CountDiscardedBytes(string(validation.SkippedBySamplingRules), req.TotalBytesUncompressed)
+			return nil
+		}
+		sampledOut = true
+		fullBytes = int64(req.Profile.SizeVT())
+		for _, s := range req.Profile.Sample {
+			s.LocationId = nil
+		}
```
aleks-p: In my tests, we still send most of the profile (e.g. the entire string table) to segment writers which means unnecessary processing and resource usage. Could we strip the profile fully here, keeping only what we know we need? Ideally we only keep one sample with the total value and remove everything else (we'd need to be mindful of negative values, like "Normalize").

Then we should in theory be able to immediately report what was discarded (newSize - oldSize), and re-calculate the profile size (size = newSize), while keeping the rest of the distributor code unchanged. 

One wrinkle is that if we truncate the string table (which we likely should), language detection would need to happen earlier.

pkg/pyroscope/modules_experimental.go
```diff
@@ -383,12 +383,14 @@ func (f *Pyroscope) initQueryBackend() (services.Service, error) {
 		return nil, err
 	}
 	logger := log.With(f.logger, "component", "query-backend")
+	blockReader := querybackend.NewBlockReader(f.logger, f.storageBucket, f.reg)
+	blockReader.KeepStrippedProfiles = f.Cfg.Distributor.KeepStrippedProfiles
```
aleks-p: I think it might be cleaner to introduce a dedicated flag for the read path, and have both levers available to use.

Distributors decide if they "keep" stripped profiles, ideally we make this a tenant override (currently it is app/process level I think).

The read path decides if it "includes" stripped profiles in (some) responses. We could call this `querier_include_sampled_stripped_profiles` or similar, also as a tenant-override.

The benefit would be that we can enable stripping and check metrics independently of changing the shape of the read path data.

pkg/distributor/distributor.go
```diff
@@ -600,13 +609,18 @@ func (d *Distributor) pushSeries(ctx context.Context, req *distributormodel.Prof
 		finalLog.addFields("detected_language", profLanguage)
 	}
 
+	if stripped {
```
aleks-p: I still think we can remove this flag and keep the stripping code contained in one area. The `pushSeries` method is complex and often visited, so it would be nice to keep it as easy to maintain as possible.

My idea is to strip the profile in the block above, calculate the new sizes and move on as we did before. Here is an example done by a clanker, but the point stands:

single commit: https://github.com/grafana/pyroscope/commit/565a7d4ed
diff with main to show that most of the code stays contained in one place: https://github.com/grafana/pyroscope/compare/565a7d4ed

Side note: this is more aggressive, it only keeps a single sample per profile - discarding sample-level labels. I think this is be acceptable in some use cases, and might not be for others. We could make the sampling strategy a tenant-level setting, deciding between "drop entirely" (what we have today), "keep a single sample" (565a7d4ed), "keep a sample per sample-label" (de182eab8).

pkg/distributor/distributor.go
```diff
@@ -578,16 +582,21 @@ func (d *Distributor) pushSeries(ctx context.Context, req *distributormodel.Prof
...
 	willSample, samplingSource := d.shouldSample(tenantID, groups.Names())
+	stripped := false
 	if !willSample {
 		finalLog.addFields(
 			"usage_group", samplingSource.UsageGroup,
 			"probability", samplingSource.Probability,
 		)
-		finalLog.msg = "skipping profile due to sampling"
+		finalLog.msg = "sampling profile"
 		validation.DiscardedProfiles.WithLabelValues(string(validation.SkippedBySamplingRules), tenantID).Add(float64(req.TotalProfiles))
```
aleks-p, line 589: I think this does need to move to the if below

pkg/distributor/distributor.go
alsoba13: Cool, now I see what you mean. I was trying to push everything after normalize, but your code is way cleaner. I prefer to "keep a sample per sample-label" and keep all the series the client sends. From the point of view of the client it can be confusing to drop some labels but keep others. 

I will make this a tenant-level setting and we can extend to drop sample level labels in the future

pkg/pyroscope/modules_experimental.go
alsoba13: done!

pkg/distributor/distributor.go
alsoba13: moved a few lines below

pkg/distributor/distributor.go
```diff
@@ -583,11 +585,28 @@ func (d *Distributor) pushSeries(ctx context.Context, req *distributormodel.Prof
...
+			validation.DiscardedBytes.WithLabelValues(string(validation.SkippedBySamplingRules), tenantID).Add(float64(req.TotalBytesUncompressed))
+			groups.CountDiscardedBytes(string(validation.SkippedBySamplingRules), req.TotalBytesUncompressed)
+			return nil
+		}
+
+		// Language detection reads the string table, which is about to be
+		// stripped; the result is cached in the request.
+		d.GetProfileLanguage(req)
+		stripProfileToTotals(req.Profile.Profile)
+		req.Labels = phlaremodel.Labels(req.Labels).InsertSorted(phlaremodel.LabelNameSampled, "true")
```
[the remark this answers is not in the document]
alsoba13: IMO this is fine as is, the increase in series number by including this label should not interfere with the tenant's limit

pkg/pyroscope/modules_experimental.go
```diff
@@ -383,12 +383,14 @@ func (f *Pyroscope) initQueryBackend() (services.Service, error) {
 		return nil, err
 	}
 	logger := log.With(f.logger, "component", "query-backend")
+	blockReader := querybackend.NewBlockReader(f.logger, f.storageBucket, f.reg)
+	blockReader.Overrides = f.Overrides
```
aleks-p: nit, for consistency we should pass f.Overrides via the constructor

pkg/validation/limits.go
```diff
@@ -39,6 +39,8 @@ type Limits struct {
 	IngestionLimit           *ingestlimits.Config `yaml:"ingestion_limit" json:"ingestion_limit" category:"advanced" doc:"hidden"`
 	IngestionBodyLimitMB     float64              `yaml:"ingestion_body_limit_mb" json:"ingestion_body_limit_mb" category:"advanced" doc:"hidden"`
 	DistributorSampling      *sampling.Config     `yaml:"distributor_sampling" json:"distributor_sampling" category:"advanced" doc:"hidden"`
+	KeepStrippedProfiles     bool                 `yaml:"keep_stripped_profiles" json:"keep_stripped_profiles"`
+	IncludeStrippedProfiles  bool                 `yaml:"include_stripped_profiles" json:"include_stripped_profiles"`
```
aleks-p: Nit, perhaps move this down to other querier flags - for added clarity. We could also add prefixes to make them more clear, but they might end up too long (current names are also fine).

aleks-p (dismissed):
I added a couple of nits but otherwise it looks good to me. Thanks for addressing my comments!

In a follow up we can add an integration test to secure the whole flow (distributor -> seg writer -> compaction -> read path) - making sure no component breaks for stripped profile.

aleks-p (dismissed):

aleks-p (approved):

PR #5360, opened 2026-07-16
fix(metastore): retain shards with recent data

simonswine:
## What this does

- interpret shard index timestamps as Unix milliseconds during retention cleanup
- retain old partitions when their shards contain recent data
- update retention test fixtures to use production timestamp units and add regression coverage

## Verification

- `go test -count=1 ./pkg/metastore/index/cleaner/retention ./pkg/metastore/index/store`


aleks-p (approved):
LGTM, thanks for fixing this!

PR #5363, opened 2026-07-16
fix(examples): repair Django rideshare startup

korniltsev-grafanista:
## What this fixes

- upgrade Gunicorn to a version compatible with the example’s Python 3.12 and setuptools versions
- run Django migrations once in a dedicated Compose service
- wait for migrations to complete before starting all three regional application services

Previously, Gunicorn 20.1 failed while importing the removed `pkg_resources` module. On a fresh Compose startup, the three regional services also ran migrations concurrently against the same database and could fail with duplicate-key errors.

## Validation

- `docker compose config --quiet`
- `sh -n examples/language-sdk-instrumentation/python/rideshare/django/app/entrypoint.sh`
- fresh `docker compose up --build --detach`
- verified the migration service exited successfully and all three regional services remained running
- verified `/bike`, `/scooter`, and `/car` returned HTTP 200 across the three regions
- verified `gunicorn --version` starts and reports 23.0.0

### DIFF
--- a/examples/examples_test.go
+++ b/examples/examples_test.go
@@ -85,8 +85,9 @@ func TestMain(m *testing.M) {
 }
 
 type env struct {
-	dir  string // project dir of docker-compose, relative to the test working directory
-	path string // path to docker-compose file
+	dir                   string // project dir of docker-compose, relative to the test working directory
+	path                  string // path to docker-compose file
+	completedSuccessfully map[string]struct{}
 }
 
 func (e *env) repoDir() string {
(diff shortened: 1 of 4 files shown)


aleks-p (approved):
LGTM. I think the CI run in this PR [only ran this example](https://github.com/grafana/pyroscope/actions/runs/29506651897/job/87649033156?pr=5363#step:3:24), not sure if you ran all of them locally. 

Outside of this PR we should make sure to run all examples when we change the test harness itself.

korniltsev-grafanista:
Follow-up opened to run all examples when the shared test harness changes: https://github.com/grafana/pyroscope/pull/5365

PR #5370, opened 2026-07-20
docs(segment-writer): document object storage cost tuning

simonswine:
## Description

Documents how to reduce object storage PUT/GET costs after upgrading to Pyroscope v2, based on a recurring support question.

After upgrading v1→v2, users see a significant increase in object storage costs, and naively raising the segment flush interval causes ingestion errors (`context deadline exceeded`, `context canceled`, and `422` responses on `/ingest`) because the segment duration exceeds the client's synchronous ingestion timeout.

### Changes

- **`segment-writer.md`**: Complete the previously stubbed `## Configuration` section with two subsections:
  - **Reduce object storage costs** — tune `-segment-writer.segment-duration` (default `500ms`) and `-segment-writer.num-tokens` (default `4`, down to `1`), with a warning about the client-timeout caveat that causes the rejection errors above; notes that replica count scales requests linearly and that compaction reduces object count over time.
  - **Trade consistency for headroom** — document `-async-ingest=true` and its read-after-write consistency trade-off.
- **`deployment-modes/index.md`**: Cross-reference the new tuning section from the Object storage resource-planning list.
- **`write_path.go`**: Un-hide the `-async-ingest` flag (remove `doc:"hidden"`) so it appears in the generated configuration reference, and improve its help text.
- **`reference-configuration-parameters/index.md`**: Regenerated via `make generate` to include `-async-ingest`.

### Verification

- `go build ./pkg/distributor/writepath/` passes
- `make generate` produces only the intended config-reference change
- Cross-reference anchor matches the heading slug

### DIFF
--- a/cmd/pyroscope/help-all.txt.tmpl
+++ b/cmd/pyroscope/help-all.txt.tmpl
@@ -42,7 +42,7 @@ Usage of ./pyroscope:
   -architecture.storage value
     	Storage architecture layer. Use 'v1' to use legacy ingester-based storage, 'v2' for new storage, and 'v1-v2-dual' to use new storage while being able to query old data. (default "v1-v2-dual")
   -async-ingest
-    	If true, the write path will not wait for the segment-writer to finish processing the request. Writes to ingester always synchronous.
+    	If true, the write path does not wait for the segment-writer to durably store and index the profile before responding. This reduces ingestion latency and allows a larger -segment-writer.segment-duration, but removes the read-after-write consistency and synchronous durability guarantee. Writes to the ingester are always synchronous.
   -auth.multitenancy-enabled
     	When set to true, incoming HTTP requests must specify tenant ID in HTTP X-Scope-OrgId header. When set to false, tenant ID anonymous is used instead.
   -blocks-storage.bucket-store.ignore-blocks-within duration
--- a/cmd/pyroscope/help-all.txt.tmpl
+++ b/cmd/pyroscope/help-all.txt.tmpl
@@ -42,7 +42,7 @@ Usage of ./pyroscope:
   -architecture.storage value
     	Storage architecture layer. Use 'v1' to use legacy ingester-based storage, 'v2' for new storage, and 'v1-v2-dual' to use new storage while being able to query old data. (default "v1-v2-dual")
   -async-ingest
-    	If true, the write path does not wait for the segment-writer to durably store and index the profile before responding. This reduces ingestion latency and allows a larger -segment-writer.segment-duration, but removes the read-after-write consistency and synchronous durability guarantee. Writes to the ingester are always synchronous.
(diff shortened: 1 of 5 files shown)


docs/sources/reference-pyroscope-v2-architecture/components/segment-writer.md
```diff
@@ -86,3 +86,25 @@ The segment-writer flush behavior can be configured to balance between:
 - **Latency**: How quickly data becomes queryable
 - **Cost**: Number of write operations to object storage
 - **Memory usage**: Amount of data held in memory before flush
+
+### Reduce object storage costs
+
+Because each flush results in write operations to object storage (and each subsequent query results in read operations), the flush frequency and the number of objects produced per flush directly affect your object storage bill. Use the following levers to reduce `PUT` and `GET` costs:
+
+- **Increase the segment duration**: `-segment-writer.segment-duration` (default `500ms`) controls how long profiles accumulate in memory before a segment is flushed. Increasing it produces fewer, larger segments and therefore fewer write operations.
```
jake-kramer: nit: Add a small comment about the costs, like increased memory usage

jake-kramer (dismissed):
LGTM. This begs the question though: should we change our defaults to accommodate less block storage operations by default?

cmd/pyroscope/help-all.txt.tmpl
```diff
@@ -42,7 +42,7 @@ Usage of ./pyroscope:
   -architecture.storage value
     	Storage architecture layer. Use 'v1' to use legacy ingester-based storage, 'v2' for new storage, and 'v1-v2-dual' to use new storage while being able to query old data. (default "v1-v2-dual")
   -async-ingest
-    	If true, the write path will not wait for the segment-writer to finish processing the request. Writes to ingester always synchronous.
+    	If true, the write path does not wait for the segment-writer to durably store and index the profile before responding. This reduces ingestion latency and allows a larger -segment-writer.segment-duration, but removes the read-after-write consistency and synchronous durability guarantee. Writes to the ingester are always synchronous.
```
knylander-grafana: We use contractions in our style. 
```suggestion
    	If true, the write path doesn't wait for the segment-writer to durably store and index the profile before responding. This reduces ingestion latency and allows a larger -segment-writer.segment-duration, but removes the read-after-write consistency and synchronous durability guarantee. Writes to the ingester are always synchronous.
```

docs/sources/configure-server/reference-configuration-parameters/index.md
```diff
@@ -3379,6 +3379,14 @@ distributor_usage_groups:
 # CLI flag: -validation.reject-newer-than
 [reject_newer_than: <duration> | default = 10m]
 
+# (advanced) If true, the write path does not wait for the segment-writer to
```
knylander-grafana:
```suggestion
# (advanced) If true, the write path doesn't wait for the segment-writer to
```

knylander-grafana (dismissed):
Thank you for updating the docs. Two very minor suggestions for contractions. Your choice as to whether use them.

jake-kramer (dismissed):

aleks-p (approved):

PR #5371, opened 2026-07-20
fix(metastore): don't let a single bad DLQ object block recovery

simonswine:
## Problem

DLQ recovery never completes (observed on v2.0.3 and v2.1.1), with a continuous stream of:

```
caller=recovery.go:141 level=warn component=metastore msg="failed to read block metadata; to be retried" err=EOF path=dlq/1/
```

A stray zero-byte key under `dlq/` (e.g. an S3 "folder placeholder" like `dlq/1/`) returns `io.EOF` on read. That error was returned from the `Iter` callback, aborting the whole sweep — so every valid entry after it was never recovered, and the stray key was never removed.

## Fix

- `recover()` only aborts the sweep for genuine stop conditions (context cancellation, raft leadership change); all per-object failures are logged/metered and iteration continues.
- Structure-aware guard `isDLQMetadataPath` only accepts `dlq/<shard>/<tenant>/<block-ulid>/meta.pb`; stray keys are skipped.
- Empty `meta.pb` objects are deleted (permanently invalid); transient errors retain the object for retry.
- New metric status values: `stray`, `empty`.
- Tests for stray placeholders, empty objects, transient errors, and path validation.

### DIFF
--- a/pkg/metastore/index/dlq/recovery.go
+++ b/pkg/metastore/index/dlq/recovery.go
@@ -5,11 +5,13 @@ import (
 	"errors"
 	"flag"
 	"io"
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
+	"github.com/oklog/ulid/v2"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/thanos-io/objstore"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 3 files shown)


pkg/metastore/raftnode/errors.go
```diff
@@ -14,7 +14,13 @@ func IsRaftLeadershipError(err error) bool {
 	return errors.Is(err, raft.ErrLeadershipLost) ||
 		errors.Is(err, raft.ErrNotLeader) ||
 		errors.Is(err, raft.ErrLeadershipTransferInProgress) ||
-		errors.Is(err, raft.ErrRaftShutdown)
+		errors.Is(err, raft.ErrRaftShutdown) ||
+		isRaftLeaderStatusError(err)
```
aleks-p, line 18: Just flagging for transparency, this has impact beyond DLQ recovery, but I can only imagine impacting things in a positive way.

pkg/metastore/index/dlq/recovery.go
```diff
@@ -133,39 +149,78 @@ func (r *Recovery) recover(ctx context.Context, path string) (err error) {
 		r.metrics.recoveryAttempts.WithLabelValues("not_found").Inc()
 		level.Warn(r.logger).Log("msg", "block metadata not found; skipping", "path", path)
 		return nil
+	case errors.Is(err, io.EOF):
```
aleks-p, line 152: Could a object storage provider return a transient EOF? Would it make sense to double check that the object is empty (e.g., via `bucket.Attributes`) before deleting it?

aleks-p (approved):

pkg/metastore/index/dlq/recovery.go
simonswine: I have sent my agent on an investigation, I think we should be good. 

> I traced the pinned providers. Go reserves plain io.EOF for graceful completion; truncated reads should be `io.ErrUnexpectedEOF`. MinIO uses the object’s reported size to enforce that distinction, and Thanos has a test showing a short S3 response returns `io.ErrUnexpectedEOF`. GCS and Azure similarly expose non-EOF read failures, with Azure explicitly retrying unexpected EOFs. So I couldn’t find a supported provider where transient failure becomes plain EOF

PR #5374, opened 2026-07-21
feat(profilecli): add replay dump/push commands

simonswine:
## Summary

Adds a two-command workflow for capturing profile data from a source cell and replaying it into a destination cell:

- `profilecli replay dump` queries the source cell's metastore for blocks matching a query/time range, reads the matching blocks directly from object storage, reconstructs individual pprof profiles (preserving original series labels and timestamps), and writes them to a standalone framed dump file.
- `profilecli replay push` reads the dump file and continuously pushes profiles to a destination cell via the pushv1 Push API, rescheduling timestamps so the recorded window repeats seamlessly in a loop (with optional speed scaling and single-pass mode).

Includes a roundtrip test for the dump file format.

## Test plan

- [x] `replay_format_test.go` roundtrip test
- [x] `replay_push_test.go` covers push scheduling/looping behavior

### DIFF
--- a/cmd/profilecli/main.go
+++ b/cmd/profilecli/main.go
@@ -145,6 +145,12 @@ func main() {
 	recordingRulesDeleteId := recordingRulesDeleteCmd.Arg("rule_id", "Recording rule Id to delete").Required().String()
 	recordingRulesParams := addRecordingRulesListParams(recordingRulesCmd)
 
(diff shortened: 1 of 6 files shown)


aleks-p (approved):
Couldn't find anything wrong. 

We might be pushing duplicate profiles at the cycle boundaries in loop mode, but this is probably fine and is likely deduped by the v2 read path.

PR #5375, opened 2026-07-21
perf(model): optimize tree construction with hybrid search and interned handles

### LINKED ISSUE #4827: Tree implementation is very slow for large trees
We have very deep and very wide flamegraphs for some services. Think 200 frames deep and 800k unique stacks with very long frame names. It can easily take 2.3s just to build a `model.Tree` instance.

Image

Our stacks already come pre-optimized with interned strings:

```rust
type Stack struct {
	Frames []unique.Handle[string]
	Value  uint64
}
```

Playing around with different implementations, the following seems to be the winner:

```go
type node struct {
	self  int64
	total int64
	edges []edge
}

type edge struct {
	name  unique.Handle[string]
	child *node
}

type Tree struct {
	root *node
}

func NewTree() Tree {
	return Tree{root: &node{}}
}

func (t *Tree) InsertStack(v int64, stack ...unique.Handle[string]) {
	if v <= 0 {
		return
	}

	n := t.root
	for _, s := range stack {
		n.total += v

		var child *node
		for _, edge := range n.edges {
			if edge.name == s {
				child = edge.child
				break
			}
		}

		if child == nil {
			child = &node{}
			n.edges = append(n.edges, edge{
				name:  s,
				child: child,
			})
		}

		n = child
	}

	n.total += v
	n.self += v
}
```

On my M3 Pro laptop:

* Baseline: 430ms
* Using `map[unique.Handle[string]]`: 340ms
* Optimized code above: 160ms

For the top 2.3s number we see in production we see a reduction to 0.5s. Here's a flamegraph to compare:

Image

Currently we carry this implementation as a full replacement, but it would be nice if it was fast upstream instead. It's very possible that there's an even faster implementation, but it probably starts with `unique.Handle[string]` for frames too.

servusdei2018:
Fixes #4827.

This PR optimizes `model.Tree` construction for services with deep and wide flamegraphs with long frame names. Roughly 2x speedup on my machine for tree insertion latency, and 50% less heap allocations.

### DIFF
--- a/pkg/model/tree.go
+++ b/pkg/model/tree.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"reflect"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


pkg/model/tree.go
```diff
@@ -123,6 +125,7 @@ type node[N NodeName] struct {
 	children    []*node[N]
 	self, total int64
 	name        N
+	handle      unique.Handle[string]
```
simonswine, line 128: Adds 8 bytes to each `Tree`, no matter if handles are used for inserting

pkg/model/tree.go
```diff
@@ -165,27 +172,192 @@ func (t *Tree[N, I]) InsertStack(v int64, stack ...N) {
...
+		return
+	}
+
+	// Validate handles upfront and pre-convert to node names to prevent path splicing.
+	var stackBuf [32]N
+	var names []N
+	if len(stack) <= len(stackBuf) {
+		names = stackBuf[:len(stack)]
+	} else {
+		names = make([]N, len(stack))
```
simonswine, line 285: Deep handle stacks will be very costly for every insertion once the depth exceeds 32. The motivating workload has roughly 200-frame stacks, so this works against the intended allocation reduction.

simonswine (commented):
@servusdei2018 Thanks for your PR.

My main concern is that the new handle-based path would only deliver end-to-end memory savings if handles were used throughout the tree lifecycle. Even then, if Pyroscope performs the interning on a hot path, I would want to measure the potential contention and overhead from `unique.Make`. I do not currently have production evidence that this would be a problem, though.

In Pyroscope, the main tree-construction path starts with a `StacktraceTree` containing integer references and constructs tree nodes directly through `TreeFromStacktraceTree`. Function names are resolved from a shared symbol string table at that point, so their underlying string data is already shared.

This PR nevertheless adds a handle to every node, increasing per-node memory usage regardless of whether the handle-based insertion path is used.

I also think the submitted benchmarks compare different workloads: the function-name case uses 50k stacks at depth 150, while the handle case uses depth 50. With matching workloads, pre-interned handles were not cheaper on my machine.

The hybrid-search idea is separate and may be worth evaluating independently in `node.insert`, using a benchmark that exercises the production `TreeFromStacktraceTree` path.

PR #5398, opened 2026-07-22
perf(compaction): Avoid deep label clones when building block indexes

jake-kramer:
Label pairs produced by the TSDB index reader are immutable; only the slice backing array is reused as the row iterator advances. A shallow copy is therefore sufficient for retention, halving resident label memory during compaction. Label cloning the largest in-use heap consumer (~28%) in a recent compaction-worker OOM.

BenchmarkDatasetIndexWriter_WriteTo: -60% allocs/op, -46% B/op.

### DIFF
--- a/pkg/block/section_profiles.go
+++ b/pkg/block/section_profiles.go
@@ -254,8 +254,12 @@ type ProfileEntry struct {
 
 	Timestamp   int64
 	Fingerprint model.Fingerprint
-	Labels      phlaremodel.Labels
-	Row         schemav1.ProfileRow
+	// Labels' backing array is reused by the row iterator on advance, but
+	// the label pairs themselves are immutable: consumers that retain
+	// labels beyond the current row must copy the slice (shallow copy is
+	// sufficient), and must never modify the pairs.
+	Labels phlaremodel.Labels
+	Row    schemav1.ProfileRow
 }
 
 func NewMergeRowProfileIterator(src []*Dataset) (iter.Iterator[ProfileEntry], error) {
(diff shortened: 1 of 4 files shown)


pkg/block/section_profiles.go
```diff
@@ -254,8 +254,12 @@ type ProfileEntry struct {
 
 	Timestamp   int64
 	Fingerprint model.Fingerprint
-	Labels      phlaremodel.Labels
-	Row         schemav1.ProfileRow
+	// Labels' backing array is reused by the row iterator on advance, but
+	// the label pairs themselves are immutable: consumers that retain
+	// labels beyond the current row must copy the slice (shallow copy is
+	// sufficient), and must never modify the pairs.
+	Labels phlaremodel.Labels
```
bryanhuhta, line 261: I'm not sure how to implement this, but I worry a comment like this is insufficient at catching potential block corruption now that we've switched from deep copies to shallow copies of the labels slice.

It would be nice to see a test that enforces the immutability of `Labels` so we don't make another change that violates this invariant and corrupts blocks.

bryanhuhta (dismissed):
Overall looks good. Once concern about the implication of switching from deep copy to shallow copy, but I'll leave it up to you to decide whether to implement a test case.

pkg/block/section_profiles.go
jake-kramer: Yea, fair point. I'm looking in to a higher-level test that verifies labels look correct on the other end. I can't think of a valid use case for mutating labels in this code path which is why it feels safe/sane to me.

simonswine: I think we can enforce this by better API design (should not have any cost, also protects the preexisting risks of modification in `Evaluate`). Take a look at #5405

simonswine (approved):

PR #5402, opened 2026-07-22
perf(metastore): skip decoding out-of-range block metadata in queries

jake-kramer:
Metadata queries decoded the full BlockMeta, including every dataset, for every block in every shard the iterator visited, and only then applied the block-level time filter. Because the shard iterator widens the query window by the 24h lookaround period on both sides, a narrow query over a large tenant materialized roughly two days of block metadata to answer a question about minutes, and pinned all of it in the block read cache.

Read min_time/max_time through a trimmed view message that decodes the same wire format as BlockMeta and skips unknown fields, including datasets, without retaining them, and drop blocks outside the query range before unmarshaling. The view is verified against UnmarshalVT with randomized metas, and a descriptor test enforces that its fields match BlockMeta in name, number, and type.


simonswine (commented):
The idea is great, I think duplicating the proto spec is unnecessary, we can do even slightly better using the underlying protowire:  #5407 

Note there is also #5362 which would build an interval index of blocks, which is related and paired with this idea can improve this part quite significantly.

jake-kramer:
@simonswine Ha my first Claude attempt had that exact same protowire implementation and I prompted it toward this solution because I thought it was more elegant/convention for how to handle this sort of issue with protobufs 😂 I'm fine with https://github.com/grafana/pyroscope/pull/5407, thanks for looking into it

PR #5405, opened 2026-07-23
refactor(compaction): isolate retained labels

simonswine:
Stacked on #5398

This isolates the shared state, so it can't be accidentally alterted by consumers.

## Benchmark
Compared with `main` using ten 1-second samples of `BenchmarkDatasetIndexWriter_WriteTo`:
- time/op: 340.5us -> 312.1us (-8.33%, p=0.000)
- bytes/op: 205.3 KiB -> 110.9 KiB (-45.99%, p=0.000)
- allocs/op: 1991 -> 790 (-60.32%, p=0.000)


jake-kramer (dismissed):

alsoba13 (approved):
lgtm

PR #5407, opened 2026-07-23
perf(metastore): scan block bounds before decoding

simonswine:
This builds on the excellent idea of #5402 but solves it more elegantly: use protowire instead of an error-prone duplicate proto specification.

## Benchmark

Cold-cache query over 1,000 large out-of-range blocks plus one matching block. Measured with 12 runs each on Apple M2 Max.

| Benchmark | `main` | PR #5402 (duplicate spec) | #5407 (protowire) | Difference vs `main` | p-value |
|---|---:|---:|---:|---:|---:|
| Metadata query | 4.20 ms | 255.15 us | 76.19 us | -98.19% | <0.0005 |
| Labels query | 4.27 ms | 252.97 us | 72.95 us | -98.29% | <0.0005 |
| Metadata B/op | 6.00 MiB | 18.22 KiB | 18.50 KiB | -99.70% | <0.0005 |
| Labels B/op | 6.00 MiB | 14.15 KiB | 14.41 KiB | -99.77% | <0.0005 |

## Test

- `go test ./pkg/metastore/index/... -count=1`
- `golangci-lint run ./pkg/metastore/index/...`

### DIFF
--- a/pkg/metastore/index/query.go
+++ b/pkg/metastore/index/query.go
@@ -140,7 +140,11 @@ func (q *blockMetadataQuerier) collectBlockMetadata(s *store.Shard) error {
 		return nil
 	}
 	for blocks.Next() {
-		md := q.shards.index.blocks.getOrCreate(s, blocks.At())
+		kv := blocks.At()
+		if minTime, maxTime, err := blockMetaTimeBounds(kv.Value); err == nil && !q.query.overlapsUnixMilli(minTime, maxTime) {
+			continue
+		}
+		md := q.shards.index.blocks.getOrCreate(s, kv)
 		if m := q.collectMatched(s.StringTable, matcher, md); m != nil {
 			q.metas = append(q.metas, m)
 		}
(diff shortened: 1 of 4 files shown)


pkg/metastore/index/query.go
```diff
@@ -256,15 +260,20 @@ func (q *metadataLabelQuerier) collectLabels(s *store.Shard) error {
...
-				}
-				if q.query.overlapsUnixMilli(ds.MinTime, ds.MaxTime) {
-					matcher.Matches(ds.Labels)
-				}
+		kv := blocks.At()
+		if minTime, maxTime, err := blockMetaTimeBounds(kv.Value); err == nil && !q.query.overlapsUnixMilli(minTime, maxTime) {
+			continue
+		}
+		md := q.shards.index.blocks.getOrCreate(s, kv)
+		if !q.query.overlapsUnixMilli(md.MinTime, md.MaxTime) {
```
aleks-p: This is not really expensive, but technically we did the check above already. Is this here as a failsafe or can it be removed?

aleks-p (dismissed):

pkg/metastore/index/query.go
jake-kramer: +1

jake-kramer (dismissed):

pkg/metastore/index/query.go
simonswine: Really failsafe, but if `err != nil` in L264, the cache very likely also can't decode it.

jake-kramer (approved):

PR #5410, opened 2026-07-23
feat(metastore): bound compaction job size by bytes, not just block count

jake-kramer:
## What

Compaction job planning previously used a fixed block-count limit per level as the sole completion criterion. This PR adds a second, independent completion criterion: accumulated input size in bytes. A job now completes once either the block count limit **or** the new byte limit is reached, whichever happens first.

- New flag `-metastore.compaction-max-job-bytes` (config key `compaction_max_job_bytes`), default 2GiB. Applies to levels 1 and above only; level 0 is exempt (level-0 blocks are segment-writer flush outputs, already small and tightly bounded by the block-count limit alone). A value of 0 disables the size-based limit entirely.
- Block size is not re-estimated at planning time — it uses the exact on-disk size recorded in `BlockMeta.Size` at block creation (segment flush or a prior compaction).
- A single block that is already larger than the configured byte limit still forms a valid, single-block job; oversized blocks are never excluded and the planner never gets stuck.
- Sizes are persisted in a new, parallel bbolt bucket (`compaction_block_queue_size`), keyed identically to the existing block-queue bucket. The existing bucket's entry format is completely unchanged.

## Why

Compaction workers process a job's input blocks in memory, and worker memory usage scales with total job input size, not block count. A recent production incident saw a 10-block, ~2.8GiB level-2 compaction job cause a compaction worker to require more than 32GiB of memory — nearly 12x the sum of its input blocks' on-disk sizes. As per-block sizes grow (e.g., high-cardinality tenants, larger segment flush intervals), a fixed block-count limit no longer bounds worker memory, and workers OOM.

Bounding job size directly by accumulated input bytes closes this gap while leaving the block-count limit in place as a floor (so small numbers of oversized blocks are still forced into a job promptly, and level 0 keeps its existing behavior unchanged).

### Upgrade / downgrade compatibility

Keeping size in a separate bucket, rather than extending the existing entry's encoded value, means the change is safe in both directions:

- **Upgrade** (pre-existing queue data, new binary): the main bucket decodes exactly as before. Entries have no corresponding record in the new size bucket, so they fall back to `Size == 0` and behave under the pre-existing count-only planning logic until they are compacted or replaced.
- **Downgrade** (queue data written by the new binary, old binary running): an old binary addresses bbolt buckets purely by name and never enumerates or rejects unknown buckets (true both at normal startup and at Raft snapshot restore, which is a byte-for-byte copy of the whole database file), so it simply never touches the size bucket. The main bucket it does read is in the unchanged pre-existing format, so it decodes correctly.

`StoreEntry`/`DeleteEntry` keep both buckets in sync during normal operation. The only caveat: if an old binary runs for a while after a downgrade (deleting entries without knowing about the size bucket) and is later upgraded again, the deleted entries' size records are left behind as harmless orphans (a stale record is indistinguishable from one never written, since Index/ID keys are never reused). This only happens across an actual downgrade window; the two buckets never diverge during normal operation.

## Testing

- Unit tests, race detector, lint, and the full test suite all pass.
- New test coverage added for:
  - Byte-cap completion scenarios, including a single block already larger than the configured limit forming its own job.
  - Level-0 exemption from the byte-based limit.
  - Mixed legacy (zero-size) and new (sized) entries within the same job/queue.
  - Store-level round-trip of the size bucket, decoding of legacy entries with no size record, and symmetric delete behavior between the main and size buckets.

## Follow-ups (deliberately deferred)

- Per-level byte-limit tuning (currently one global `MaxJobBytes` value applied uniformly across levels >= 1).
- Metrics exposing queue byte totals (currently only block counts are exported).

### DIFF
--- a/cmd/pyroscope/help-all.txt.tmpl
+++ b/cmd/pyroscope/help-all.txt.tmpl
@@ -533,6 +533,8 @@ Usage of ./pyroscope:
     	 (default 15s)
   -metastore.compaction-max-failures uint
     	 (default 3)
+  -metastore.compaction-max-job-bytes uint
+    	Maximum estimated total input size, in bytes, of a compaction job for levels above 0. A job completes once either the block count or this byte limit is reached, whichever happens first; a single block already larger than this limit still forms a valid, single-block job. Does not apply to level 0. 0 disables the size-based limit. (default 2147483648)
   -metastore.compaction-max-job-queue-size uint
     	 (default 10000)
   -metastore.data-dir string
--- a/docs/sources/configure-server/reference-configuration-parameters/index.md
+++ b/docs/sources/configure-server/reference-configuration-parameters/index.md
@@ -1334,6 +1334,14 @@ index:
 
 [levels: <list of LevelConfigs> | default = ]
 
+# (advanced) Maximum estimated total input size, in bytes, of a compaction job
+# for levels above 0. A job completes once either the block count or this byte
+# limit is reached, whichever happens first; a single block already larger than
+# this limit still forms a valid, single-block job. Does not apply to level 0. 0
+# disables the size-based limit.
+# CLI flag: -metastore.compaction-max-job-bytes
+[compaction_max_job_bytes: <int> | default = 2147483648]
+
 [cleanupbatchsize: <int> | default = ]
 
 [cleanupdelay: <duration> | default = ]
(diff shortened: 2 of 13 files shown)


jake-kramer:
Closing in favor of https://github.com/grafana/pyroscope-squad/issues/860

PR #5414, opened 2026-07-24
feat(metastore): online BoltDB compaction to reclaim disk space without restart (#5279)

### LINKED ISSUE #5279: Reclaim metastore BoltDB disk space without requiring restart
## Problem

Metastore index cleanup can delete old index buckets and add tombstones, but the underlying BoltDB file does not shrink in a long-running process. Disk space is only reclaimed when a snapshot is restored and `metastore.snapshot-compact-on-restore=true`.

In practice, this means long-running single-binary/self-hosted deployments can keep a large `metastore.boltdb` file even after retention cleanup has removed old data from the logical index.

This is especially problematic for local filesystem deployments with finite disk, where the process may run until disk exhaustion and crashloop before compaction ever happens.

## Code pointers

- `pkg/metastore/fsm/boltdb.go`
  - `bbolt.Compact` is only called from the snapshot restore path.
  - compaction is gated by `SnapshotCompactOnRestore`.
- `pkg/metastore/fsm/fsm.go`
  - `metastore.snapshot-compact-on-restore` defaults to `false`.
  - snapshots are persisted from the running BoltDB state, but regular online compaction is not performed.
- `pkg/metastore/index_raft_handler.go`
  - `TruncateIndex` deletes logical index shards, but this does not shrink the BoltDB file.

## Proposal

Add a mechanism to reclaim metastore BoltDB disk space for long-running metastores without requiring a restart/snapshot restore cycle.

Possible approaches:

- Periodically compact the BoltDB database in the background when safe.
- Compact after successful snapshots outside of the restore path.
- Provide an explicit admin/API operation to trigger metastore DB compaction.
- Track BoltDB file size vs logical/live size and compact only above a threshold.

## Acceptance criteria

- A long-running metastore can reclaim BoltDB disk space after retention cleanup without process restart.
- The mechanism does not block raft apply/read paths for excessive time.
- The mechanism is safe for single-node and multi-node raft deployments.
- Metrics/logging expose when compaction runs, how long it takes, and compaction ratio.
- Add tests for compaction behavior or at least the triggering/gating logic.

AdarshJ173:
## Problem

Long-running self-hosted / single-binary Pyroscope deployments accumulate free pages in the metastore BoltDB file after retention cleanup. The physical file size never shrinks during process lifetime — disk space is only reclaimed today if a snapshot is restored with `metastore.snapshot-compact-on-restore=true`. On finite-disk systems this leads to disk exhaustion and crashloops before compaction ever happens.

Fixes #5279.

---

## Solution: Background Online Compaction

A new `BackgroundCompactor` runs as a lightweight goroutine alongside the FSM. It periodically inspects the BoltDB free-page ratio and, when it exceeds a configurable threshold, runs `bbolt.Compact()` and performs an atomic hot-swap of the database file — **without requiring a process restart or a Raft snapshot cycle**.

### Design highlights

| Property | Detail |
|---|---|
| **Locking** | Acquires `fsm.mu.Lock()` + `fsm.txns.Wait()` during the rename-only hot-swap phase — identical to `FSM.Restore` — safe for single-node and multi-node Raft. |
| **Compaction path** | Reuses the existing `boltdb.compact()` + `boltdb.openPath()` helpers already proven by snapshot-restore. No new BoltDB code. |
| **Threshold gating** | Compaction fires only when `FreePageN + PendingPageN / TotalPages ≥ BoltDBCompactMinFreeRatio` (default `0.30`), so a dense database is never needlessly rewritten. |
| **Storm prevention** | Minimum interval between runs (default `4h`) prevents compaction storms on deployments that continuously delete data. |
| **Opt-out** | Fully disabled with `--metastore.boltdb-compact-enabled=false`. |
| **Zero data loss** | `compact()` writes to a temp file; `openPath()` does `os.Rename` (atomic on POSIX) then reopens. The live file is replaced only after the compacted copy is fully written and fsync'd. |

---

### `pkg/metastore/fsm/compaction.go` _(new)_
`BackgroundCompactor` with `Start()` / `Stop()` lifecycle, ticker-based loop, `freePageRatio()` sampler, and `maybeCompact()` which does threshold check → compact → lock → hot-swap → unlock.

### `pkg/metastore/fsm/metrics.go` _(extended)_
Four new Prometheus metrics:
```
pyroscope_boltdb_online_compaction_runs_total{result="success"|"error"}
pyroscope_boltdb_online_compaction_duration_seconds
pyroscope_boltdb_online_compaction_size_before_bytes
pyroscope_boltdb_online_compaction_ratio
```

### `pkg/metastore/fsm/fsm.go` _(extended)_
Three new `Config` fields + CLI flags:
```
--metastore.boltdb-compact-enabled     (default: true)
--metastore.boltdb-compact-interval    (default: 4h)
--metastore.boltdb-compact-min-ratio   (default: 0.30)
```
`FSM.New()` starts the compactor when enabled; `FSM.Shutdown()` stops it gracefully before closing BoltDB.

### `pkg/metastore/fsm/compaction_test.go` _(new)_
Unit tests:
- `TestBackgroundCompactor_SkipsWhenBelowThreshold` — file size unchanged when ratio < threshold
- `TestBackgroundCompactor_CompactsWhenAboveThreshold` — file shrinks when ratio ≥ threshold
- `TestBackgroundCompactor_StopIsIdempotent` — no deadlock or panic
- `TestBackgroundCompactor_MetricsEmittedOnSuccess` — counters and ratio gauge are set

---

## Acceptance criteria (from #5279)

- [x] Long-running metastore can reclaim BoltDB disk space after retention cleanup without process restart.
- [x] The mechanism does not block Raft apply/read paths for excessive time (hot-swap is O(ms): just fsync + rename, not the compaction I/O itself).
- [x] Safe for single-node and multi-node Raft deployments (uses the same `mu.Lock()` + `txns.Wait()` strategy as `FSM.Restore`).
- [x] Metrics/logging expose when compaction runs, how long it takes, and compaction ratio.
- [x] Tests for compaction behaviour and triggering/gating logic.

---

> Contribution developed with AI assistance (Perplexity). Happy to revise based on maintainer feedback — particularly around the `freePageRatio()` heuristic or the default threshold values.

### DIFF
--- a/pkg/metastore/fsm/compaction.go
+++ b/pkg/metastore/fsm/compaction.go
@@ -1,0 +1,162 @@
+package fsm
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+)
+
+// BackgroundCompactor periodically compacts the BoltDB database in-place
+// to reclaim disk space accumulated by free pages after retention cleanup.
+//
+// Design constraints:
+//   - Uses the existing bbolt.Compact path (boltdb.compact + boltdb.openPath)
+//     that is already proven by the snapshot-restore flow.
+//   - Hot-swap is gated behind fsm.mu.Lock() + fsm.txns.Wait(), which is
+//     identical to the FSM.Restore locking strategy and therefore safe for
+//     both single-node and multi-node Raft deployments.
+//   - Compaction is skipped when the free-page ratio is below the configured
+//     threshold so that already-dense databases are never needlessly rewritten.
+type BackgroundCompactor struct {
+	fsm    *FSM
+	logger log.Logger
+
+	interval        time.Duration
+	minFreeRatio    float64
+
+	ctx    context.Context
+	cancel context.CancelFunc
+	done   chan struct{}
+}
+
+func newBackgroundCompactor(fsm *FSM) *BackgroundCompactor {
+	ctx, cancel := context.WithCancel(context.Background())
+	return &BackgroundCompactor{
+		fsm:          fsm,
+		logger:       fsm.logger,
+		interval:     fsm.config.BoltDBCompactInterval,
+		minFreeRatio: fsm.config.BoltDBCompactMinFreeRatio,
+		ctx:          ctx,
+		cancel:       cancel,
+		done:         make(chan struct{}),
+	}
+}
+
+// Start launches the background compaction loop. It is safe to call Start
+// multiple times; subsequent calls are no-ops.
+func (c *BackgroundCompactor) Start() {
+	go c.loop()
+}
+
+// Stop signals the compaction loop to stop and waits for it to exit.
+func (c *BackgroundCompactor) Stop() {
+	c.cancel()
+	<-c.done
+}
+
+func (c *BackgroundCompactor) loop() {
+	defer close(c.done)
+	ticker := time.NewTicker(c.interval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-c.ctx.Done():
+			return
+		case <-ticker.C:
+			if err := c.maybeCompact(); err != nil {
+				level.Error(c.logger).Log("msg", "online boltdb compaction failed", "err", err)
+			}
+		}
+	}
+}
+
+// freePageRatio returns the fraction of the database's allocated pages that
+// are currently unused (free). A value of 0.30 means 30% of allocated pages
+// are free and could be reclaimed by compaction.
+//
+// We derive page count from bbolt.Stats() which is always up-to-date without
+// requiring an additional read transaction.
+func (c *BackgroundCompactor) freePageRatio() float64 {
+	db := c.fsm.db.boltdb
+	if db == nil {
+		return 0
+	}
+	stats := db.Stats()
+	totalPages := stats.TxStats.PageCount + int64(stats.FreePageN) + int64(stats.PendingPageN)
+	if totalPages <= 0 {
+		return 0
+	}
+	freePages := int64(stats.FreePageN) + int64(stats.PendingPageN)
+	return float64(freePages) / float64(totalPages)
+}
+
+// maybeCompact checks the free-page ratio and, if it exceeds the configured
+// threshold, performs an online compaction and hot-swaps the database file.
+func (c *BackgroundCompactor) maybeCompact() error {
+	ratio := c.freePageRatio()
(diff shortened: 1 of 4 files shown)


bryanhuhta:
I filed #5574 about the metastore index's in-memory caches not being invalidated when `FSM.Restore` swaps the bbolt database. While checking whether this PR interacts with that, I ran into a few things that seem worth raising here. Apologies for arriving with a pile of concerns on an otherwise self-contained change.

**`fsm.mu` and `fsm.txns` guard reads, not writes.** The hot-swap comment says it uses the same locking strategy as `FSM.Restore`, but the two aren't equivalent in the way that matters. `FSM.Read` takes `fsm.mu.RLock()` and registers in `fsm.txns` (`fsm.go:309-324`); `applyCommand` does neither and calls `fsm.db.boltdb.Begin(true)` directly (`fsm.go:279`). `FSM.Restore` is safe against a concurrent write not because of those primitives but because hashicorp/raft invokes `Restore` and `Apply` on the same goroutine. `BackgroundCompactor.loop()` is a separate goroutine, so `fsm.mu.Lock()` + `fsm.txns.Wait()` will drain in-flight reads but will not exclude an in-flight raft apply.

**Writes committed during compaction look like they would be discarded.** `bbolt.Compact` copies from a point-in-time read view of the source, so the compacted file reflects the database as of the start of `compact()`. Anything applied between that point and `openPath()` lands in the live file only, and the swap replaces the live file with the older copy. The existing `compact()` caller runs inside `FSM.Restore`, where the database is quiescent and this can't arise; on a live database it seems like it could. The persisted applied index would move backwards with it, which may or may not be recoverable depending on log retention.

**The swap runs no restorers and invalidates no caches.** `FSM.Restore` follows `db.restore` with `fsm.init()` and `fsm.restore()`; this path calls neither. The metastore index keeps `Index.shards` and `Index.blocks` as process-lifetime caches that sit outside bbolt's transaction isolation, so after the swap they still describe the pre-swap database. That is the same structural gap as #5574, though in the opposite direction: there the caches end up behind the database, here they would end up ahead of it, which is why it wouldn't produce that issue's panic. Either way, a cache-invalidation hook is likely a prerequisite for this feature, and #5574 will probably need to add one that this can reuse.

The first two look addressable by sequencing the swap through the raft apply goroutine, or by adding explicit write exclusion, rather than by anything structural in the compaction itself. Happy to be told I've misread the locking here.

PR #5427, opened 2026-07-27
feat(compaction): optionally read block metadata from object storage

aleks-p:
The compaction worker fetches full source block metadata from the metastore (`GetBlockMetadata`), even though the same metadata is stored in each block's footer. This keeps the worker coupled to the metastore on a path where the blocks are downloaded from object storage right after anyway - and it stands in the way of shrinking the metastore index.

This PR adds `-compaction-worker.metadata-source` (default `metastore`, opt-in `object-storage`) to read the metadata from the block object footers instead. Object paths are derived from the job fields (tenant, shard, level), which are guaranteed to match the source blocks since the compaction queue is keyed by them. Reads run in parallel with bounded concurrency.

Notes:
- A missing object drops the block from the job, mirroring the metastore lookup filtering; all objects missing lands on the existing `metadata_not_found` path.
- The `pyroscope_instance` label of exported metrics hashes the `CreatedBy` string table index, which happens to be identical in the footer and metastore representations of the metadata - a test pins this so series identity can't silently change with the metadata source.

### DIFF
--- a/pkg/compactionworker/worker.go
+++ b/pkg/compactionworker/worker.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
@@ -25,6 +26,7 @@ import (
 	"github.com/prometheus/prometheus/model/labels"
 	thanosstore "github.com/thanos-io/objstore"
 	_ "go.uber.org/automaxprocs"
+	"golang.org/x/sync/errgroup"
 
 	metastorev1 "github.com/grafana/pyroscope/api/gen/proto/go/metastore/v1"
 	"github.com/grafana/pyroscope/v2/pkg/block"
@@ -34,6 +36,16 @@ import (
 	"github.com/grafana/pyroscope/v2/pkg/util"
 )
 
+// Sources of the compaction job source block metadata.
+const (
+	MetadataSourceMetastore     = "metastore"
+	MetadataSourceObjectStorage = "object-storage"
+)
+
+// metadataReadConcurrency bounds the parallel block metadata reads
+// from object storage within a single compaction job.
+const metadataReadConcurrency = 8
+
 type Config struct {
 	JobConcurrency     int            `yaml:"job_capacity" category:"advanced"`
 	JobPollInterval    time.Duration  `yaml:"job_poll_interval" category:"advanced"`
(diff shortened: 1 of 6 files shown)


pkg/compactionworker/worker.go
```diff
@@ -533,17 +570,44 @@ func (w *Worker) getBlockMetadata(logger log.Logger, job *compactionJob) error {
...
+	for i, b := range job.SourceBlocks {
+		g.Go(func() error {
+			ctx, cancel := context.WithTimeout(ctx, w.config.RequestTimeout)
+			defer cancel()
+			path := block.BuildObjectPath(job.Tenant, job.Shard, job.CompactionLevel, b)
+			obj, err := block.NewObjectFromPath(ctx, w.storage, path)
+			switch {
+			case err == nil:
+				blocks[i] = obj.Metadata()
+			case objstore.IsNotExist(w.storage, err):
```
marcsanmi, line 605: The index is deleted synchronously at retention, but the objects are deleted asynchronously so checking object existence instead of the index means a job could still find blocks the metastore already dropped. Could we be at risk of resurrecting data past retention via `ReplaceBlocks`?

pkg/compactionworker/worker.go
```diff
@@ -533,17 +570,44 @@ func (w *Worker) getBlockMetadata(logger log.Logger, job *compactionJob) error {
...
+// block object footers, bypassing the metastore. The object paths are
+// derived from the job fields: the compaction queue is keyed by tenant,
+// shard, and level, so they are guaranteed to match the source blocks.
+func (w *Worker) readBlockMetadataFromStorage(logger log.Logger, job *compactionJob) ([]*metastorev1.BlockMeta, error) {
+	blocks := make([]*metastorev1.BlockMeta, len(job.SourceBlocks))
+	g, ctx := errgroup.WithContext(job.ctx)
+	g.SetLimit(metadataReadConcurrency)
+	for i, b := range job.SourceBlocks {
+		g.Go(func() error {
+			ctx, cancel := context.WithTimeout(ctx, w.config.RequestTimeout)
```
marcsanmi: Do you think it would make sense to have a separate timeout here? Reusing RequestTimeout (5s default, if I'm not wrong) might be tight (?); it was sized for a single metastore RPC, but NewObjectFromPath can do up to 3 sequential storage round-trips.

pkg/compactionworker/worker.go
aleks-p: I think you are referring to a case where a compaction is planned for a block that is about to be removed due to the retention policy. This would normally indicate a misconfigured retention policy or poorly sized compaction workloads such that the job queue gets longer than the retention period.

This is something that can already happen on main - we check the blocks at the start of the job but they can be removed during the job. This PR widens that gap but the fix would be the same, the metastore would need to reject job outputs for deleted blocks. I've created an issue to track that - https://github.com/grafana/pyroscope/issues/5436 - I'd consider it a low priority one.

pkg/compactionworker/worker.go
aleks-p: Good point, added in ae5496b78

marcsanmi (approved):
LGTM

PR #5434, opened 2026-07-28
fix: clean up stale ring members on shutdown and via auto-forget

aleks-p:
When an instance shuts down cleanly, both the overrides-exporter and the segment-writer keep their ring entries, leaving unhealthy members behind:

- The overrides-exporter hardcodes `KeepInstanceInTheRingOnShutdown: true`. Query-frontends typically run as Deployments, so a departed pod never comes back to reclaim its entry; it lingers as `Unhealthy` until auto-forget removes it 4 heartbeat timeouts later. Every rollout, node drain or eviction produces a few minutes of unhealthy ring members (and alert noise) with no benefit in return.
- The segment-writer disables `unregister-on-shutdown` on purpose (a restarted pod reclaims its entry, keeping shard placement undisturbed for restarts shorter than the heartbeat timeout), but the classic lifecycler has no auto-forget: entries left behind by a scale-down stay in the ring indefinitely and must be forgotten manually on the ring page. They also block readiness of any segment-writer that starts afterwards, because `readiness-check-ring-health` (enabled by default) requires every ring member to be healthy.

Changes:

- overrides-exporter: new `-overrides-exporter.ring.unregister-on-shutdown` option, `true` by default. Instances now remove themselves from the ring on clean shutdown and hand leadership over immediately. Set it to `false` to restore the previous behavior.
- segment-writer: new opt-in auto-forget janitor (`-segment-writer.auto-forget-unhealthy-periods`, disabled by default) that removes ring members whose heartbeat has been stale for the given number of heartbeat timeouts. This is the classic-lifecycler equivalent of dskit's `AutoForgetDelegate`, modeled on Loki's `-ingester.autoforget-unhealthy`, including the guard that skips a round when the instance's own heartbeat looks stale (its view of the ring is then likely the problem).

Notes:

- Unlike Loki, a round is not skipped when all other members look unhealthy: forgetting a live segment-writer is self-healing (it re-registers on its next heartbeat, and shard placement depends on instance IDs and token counts rather than token values), whereas skipping would leave a scale-down from N to 1 never cleaned up.
- Auto-forget ships disabled to keep upgrades behavior-neutral. We plan to enable it with 4 periods (forget after ~4m at the default 1m heartbeat timeout) in our own deployments, matching what the overrides-exporter ring already does.

### DIFF
--- a/pkg/segmentwriter/autoforget.go
+++ b/pkg/segmentwriter/autoforget.go
@@ -61,6 +61,7 @@ func (f *autoForget) iteration(ctx context.Context) error {
 				// the problem is more likely on our side (e.g. a network
 				// partition), so we must not judge the other members.
 				level.Warn(f.logger).Log("msg", "auto-forget sees its own instance as unhealthy, the network may be partitioned, skipping this round", "instance", id)
+				forgotten = forgotten[:0]
 				return nil, false, nil
 			}
 			forgotten = append(forgotten, id)
--- a/pkg/segmentwriter/autoforget.go
+++ b/pkg/segmentwriter/autoforget.go
@@ -1,0 +1,84 @@
+package segmentwriter
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/grafana/dskit/kv"
+	"github.com/grafana/dskit/ring"
+	"github.com/grafana/dskit/services"
+)
+
+// autoForget periodically removes ring members whose heartbeat has been stale
+// for longer than the forget period. The segment-writer lifecycler keeps its
+// ring entry on shutdown (-segment-writer.unregister-on-shutdown defaults to
+// false) so that restarts do not disturb shard placement, but instances that
+// never come back, e.g. after a scale-down, would otherwise stay in the ring
+// until forgotten manually. This is the classic-lifecycler equivalent of
+// dskit's AutoForgetDelegate.
+type autoForget struct {
+	services.Service
+
+	kv           kv.Client
+	instanceID   string
+	forgetPeriod time.Duration
+	logger       log.Logger
+}
+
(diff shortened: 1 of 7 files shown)


pkg/segmentwriter/autoforget.go
```diff
@@ -0,0 +1,84 @@
...
+		for id, instance := range desc.Ingesters {
+			if now.Sub(time.Unix(instance.GetTimestamp(), 0)) <= f.forgetPeriod {
+				continue
+			}
+			if id == f.instanceID {
+				// If our own heartbeat looks stale in our view of the ring,
+				// the problem is more likely on our side (e.g. a network
+				// partition), so we must not judge the other members.
+				level.Warn(f.logger).Log("msg", "auto-forget sees its own instance as unhealthy, the network may be partitioned, skipping this round", "instance", id)
+				return nil, false, nil
```
jake-kramer, line 65: This is required to get a correct log message below when this case is hit
```suggestion
                forgotten = forgotten[:0]
                return nil, false, nil
```

jake-kramer (dismissed):

pkg/segmentwriter/autoforget.go
aleks-p: Done in 4bd2fc4cb, thanks!

jake-kramer (approved):

Sparsh752:
Hi @jake-kramer @aleks-p @marcsanmi, we have been seeing this same problem with Pyroscope V1 as well, where the ingester (and for other components as well) hash rings don't get updated when a pyroscope pod goes down with ungraceful exit. Is there any plan to fix that as well, and what is the ETA for that?

jake-kramer:
@Sparsh752 We don't have plans to fix this for v1

Sparsh752:
@jake-kramer Could you take this up and fix it for V1 as well, if it is a small change?

PR #5442, opened 2026-07-29
feat(experimental): make async queries resumable

jake-kramer:
# What

Async query execution currently lives entirely on the query frontend that accepted the request: the request spec exists only in that process's memory, and progress is tracked via an in-memory heartbeat. If that frontend pod dies
mid-query (OOM, rollout, node eviction), the query becomes permanently unrecoverable — the persisted metadata record stays `IN_PROGRESS` forever, and any client polling for the result does so indefinitely with no way to detect or recover from the loss.

This change makes async queries resumable across frontend restarts:

- **Spec persistence at submit time.** The full request spec is written to
  object storage (`spec.pb`) as part of `Submit`, before dispatch, so any
  frontend — not just the one that accepted the request — has everything
  needed to re-run the query later.
- **Heartbeat as an ownership lease.** The existing heartbeat mechanism is
  reinterpreted as a lease: as long as heartbeats keep landing within
  `leaseTimeout`, the current owner is presumed alive and the record is left
  alone.
- **Adoption scan.** The async store's background loop gained a periodic scan
  that looks for in-progress records whose lease has expired. Any frontend
  that observes one claims ownership and redispatches the query from the
  persisted `spec.pb`, transparently to the polling client. Note that the adoption mechanism triggers only on silent owner death (stale lease + in progress query), never on query failure.

## Why `result.pb` is the source of truth for success

`metadata.json` is a mutable record that several frontends can rewrite concurrently once adoption exists — a slow-but-alive owner and a new adopter, for example — and object storage gives writers no way to notice that someone else wrote first: the last write simply wins. A stale write can therefore overwrite a terminal status, which makes the mutable record unreliable as the sole source of truth.

`result.pb` is different: only successful executions ever write it, so its existence proves success no matter what the metadata says. Racing duplicate executions may replace it with another valid result: existence, not content, is the invariant. Readers check for it first and repair the metadata when the two disagree, and the adoption scan skips any record that already has a result.

## Deliberate behavior change

**A real success now overwrites a previously persisted stale failure status.**

Before this change, a terminal status was never revisited. With adoption, it's possible for a stale, incorrect `FAILURE` to have been written (e.g. by a duplicate/racing execution that failed) before a genuinely successful result lands. Since `result.pb`'s existence is authoritative, a confirmed success is now allowed to correct a previously recorded failure. 

Closes https://github.com/grafana/pyroscope-squad/issues/838

### DIFF
--- a/pkg/frontend/async/coordinator.go
+++ b/pkg/frontend/async/coordinator.go
@@ -6,13 +6,16 @@ import (
 	"sync"
 	"time"
 
+	"connectrpc.com/connect"
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/google/uuid"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	"google.golang.org/protobuf/proto"
 
 	querierv1 "github.com/grafana/pyroscope/api/gen/proto/go/querier/v1"
+	"github.com/grafana/pyroscope/api/gen/proto/go/querier/v1/querierv1connect"
 	"github.com/grafana/pyroscope/v2/pkg/tenant"
 )
 
@@ -20,11 +23,11 @@ type Limits interface {
 	MaxAsyncQueryConcurrency(tenantID string) int
 }
 
-// QueryResult is the result of a query execution sent over a channel.
+// queryResult is the result of a query execution sent over a channel.
 // On success, Response carries the raw SelectMergeStacktracesResponse
 // from the wrapped handler (its Async field is unset); the
 // coordinator/store own request_id and status.
-type QueryResult struct {
+type queryResult struct {
 	Response *querierv1.SelectMergeStacktracesResponse
 	Err      error
 }
@@ -33,6 +36,7 @@ type Coordinator struct {
 	logger log.Logger
 	store  *Store
 	limits Limits
+	next   querierv1connect.QuerierServiceHandler
 
 	mu       sync.Mutex
 	inFlight map[string]int // tenantID -> count
@@ -41,11 +45,12 @@ type Coordinator struct {
 	asyncQueriesMax     *prometheus.GaugeVec
 }
 
-func NewCoordinator(logger log.Logger, store *Store, limits Limits, reg prometheus.Registerer) *Coordinator {
+func NewCoordinator(logger log.Logger, store *Store, limits Limits, next querierv1connect.QuerierServiceHandler, reg prometheus.Registerer) *Coordinator {
 	return &Coordinator{
 		logger:   logger,
 		store:    store,
 		limits:   limits,
+		next:     next,
 		inFlight: make(map[string]int),
 		asyncQueriesCurrent: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
 			Name: "pyroscope_async_queries_in_progress",
@@ -77,30 +82,66 @@ func (c *Coordinator) tryAcquire(tenantID string) error {
 	return nil
 }
 
-// Register reserves the per-tenant concurrency slot, persists the
-// in-progress metadata, and starts watching resultCh. The caller is
-// responsible for dispatching the query and sending its outcome on
-// resultCh; the coordinator owns reads from the channel, store writes,
-// and the heartbeat loop. Returns the assigned request ID, or an error
-// if the slot could not be acquired.
-func (c *Coordinator) Register(ctx context.Context, tenantID string, resultCh <-chan QueryResult) (string, error) {
+// Submit reserves the tenant's concurrency slot, strips the Async marker
+// from req, persists the resulting spec as a new in-progress query, and
+// dispatches it in the background. Returns the assigned request ID.
+func (c *Coordinator) Submit(ctx context.Context, tenantID string, req *querierv1.SelectMergeStacktracesRequest) (string, error) {
 	if err := c.tryAcquire(tenantID); err != nil {
 		return "", err
 	}
 
 	requestID := uuid.New().String()
+	spec := proto.Clone(req).(*querierv1.SelectMergeStacktracesRequest)
+	spec.Async = nil
 
-	if err := c.store.create(ctx, tenantID, requestID); err != nil {
+	if err := c.store.create(ctx, tenantID, requestID, spec); err != nil {
 		c.decrement(tenantID)
 		return "", fmt.Errorf("failed to create async query: %w", err)
 	}
 
-	go c.awaitResult(tenantID, requestID, resultCh)
-
+	c.dispatch(tenantID, requestID, spec)
+	level.Info(c.logger).Log("msg", "async query submitted", "tenant", tenantID, "request_id", requestID)
 	return requestID, nil
 }
 
-func (c *Coordinator) awaitResult(tenantID, requestID string, resultCh <-chan QueryResult) {
+// Dispatch implements Store's Dispatcher interface: it (re-)runs a query
+// whose record and spec are already persisted. A declined dispatch (tenant
+// at its concurrency limit) never rolls back the store's claim on the
+// record; it is simply retried by a later adoption scan once the lease
+// expires again. The context is intentionally unused: like Submit, a
+// dispatched query always runs detached from its caller so it survives
+// independently of whatever triggered it.
+func (c *Coordinator) Dispatch(_ context.Context, tenantID, requestID string, spec *querierv1.SelectMergeStacktracesRequest) {
+	if err := c.tryAcquire(tenantID); err != nil {
+		level.Warn(c.logger).Log("msg", "skipping async query adoption: concurrency limit reached", "tenant", tenantID, "request_id", requestID, "err", err)
+		return
+	}
+	c.dispatch(tenantID, requestID, spec)
+}
+
+// dispatch runs spec against next on a background context, detached from the
+// caller's context, so the query survives client disconnects and outlives
+// the scan tick that triggered an adoption. Submit and Dispatch both share
+// this path once their concurrency slot is reserved.
+func (c *Coordinator) dispatch(tenantID, requestID string, spec *querierv1.SelectMergeStacktracesRequest) {
+	queryCtx := tenant.InjectTenantID(context.Background(), tenantID)
+	resultCh := make(chan queryResult, 1)
+
+	// query goroutine: runs the spec end-to-end.
+	go func() {
+		resp, err := c.next.SelectMergeStacktraces(queryCtx, connect.NewRequest(spec))
+		if err != nil {
+			resultCh <- queryResult{Err: err}
+			return
+		}
+		resultCh <- queryResult{Response: resp.Msg}
+	}()
+
+	// supervisor goroutine: renews the lease and records the outcome.
+	go c.awaitResult(tenantID, requestID, resultCh)
+}
+
+func (c *Coordinator) awaitResult(tenantID, requestID string, resultCh <-chan queryResult) {
 	defer c.decrement(tenantID)
 
 	ctx := tenant.InjectTenantID(context.Background(), tenantID)
@@ -120,6 +161,8 @@ func (c *Coordinator) awaitResult(tenantID, requestID string, resultCh <-chan Qu
 			}
 			if err := c.store.complete(ctx, tenantID, requestID, res.Response); err != nil {
 				level.Error(c.logger).Log("msg", "failed to store async query result", "tenant", tenantID, "request_id", requestID, "err", err)
+			} else {
+				level.Info(c.logger).Log("msg", "async query completed", "tenant", tenantID, "request_id", requestID)
 			}
 			return
 		case <-ticker.C:
--- a/pkg/frontend/async/coordinator.go
+++ b/pkg/frontend/async/coordinator.go
@@ -104,9 +104,9 @@ func (c *Coordinator) Submit(ctx context.Context, tenantID string, req *querierv
 	return requestID, nil
 }
 
-// CanDispatch implements Store's Dispatcher interface: a read-only peek at
+// HasCapacity implements Store's Dispatcher interface: a read-only peek at
 // whether tenantID currently has spare capacity.
-func (c *Coordinator) CanDispatch(tenantID string) bool {
+func (c *Coordinator) HasCapacity(tenantID string) bool {
 	maxConcurrent := c.limits.MaxAsyncQueryConcurrency(tenantID)
 	c.mu.Lock()
 	defer c.mu.Unlock()
--- a/pkg/frontend/async/coordinator.go
+++ b/pkg/frontend/async/coordinator.go
@@ -108,10 +108,8 @@ func (c *Coordinator) Submit(ctx context.Context, tenantID string, req *querierv
 // whose record and spec are already persisted. A declined dispatch (tenant
 // at its concurrency limit) never rolls back the store's claim on the
 // record; it is simply retried by a later adoption scan once the lease
-// expires again. The context is intentionally unused: like Submit, a
-// dispatched query always runs detached from its caller so it survives
-// independently of whatever triggered it.
-func (c *Coordinator) Dispatch(_ context.Context, tenantID, requestID string, spec *querierv1.SelectMergeStacktracesRequest) {
+// expires again.
+func (c *Coordinator) Dispatch(tenantID, requestID string, spec *querierv1.SelectMergeStacktracesRequest) {
 	if err := c.tryAcquire(tenantID); err != nil {
 		level.Warn(c.logger).Log("msg", "skipping async query adoption: concurrency limit reached", "tenant", tenantID, "request_id", requestID, "err", err)
 		return
--- a/pkg/frontend/async/coordinator.go
+++ b/pkg/frontend/async/coordinator.go
@@ -104,6 +104,15 @@ func (c *Coordinator) Submit(ctx context.Context, tenantID string, req *querierv
 	return requestID, nil
 }
 
+// CanDispatch implements Store's Dispatcher interface: a read-only peek at
+// whether tenantID currently has spare capacity.
+func (c *Coordinator) CanDispatch(tenantID string) bool {
+	maxConcurrent := c.limits.MaxAsyncQueryConcurrency(tenantID)
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return maxConcurrent > 0 && c.inFlight[tenantID] < maxConcurrent
+}
+
 // Dispatch implements Store's Dispatcher interface: it (re-)runs a query
 // whose record and spec are already persisted. A declined dispatch (tenant
 // at its concurrency limit) never rolls back the store's claim on the
--- a/pkg/frontend/async/store.go
+++ b/pkg/frontend/async/store.go
@@ -4,8 +4,10 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
+	"os"
 	"strings"
 	"time"
 
@@ -13,6 +15,8 @@ import (
 	"github.com/go-kit/log/level"
 	"github.com/google/uuid"
 	"github.com/grafana/dskit/services"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/thanos-io/objstore"
 	"google.golang.org/protobuf/proto"
 
@@ -20,11 +24,12 @@ import (
 )
 
 const (
-	storagePrefix            = "async-queries/"
-	defaultTTL               = 30 * time.Minute
-	defaultHeartbeatInterval = 15 * time.Second
-	defaultHeartbeatTimeout  = 45 * time.Second
-	cleanupInterval          = 5 * time.Minute
+	storagePrefix               = "async-queries/"
+	defaultTTL                  = 30 * time.Minute
+	defaultHeartbeatInterval    = 15 * time.Second
+	defaultLeaseTimeout         = 45 * time.Second
+	defaultAdoptionScanInterval = 30 * time.Second
+	cleanupInterval             = 5 * time.Minute
 )
 
 // Status represents the lifecycle state of an async query.
@@ -48,6 +53,7 @@ type Metadata struct {
 	CreatedAt     time.Time `json:"created_at"`
 	LastHeartbeat time.Time `json:"last_heartbeat,omitempty"`
 	ErrorMessage  string    `json:"error_message,omitempty"`
+	Owner         string    `json:"owner,omitempty"`
 }
 
 // Result bundles a query's Metadata with its decoded Response. Response is only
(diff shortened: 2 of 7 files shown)


pkg/frontend/async/store.go
```diff
@@ -125,8 +236,19 @@ func (s *Store) complete(ctx context.Context, tenantID, requestID string, resp *
 	if err := s.readJSON(ctx, base+"metadata.json", &meta); err != nil {
 		return fmt.Errorf("failed to read metadata: %w", err)
 	}
+	if meta.Status == StatusSuccess {
+		// Already recorded success (this execution's own write racing itself,
+		// or a duplicate's); nothing left to do.
+		return nil
```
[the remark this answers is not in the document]
jake-kramer: This isn't a high severity bug in the sense that writing a new `result.pb` is still "correct" given it's a valid result. Let me update the documentation/code comments to clarify we don't actually need "exactly once" semantics

jake-kramer:
### Async query lifecycle — state changes by scenario

Per-query storage: `async-queries/{tenant}/{request_id}/` — `metadata.json` (mutable, multi-writer), `spec.pb` (written once at submit), `result.pb` (written only by a successful execution; its **existence** is the success anchor). Timings: heartbeat 15s · lease 45s · scan 30s · TTL 30m. The *adoption scan* lane below is the async store's background loop — it runs inside every frontend process; there is no separate adoption deployable. Ownership of a stale lease is optimistic: the first claim to land resets the lease and other scanners back off, but two can race — the cost is a duplicate execution, never a wrong result.

#### Record state machine (all scenarios play out on this)

```mermaid
stateDiagram-v2
    [*] --> InProgress : submit -- spec.pb + metadata written, owner dispatches
    InProgress --> InProgress : heartbeat every 15s renews lease
    InProgress --> Success : complete() -- result.pb first, then metadata
    InProgress --> Failure : fail() -- query returned an error
    InProgress --> StaleLease : owner silent > 45s (died, or paused)
    StaleLease --> InProgress : adoption -- scan claims record,<br/>new owner re-runs spec from scratch
    StaleLease --> Failure : spec.pb missing or corrupt<br/>(unrecoverable, persisted)
    StaleLease --> Success : scan finds result.pb already exists<br/>(heal metadata, nothing re-run)
    Failure --> Success : late complete() -- a real success<br/>always wins over a stale failure
    Success --> [*] : TTL cleanup (~30m)
    Failure --> [*] : TTL cleanup (~30m)
    note right of Success
        result.pb existence is authoritative.
        Stale metadata writes can flip the JSON,
        but get() checks the anchor first and
        repairs the record, so a reported
        Success can never be lost.
    end note
```

<details>
<summary><b>Well-behaved client (happy path)</b></summary>

```mermaid
sequenceDiagram
    participant C as Client
    participant A as Frontend A (owner)
    participant B as Frontend B
    participant S as Object storage
    C->>A: SelectMergeStacktraces (Async: FORCE)
    A->>S: put spec.pb, then metadata{in_progress, owner: A}
    A-->>C: request_id (IN_PROGRESS)
    loop every 15s while running
        A->>S: heartbeat -- renew lease
    end
    C->>B: poll(request_id)
    B->>S: read metadata (lease fresh)
    B-->>C: IN_PROGRESS
    A->>S: put result.pb, then metadata{success}
    C->>B: poll(request_id)
    B->>S: metadata success -> read result.pb
    B-->>C: SUCCESS + payload
    Note over S: ~30m later the janitor deletes all three objects
```

</details>

<details>
<summary><b>Client dies (abandons the query)</b></summary>

```mermaid
sequenceDiagram
    participant C as Client
    participant A as Frontend A (owner)
    participant S as Object storage
    C->>A: submit (Async: FORCE)
    A->>S: spec.pb + metadata{in_progress}
    A-->>C: request_id
    Note over C: client crashes -- never polls again
    Note over A: execution continues on a detached context
    A->>S: heartbeats every 15s
    A->>S: result.pb + metadata{success}
    Note over S: nobody ever reads the result
    Note over S: janitor deletes everything at the 30m TTL
    Note over A,S: cost: one wasted execution + 30m of storage.<br/>Detecting abandonment and cancelling/pausing is issue #840.
```

</details>

<details>
<summary><b>Owner pod dies (the case this PR exists for)</b></summary>

```mermaid
sequenceDiagram
    participant C as Client
    participant A as Frontend A (owner)
    participant B as Frontend B (poll path)
    participant AS as Adoption scan (bg loop, every frontend)
    participant S as Object storage
    C->>A: submit
    A->>S: spec.pb + metadata{in_progress, owner: A}
    A-->>C: request_id
    Note over A: T+20s -- pod OOMs.<br/>Heartbeats stop, in-flight execution is lost.
    C->>B: poll
    B->>S: get(): in_progress, lease stale, spec present
    B-->>C: IN_PROGRESS (truthful -- rescue is coming)
    Note over AS: T+45..75s -- 30s ticker fires<br/>(this instance runs inside pod B)
    AS->>S: list async-queries/ -> lease stale -> candidate
    AS->>S: adopt(): result.pb? absent. spec.pb? ok.<br/>re-read: still stale.
    AS->>S: claim -- metadata{owner: B, lease: now}
    Note over AS: claim resets the 45s lease --<br/>every other frontend's scan now sees<br/>it fresh and backs off
    AS->>B: Dispatch(spec) -- coordinator on the same pod
    Note over B: re-runs the query from spec.pb, from scratch
    B->>S: result.pb + metadata{success}
    C->>B: poll
    B-->>C: SUCCESS + payload
```

</details>

<details>
<summary><b>Slow-but-alive owner (duplicate execution, the anchor at work)</b></summary>

```mermaid
sequenceDiagram
    participant C as Client
    participant A as Frontend A (paused, alive)
    participant B as Frontend B (executor)
    participant AS as Adoption scan (bg loop, on pod B)
    participant S as Object storage
    Note over A: GC pause > 45s -- lease expires<br/>but A's execution is still running
    AS->>S: scan: lease stale -> result.pb? absent.<br/>spec.pb? ok. re-read: still stale. -> claim
    AS->>B: Dispatch(spec) -- coordinator on the same pod
    B->>S: result.pb + metadata{success}
    A->>S: A's stale heartbeat write lands late --<br/>metadata flips back to in_progress (!)
    A->>S: A's own execution fails -> metadata{failure}
    Note over S: metadata.json now lies: FAILURE.<br/>But result.pb exists.
    C->>B: poll
    B->>S: status != success -> check anchor: result.pb EXISTS
    B->>S: heal metadata -> success
    B-->>C: SUCCESS + payload (the computed answer wins)
    Note over A,B: cost of the race: one duplicate execution.<br/>correctness: unaffected.
```

</details>

<details>
<summary><b>Poison pill — a query that kills every pod it lands on</b></summary>

```mermaid
sequenceDiagram
    participant C as Client
    participant F as Frontends (rotating adopters)
    participant AS as Adoption scans (every frontend)
    participant S as Object storage
    C->>F: submit massive query
    F->>S: spec.pb + metadata{in_progress}
    Note over F: owner OOMs
    loop every ~45-75s, until T+30m
        AS->>S: scan: stale + spec present -> claim (lease reset)
        AS->>F: Dispatch(spec)
        Note over F: adopter re-runs the query... and OOMs too
    end
    Note over S: T+30m -- janitor deletes spec.pb:<br/>it was written once and never refreshed,<br/>while metadata.json stayed fresh (claims rewrite it)
    AS->>S: next scan: stale + spec ABSENT
    AS->>S: markUnadoptable -> metadata{failure} (persisted)
    C->>F: poll
    F-->>C: FAILURE ("query orphaned ... no request spec available")
    Note over F,S: bounded at ~TTL / (lease + scan) = ~two dozen runs,<br/>each holding one of the tenant's concurrency slots.<br/>Operational signature: adoption_attempts_total climbing<br/>for one tenant while successes track it.
```

</details>

pkg/frontend/async/store.go
```diff
@@ -57,47 +63,141 @@ type Result struct {
...
 	s.Service = services.NewBasicService(s.starting, s.running, s.stopping)
 	return s
 }
 
+// SetDispatcher wires the adoption scan to a Dispatcher. Must be called before
+// the Store's Service is started: s.dispatcher is a plain field, never
+// synchronized against running()'s reads of it, so calling this after the
+// Service has started is a genuine data race, not merely a logic error. nil is
+// safe (adoption is then a no-op), e.g. when Store runs standalone or in tests.
+func (s *Store) SetDispatcher(d Dispatcher) { s.dispatcher = d }
```
bryanhuhta: Is there a reason we need a setting to set a dispatcher? Can this be done instead through the constructor? If we need a setter, we should make sure we only set it when the service is in the `New` state:

```suggestion
func (s *Store) SetDispatcher(d Dispatcher) {
	if s.State() == services.New {
		s.dispatcher = d 
	}
}
```

pkg/frontend/async/store.go
```diff
@@ -57,47 +63,141 @@ type Result struct {
...
+
+	data, err := proto.Marshal(spec)
+	if err != nil {
+		return fmt.Errorf("failed to marshal spec: %w", err)
+	}
+	// spec.pb must land before metadata.json: metadata.json existing is what a
+	// poller/scanner treats as "this record is real", so writing spec first means
+	// a crash between the two uploads can never leave a spec-less in_progress
+	// record that this code path created.
+	if err := s.bucket.Upload(ctx, base+"spec.pb", bytes.NewReader(data)); err != nil {
```
bryanhuhta:
```suggestion
	if err := s.bucket.Upload(ctx, path.Join(base, "spec.pb"), bytes.NewReader(data)); err != nil {
```

There's likely a lot of places that use string concatenation instead of `path.Join` left over from the initial implementation, but it would be nice to at least begin switching over.

pkg/frontend/async/store.go
```diff
@@ -136,12 +259,60 @@ func (s *Store) fail(ctx context.Context, tenantID, requestID string, queryErr e
...
 	meta.ErrorMessage = queryErr.Error()
 	meta.LastHeartbeat = time.Now().UTC()
+	meta.Owner = s.ownerID
 	return s.saveJSON(ctx, base+"metadata.json", &meta)
 }
 
+// errOrphanedNoSpec is the single, shared definition of "unrecoverable" used by
+// both get() and the adoption scan: a stale in-progress record whose spec is
+// gone can never be resumed.
+var errOrphanedNoSpec = errors.New("query orphaned: owner heartbeat expired and no request spec is available for adoption")
```
bryanhuhta: nit: It's nice to have all the sentinel errors defined together at the top of the file so it's clear what kind errors we're dealing with.

pkg/frontend/async/store.go
```diff
@@ -162,26 +333,54 @@ func (s *Store) get(ctx context.Context, tenantID, requestID string) (*Result, e
 		return nil, nil
 	}
 
-	if meta.Status == StatusInProgress && !meta.LastHeartbeat.IsZero() && time.Since(meta.LastHeartbeat) > s.heartbeatTimeout {
-		meta.Status = StatusFailure
-		meta.ErrorMessage = "query appears to have been orphaned (heartbeat timed out)"
+	// Object storage has no compare-and-swap, so a stale writer can always
+	// overwrite metadata.json and make its Status lie. result.pb is different:
```
bryanhuhta, line 363: Not strictly true as you pointed out earlier: https://github.com/thanos-io/objstore/pull/178! That said, some implementations (e.g. Swift doesn't have CAS support).

pkg/frontend/async/store.go
```diff
@@ -162,26 +333,54 @@ func (s *Store) get(ctx context.Context, tenantID, requestID string) (*Result, e
...
-		if err := proto.Unmarshal(data, &resp); err != nil {
-			return nil, fmt.Errorf("failed to unmarshal result: %w", err)
+			// Inconclusive (e.g. a transient object-storage error): leave the
+			// record in progress rather than risk permanently failing a
+			// live, adoptable query. A later poll or the adoption scan will
+			// re-evaluate it.
+			level.Warn(s.logger).Log("msg", "failed to check spec existence; leaving query in progress", "request_id", requestID, "err", err)
+		} else if absent {
+			if err := s.markUnadoptable(ctx, tenantID, requestID); err != nil {
+				level.Warn(s.logger).Log("msg", "failed to persist orphaned query state", "request_id", requestID, "err", err)
```
bryanhuhta: Maybe log tenant here too? That could help speed up debugging if we get into this state.

pkg/frontend/async/store.go
```diff
@@ -277,3 +481,105 @@ func (s *Store) runCleanup(ctx context.Context) {
...
+// queries for many tenants. A claim whose dispatch is declined does no
+// further work.
+//
+// A query that repeatedly kills its owner (e.g. one that OOMs the frontend)
+// is re-adopted at most once per leaseTimeout+scanInterval, and cleanup()
+// bounds the loop: spec.pb is written once and never rewritten, so it ages
+// out after the TTL even while metadata.json stays fresh, at which point the
+// next scan finds no spec and persists a terminal failure. Worst case is
+// roughly ttl/(leaseTimeout+scanInterval) re-executions, each still subject
+// to the tenant's concurrency limit.
```
bryanhuhta: I don't think this a great mechanism at bounding how many times a query can swap owners. If we have a poison pill query that always OOMs the owning query frontend, we want a better limit one that's indirectly established via various timeout and knobs. IMO far better to have an explicit retry limit like compaction jobs have.

Using the defaults, if we run the math:

```
maxRetries = ttl / (leaseTimeout + scanInterval)
           = 30m / (45s + 30s)
           = 24
```

Even if we didn't implement an explicit limit on retries, the defaults established here means query frontends could OOM 24 times before settling. This seems like an unreasonably high number.

pkg/frontend/async/store.go
```diff
@@ -277,3 +481,105 @@ func (s *Store) runCleanup(ctx context.Context) {
...
+		if err := s.readJSON(ctx, name, &meta); err != nil {
+			level.Warn(s.logger).Log("msg", "failed to read metadata during adoption scan", "object", name, "err", err)
+			return nil
+		}
+		if !s.leaseExpired(&meta) {
+			return nil
+		}
+		s.adopt(ctx, meta)
+		return nil
+	}, objstore.WithRecursiveIter())
```
bryanhuhta, line 538: Is a recursive search the right approach here? We won't accidentally start scanning a bunch of files unrelated to finding `metadata.json` files?

pkg/frontend/async/store.go
```diff
@@ -277,3 +481,105 @@ func (s *Store) runCleanup(ctx context.Context) {
...
+// A query that repeatedly kills its owner (e.g. one that OOMs the frontend)
+// is re-adopted at most once per leaseTimeout+scanInterval, and cleanup()
+// bounds the loop: spec.pb is written once and never rewritten, so it ages
+// out after the TTL even while metadata.json stays fresh, at which point the
+// next scan finds no spec and persists a terminal failure. Worst case is
+// roughly ttl/(leaseTimeout+scanInterval) re-executions, each still subject
+// to the tenant's concurrency limit.
+func (s *Store) runAdoption(ctx context.Context) {
+	err := s.bucket.Iter(ctx, storagePrefix, func(name string) error {
+		if !strings.HasSuffix(name, "metadata.json") {
```
bryanhuhta: It hasn't been done yet, but it would be nice to see all these special file names be set as constants at the top of this file (metadata.json, spec.pb, etc). It would help in both understanding which file names are hardcoded and also make referencing them less error prone.

bryanhuhta:
```suggestion
		if path.Base(name) != "metadata.json" {
```

`path.Base` feels like a better API to use here.

pkg/frontend/async/store.go
```diff
@@ -277,3 +481,105 @@ func (s *Store) runCleanup(ctx context.Context) {
...
+	}
+}
+
+func (s *Store) adopt(ctx context.Context, meta Metadata) {
+	if exists, err := s.resultExists(ctx, meta.TenantID, meta.RequestID); err != nil {
+		level.Warn(s.logger).Log("msg", "failed to check result existence before adoption; proceeding", "request_id", meta.RequestID, "err", err)
+	} else if exists {
+		s.healToSuccess(ctx, meta.TenantID, meta.RequestID, meta)
+		return
+	}
```
bryanhuhta, line 554: If the result exists already but we fail to read it due to a transient object storage error, it could be possible to adopt this query and restart its execution. Should we just skip the adoption attempt if this call fails?

pkg/pyroscope/modules_experimental.go
```diff
@@ -169,8 +171,10 @@ func (f *Pyroscope) initQueryFrontendV12() (services.Service, error) {
 			log.With(f.logger, "component", "async-query-coordinator"),
 			f.asyncQueryStore,
 			f.Overrides,
+			querierHandler,
 			f.reg,
 		)
+		f.asyncQueryStore.SetDispatcher(coordinator)
```
bryanhuhta: I don't think we should register async query support for v12. I think it should be only v2. Its a mistake that we're already registering async query stuff here.

pkg/frontend/async/coordinator.go
```diff
@@ -77,30 +82,66 @@ func (c *Coordinator) tryAcquire(tenantID string) error {
...
 
-func (c *Coordinator) awaitResult(tenantID, requestID string, resultCh <-chan QueryResult) {
+// Dispatch implements Store's Dispatcher interface: it (re-)runs a query
+// whose record and spec are already persisted. A declined dispatch (tenant
+// at its concurrency limit) never rolls back the store's claim on the
+// record; it is simply retried by a later adoption scan once the lease
+// expires again. The context is intentionally unused: like Submit, a
+// dispatched query always runs detached from its caller so it survives
+// independently of whatever triggered it.
+func (c *Coordinator) Dispatch(_ context.Context, tenantID, requestID string, spec *querierv1.SelectMergeStacktracesRequest) {
```
bryanhuhta: If `dispatch()` always runs on a background context, we shouldn't need `context.Context` in the function signature of the `Dispatcher.Dispatch` interface.

bryanhuhta (commented):
Generally LGTM! I like the direction and I left a few remarks.

pkg/frontend/async/store.go
jake-kramer: Yea this is the way to search through object storage down multiple "directory levels". It starts with prefix `async-queries/` so will scan through only async query objects.

pkg/frontend/async/store.go
jake-kramer: I've settled on a follow up PR for the CAS support. It's a bit more complex and clutters this work imo [checks need to go in multiple places; handling unsupported object stores is a bit hairy].

pkg/frontend/async/store.go
jake-kramer: Can't use the constructor because this is all set up through dskit modules. Dependencies here are `Storage` > `AsyncQueryStore` > `QueryFrontend`. So since dispatcher is a query frontend component, we need to set it on the async query store after initialization.

I'd rather not silently skip not setting dispatcher if the store is not in the write state. I'll update to hard fail if state isn't new though.

jake-kramer:
@bryanhuhta Thanks for your comments -- addressed with some follow up commits.

pkg/frontend/async/store.go
```diff
@@ -277,3 +512,117 @@ func (s *Store) runCleanup(ctx context.Context) {
...
+	fresh.Owner = s.ownerID
+	fresh.LastHeartbeat = time.Now().UTC()
+	fresh.AdoptionCount++
+	if err := s.saveJSON(ctx, metaPath, &fresh); err != nil {
+		level.Warn(s.logger).Log("msg", "failed to claim orphaned query", "tenant", meta.TenantID, "request_id", meta.RequestID, "err", err)
+		return
+	}
+	s.adoptionSuccesses.WithLabelValues(meta.TenantID).Inc()
+	level.Info(s.logger).Log("msg", "adopted orphaned async query", "tenant", meta.TenantID, "request_id", meta.RequestID, "previous_owner", previousOwner, "lease_age", leaseAge)
+	s.dispatcher.Dispatch(meta.TenantID, meta.RequestID, spec)
```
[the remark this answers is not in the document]
jake-kramer: This is a good catch, will address

pkg/frontend/async/store.go
```diff
@@ -57,100 +86,264 @@ type Result struct {
...
 	}
+	if meta.Status != StatusInProgress {
+		return nil
+	}
 	meta.Status = StatusFailure
 	meta.ErrorMessage = queryErr.Error()
 	meta.LastHeartbeat = time.Now().UTC()
-	return s.saveJSON(ctx, base+"metadata.json", &meta)
+	meta.Owner = s.ownerID
+	return s.saveJSON(ctx, metaPath, &meta)
```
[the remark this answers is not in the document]
jake-kramer: Known consideration of current approach, will be resolved by CAS object store semantics

pkg/frontend/async/coordinator.go
```diff
@@ -77,30 +82,73 @@ func (c *Coordinator) tryAcquire(tenantID string) error {
...
 }
 
-func (c *Coordinator) awaitResult(tenantID, requestID string, resultCh <-chan QueryResult) {
+// CanDispatch implements Store's Dispatcher interface: a read-only peek at
+// whether tenantID currently has spare capacity.
+func (c *Coordinator) CanDispatch(tenantID string) bool {
+	maxConcurrent := c.limits.MaxAsyncQueryConcurrency(tenantID)
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return maxConcurrent > 0 && c.inFlight[tenantID] < maxConcurrent
```
bryanhuhta, line 113: I suppose this could be RWLock and we could do RLock/RUnlock here but I'm honestly not sure that really matter performance-wise in this case. Anyways, RWLocks aren't always a shoe-in for better throughput anyhow.

pkg/frontend/async/store.go
```diff
@@ -57,100 +86,268 @@ type Result struct {
 	Response *querierv1.SelectMergeStacktracesResponse
 }
 
+// Dispatcher re-runs a query that Store has adopted after its owner's lease
+// expired. Implemented by Coordinator; wired into Store via SetDispatcher.
+type Dispatcher interface {
+	// CanDispatch is a read-only capacity peek, not a reservation: it makes
+	// no guarantee that a following Dispatch call will actually run the
+	// query (Dispatch has its own gate and may still decline).
+	CanDispatch(tenantID string) bool
```
bryanhuhta:
```suggestion
	MaybeCanDispatch(tenantID string) bool
```

Perhaps slightly more self-documenting?

pkg/frontend/async/store.go
```diff
@@ -450,23 +451,32 @@ func (s *Store) readRaw(ctx context.Context, path string) ([]byte, error) {
...
 	for {
 		select {
 		case <-ctx.Done():
 			return nil
 		case <-cleanupTicker.C:
 			s.runCleanup(ctx)
-		case <-adoptionTicker.C:
+		case <-adoptionTimer.C:
 			s.runAdoption(ctx)
+			adoptionTimer.Reset(jitteredInterval(s.scanInterval))
```
bryanhuhta, line 504: Nice!

bryanhuhta (dismissed):
This LGTM! I'm happy where this is at.

pkg/frontend/async/coordinator.go
jake-kramer: Yea I think hit rate here is pretty low

pkg/frontend/async/store.go
jake-kramer: I like `HasCapacity` instead.. more explicit

jake-kramer: Addressed in https://github.com/grafana/pyroscope/pull/5442/commits/fc32770b7ae751c07a6c9d216bff7a160d701501

bryanhuhta (approved):

PR #5449, opened 2026-07-31
feat(symbolizer): improve debuginfod client resilience and observability

jake-kramer:
Several commits improving the symbolizer's debuginfod client

**1. `fix(symbolizer): retry connection-level debuginfod failures and count attempts`**
The failure message counts attempts actually made instead of the backoff wait counter.

**2. `fix(symbolizer): stop dropping not-found cache config in the client constructor`**
The constructor rebuilt its config field-by-field and silently dropped `NotFoundCacheTTL`/`NotFoundCacheMaxItems`, so cached 404s never expired instead of expiring after the configured seven days.

**3. `feat(symbolizer): circuit-break debuginfod after consecutive network failures`**
Five consecutive network-level failures open the breaker for a minute, then a single half-open probe decides. Trips are logged and counted (`pyroscope_symbolizer_debuginfod_circuit_breaker_trips_total`); open-state rejections surface with an `error:unavailable` metric status.

**4. `fix(symbolizer): trace debuginfod fetches, classify cache-probe misses`**
The fetch gets a `debuginfod_fetch` span via dskit tracing, carrying build ID and outcome (404s and breaker skips are expected outcomes, not span errors) (previously the dominant cost of a slow resolve appeared only as unexplained gaps between bucket operations). The bucket cache probes now classify a missing object as a distinct `errObjectNotFound` instead of treating every `Get` error as a miss.

**5. `refactor(symbolizer): budget upstream fetches separately from cache probes`**
`max-debuginfod-concurrency` throttled the entire per-binary resolution pipeline, coupling cheap object-store cache probes to the politeness cap for a third-party service. The cap now lives inside the client as a semaphore around actual upstream fetches (acquired inside the breaker, so an open breaker never consumes a slot), and the frontend pipeline fans out four resolves per fetch slot so cache lookups drain at bucket speed.

**6. `feat(symbolizer): send an identifying User-Agent to debuginfod servers`**
`pyroscope-symbolizer/<version> (github.com/grafana/pyroscope)` instead of Go's default `Go-http-client`, so upstream operators can attribute the traffic and contact us rather than throttle by IP. The reference libdebuginfod client identifies itself the same way.


pkg/symbolizer/symbolizer.go
```diff
@@ -472,6 +474,9 @@ func (s *Symbolizer) getLidiaBytes(ctx context.Context, buildID string) ([]byte,
 func (s *Symbolizer) fetchLidiaFromObjectStore(ctx context.Context, tenantID, buildID string) ([]byte, error) {
 	objstoreReader, err := s.bucket.Get(ctx, lidiaObjectPath(tenantID, buildID))
 	if err != nil {
+		if objstore.IsNotExist(s.bucket, err) {
+			return nil, errObjectNotFound
```
marcsanmi, line 506: I think this is not consumed anywhere; `getLidiaBytes` still treats every error as a miss

pkg/symbolizer/debuginfod_client.go
```diff
@@ -117,8 +166,17 @@ func NewDebuginfodClientWithConfig(logger log.Logger, cfg DebuginfodClientConfig
 func (c *DebuginfodHTTPClient) FetchDebuginfo(ctx context.Context, buildID string) (io.ReadCloser, error) {
 	start := time.Now()
 	status := statusSuccess
+	span, ctx := tracing.StartSpanFromContext(ctx, "debuginfod_fetch")
 	defer func() {
 		c.metrics.debuginfodRequestDuration.WithLabelValues(status).Observe(time.Since(start).Seconds())
+		span.SetTag("build_id", buildID)
```
marcsanmi: Should we sanitze the buildID before setting it as a span tag?

pkg/symbolizer/debuginfod_client_test.go
```diff
@@ -432,3 +439,226 @@ func TestDebuginfodClientNotFoundCache(t *testing.T) {
 
 	assert.Equal(t, 2, requestCount)
 }
+
```
marcsanmi, line 443: Do you think is worth adding a test asserting 404s don't count toward tripping?

pkg/frontend/readpath/queryfrontend/symbol_ref_resolve.go
```diff
@@ -107,7 +113,7 @@ func (q *QueryFrontend) resolveBinaries(ctx context.Context, tenantIDs []string,
 	if timeout <= 0 {
 		timeout = defaultResolveTimeout
 	}
-	concurrency := q.symbolizer.ResolveConcurrency()
+	concurrency := q.symbolizer.ResolveConcurrency() * resolvesPerFetchSlot
```
marcsanmi: With 40 concurrent resolves (4 per fetch slot) but only 10 upstream fetch slots, each binary's resolve-timeout clock (the `binCtx` below) now starts while it may still be queued on the client's fetch semaphore; before, the errgroup limit equaled the fetch cap, so a binary's timer only started once it could actually fetch.

Under a slow-but-alive upstream, binaries beyond the fetch cap can burn most of their 20s in the queue and degrade to fallbacks that previously would have succeeded. Probably an acceptable trade, but I wanted to know your thoughts on this.

jake-kramer: It's a good point and definitely worth an explicit call out in the documentation. I guess the question is whether or not `symbolizer.resolve-timeout` should represent a user-facing timeout. I would say it's "kind of" a user-visible latency in the sense that 20s here adds a real 20s to a user's request. 

Maybe what would be better is a single timeout around all of `resolveBinaries` (global, not per-tenant). It would be more representative of user-latency. Thoughts?

marcsanmi: Agree; a global budget is a better knob.

jake-kramer:
Some more changes:

1. Made resolve timeout in to global config as discussed (this removed per-tenant config without deprecation, but since this is gated behind experimental symbolization config anyways, I'm ok to outright remove)
2. Removed the per fetch slot concurrency
3. Updated semaphore acquisition to before circuit breaker (circuit breaker should only be exercised on goroutines "ready" to go)
4. Fall back to debuginfod when the bucket probe errors

pkg/symbolizer/symbolizer.go
```diff
@@ -451,7 +470,14 @@ func (s *Symbolizer) getLidiaBytes(ctx context.Context, buildID string) ([]byte,
...
 		return lidiaBytes, nil
 	}
-	s.metrics.cacheOperations.WithLabelValues("object_storage", "get", "miss").Inc()
+	if errors.Is(err, errObjectNotFound) {
+		s.metrics.cacheOperations.WithLabelValues("object_storage", "get", "miss").Inc()
+	} else {
+		// The cache probe is best-effort: debuginfod can still serve the
+		// build ID during a bucket outage.
+		s.metrics.cacheOperations.WithLabelValues("object_storage", "get", "error").Inc()
+		level.Warn(s.logger).Log("msg", "lidia cache probe failed, falling back to debuginfod", "buildID", buildID, "err", err)
```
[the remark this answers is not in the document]
marcsanmi: Worth Checking. In `FetchDebuginfo`, I'd refuse to spawn work for a dead caller

marcsanmi (commented):
LGTM, I'd just check the thing about orphans (spawning work for dead callers)

pkg/symbolizer/symbolizer.go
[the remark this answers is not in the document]
jake-kramer: Addressed by https://github.com/grafana/pyroscope/pull/5449/changes/7c8cfc501974755bbb24ad348394ac527c190a1f

marcsanmi (approved):
LGTM 👍

PR #5450, opened 2026-08-01
fix(adaptiveplacement): key distribution stats by tenant and dataset

aleks-p:
The stats builder in the adaptive placement manager keyed datasets by name only. When two tenants ingest a dataset with the same name, all of its usage is attributed to whichever tenant sorts first: the other tenant never appears in the stats, no placement rule is ever built for it, and its ingest stays pinned to the default single shard regardless of load, while the first tenant's shard limit is inflated by usage that isn't its own.

Key the stats by tenant+dataset instead.


simonswine (approved):
Thanks for spotting this. LGTM

PR #5462, opened 2026-08-04
feat(distributor): add max-inflight-bytes instance limit

### LINKED ISSUE #5437: Add an instance-level inflight bytes limit to the distributor
An application-level flag such as `-distributor.max-inflight-bytes` can protect distributors from being `OOMKilled` when segment writers experience slowdowns or outages.

Prior art:
- Mimir: [`-distributor.instance-limits.max-inflight-push-requests(-bytes)`](https://github.com/grafana/mimir/blob/885eae7283962bc820b5fdcbfae12c26accfeb6d/pkg/distributor/instance_limits.go#L27-L37), enforced at [distributor.go#L1443](https://github.com/grafana/mimir/blob/885eae7283962bc820b5fdcbfae12c26accfeb6d/pkg/distributor/distributor.go#L1443)
- Loki: [`-distributor.max-inflight-bytes`](https://github.com/grafana/loki/blob/8d2011726173097c479cd2d2a982013ac312f536/pkg/distributor/distributor.go#L133), enforced at [distributor.go#L661-L669](https://github.com/grafana/loki/blob/8d2011726173097c479cd2d2a982013ac312f536/pkg/distributor/distributor.go#L661-L669)

Should be opt-in and apply for both the regular (sync) and the async write path.

mmustafasenoglu:
## Summary

Add a per-distributor inflight bytes limit to protect against OOM kills when segment writers experience slowdowns or outages.

## Motivation

When segment writers slow down or become unavailable, distributors accumulate inflight push requests that can grow large enough to cause OOM kills. An application-level limit on total inflight bytes allows operators to protect distributor instances without relying on external safeguards.

Closes #5437

## Changes

- ****: 
  - Added `MaxInflightBytes` field to `Config` with `-distributor.max-inflight-bytes` flag (default 0 = unlimited)
  - Added `inflightBytes` atomic counter to `Distributor` struct
  - Added inflight bytes check at the beginning of `Push()`: rejects with `resource-exhausted` when limit is exceeded

- ****: Added `InflightBytesLimit` discard reason

## Testing

- The flag is opt-in (default 0 = unlimited), so existing behavior is unchanged
- When enabled, exceeding the limit returns a clear error message with human-readable byte sizes

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -83,6 +83,12 @@ type Config struct {
 	PushTimeout time.Duration
 	PoolConfig  clientpool.PoolConfig `yaml:"pool_config,omitempty"`
 
+	// MaxInflightBytes is the max total size in bytes of inflight push requests
+	// handled by this distributor instance. This limit is per-distributor, not
+	// per-tenant. When the limit is reached, new push requests are rejected with
+	// a resource-exhausted error. 0 means unlimited.
+	MaxInflightBytes int64 `yaml:"max_inflight_bytes" category:"advanced"`
+
 	// Distributors ring
 	DistributorRing util.CommonRingConfig `yaml:"ring"`
 }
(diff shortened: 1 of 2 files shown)


mmustafasenoglu:
Fixed the race condition flagged by Cursor Bugbot. Replaced the check-then-add pattern with a CAS loop that atomically checks and reserves the inflight bytes under concurrent Push calls.

PR #5463, opened 2026-08-04
feat(admin): Add compaction admin page

bryanhuhta:
This adds a compaction admin page that describes:

* What jobs are in-flight, unassigned, expired leases, and failed
* How many jobs per worker
* Compaction queue

https://github.com/user-attachments/assets/6f7ce56a-c985-4caf-93bd-85d02bac1817


pkg/metastore/compaction_service.go
```diff
@@ -90,7 +126,7 @@ func (svc *CompactionService) PollCompactionJobs(
 	}
 
 	cmd := fsm.RaftLogEntryType(raft_log.RaftCommand_RAFT_COMMAND_GET_COMPACTION_PLAN_UPDATE)
-	proposeResp, err := svc.raft.Propose(ctx, cmd, req)
+	proposeResp, err := svc.raft.Propose(ctx, cmd, request)
```
bryanhuhta: I think this was actually a bug. Previously, we built up `request` which trimmed down the payload to just status updates, but then we never actually used the request. We passed `req` instead which worked because it had the same wire format, but also contained more information that necessary.

Correct me if I'm wrong here though!

api/metastore/v1/raft_log/raft_log.proto
```diff
@@ -23,16 +23,23 @@ message AddBlockMetadataResponse {}
...
 }
 
 message CompactionJobStatusUpdate {
   string name = 1;
   uint64 token = 2;
   metastore.v1.CompactionJobStatus status = 3;
+  // Reserved field represents `compacted_blocks`. The message must remain
+  // wire-compatible with metastore.v1.CompactionJobStatusUpdate: raft log
+  // entries written by older versions may contain the raw worker request.
+  reserved 4;
```
bryanhuhta, line 43: We need to add `reserved` here as a result of the change in `pkg/metastore/compaction_service.go`. That function was previously submitting this record, which is wire-compatible with the "correct" message type. Since we're both changing the correct message type _and_ actually using the correct message type, I think it's prudent to keep the two structures wire compatible (for now).

pkg/metastore/compaction_service.go
simonswine: Can we split this out in a separate PR?

simonswine (commented):
I wonder how the endpoint and UI is handling tenant heavy cells, but otherwise looks very useful for debugging.

aleks-p:
I second @simonswine's remark on making sure we can handle installations with 100s of tenants. I think this page can be the entry point for compaction introspection but some of the tables that go into the details might need to be split out into separate pages (tenant / shard level).

Also, for the next iteration it would be good to link to the existing block explorer ("Bucket Blocks Explorer") so that we can not only list jobs but also inspect the blocks in any job.

pkg/metastore/compaction_service.go
bryanhuhta: 👍 https://github.com/grafana/pyroscope/pull/5465

bryanhuhta:
Moving this to draft as I test cells with lots of tenants. I will incorporate some of @aleks-p's ideas about creating sub pages to help improve the information density on each page

PR #5465, opened 2026-08-05
fix(compaction): Use `raft_log.CompactionJobStatusUpdate` instead of `metastorev1.CompactionJobStatusUpdate`

bryanhuhta:
This code previously used `req` to prepare compaction updates to the metastore Raft log. We're supposed to be using `proposeReq` to prepare updates.

The bug sequence:

1. Worker finishes a job and calls `PollCompactionJobs` with a status update carrying `CompactedBlocks`.
2. The service builds the trimmed request (name/token/status only), but then (incorrectly) proposes req, the full worker request.
3. `Propose()` serializes the full request, including `CompactedBlocks`, and that byte blob is fsync'd into the leader's log and replicated to all peers.
4. The FSM unmarshals the entry as `GetCompactionPlanUpdateRequest`. Since field 4 doesn't exist on that type, the CompactedBlocks` bytes land in protobuf unknown fields and are ignored.
5. Then the service does the second proposal (`UpdateCompactionPlanRequest`), which intentionally includes `CompactedBlocks` for completed jobs, so the same block metadata got written to the raft log a second time.

### DIFF
--- a/api/metastore/v1/raft_log/raft_log.proto
+++ b/api/metastore/v1/raft_log/raft_log.proto
@@ -33,6 +33,11 @@ message CompactionJobStatusUpdate {
   string name = 1;
   uint64 token = 2;
   metastore.v1.CompactionJobStatus status = 3;
+  // Reserved field represents `compacted_blocks`. The message must remain
+  // wire-compatible with metastore.v1.CompactionJobStatusUpdate: raft log
+  // entries written by older versions may erroneously contain field data for
+  // metastore.v1.CompactionJobStatusUpdate#compacted_blocks.
+  reserved 4;
 }
 
 // GetCompactionPlanUpdateResponse includes the planned change.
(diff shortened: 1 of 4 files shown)


api/metastore/v1/raft_log/raft_log.proto
```diff
@@ -33,6 +33,11 @@ message CompactionJobStatusUpdate {
   string name = 1;
   uint64 token = 2;
   metastore.v1.CompactionJobStatus status = 3;
+  // Reserved field represents `compacted_blocks`. The message must remain
+  // wire-compatible with metastore.v1.CompactionJobStatusUpdate: raft log
+  // entries written by older versions may erroneously contain field data for
+  // metastore.v1.CompactionJobStatusUpdate#compacted_blocks.
+  reserved 4;
```
bryanhuhta, line 40: There will be old log entries written with this proto object:

https://github.com/grafana/pyroscope/blob/98fb7e138d01a9cda14aa0b05542ec2233fb3e08/api/metastore/v1/compactor.proto#L61-L67

By switching to writing/reading a different object, we need to make sure this is wire-compatible so we don't get erroneous crashes. Hypothetically this field could be removed in the future once we know that all raft logs are now writing the `raft_log.proto` object and not the `compactor.proto` one (though I don't think this is necessary).

aleks-p (dismissed):
LGTM. I wonder if we can add a test that hardens this a bit?

bryanhuhta:
> LGTM. I wonder if we can add a test that hardens this a bit?

In https://github.com/grafana/pyroscope/pull/5465/commits/9bca9dc3beadab74a62cc20652eb47707044c403 and https://github.com/grafana/pyroscope/pull/5465/commits/f552fe55024bf95ecb271047f6d2897b1bd0d515 I added a test case to verify that the raft log uses the correct protobuf type. I also added a test case to make sure we can still safely unmarshal the old entries safely.

aleks-p (approved):

PR #5475, opened 2026-08-06
feat(experimental): fence async query state with conditional writes

jake-kramer:
Builds on async query resumability (#5442).

## What

Async query metadata is now guarded by a single mechanism: the object version of `metadata.json`. We use object storage CAS semantics: owners write with `If-Match` against the version of their last write; a mismatch means another owner took over, so the execution cancels and frees its concurrency slot. Racing adoption claims pass through the same condition, so exactly one wins. `result.pb` is uploaded before the success mark and readers trust its existence over metadata, so a finished query is never lost or masked by a stale failure.

Startup probes the backend with real writes. A backend that does not enforce `If-Match` degrades gracefully: plain writes, adoption disabled. There are only a few exceptions of backends that do not support CAS.

## Why

#5442 tolerated two races: two frontends could adopt and run the same query, and a superseded owner could stamp a stale `FAILURE` that a polling client sees before giving up. Conditional writes close both.

## Commits

Each commit builds and tests on its own:

1. `chore(deps)` — objstore bump with the conditional-upload API
2. `refactor(objstore)` — in-memory bucket delegates to upstream, so tests get real If-Match behavior
3. `fix(objstore)` — S3 reports racing conditional PUTs as 409, not only 412
4. `fix(blocks)` — the dual-write markers path rejects upload options
5. `feat` — the enforcement probe
6. `feat` — the fenced store (`fencing.go` holds the write machinery, `store.go` the lifecycle)

## Testing

- Local smoke test: v2 monolith + `profilecli query profile --async`, fenced mode verified on filesystem storage (xattr versions).

## Manual verification

Ran this build in dev and checked the following by hand:

- **Startup probe.** Deployed the build and confirmed every pod passed the new conditional-write probe at startup.
- **Basic submit and poll.** Submitted an async query, got a request ID back, and polled until it returned a result. Confirmed from the logs that the pod which accepted the query was also the one that executed and completed it. The query took about 60 seconds (longer than the 45-second lease timeout) so the owner had to have been renewing its lease throughout.
- **Polling from a different pod.** Submitted a query to one pod, then polled it from a different one and got the completed result back. Checked the logs and only the first pod ever reported owning or executing it.
- **Killing the owner mid-query.** Submitted a query and hard-killed the pod running it a few seconds later. Kept polling from a surviving pod: the query stayed in progress for roughly 90 seconds, then came back with a result. Inspected the stored record afterwards and saw a different pod listed as the owner, with an adoption count of 1.
- **Forcing a takeover race.** Wrote 24 query records into storage at once, each with an already-expired lease, so every pod's takeover scan saw them simultaneously. Let them race, then read the adoption counters off all six pods: 53 claim attempts, 24 succeeded (exactly one per query), 29 were rejected by the version check. Checked all 24 stored records afterwards; each had a single owner and an adoption count of 1, and the pods that lost races hadn't executed anything.

Follow up to #5442


pkg/frontend/async/coordinator.go
```diff
@@ -23,10 +24,7 @@ type Limits interface {
 	MaxAsyncQueryConcurrency(tenantID string) int
 }
 
-// queryResult is the result of a query execution sent over a channel.
-// On success, Response carries the raw SelectMergeStacktracesResponse
-// from the wrapped handler (its Async field is unset); the
-// coordinator/store own request_id and status.
+// queryResult is a query execution's outcome, sent over resultCh.
```
bryanhuhta, line 27: `resultCh` seems to be an artifact of some intermediate state and not applicable here.

pkg/pyroscope/modules_experimental.go
```diff
@@ -191,13 +192,41 @@ func (f *Pyroscope) getFrontendAddress() (addr string, err error) {
...
+		}
+		level.Warn(f.logger).Log("msg", "object storage does not enforce conditional writes; async query adoption is disabled and metadata writes are unconditional", "err", err)
+		f.asyncQueryStore = asyncquery.NewStore(
+			log.With(f.logger, "component", "async-query-store"),
+			f.storageBucket,
+			f.reg,
+		)
+		f.asyncQueryStore.DisableConditionalWrites()
+		return f.asyncQueryStore, nil
+	}
```
bryanhuhta, line 228: It would be nice to be able to always disable conditional writes with a config value, that way an oncall has a lever at their disposal if conditional writes are causing a problem.

pkg/frontend/async/store.go
```diff
@@ -219,28 +253,43 @@ func (s *Store) specAbsent(ctx context.Context, tenantID, requestID string) (boo
...
+		// self-healing), while an unconditional fallback could stomp a new
+		// owner's claim.
+		level.Warn(s.logger).Log("msg", "failed to refresh lease version after heartbeat; a future tick may self-orphan", "tenant", tenantID, "request_id", requestID, "err", err)
 		return nil
 	}
-	meta.LastHeartbeat = time.Now().UTC()
-	meta.Owner = s.ownerID
-	return s.saveJSON(ctx, metaPath, &meta)
+	lease.version = attrs.Version
+	return nil
```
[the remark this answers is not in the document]
bryanhuhta: @jake-kramer is case this actually possible?

pkg/frontend/async/store.go
```diff
@@ -258,57 +310,109 @@ func (s *Store) complete(ctx context.Context, tenantID, requestID string, resp *
...
-		return nil
+	if err := s.saveJSONConditional(ctx, metaPath, &meta, lease.version); err != nil {
+		// result.pb is durable and served via the anchor even if this mark
+		// never lands. Losing to a new owner is routine contention, so Debug.
+		msg := "failed to record successful result in metadata; result is durable and served via the anchor"
+		if s.bucket.IsConditionNotMetErr(err) {
+			level.Debug(s.logger).Log("msg", msg, "tenant", tenantID, "request_id", requestID, "err", err)
+		} else {
+			level.Warn(s.logger).Log("msg", msg, "tenant", tenantID, "request_id", requestID, "err", err)
+		}
```
bryanhuhta, line 326: Maybe something like this so we can reuse the log fields?

```suggestion
		msg := "failed to record successful result in metadata; result is durable and served via the anchor"
		logger := level.Warn(s.logger)
		if s.bucket.IsConditionNotMetErr(err) {
			// result.pb is durable and served via the anchor even if this mark
			// never lands. Losing to a new owner is routine contention, so Debug.
			logger = level.Debug(s.logger)
		}
		logger.Log("msg", msg, "tenant", tenantID, "request_id", requestID, "err", err)
```

bryanhuhta (approved):
This LGTM! Let's get this into dev/ops and test it out. I suspect it will be challenging to tell if its markedly better than the non-CAS version, but I think it will help operationally as async queries start to scale out in adoption.

aleks-p (changes requested):
Sorry I looked at this a bit late, but I'd like to challenge both this change and #5442. 

The premise of #5442 is that "a client polling for the result does so indefinitely with no way to detect or recover from the loss", when a query frontend owning a query goes down (e.g. gets OOMKilled or just re-scheduled).

There are two problems we are trying to solve here, "detect" and "recover". The second one is already a problem for regular (sync) queries and we don't solve it there - the clients decide if they want to retry or not (currently the ones we control don't). I don't think it is the right time to try to solve it for async queries either - considering this is still being experimented with (also retrying OOMKiller queries can actually be harmful if not done right). 

Now, for the first problem - detecting that an async query is abandoned - this is a valid missing piece but I think we can simplify how we solve it, and make it more complex later if needed. Concretely, we could make abandonment a read-side judgement rather than a distributed protocol. That lets us drop spec.pb, the Dispatcher, the adoption scan and the adoption counters - and with them this PR: no CAS, no startup probe, no degrade mode, no objstore bump.

None of the fencing work is wasted if we later decide recovery is worth it - it's the right foundation for it. I just don't think we should pay for it before we know we need it.

PR #5483, opened 2026-08-10
fix(v2): serve profile-type Series queries from metadata when the selector is empty

aleks-p:
`Series` has a fast path that answers the "list all services and profile types" query from block metadata without touching blocks. `isProfileTypeQuery` guards it with `len(matchers) > 0`, but clients spell an unfiltered query either as an empty list or as a single empty selector `["{}"]`. Both parse to zero matchers, so the guard was rejecting on notation alone and the query fell through to a full TSDB index scan of every block in the range.

Dispatch on the parsed matchers instead, which the handler parses anyway a few lines later.

Notes:
- `LabelNames` has the same symptom but no fast path at all, and the metadata index only holds `service_name`/`__profile_type__`, so it genuinely needs the block scan. Not addressed here.
- Worth also fixing in Profiles Drilldown so it sends an empty list, but the server shouldn't be sensitive to the spelling.

### DIFF
--- a/pkg/frontend/readpath/queryfrontend/query_series_labels.go
+++ b/pkg/frontend/readpath/queryfrontend/query_series_labels.go
@@ -2,6 +2,7 @@ package queryfrontend
 
 import (
 	"context"
+	"fmt"
 
 	"connectrpc.com/connect"
 	"github.com/go-kit/log/level"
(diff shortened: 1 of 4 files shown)


pkg/frontend/readpath/queryfrontend/query_series_labels.go
```diff
@@ -72,15 +73,17 @@ func (q *QueryFrontend) Series(
 		return connect.NewResponse(&querierv1.SeriesResponse{}), nil
 	}
 
-	if q.isProfileTypeQuery(c.Msg.LabelNames, c.Msg.Matchers) {
+	matchers, err := parseMatchers(c.Msg.Matchers)
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("parsing label selector: %w", err))
```
marcsanmi: Nit: this outer wrap came over from `buildLabelSelectorFromMatchers`, but `parseMatchers` already wraps with "failed to parse label selector ...", so the message doubles up: `parsing label selector: failed to parse label selector []: ...`. Since the call is inlined now, could just pass `err` (?)

marcsanmi (dismissed):
LGTM, thanks for fixing this!

pkg/frontend/readpath/queryfrontend/query_series_labels.go
aleks-p: good catch, addressed in 045133a

marcsanmi (approved):

PR #5498, opened 2026-08-12
docs: remove duplicate self vs total screenshot

6fears7:
The annotated screenshot was added twice, the first of which is in the wrong section and repeated below in the correct section


marcsanmi (dismissed):
Hi @6fears7,

I think this removes the wrong image `self_vs_total_annotated.png` only appears once in the file; the actual duplicate is `self_vs_total_explainer_diagram.png`.

jake-kramer (approved):
Thanks for your contribution

PR #5518, opened 2026-08-14
fix(metastore): disable raft pipeline replication

aleks-p:
A metastore replica in one of our dev cells stops being replicated to and never recovers. Only restarting the *leader* clears it. #4892 and #4935 helped other symptoms but not this one, and it still happens with #5307 deployed.

One theory is that it is a deadlock in hashicorp/raft's pipeline replication. 

<details>
  <summary>Technical details</summary>
The transport sizes both `netPipeline` channels at `MaxRPCsInFlight-2`, and we leave `MaxRPCsInFlight` at the default 2, so both are unbuffered. When a follower answers `Success=false`, `pipelineDecode` returns and closes `finishCh`, leaving `doneCh` with no reader; the SEND loop then races between the closed `finishCh` and a ready `triggerCh`, and picking `triggerCh` issues one more send that parks on `inprogressCh` while `decodeResponses` parks on `doneCh`. `pipeline.Close` is deferred inside `pipelineReplicate`, which is stuck in that select, so `shutdownCh` never closes.

`replicate()` never returns either, so its deferred `close(stopHeartbeat)` never runs. That is why the victim looks healthy from the outside: heartbeats keep arriving, so it never times out and forces an election that would have healed things — it just never receives another log entry.

Our own `timeoutLogStore` (#4892) is one way to produce the triggering `Success=false`, since raft's `appendEntries` handler returns without setting `Success` when `StoreLogs` fails. A rejected `appendEntries` at leader election does it too.

</details>

The theory was confirmed on a live cluster: the leader had four pipelines but only two `pipelineDecode` goroutines, with two senders parked in `netPipeline.AppendEntries` - blocked 4287 and 15948 minutes, one per unready replica. The 4287-minute one matched that replica's idle `runFSM` and its single log-store write timeout to within a few minutes.

Notes:

- This is a stop-gap; the real fix belongs upstream. hashicorp/raft#612 is the same bug, open since 2024. https://github.com/hashicorp/raft/pull/683 would at least surface the stall in the leader's logs, but its remedy — suppressing that peer's heartbeats so it forces an election — cannot fire here: with pre-vote on, the other peers reject the victim's pre-vote while they still have a leader.
- `TestRaftPipelineDeadlock` asserts that the follower catches up again once its log store is healthy. It fails on round 1 without this change and passes with it, 10/10 runs either way.
- This and #5307 are two halves of the same production failure: this fixes the deadlock, #5307 the `non-monotonic log entries` livelock. The test itself no longer depends on #5307.
- Running in a dev cell since 14 Aug. Both triggers have fired since — a follower log-store timeout and a full rolling update — with no recurrence and no change in replication throughput.

cc @simonswine

### DIFF
--- a/.idea/pyroscope.iml
+++ b/.idea/pyroscope.iml
@@ -1,7 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <module type="WEB_MODULE" version="4">
   <component name="Go" enabled="true" />
+  <component name="GoModuleSettings" enabled="true" />
   <component name="NewModuleRootManager">
     <content url="file://$MODULE_DIR$" />
+    <orderEntry type="sourceFolder" forTests="false" />
   </component>
 </module>
--- a/.idea/pyroscope.iml
+++ b/.idea/pyroscope.iml
@@ -1,9 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <module type="WEB_MODULE" version="4">
   <component name="Go" enabled="true" />
-  <component name="GoModuleSettings" enabled="true" />
   <component name="NewModuleRootManager">
     <content url="file://$MODULE_DIR$" />
-    <orderEntry type="sourceFolder" forTests="false" />
   </component>
 </module>
--- a/pkg/metastore/raftnode/node.go
+++ b/pkg/metastore/raftnode/node.go
@@ -189,9 +189,6 @@ func NewNode(
 	return &n, nil
 }
 
-// newTransport creates the raft TCP transport. A nil Logger makes raft build
-// the same "raft-net" hclog on stderr that the plain NewTCPTransport
-// constructor would.
 func newTransport(config Config, advertise net.Addr) (*raft.NetworkTransport, error) {
 	return raft.NewTCPTransportWithConfig(config.BindAddress, advertise,
 		&raft.NetworkTransportConfig{
--- a/pkg/metastore/raftnode/node.go
+++ b/pkg/metastore/raftnode/node.go
@@ -74,17 +74,18 @@ const (
 	defaultTransportTimeout      = 10 * time.Second
 	defaultLogStoreTimeout       = 10 * time.Second
 
+	// Workaround for upstream issue https://github.com/hashicorp/raft/issues/612.
 	// Anything below raft's minInFlightForPipelining (2) makes
 	// AppendEntriesPipeline return ErrPipelineReplicationNotSupported, which
 	// replicate() handles by staying in synchronous RPC mode.
 	//
-	// Pipelining is unusable on raft v1.7.3: it sizes both netPipeline
+	// Pipelining is unstable on raft v1.7.3: it sizes both netPipeline
 	// channels at MaxRPCsInFlight-2, so the default of 2 leaves them
 	// unbuffered, and a follower that fails a single AppendEntries can
-	// deadlock replication to itself permanently. See
-	// TestRaftPipelineDeadlock for the full interlock. Raising this instead
-	// of lowering it would only widen the window, since both channels stay
-	// bounded. Revert once the upstream bug is fixed.
+	// deadlock replication to itself permanently. TestRaftPipelineDeadlock reproduces the issue.
+	//
+	// Raising the value instead of lowering it would only widen the window, since both channels stay
+	// bounded.
 	//
 	// The cost is small: replication still ships up to MaxAppendEntries
 	// entries per round trip, and a single entry commits in one round trip
--- a/pkg/metastore/raftnode/node.go
+++ b/pkg/metastore/raftnode/node.go
@@ -73,6 +73,23 @@ const (
 	defaultTransportConnPoolSize = 10
 	defaultTransportTimeout      = 10 * time.Second
 	defaultLogStoreTimeout       = 10 * time.Second
+
+	// Anything below raft's minInFlightForPipelining (2) makes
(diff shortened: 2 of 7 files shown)


pkg/metastore/test/create.go
```diff
@@ -141,6 +144,33 @@ func NewMetastoreSet(t *testing.T, cfg *metastore.Config, n int, bucket objstore
 	return res
 }
 
+// freeLocalPorts picks n ports that are free right now. Raft needs real TCP
+// listeners and every peer address has to be known before any node starts, so
+// they cannot simply be bound as :0. Fixed ports are not an option either:
+// this helper is used from more than one package, and go test runs packages
+// in parallel, so they would intermittently collide with "address already in
+// use".
+func freeLocalPorts(t *testing.T, n int) []int {
```
simonswine: Let's reuse `pkg/test/integration.GetFreePorts`

pkg/metastore/raftnode/node.go
```diff
@@ -73,6 +73,23 @@ const (
...
+	// channels at MaxRPCsInFlight-2, so the default of 2 leaves them
+	// unbuffered, and a follower that fails a single AppendEntries can
+	// deadlock replication to itself permanently. See
+	// TestRaftPipelineDeadlock for the full interlock. Raising this instead
+	// of lowering it would only widen the window, since both channels stay
+	// bounded. Revert once the upstream bug is fixed.
+	//
+	// The cost is small: replication still ships up to MaxAppendEntries
+	// entries per round trip, and a single entry commits in one round trip
+	// either way.
```
simonswine, line 92: Worth linking an upstream report.

https://github.com/hashicorp/raft/issues/612 might do, or eventually reporting one specific to MaxRPCsInFlight instead of it.

.idea/pyroscope.iml
```diff
@@ -1,7 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <module type="WEB_MODULE" version="4">
   <component name="Go" enabled="true" />
+  <component name="GoModuleSettings" enabled="true" />
   <component name="NewModuleRootManager">
     <content url="file://$MODULE_DIR$" />
+    <orderEntry type="sourceFolder" forTests="false" />
```
aleks-p: This snuck in and I've been carrying it locally for a few weeks now :D I can keep it or remove it

pkg/metastore/raftnode/node.go
aleks-p, line 92: This is verbose but is the main explanation and hints towards it being a workaround until Raft addresses the problem. The expectation is that https://github.com/hashicorp/raft/pull/683 might help and we could remove all the workarounds, but it is yet to be confirmed.

pkg/metastore/test/create.go
aleks-p: this is duplicated from integration tests, I'll try to move it out to a util

pkg/metastore/test/metastore_pipeline_deadlock_test.go
```diff
@@ -0,0 +1,337 @@
...
+	require.Eventually(t, func() bool {
+		return followerCaughtUp(t, ms, leaderIdx, followerIdx)
+	}, 20*time.Second, 250*time.Millisecond, "follower did not catch up with the leader")
+	t.Log("cluster healthy; replication would be pipelining if it were enabled")
+
+	for round := 1; round <= rounds; round++ {
+		faults[followerIdx].fail()
+		time.Sleep(failFor)
+		faults[followerIdx].clear()
+
```
simonswine, line 137: In this loop we might not notice if there were any successful writes (`AddBlock` have no error check), between `fail` and `clear`.

Probably worth getting the commit index and making sure it increased.

.idea/pyroscope.iml
simonswine: Not too sure if you want those changes to be there or not

pkg/metastore/test/create.go
```diff
@@ -26,6 +26,8 @@ import (
 	"github.com/grafana/pyroscope/v2/pkg/test/mocks/mockdiscovery"
 	"github.com/grafana/pyroscope/v2/pkg/util/health"
 	"github.com/grafana/pyroscope/v2/pkg/validation"
+
+	metastorev1 "github.com/grafana/pyroscope/api/gen/proto/go/metastore/v1"
```
simonswine, line 29: Not clear why this reorders, maybe because of the "v2"

simonswine (dismissed):
LGTM. I don't see how we benefit from the pipelining, so this seems like a reasonable workaround. 

A couple of nits in the comments

pkg/metastore/test/create.go
aleks-p: yeah, GoLand applies `goimports` on save, I ran it manually via the CLI as well and it does the same

.idea/pyroscope.iml
aleks-p: c56038b

pkg/metastore/test/metastore_pipeline_deadlock_test.go
aleks-p: 6786254

pkg/metastore/raftnode/node.go
aleks-p: 513b76d

pkg/metastore/test/create.go
aleks-p: c56038b

simonswine (approved):

PR #5519, opened 2026-08-15
fix(segmentwriter): allow shard shuffle table to grow past 4096 entries

luweglarz:
While playing with tokens on segment-writer, I hit a panic during shard shuffling when the number of shards in the ring exceeds the fixed steps[4096] table:
```
distributor-1  | panic: runtime error: index out of range [4096] with length 4096
distributor-1  | goroutine 601 [running]:
distributor-1  | github.com/grafana/pyroscope/v2/pkg/util.PanicError({0x34820c0, 0xc000e14180})
distributor-1  | 	/home/lweglarz/pyroscope/pkg/util/recovery.go:46 +0x4d
distributor-1  | github.com/grafana/pyroscope/v2/pkg/distributor/writepath.(*Router).sendToSegmentWriterOnly.(*Router).send.func1.1()
distributor-1  | 	/home/lweglarz/pyroscope/pkg/distributor/writepath/router.go:230 +0x67
distributor-1  | panic({0x34820c0?, 0xc000e14180?})
...
distributor-1  | github.com/grafana/pyroscope/v2/pkg/segmentwriter/client/distributor.(*perm).resize(...)
distributor-1  | 	.../distributor.go:324 +0x15d
```

I hit it with `-segment-writer.num-tokens=1024` (5 instances × 1024 tokens = 5120 total shards). 
The panic triggers whenever instances × num-tokens > 4096.

The fix lets steps grow to fit the ring's actual shard count instead of being capped at a fixed size.

### DIFF
--- a/pkg/segmentwriter/client/distributor/distributor.go
+++ b/pkg/segmentwriter/client/distributor/distributor.go
@@ -313,6 +313,7 @@ func (i *iterator) At() ring.InstanceDesc {
 type perm struct{ v []uint32 }
 
 func (p *perm) resize(n int) {
+	s := loadSteps(n)
 	d := max(0, n-len(p.v))
 	p.v = slices.Grow(p.v, d)[:n]
 	// We do want to start with 0 (in contrast to the standard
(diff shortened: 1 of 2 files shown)


pkg/segmentwriter/client/distributor/distributor.go
```diff
@@ -321,26 +322,39 @@ func (p *perm) resize(n int) {
 	// Although, it's possible to make the change incrementally,
 	// for simplicity, we just rebuild the permutation.
 	for i := 0; i < n; i++ {
-		j := steps[i]
```
simonswine: I would prefer to keep this simpler: `steps[i%4<<10]`

Wdyt?

pkg/segmentwriter/client/distributor/distributor.go
```diff
@@ -321,26 +322,39 @@ func (p *perm) resize(n int) {
...
+	stepsMu sync.Mutex
+	steps   []uint32
+)
+
+// The seed impacts mapping of shards to nodes.
+// TODO(kolesnikovae):
+//
+//	Stochastic approach does not work well
+//	in all the cases; it should be replaced
+//	with a deterministic one.
```
simonswine: That TODO is no longer in the right place, let's maybe put it before `loadSteps(...)`

simonswine (dismissed):
Thanks for catching this! LGTM

simonswine (approved):

PR #5524, opened 2026-08-18
Add security reporting guidance

KristianGrafana:
## What changed

- Added guidance to the repository's agent instructions directing security issues to Grafana's security reporting page instead of the repository.

## Why

This makes the expected reporting channel explicit for coding agents working in the repository.

## Validation

- `git diff --check`
- Confirmed the commit changes only one agent instruction file with one added line.


jake-kramer:
@KristianGrafana Is this ready to be looked at?

PR #5525, opened 2026-08-18
docs: document authentication options and proxy examples

### LINKED ISSUE #2442: Document authentication options and provide an example setup
Now that we don't have the means to manage users and API keys in Pyroscope, we need to provide authentication and hardening guidelines to users. I think we should also have an example of authenticating agent/SDKs via proxy such as nginx (ideally for both single-/multi-tenant deployments).

VedantMadane:
## Summary

Documents how to authenticate and harden Grafana Pyroscope now that the server has no built-in user/API key management. Covers multitenancy (`X-Scope-OrgID`), reverse-proxy patterns, and nginx examples for single- and multi-tenant setups.

## Changes

- New configure-server page: Authentication and access control.
- Includes client, Grafana, and hardening guidance.

Fixes #2442

### DIFF
--- a/docs/sources/configure-server/authentication.md
+++ b/docs/sources/configure-server/authentication.md
@@ -1,0 +1,237 @@
+---
+aliases:
+  - /docs/pyroscope/latest/configure-server/authentication/
+description: Authenticate and harden access to Grafana Pyroscope with multitenancy and reverse proxies.
+keywords:
+  - authentication
+  - authorization
+  - multitenancy
+  - security
+  - nginx
+menuTitle: Authentication
+title: Authentication and access control
+weight: 150
+---
+
+# Authentication and access control
+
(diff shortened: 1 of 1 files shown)


knylander-grafana:
@VedantMadane Thank you for your contribution to the docs! I was looking through your sections and realized we might have some overlap with some of our existing documentation. 

For example: 
* docs/sources/configure-server/about-tenant-ids.md: covers multitenancy enable/disable, X-Scope-OrgID, default anonymous tenant, flag name, tenant ID format restrictions
* docs/sources/reference-server-api/index.md: Mentions that Pryoscope doesn't include an authentication layer. 

Here are some suggestions: 
- Consider removing the "removal of the legacy self-hosted user store" claim
- Trim the "What Pyroscope authenticates" / "Multitenancy" sections to brief summaries with links to about-tenant-ids.md instead of restating them
- Keep the `nginx` examples and hardening checklist — these are great additions!

PR #5532, opened 2026-08-19
chore(ci): enable Renovate updates for shared workflows

korniltsev-grafanista:
## Summary
- enable regular Renovate updates for GitHub Actions from `grafana/shared-workflows`
- keep the existing security-only policy for all other dependencies

This will cover tagged shared-workflow actions. The existing `trigger-argo-workflow` pins remain unmanaged because they point to an untagged commit without a version annotation.

## Test plan
- [x] Validate `renovate.json` with `jq`
- [ ] Confirm the Renovate config validation workflow passes

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -4,5 +4,13 @@
     "github>grafana/grafana-renovate-config//presets/base",
     "security:only-security-updates"
   ],
-  "baseBranchPatterns": ["$default", "release/v2.2", "release/v2.1"]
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -4,5 +4,13 @@
     "github>grafana/grafana-renovate-config//presets/base",
     "security:only-security-updates"
```
marcsanmi, line 5: Might be worth also extending `github>grafana/grafana-renovate-config//presets/shared-workflows`?

PR #5536, opened 2026-08-19
feat(metastore): optionally exit when a log store write times out

aleks-p:
Follow-up to #5518, which stopped a follower disk stall from deadlocking raft's pipeline replication. The other half of that failure is still open: `timeoutLogStore` stops waiting for a slow write but cannot cancel it, so the goroutine keeps running and may still land its entries. Raft believes the append failed, the disk disagrees, and raft's retry for the same index hits raft-wal's monotonicity check.

#5307 reconciles that on every later write. This adds the opposite option: don't reconcile, terminate. raft-wal truncates any partially written tail when it reopens the segment, so the node restarts with a valid prefix and raft rebuilds its view from disk. A replica returning with uncommitted entries the leader lacks is an ordinary raft state, resolved by the usual conflict truncation.

Off by default (`-metastore.raft.exit-on-log-store-timeout`); nothing changes unless it is set.

Notes:

- The trade is availability for provability. Each timeout costs a restart plus a snapshot restore instead of being ridden out in place; a persistently bad disk becomes `CrashLoopBackOff`
- No write serialisation is needed here, unlike #5307: after the first timeout the process is leaving, so there is no second write to race the abandoned one.
- This and #5307 are alternatives, not companions. Enabling this flag makes #5307 unnecessary; leaving it off makes #5307 the fix.
- If hashicorp/raft#683 lands, `timeoutLogStore` and everything built on it can go away.

cc @simonswine

### DIFF
--- a/pkg/metastore/raftnode/logstore.go
+++ b/pkg/metastore/raftnode/logstore.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
 	"github.com/hashicorp/raft"
 	"github.com/prometheus/client_golang/prometheus"
 )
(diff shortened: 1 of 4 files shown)


pkg/metastore/raftnode/logstore.go
```diff
@@ -89,6 +110,14 @@ func (s *timeoutLogStore) withTimeout(fn func() error) error {
 		}
 		s.writeLatency.Observe(time.Since(start).Seconds())
 		s.timeouts.Inc()
+		if s.exitOnTimeout {
+			level.Error(s.logger).Log(
+				"msg", "log store write timed out, exiting: the write cannot be cancelled, "+
+					"so raft's view of the log may no longer match what is on disk",
+				"timeout", s.timeout,
+			)
+			s.exit(1)
```
marcsanmi, line 119: I think `os.Exit` might drop the log (the app logger buffers through `util.NewAsyncWriter`)?

simonswine: I think this is actually wanted behaviour, if our IO blocks, our log writing might block too.  Which is obviously not great with diagnostics.

What we could do we could exit with something not 1, so we know this is an exit coming from this timeout guard.

PR #5537, opened 2026-08-19
chore: Add more logging to async queries

bryanhuhta:
We don't have great logging visibility into the lifecycle of a async query. This adds some logging points to make it easier to trace a query across polls.

### DIFF
--- a/pkg/frontend/async/handler.go
+++ b/pkg/frontend/async/handler.go
@@ -131,7 +131,7 @@ func (h *Handler) poll(
 	case StatusInProgress:
 		level.Debug(logger).Log("msg", "async query polled, still in progress")
 	case StatusFailure:
-		level.Info(logger).Log("msg", "async query result returned to client", "err", result.Metadata.ErrorMessage)
+		level.Error(logger).Log("msg", "async query result returned to client", "err", result.Metadata.ErrorMessage)
 	default:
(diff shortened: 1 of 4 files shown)


pkg/frontend/async/handler.go
```diff
@@ -109,5 +123,17 @@ func (h *Handler) poll(
...
 
+	logger = log.With(logger,
+		"status", string(result.Metadata.Status),
+		"age", time.Since(result.Metadata.CreatedAt).Round(time.Millisecond),
+	)
+	switch result.Metadata.Status {
+	case StatusInProgress:
+		level.Debug(logger).Log("msg", "async query polled, still in progress")
+	case StatusFailure:
+		level.Info(logger).Log("msg", "async query result returned to client", "err", result.Metadata.ErrorMessage)
```
jake-kramer: `warn` or `error`?

jake-kramer (dismissed):

pkg/frontend/async/handler.go
bryanhuhta: Good point, fixed in https://github.com/grafana/pyroscope/pull/5537/commits/b62fb762f3ded1b84db79a0a06c2549ec75a333d

jake-kramer (approved):

PR #5538, opened 2026-08-20
docs(config): document the retention_period limit

### LINKED ISSUE #5523: Retention period is not documented
Hi, I was looking for the retention period in the v2 [server configuration guide](https://grafana.com/docs/pyroscope/latest/configure-server/reference-configuration-parameters/#limits) and did not see any mention of retention.

It seems there _is_ such an option, but it's marked as "hidden".

https://github.com/grafana/pyroscope/blob/cd724d81160a31a1f6ff4cf55851f99643f89ece/pkg/metastore/index/cleaner/retention/retention.go#L26-L28

It is in the help text, though

https://github.com/grafana/pyroscope/blob/cd724d81160a31a1f6ff4cf55851f99643f89ece/cmd/pyroscope/help.txt.tmpl#L218-L219

Can this be documented, or is it not ready for prime-time?

arcusbuilds:
Fixes #5523

`retention_period` is tagged `doc:"hidden"`, so `tools/doc-generator` skips it and it has never appeared in the configuration reference. The reporter found it in the source and asked whether that was deliberate.

It reads as an oversight rather than a knob you meant to keep private:

* it is inlined into `Limits` and read through `Overrides.Retention()`, like every other tenant limit
* it already shows up in the basic `-help` output, not only in `-help-all`
* its sibling `compactor_blocks_retention_period` is documented with no such tag
* per-tenant overrides are already covered by `TestRetentionOverridesYAML`

So this drops the tag and regenerates the reference page and the example config. It also names the parameter in the metadata index architecture page, whose time-based retention section explained the behavior without ever saying which setting controls it.

No behavior change. The help printer classifies flags by the `category` tag rather than the `doc` tag, so `help.txt.tmpl` and `help-all.txt.tmpl` are byte identical and `TestHelp` passes unchanged.

**Changes**

* `pkg/metastore/index/cleaner/retention/retention.go`: remove `doc:"hidden"`
* `docs/sources/configure-server/reference-configuration-parameters/index.md`: regenerated
* `cmd/pyroscope/pyroscope.yaml`: regenerated
* `docs/sources/reference-pyroscope-v2-architecture/metadata-index/index.md`: link the parameter from the retention section

**Verification**

* both `doc-generator` invocations from the `generate` target re-run; the diff is the 4 and 3 added lines above and nothing else
* `go test ./cmd/pyroscope/` passes, both help golden files unchanged
* `go test ./pkg/validation/... ./pkg/metastore/index/cleaner/...` passes

I left the flag's description string as it is to keep the diff mechanical. Happy to reword it here if you would rather it read differently now that it is published.

### DIFF
--- a/cmd/pyroscope/pyroscope.yaml
+++ b/cmd/pyroscope/pyroscope.yaml
@@ -481,6 +481,9 @@ limits:
   # limit is enforced in the distributor. 0 to disable, defaults to 10m.
   # reject_newer_than: 10m
 
+  # Retention period for the data. 0 means data never deleted.
+  # retention_period: 31d
+
   # Maximum number of recording rules a tenant can create. 0 to disable.
   # max_recording_rules: 25
 
--- a/docs/sources/configure-server/reference-configuration-parameters/index.md
+++ b/docs/sources/configure-server/reference-configuration-parameters/index.md
@@ -3397,6 +3397,10 @@ distributor_usage_groups:
 # CLI flag: -validation.reject-newer-than
 [reject_newer_than: <duration> | default = 10m]
 
+# Retention period for the data. 0 means data never deleted.
+# CLI flag: -retention-period
+[retention_period: <duration> | default = 31d]
+
 # Maximum number of recording rules a tenant can create. 0 to disable.
 # CLI flag: -recording-rules.max-rules-per-tenant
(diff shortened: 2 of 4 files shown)


jake-kramer:
Thank you for your contribution!

jake-kramer (approved):

PR #5546, opened 2026-08-21
feat(distributor): add an instance-level inflight bytes limit

### LINKED ISSUE #5437: Add an instance-level inflight bytes limit to the distributor

aleks-p:
Distributors can be OOMKilled when segment writers slow down or go away. `-distributor.max-inflight-bytes` caps the total uncompressed profile bytes one distributor holds, rejecting further pushes with 503 until capacity frees up. Off by default.

Closes #5437

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -41,6 +41,7 @@ import (
 	connectapi "github.com/grafana/pyroscope/v2/pkg/api/connect"
 	"github.com/grafana/pyroscope/v2/pkg/clientpool"
 	"github.com/grafana/pyroscope/v2/pkg/distributor/aggregator"
+	"github.com/grafana/pyroscope/v2/pkg/distributor/inflight"
 	"github.com/grafana/pyroscope/v2/pkg/distributor/ingestlimits"
 	distributormodel "github.com/grafana/pyroscope/v2/pkg/distributor/model"
 	"github.com/grafana/pyroscope/v2/pkg/distributor/sampling"
@@ -83,6 +84,8 @@ type Config struct {
 	PushTimeout time.Duration
 	PoolConfig  clientpool.PoolConfig `yaml:"pool_config,omitempty"`
 
+	MaxInflightBytes int64 `yaml:"max_inflight_bytes" category:"advanced"`
+
 	// Distributors ring
 	DistributorRing util.CommonRingConfig `yaml:"ring"`
 }
@@ -91,6 +94,7 @@ type Config struct {
 func (cfg *Config) RegisterFlags(fs *flag.FlagSet, logger log.Logger) {
 	cfg.PoolConfig.RegisterFlagsWithPrefix("distributor", fs)
 	fs.DurationVar(&cfg.PushTimeout, "distributor.push.timeout", 5*time.Second, "Timeout when pushing data to ingester.")
+	fs.Int64Var(&cfg.MaxInflightBytes, "distributor.max-inflight-bytes", 0, "Maximum total size, in bytes, of the uncompressed profiles the distributor may hold in memory at a time. Requests exceeding it are rejected with 503. 0 to disable.")
 	cfg.DistributorRing.RegisterFlags("distributor.ring.", "collectors/", "distributors", fs, logger)
 }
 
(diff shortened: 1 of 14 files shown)


AustinKobayashi (dismissed):
~LGTM~

pkg/distributor/distributor.go
```diff
@@ -979,15 +1040,67 @@ func profileSizeBytes(p *profilev1.Profile, fullSize int64) (symbols, samples in
...
+
+	mu          sync.Mutex
+	reservation *inflight.Reservation
+	released    bool
+}
+
+// charge accounts for a contribution of size bytes. Merge interns its input,
+// so the accumulator grows by at most that much: the reservation is an upper
+// bound on what the aggregate holds, never an underestimate.
+func (a *pendingAggregate) charge(l *inflight.Limiter, size int64) {
```
AustinKobayashi, line 1059: Could this upper bound become too conservative for high-rate series with repeated stacks, causing capacity rejections even when the aggregate barely grows?

aleks-p: It definitely could! I chose this approach as a compromise between being fully accurate (for aggregations in particular, not for the rest of the feature) and recalculating the size on every addition to an aggregate which is not free. I can run some numbers to quantify this and take it from there.

In general, aggregations today are opt-in (per tenant) and in my knowledge severely underused - we could consider dropping them as well (separately to this PR).

aleks-p: Addressed in 91238c821 by recalculating the aggregate’s size every 10 additions, while summing input sizes between recalculations.

CPU overhead

| Measurement | No reconciliation | Every 10 additions | Always Reconcile |
|---|---:|---:|---:|
| Repeated profiles | 0% | 0-4% | 3-15% |
| Growing aggregates | 0% | 7-40% | 49-369%|

Growing aggregates are ones where the stack traces for each appended profile are unique, making the aggregate grow in size and therefore the `Size()` call becoming more expensive.

Over-reservation drops from 97-317x wth no reconciliation to around 10-30x with reconciliation at every 10 additions (measured right before the reconciliation).

pkg/distributor/distributor.go
```diff
@@ -795,8 +842,15 @@ func (d *Distributor) aggregate(ctx context.Context, req *distributormodel.Profi
...
-	r, ok, err := a.Aggregate(labels.Hash(), profile.TimeNanos, mergeProfile(profile))
+	// pending is assigned by the merge below, which the aggregator runs on this
+	// goroutine, and stays nil if the aggregate was never touched by us.
+	var pending *pendingAggregate
+	merge := d.mergeProfile(profile, profileInflightBytes(series.Profile), &pending)
+	r, ok, err := a.Aggregate(labels.Hash(), profile.TimeNanos, merge)
 	if err != nil {
+		// A failed aggregate is never handed to an owner, so the contributor
+		// that broke it returns the bytes.
+		pending.release()
```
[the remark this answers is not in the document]
aleks-p: Valid observation: a failed contributor can release the aggregate’s reservation before the existing owner finishes cleaning up. The failure wakes the waiting owner, so this is normally a short accounting gap during error cleanup, rather than undercounting for the remaining aggregation window.

PR #5551, opened 2026-08-24
feat(querybackend): cache metadata query results

simonswine:
Adds an opt-in, per-tenant result cache for V2 `LabelNames`, `LabelValues`, and `SeriesLabels` queries. Eligible ranges are split into aligned duration tiers; cache entries use a canonical query, sorted block IDs, and a tenant generation so block changes and generation bumps invalidate results safely.

The cache is fail-open in a dedicated object-storage bucket, with asynchronous bounded writes and low-cardinality lookup/write metrics. This includes runtime overrides, result-cache configuration and API generation, unit tests, and a draft design with rollout guidance.

## Cache Time Slots

The query range uses inclusive millisecond boundaries. Starting at the query start, the backend chooses the largest aligned tier that fits (`24h`, then `2h`, then `15m`). An unaligned remainder runs normally until the next smallest-tier boundary.

```text
Query range
  start                                                                     end
    |                                                                         |
    +-----------------+------------------------+--------------+--------------+-----------+
    | unaligned edge  |        24h slot        |   2h slot    |   15m slot   |   edge    |
    |    uncached     |       cacheable        |  cacheable   |  cacheable   | uncached  |
    +-----------------+------------------------+--------------+--------------+-----------+
```

## Benchmark

`BenchmarkQueryMetadata` against a six-day query range on macOS/arm64 (Apple M2 Max). Each cached response had the same SHA-256 as its uncached response. Cold-cache results contain one iteration; use the warm-cache measurements for comparison.

| Query | No cache | Cold cache | Warm cache | Warm speedup |
| --- | ---: | ---: | ---: | ---: |
| LabelNames | 8.576s | 11.44s | 1.089s | 7.9x |
| LabelValues (`namespace`) | 8.276s | 8.942s | 0.782s | 10.6x |
| LabelValues (`cluster`) | 8.343s | 8.667s | 0.846s | 9.9x |
| LabelValues (`pod`) | 9.667s | 10.61s | 1.918s | 5.0x |
| Series (`service_name`, `profile_type`) | 12.37s | 13.20s | 0.859s | 14.4x |
| Series (`cluster`, `namespace`, `pod`) | 29.56s | 28.09s | 4.386s | 6.7x |

### DIFF
--- a/pkg/querybackend/result_cache.go
+++ b/pkg/querybackend/result_cache.go
@@ -37,6 +37,7 @@ type ResultCacheOverrides interface {
 	ResultCacheEnabled(tenantID string) bool
 	ResultCacheGeneration(tenantID string) uint32
 	ResultCacheFragmentDurations(tenantID string) []time.Duration
+	ResultCacheMetadataServiceNameMinQueryDuration(tenantID string) time.Duration
 }
 
 type resultCacheMetrics struct {
@@ -328,7 +329,24 @@ func (q *QueryBackend) resultCacheEligible(req *queryv1.InvokeRequest) bool {
 	return q.resultCacheBucket != nil && q.resultCacheOverrides != nil && len(req.Tenant) == 1 &&
 		validQuery &&
 		!req.GetOptions().GetCollectDiagnostics() && q.resultCacheOverrides.ResultCacheEnabled(req.Tenant[0]) &&
-		len(q.resultCacheOverrides.ResultCacheFragmentDurations(req.Tenant[0])) > 0
+		len(q.resultCacheOverrides.ResultCacheFragmentDurations(req.Tenant[0])) > 0 &&
+		!resultCacheMetadataServiceNameShortRange(req, q.resultCacheOverrides.ResultCacheMetadataServiceNameMinQueryDuration(req.Tenant[0]))
+}
+
+func resultCacheMetadataServiceNameShortRange(req *queryv1.InvokeRequest, minQueryDuration time.Duration) bool {
+	if minQueryDuration <= 0 || req.EndTime-req.StartTime >= minQueryDuration.Milliseconds() {
+		return false
+	}
+	matchers, err := phlaremodel.ParseMetricSelector(req.LabelSelector)
+	if err != nil {
+		return false
+	}
+	for _, matcher := range matchers {
+		if matcher.Name == "service_name" {
+			return true
+		}
+	}
+	return false
 }
 
 // readResultCache returns a cache hit. Its error is non-nil only when a read's
--- a/pkg/querybackend/result_cache.go
+++ b/pkg/querybackend/result_cache.go
@@ -24,6 +24,7 @@ import (
 const (
 	resultCacheLabelNames   = "label_names"
 	resultCacheLabelValues  = "label_values"
+	resultCacheSeriesLabels = "series_labels"
 	resultCacheWorkers      = 2
 	resultCacheQueueSize    = 128
 	resultCacheWriteTimeout = 30 * time.Second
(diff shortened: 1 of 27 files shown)


pkg/validation/limits.go
```diff
@@ -269,6 +317,38 @@ func (l *Limits) Validate() error {
 	return nil
 }
 
+func validateResultCacheFragmentDurations(values []model.Duration) error {
```
bryanhuhta, line 320: This validation logic is pretty subtle. In your testing, did you find it favorable to need manually pick each fragment duration?

If there's no performance benefit, could we instead give three values for the user to specify:

- Minimum fragment duration
- Maximum fragment duration
- Fragment count

Then we could then basically use this algorithm:

1. Using min, max, and fragment count, generate ideal logarithmic steps for each fragment count
2. Round each duration up/down such that its a multiple of the prior duration

Worked out, it might look like this:

```
min = 15m
max = 24h
count = 4

+---------+-----+-------+--------+-----+
|         | t0  |  t1   |   t2   | t3  |
+---------+-----+-------+--------+-----+
| ideal   | 15m | ~1h9m | ~5h13m | 24h |
| rounded | 15m | 1h    | 6h     | 24h |
+---------+-----+-------+--------+-----+
```

This might be easier to configure than having to describe the logic of how to chose appropriate fragment durations. But maybe its just as finicky to write such an algorithm 😄 

Here's an AI-generated diagram plotting the idea:

Image

pkg/querybackend/result_cache.go
```diff
@@ -0,0 +1,478 @@
...
+		q.resultCacheMetrics.writes.WithLabelValues(job.queryType, job.duration, "dropped").Inc()
+	}
+}
+
+func (q *QueryBackend) runResultCacheWriter(ctx context.Context) {
+	defer q.resultCacheWorkers.Done()
+	for {
+		select {
+		case <-ctx.Done():
+			return
```
bryanhuhta, line 455: In this case we probably want to drain `q.resultCacheWrites`.

pkg/querybackend/result_cache.go
```diff
@@ -0,0 +1,478 @@
...
+}
+
+func (q *QueryBackend) executeWithResultCacheBypassed(ctx context.Context, req *queryv1.InvokeRequest) (*queryv1.InvokeResponse, error) {
+	clone := req.CloneVT()
+	clone.Options = clone.GetOptions().CloneVT()
+	if clone.Options == nil {
+		clone.Options = &queryv1.InvokeOptions{}
+	}
+	clone.Options.BypassResultCache = true
+	return q.invokeUncached(ctx, clone)
```
bryanhuhta, line 324: Do we need to clone here? Can we set `BypassResultCache` directly instead?

bryanhuhta (commented):
Took an initial pass at this and I think it looks good. I like the direction this is going and I _really_ like the perf numbers you benchmarked. I would love to see this go further and hopefully help improve some read path queries.

I have some meta-thoughts:

1. Is splitting cache fragments by time the best option? Could we instead cache the result per dataset instead and minimize cold cache read amplification?
2. We could theoretically cache some results on each query backend's disk as a slightly faster cache which avoids a lookup into object storage. If the caches get really noisy, this could save on some cost.

pkg/querybackend/result_cache.go
```diff
@@ -0,0 +1,478 @@
...
+// readResultCache returns a cache hit. Its error is non-nil only when a read's
+// object state is unknown; corrupt entries are safe to replace after execution.
+func (q *QueryBackend) readResultCache(ctx context.Context, queryType, duration, key string, expected *queryv1.ResultCacheKey, aggregator *reportAggregator) (bool, error) {
+	span, ctx := tracing.StartSpanFromContext(ctx, "QueryBackend.ResultCacheLookup")
+	defer span.Finish()
+	span.SetTag("query_type", queryType)
+	span.SetTag("fragment_duration", duration)
+	span.SetTag("fragment_start", time.UnixMilli(expected.GetQuery().GetStartTime()).UTC().Format(time.RFC3339))
+
+	r, err := q.resultCacheBucket.Get(ctx, key)
```
AustinKobayashi, line 361: Since this passes the request context directly to an optional cache lookup, could provider retries or a slow cache consume most or all of the query deadline? If the lookup returns because `ctx` expired, the fail open path will call `invokeUncached` with an already cancelled context. Would it make sense to use a shorter cache-specific timeout while preserving the original context for fallback?

pkg/usage/usage.go
```diff
@@ -219,6 +220,9 @@ func getFlagName(fl *flag.Flag) string {
 		case "*tsdb.DurationList":
 			return "comma-separated list of durations"
 		}
+		if v.Kind() == reflect.Ptr && v.Elem().Kind() == reflect.Slice && v.Elem().Type().Elem() == reflect.TypeOf(model.Duration(0)) {
```
jake-kramer: I think makes sense to reuse `DurationList`

PR #5559, opened 2026-08-26
docs: Include the client flag for utf-8 label names

simonswine:
Was missed when originally implemented

### DIFF
--- a/docs/sources/reference-server-api/index.md
+++ b/docs/sources/reference-server-api/index.md
@@ -113,7 +113,10 @@ print(resp.content)
 
 ### Querying profiling data
 
-#### UTF-8 label names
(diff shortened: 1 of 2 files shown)


docs/sources/reference-server-api/index.md
```diff
@@ -113,6 +113,32 @@ print(resp.content)
 
 ### Querying profiling data
 
+#### UTF-8 label names
```
marcsanmi: Nit: the level of this block sits at the same level as endpoint, should it be `###` instead?

marcsanmi (dismissed):
LGTM

docs/sources/reference-server-api/index.md
simonswine: I think you are right it is odd, but still your solution would generate an empty `# Query profiling data`

I went with an admonition instead

simonswine: image

marcsanmi (approved):

PR #5560, opened 2026-08-27
fix(segmentwriter): probe the bucket with Upload instead of Iter

AustinKobayashi:
Follow-up to [#4453](https://github.com/grafana/pyroscope/pull/4453), which added the bucket health check but only warned when it failed. Ref [grafana/pyroscope-squad#587](https://github.com/grafana/pyroscope-squad/issues/587).

## Changes

- **Probe with `Upload` instead of `Iter`.** `Iter` tested LIST, which the write path never uses, so a list-only credential passed the check and then failed every write. `Upload` tests the permission `Push` needs, at the cost of one small PUT.
- **A failed upload prevents startup** instead of logging a warning. One attempt only: a restart is the retry, and the supervisor backs it off.
- **A fresh key per start**, `segments/_health/<instance-id>/<uuid>`. GCS grants such as `roles/storage.objectCreator` permit creating an object and deny replacing one, so a fixed key would fail every start after the first. It stays on `segments/` because a policy granting some other prefix while denying this one would pass the check and fail every real write.
- **Cleanup is best effort**, warning and incrementing `pyroscope_segment_writer_bucket_health_check_cleanup_failures_total`. It is skipped on the filesystem backend, whose `Delete` prunes empty parents and can race a concurrently written segment. A bucket denying deletes keeps one small object per start; expire those with a lifecycle rule scoped to `segments/_health/`.

## Testing

Unit tests cover the disabled short-circuit, a successful probe, exactly one upload attempt on failure, three fatal cases, two non-fatal cleanup failures, the filesystem backend skipping cleanup, two checks producing distinct keys, and the startup error reaching the caller.

Live on `fire-dev-001` (GCS), canaried onto one pod behind a Flux ignore:

- The probe ran in 146ms and 191ms, a single attempt each, uploading and removing its object. `..._cleanup_failures_total` stayed 0 across three starts.
- Two restarts produced two distinct keys under the same instance prefix, confirming each probe creates rather than replaces.
- Zero upload errors, and all four distributors held `ACTIVE=6, Unhealthy=0` throughout.

### DIFF
--- a/pkg/segmentwriter/service.go
+++ b/pkg/segmentwriter/service.go
@@ -41,6 +41,11 @@ const (
 	defaultHedgedRequestBurst   = 10 // allow bursts of 10 hedged requests
 )
 
+// errStopIteration aborts the bucket health check iteration: we only care that
+// the bucket is reachable, not about its contents. It is matched with errors.Is
+// so that the check keeps working if a bucket implementation wraps it.
+var errStopIteration = errors.New("stop iteration")
+
 type Config struct {
 	GRPCClientConfig           grpcclient.Config     `yaml:"grpc_client_config" doc:"description=Configures the gRPC client used to communicate with the segment writer."`
 	LifecyclerConfig           ring.LifecyclerConfig `yaml:"lifecycler,omitempty"`
@@ -97,7 +102,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.UintVar(&cfg.UploadHedgeRateBurst, prefix+".upload-hedge-rate-burst", defaultHedgedRequestBurst, "Maximum number of hedged requests in a burst.")
 	f.BoolVar(&cfg.MetadataDLQEnabled, prefix+".metadata-dlq-enabled", true, "Enables dead letter queue (DLQ) for metadata. If the metadata update fails, it will be stored and updated asynchronously.")
 	f.DurationVar(&cfg.MetadataUpdateTimeout, prefix+".metadata-update-timeout", 2*time.Second, "Timeout for metadata update requests.")
-	f.BoolVar(&cfg.BucketHealthCheckEnabled, prefix+".bucket-health-check-enabled", true, "Enables bucket health check on startup. This both validates credentials and warms up the connection to reduce latency for the first write.")
+	f.BoolVar(&cfg.BucketHealthCheckEnabled, prefix+".bucket-health-check-enabled", true, "Enables bucket health check on startup. This both validates credentials and warms up the connection to reduce latency for the first write. If the check fails, the segment writer fails to start.")
 	f.DurationVar(&cfg.BucketHealthCheckTimeout, prefix+".bucket-health-check-timeout", 10*time.Second, "Timeout for bucket health check operations.")
 }
 
@@ -194,6 +199,17 @@ func New(
 // performBucketHealthCheck performs a lightweight bucket operation to warm up the connection
 // and detect any object storage issues early. This serves the dual purpose of validating
 // bucket accessibility and reducing latency for the first actual write operation.
+//
+// A failure is fatal: the segment writer cannot serve writes without the bucket, as Push
+// blocks until the segment has been uploaded, and a definitive upload error surfaces as
+// codes.Unknown, which the client does not treat as retryable - so writes routed to a
+// write-broken instance are lost rather than failed over. Returning an error here prevents
+// the instance from starting, instead of letting it report itself ready and silently drop
+// a share of the cell's writes.
+//
+// This completes the plan agreed when the check was introduced in #4453, where warning and
+// continuing was explicitly a staging step until the behaviour had been verified against a
+// live cell.
 func (i *SegmentWriterService) performBucketHealthCheck(ctx context.Context) error {
 	if !i.config.BucketHealthCheckEnabled {
 		return nil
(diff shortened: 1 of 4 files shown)


simonswine (commented):
Thanks for looking at that

I've also mentioned that in the reviewed of https://github.com/grafana/pyroscope/pull/4453, just repeat it here and add some :)

- `Iter` might be come at a latency if there are many objects in the root. (we still fetch one page despite the early skip)
- `Iter` permissions is not a permission we need other than for this health check

I think what we should do: Test the `Upload`, this will ensure we can write to the bucket, which is the crucial permission we need.

But then we have the problem of having many small random small objects laying around, we should try to delete it right away, in the case of a delete permission error, we should log a warning, that it couldn't clean up, but it functions as normal and this is the flag to disable the health check.

Wdyt?

In general the PR is very verbose on comments and PR description for what the feature is, try to compress that a bit more

AustinKobayashi:
> Thanks for looking at that
> 
> I've also mentioned that in the reviewed of #4453, just repeat it here and add some :)
> 
>     * `Iter` might be come at a latency if there are many objects in the root. (we still fetch one page despite the early skip)
> 
>     * `Iter` permissions is not a permission we need other than for this health check
> 
> 
> I think what we should do: Test the `Upload`, this will ensure we can write to the bucket, which is the crucial permission we need.
> 
> But then we have the problem of having many small random small objects laying around, we should try to delete it right away, in the case of a delete permission error, we should log a warning, that it couldn't clean up, but it functions as normal and this is the flag to disable the health check.
> 
> Wdyt?
> 
> In general the PR is very verbose on comments and PR description for what the feature is, try to compress that a bit more

I agree that `Upload` is the better path, but I was initially concerned about `Delete` failing and creating too many stale objects. I keyed the object on the instance ID, so that should constrain it to one stale object per replica.

It writes under `segments/` rather than the internals prefix. That way a policy scoped to `segments/*` already covers the probe, and we're not asking for write access to a prefix the writer never touches otherwise. I also gave the upload the same retry/backoff as a real segment upload, since I don't think the probe should be more strict than a real upload.

I also added a metric for the `Delete` failures since I'm curious to see if that's a common occurrence.

pkg/segmentwriter/service.go
```diff
@@ -173,41 +189,64 @@ func New(
...
-		// Return an error to stop iteration after first item (if any)
-		return errors.New("stop iteration")
+	// An already-expired context skips the loop entirely, so err must start out non-nil.
+	err := uploadCtx.Err()
+	retries := backoff.New(uploadCtx, backoff.Config{
+		MinBackoff: i.config.UploadMinBackoff,
+		MaxBackoff: i.config.UploadMaxBackoff,
+		MaxRetries: i.config.UploadMaxRetries,
 	})
+	for retries.Ongoing() {
```
simonswine: Nit: I would not retry here to keep it simpler, the retry happens anyhow through a kubernetes container backoff restart.

pkg/segmentwriter/service.go
```diff
@@ -173,41 +189,64 @@ func New(
...
+		if err = i.storageBucket.Upload(uploadCtx, name, bytes.NewReader(payload)); err == nil {
```
simonswine: In terms of GCS permissions create and overwrite are treated differently, which impacts the healthcheck with this role: https://cloud.google.com/storage/docs/access-control/iam-roles#storage.objectCreator

Maybe the better tradeoff is having more objects polluting the bucket?

pkg/segmentwriter/service.go
```diff
@@ -173,41 +189,64 @@ func New(
...
-	// Ignore the "stop iteration" error we intentionally return
-	// and any "object not found" type errors as they indicate the bucket is accessible
-	if err == nil || i.storageBucket.IsObjNotFoundErr(err) || err.Error() == "stop iteration" {
-		level.Debug(i.logger).Log("msg", "bucket health check succeeded")
-		return nil
+	// Best effort: the key is per instance, so a bucket denying deletes keeps one object per
+	// replica, not one per restart.
+	deleteCtx, cancelDelete := context.WithTimeout(ctx, i.config.BucketHealthCheckTimeout)
+	defer cancelDelete()
+	if err := i.storageBucket.Delete(deleteCtx, name); err != nil {
```
simonswine, line 252: My agent discovered that with the filesystem backend it is possible that it will delete a concurrently uploaded segment, if it happens within a small window. I think pretty unlikely, but worth flagging.

> High: Health-check cleanup can delete a concurrently uploaded segment on filesystem storage  
> pkg/segmentwriter/service.go:47-48 and pkg/segmentwriter/service.go:227-229
> The probe is placed inside the real segments/ tree. The Thanos filesystem bucket’s Delete implementation recursively removes empty parent directories. It checks whether a parent is empty and then calls os.RemoveAll on that parent, creating a TOCTOU window.
> If another replica creates its first real segment after the emptiness check but before RemoveAll("segments"), the health-check cleanup deletes that segment. This is possible with shared filesystem storage and concurrent replicas. The probe should not be placed beneath a production data prefix when using this deletion implementation.

simonswine (commented):
Thanks for taking another look at it.

I think the GCS issues should be addressed, otherwise I think we can merge this.

pkg/segmentwriter/service.go
AustinKobayashi: I dropped the retry so it's a single attempt now

pkg/segmentwriter/service.go
AustinKobayashi: Good catch, I didn't encounter this when I tested in `profiles-dev-002`. I took the pollution tradeoff you suggested, the key is now `segments/_health/<instance-id>/<uuid>`. Two restarts in `fire-dev-001` (GCS) produced two distinct keys.

pkg/segmentwriter/service.go
AustinKobayashi: Nice catch. I looked into the Thanos filesystem implementation and confirmed the race. I kept the probe under `segments/` so it exercises the production write prefix, but cleanup is now skipped for the filesystem backend. Other backends still use best-effort deletion. I added a test for the filesystem behavior, accepting one small retained probe object per startup as the safer tradeoff.

simonswine (approved):
Thank you for iterating on it! LGTM

PR #5562, opened 2026-08-27
docs: Document the top-n flag in profilescli

simonswine:
This is improves the section about `profilescli query top` command.

Added in #4889


marcsanmi (approved):
LGTM

PR #5575, opened 2026-08-29
fix(segmentwriter): delete client metrics for removed segment writer addresses

### LINKED ISSUE #5517: `pyroscope_segment_writer_client_sent_bytes` metric has unbounded cardinality
#### Describe the bug

The cardinality of the `pyroscope_segment_writer_client_sent_bytes` metric can grow without bounds. The total number of metrics emitted by the `/metrics` endpoint requires ever-increasing amounts of memory to process from standard collectors (like otel-collector).

The histogram metric has dimensions on `shard`, `tenant`, and `addr`.

There is no code for clearing / deleting old tracked series from the metric.

As downstream segment-writer Pods restart, they can get a new IP address, creating new series for `addr`. The old series under the now-unused IP address linger.

If tenants are spread across many shards and/or adaptive placement is making significant changes to things, we can also see orphaned series as a result.

#### To Reproduce

Default behavior. Restart some segment-writer Pods and watch series count of `pyroscope_segment_writer_client_sent_bytes` grow.

#### Expected behavior

Metrics series shouldn't grow without bound.

How you achieve this is up for debate.

But it usually entails a TTL or similar orphan pruning mechanism on metrics series.

For the github.com/prometheus/client_golang/prometheus Go module, there are `DeleteLabelValues()`, `Delete()`, `DeletePartialMatch()`, and `Reset()` methods to remove series.

There is already precedence for this pattern in the Pyroscope code base. See e.g. https://github.com/grafana/pyroscope/blob/b4d74078918e396eaffa30273b913d18f9c0a42c/pkg/scheduler/scheduler.go#L622

#### Environment

k8s

iamsharduld:
Fixes #5517

## Problem
`pyroscope_segment_writer_client_sent_bytes` is a histogram labelled by `shard`, `tenant` and
`addr`. Segment writer addresses churn with pod IPs and shard relocation, and nothing ever removed
the series of an address that stopped being used, so the number of series exposed by the distributor
grows without bound (every restarted segment writer pod adds a new set of `addr` series).

## Solution
The ring client pool already closes an instance's connection once the instance is no longer in the
ring (`removeStaleClients`, every 15s). This change hooks into that: `connpool.NewConnPoolFactory`
takes an `onClose(addr)` callback which the segment writer client uses to delete all series of the
retired address with `DeletePartialMatch`. No new state, goroutines or configuration.

Notes:
- Deletion runs in the pool's asynchronous `Close`. A push whose response lands just before the
  close can observe after the deletion, leaving at most one set of series for that address until
  it is dialled and retired again, so growth stays bounded.
- An instance that leaves the ring transiently (rollout, unhealthy) has its series deleted and
  re-created on the next push, i.e. a histogram reset; the hook also fires when the pool shuts
  down. Dashboards use `rate()` / `histogram_sum(rate())` by `shard`, which absorb resets.
- The dashboards in `operations/monitoring` group this metric by `shard` only, so they are unaffected.
- Pruning by `tenant` (on tenant deletion) is not covered here; `tenant` and `shard` are bounded
  per live address.

## Tests
- `Test_Push_MetricsDeletedOnConnRemoval`: push, assert one series, remove the client for the
  address from the pool (what `removeStaleClients` calls), assert the series is gone.
- `TestConnFactory_Close`: the hook receives the instance address after `Close`; a nil hook is safe.

### DIFF
--- a/pkg/segmentwriter/client/client.go
+++ b/pkg/segmentwriter/client/client.go
@@ -160,13 +160,14 @@ func NewSegmentWriterClient(
 	placement placement.Placement,
 	dialOpts ...grpc.DialOption,
 ) (*Client, error) {
-	pool, err := newConnPool(ring, logger, grpcClientConfig, dialOpts...)
+	m := newMetrics(registry)
+	pool, err := newConnPool(ring, logger, grpcClientConfig, m, dialOpts...)
 	if err != nil {
 		return nil, err
 	}
 	c := &Client{
 		logger:      logger,
-		metrics:     newMetrics(registry),
+		metrics:     m,
 		distributor: distributor.NewDistributor(placement, ring),
 		pool:        pool,
 		ring:        ring,
@@ -318,6 +319,7 @@ func newConnPool(
 	rring ring.ReadRing,
 	logger log.Logger,
 	grpcClientConfig grpcclient.Config,
+	m *metrics,
 	dialOpts ...grpc.DialOption,
 ) (*connpool.Pool, error) {
 	options, err := grpcClientConfig.DialOption(nil, nil, nil)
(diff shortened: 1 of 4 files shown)


simonswine (commented):
Thanks for the report and for looking into this. I think while this will reduce the cardinality through churn, but I would go further and question the value of that detailed metrics all together.

I suggest to remove both addr and tenant from this histogram. But I think doing this might be disruptive, for now we can have a default off distributor flag, when set to `true` it should fix both addr and tenant name to `""` or `"<removed>"`.

Wdyt?

PR #5578, opened 2026-09-01
docs: fix error in helm command

tebriel:
`helm upgrade` command was wrong (likely copied from grafana)

### DIFF
--- a/docs/sources/deploy-kubernetes/helm.md
+++ b/docs/sources/deploy-kubernetes/helm.md
@@ -181,7 +181,7 @@ datasources:
 Modify the Helm deployment by running:
 
 ```bash
-   helm upgrade -n pyroscope-test --reuse-values grafana grafana/grafana \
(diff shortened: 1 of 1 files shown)


marcsanmi (dismissed):

marcsanmi:
Hi @tebriel,

Thanks for the PR, and sorry for the approve-then-dismiss; I looked too fast. This is correct as written: as the [sentence](https://github.com/tebriel/pyroscope/blob/8328d299ea3d4c22958288dcf9958ca1cb1c6556/docs/sources/deploy-kubernetes/helm.md?plain=1#L166) above notes, this provisions the data source into the Grafana deployment, and `datasources:` is the `grafana/grafana` chart's value, so the upgrade has to target that release. Closing, but thanks for taking the time to flag it.

If you think we could make this more obvious, feel free to reopen or open a new PR.

PR #5590, opened 2026-09-05
fix(querybackend): do not retry responses gRPC cannot deliver

marcsanmi:
Closes grafana/pyroscope-squad#911.

The query-backend client retries RESOURCE_EXHAUSTED, which the server returns when shedding load. Grpc-go uses the same code when a response exceeds the message size limit, and that failure is deterministic: the query is re-run until the 30s deadline and the caller sees a 504 instead of the real error.

The status code cannot tell the two apart, so the verdict travels out of band. Once a handler has built a response it sets `grpc-retry-pushback-ms: -1`, the trailer gRPC reserves for this. Grpc-go reads it only when deciding a retry, so it is inert on success and only takes effect when the send fails. Queries that fail return before that point and keep the retry policy, load shedding included. Clients forward the verdict up the query plan, so the frontend does not retry a subtree that cannot deliver.

The **trade-off** is that the scope is positional rather than by error identity: any retryable failure raised between the handler returning and the first byte on the wire counts as permanent. That window holds only encoding and the size check (grpc-go `Server.sendResponse`), so nothing transient can occur there, and it is worth re-checking on a major grpc-go bump. Doing the size check ourselves would name the case explicitly, but it duplicates grpc-go, needs the server message size limit plumbed into the module, and over-rejects when compression is enabled.

`backoff_on_ratelimits` is now rejected for this client, since dskit's retrier retries every `RESOURCE_EXHAUSTED` without honoring the pushback.

### DIFF
--- a/pkg/querybackend/backend.go
+++ b/pkg/querybackend/backend.go
@@ -19,12 +19,13 @@ import (
 
 	metastorev1 "github.com/grafana/pyroscope/api/gen/proto/go/metastore/v1"
 	queryv1 "github.com/grafana/pyroscope/api/gen/proto/go/query/v1"
+	"github.com/grafana/pyroscope/v2/pkg/querybackend/internal/pushback"
 	"github.com/grafana/pyroscope/v2/pkg/util"
 )
 
 type Config struct {
 	Address          string            `yaml:"address" category:"advanced"`
-	GRPCClientConfig grpcclient.Config `yaml:"grpc_client_config" doc:"description=Configures the gRPC client used to communicate between the query-frontends and the query-schedulers."`
+	GRPCClientConfig grpcclient.Config `yaml:"grpc_client_config" doc:"description=Configures the gRPC client used to communicate with query-backends. backoff_on_ratelimits must be disabled: its retries ignore the server's pushback."`
 	ClientTimeout    time.Duration     `yaml:"client_timeout" category:"advanced"`
 }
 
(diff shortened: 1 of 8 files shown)


pkg/querybackend/backend.go
```diff
@@ -38,6 +39,11 @@ func (cfg *Config) Validate() error {
 	if cfg.Address == "" {
 		return fmt.Errorf("query-backend.address is required")
 	}
+	// dskit's rate-limit retrier retries every RESOURCE_EXHAUSTED error without
+	// honoring grpc-retry-pushback-ms, which would retry oversized responses.
+	if cfg.GRPCClientConfig.BackoffOnRatelimits {
+		return fmt.Errorf("query-backend.grpc-client-config.backoff-on-ratelimits must be disabled: its retries ignore grpc-retry-pushback-ms")
+	}
```
simonswine: Not too sure if we could force it false instead and log a warning.

Also some docs override about it would be good here:

https://grafana.com/docs/pyroscope/latest/configure-server/reference-configuration-parameters/

Image

pkg/querybackend/backend_test.go
```diff
@@ -0,0 +1,147 @@
...
+	undeliverable := pushback.Mark(status.Error(codes.ResourceExhausted, "trying to send message larger than max"))
+	scheduled := make(chan struct{}, 1)
+
+	handler := &testQueryHandler{
+		invoke: func(context.Context, *queryv1.InvokeRequest) (*queryv1.InvokeResponse, error) {
+			select {
+			case scheduled <- struct{}{}:
+				return nil, first
+			default:
+				time.Sleep(50 * time.Millisecond)
```
simonswine: I think we can avoid using `time.Sleep` with channels/counter. Would benefit it's speed and avoid flakiness

simonswine (dismissed):
Thanks foor looking at this, two small asks before we merge

simonswine (approved):

PR #5591, opened 2026-09-06
fix(profilecli): stop replay dump from closing the dataset twice

### LINKED ISSUE #5585: profilecli: replay dump double-closes the Dataset via profileRowIterator.Close, corrupting the shared Object refcount (panic: negative reference counter)
## Summary

`dumpDataset` in `cmd/profilecli/replay_dump.go` (introduced in #5374) closes the same `Dataset` through two ownership paths: it defers `ds.Close()` after opening the dataset, and later defers `it.Close()` - but `profileRowIterator.Close` closes its reader, which is that same dataset (`pkg/block/section_profiles.go:449`):

```go
ds := block.NewDataset(dsMeta, obj)
if err := ds.Open(ctx, ...); err != nil { ... }
defer ds.Close()

it, err := block.NewProfileRowIterator(ds)
if err != nil { ... }
defer it.Close()
```

```go
func (p *profileRowIterator) Close() error {
	return p.reader.Close()
}
```

Defers run LIFO, so the dataset is closed twice on every successful `dumpDataset`. The duplicate close goes undetected because `refctr.Counter.Dec` checks the invariant one step late (`pkg/util/refctr/refctr.go:63`) - at `c == 0` it does not panic; it decrements to `-1` and runs `release()` a second time:

```go
if r.c < 0 {
	panic("bug: negative reference counter")
}
if r.c--; r.c < 1 {
	release()
}
```

The second release runs `Dataset.closeErr` again, which double-releases the shared block `Object`. When the block contains a **second** matching dataset, its open/close cycle finds the Object counter already negative and panics:

```
panic: bug: negative reference counter

github.com/grafana/pyroscope/v2/pkg/util/refctr.(*Counter).Dec  refctr.go:68
github.com/grafana/pyroscope/v2/pkg/block.(*Object).CloseWithError  object.go:247
github.com/grafana/pyroscope/v2/pkg/block.(*Dataset).closeErr  dataset.go:238
github.com/grafana/pyroscope/v2/pkg/block.(*Dataset).Close  dataset.go:208
main.dumpDataset  cmd/profilecli/replay_dump.go:280
main.dumpBlock    cmd/profilecli/replay_dump.go:218
main.replayDump   cmd/profilecli/replay_dump.go:177
```

The trigger is therefore ≥2 matching datasets in one block, not dump size: narrow single-service dumps appear to succeed while still silently corrupting each discarded Object's counter.

## Impact

- `profilecli replay dump` crashes deterministically on any multi-tenant/multi-service cell when the query matches more than one dataset in a block (e.g. the default `--query={}`). Single-dataset-per-block dumps complete but corrupt discarded refcounters silently.
- The same failed-open cleanup pattern exists in `pkg/querybackend/query.go` (`lookupDatasets` unconditionally defers `Close` on all constructed datasets before knowing every `Open` succeeded) and in `pkg/block/compaction.go` (closing all datasets after a `g.Wait()` error, including ones whose failed `Open` already rolled back and released the Object). Those are error-path-only and wrapped in `RecoverPanic` in the query backend, so the expected impact there is a failed request rather than a process crash - but a real storage error can be masked by the secondary refcount panic.

## Reproduction

1. Run a v2 cell with multiple services under one tenant (observed on 2.3.0, revision 6db589a; S3 backend; ~100 services; mix of fresh segments and compacted blocks).
2. `profilecli replay dump --storage.backend=s3 --storage.s3.bucket-name=... --storage.s3.region=... --storage.prefix=... --metastore.address=<metastore>:9095 --tenant-id=anonymous --query={} --from=now-30m --to=now --output=dump.bin`
3. Panics after "found blocks matching query" once it reaches a block with a second matching dataset.
4. Same command with `--query={service_name="<one-service>"}` completes, and the dump replays fine with `replay push --loop`.

Environment: Kubernetes (EKS), helm chart 2.2.1, image 2.3.0, microservices mode, v2 storage.

## Suggested fix

1. In `dumpDataset`, transfer ownership to the iterator once it exists: keep an explicit `ds.Close()` only on the iterator-construction error path and drop the unconditional `defer ds.Close()`.
2. Harden `Counter.Dec` to reject a decrement at zero before running release (`if r.c <= 0 { panic(...) }`), so duplicate closes are caught at the first offense instead of corrupting shared state.
3. In the query-backend/compaction error paths, close only datasets whose `Open` succeeded.

A regression test that dumps two datasets sharing one `Object` (asserting exactly one release per successful open), plus a `Counter.Dec`-at-zero test, would lock this in.

lazerg:
`dumpDataset` deferred `ds.Close()` and then `it.Close()`, but `profileRowIterator.Close` closes its reader, which is that same dataset. Defers run LIFO, so every successful dump closed the dataset twice and released the shared block `Object` one time too many. With a second matching dataset in the block, the next open/close cycle finds a negative refcounter and panics.

The iterator owns the dataset once it exists, so the unconditional defer is gone and `ds.Close()` now only runs on the iterator-construction error path. I put that contract in a doc comment on `NewProfileRowIterator` too, since it wasn't written down anywhere.

`readProfilesFromDataset` and `resolveProfile` in `pkg/operations/v2/blocks/profile_handlers.go` have the same pattern. They keep their own reference on the `Object`, so the counter goes negative without anything panicking today. Happy to fix those here as well if you'd prefer it in one change.

Fixes #5585

### DIFF
--- a/pkg/block/section_profiles.go
+++ b/pkg/block/section_profiles.go
@@ -378,8 +378,9 @@ type profileRowIterator struct {
 	labels           immutableLabels
 }
 
-// NewProfileRowIterator takes ownership of the dataset: closing the
-// iterator closes it. The caller must not close the dataset itself.
+// NewProfileRowIterator takes ownership of one open dataset reference on
+// success. Closing the iterator releases that reference. On error, ownership
+// remains with the caller.
 func NewProfileRowIterator(s *Dataset) (iter.Iterator[ProfileEntry], error) {
 	k, v := index.AllPostingsKey()
 	tsdb := s.Index()
--- a/pkg/block/section_profiles.go
+++ b/pkg/block/section_profiles.go
@@ -378,6 +378,8 @@ type profileRowIterator struct {
 	labels           immutableLabels
 }
 
+// NewProfileRowIterator takes ownership of the dataset: closing the
+// iterator closes it. The caller must not close the dataset itself.
 func NewProfileRowIterator(s *Dataset) (iter.Iterator[ProfileEntry], error) {
 	k, v := index.AllPostingsKey()
 	tsdb := s.Index()
--- a/cmd/profilecli/replay_dump.go
+++ b/cmd/profilecli/replay_dump.go
@@ -236,12 +236,13 @@ func dumpDataset(
 	if err := ds.Open(ctx, block.SectionTSDB, block.SectionProfiles, block.SectionSymbols); err != nil {
 		return 0, fmt.Errorf("failed to open dataset: %w", err)
 	}
-	defer ds.Close()
 
 	it, err := block.NewProfileRowIterator(ds)
 	if err != nil {
+		_ = ds.Close()
 		return 0, fmt.Errorf("failed to create profile row iterator: %w", err)
 	}
+	// The iterator takes ownership of the dataset and closes it.
 	defer it.Close()
 
 	var count int
--- a/cmd/profilecli/replay_dump.go
+++ b/cmd/profilecli/replay_dump.go
@@ -242,7 +242,6 @@ func dumpDataset(
 		_ = ds.Close()
 		return 0, fmt.Errorf("failed to create profile row iterator: %w", err)
 	}
-	// The iterator takes ownership of the dataset and closes it.
 	defer it.Close()
 
 	var count int
--- a/cmd/profilecli/replay_dump_test.go
+++ b/cmd/profilecli/replay_dump_test.go
@@ -12,6 +12,7 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 
 	metastorev1 "github.com/grafana/pyroscope/api/gen/proto/go/metastore/v1"
+	"github.com/grafana/pyroscope/v2/pkg/block"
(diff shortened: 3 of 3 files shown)


lazerg:
recheck

pkg/block/section_profiles.go
```diff
@@ -378,6 +378,8 @@ type profileRowIterator struct {
 	labels           immutableLabels
 }
 
+// NewProfileRowIterator takes ownership of the dataset: closing the
+// iterator closes it. The caller must not close the dataset itself.
```
simonswine: I don't think this is fully correct, as on error that is not true per your fix.

Maybe something like this:


```suggestion
// NewProfileRowIterator takes ownership of one open dataset reference on
// success. Closing the iterator releases that reference. On error, ownership
// remains with the caller.
```

simonswine (dismissed):
Thanks for tracking this down. LGTM

pkg/block/section_profiles.go
lazerg: Good catch, the error path leaves the dataset with the caller. Took your wording in 77e9748.

simonswine (approved):
Thanks again! LGTM for real

PR #5594, opened 2026-09-07
fix(blocks): close profile datasets through iterators

simonswine:
## Summary
- let profile row iterators own the dataset lifetime
- materialize pprof profiles before releasing their symbol resolver
- add a regression test that detects double-close reader calls

## Related
- #5591


aleks-p (approved):
LGTM

PR #5597, opened 2026-09-08
fix(compactionworker): synchronize worker lifecycle tests

simonswine:
Synchronize compaction worker lifecycle tests with observable worker events instead of stopping the service after a fixed 500 ms delay.\n\nPreviously, slower or race-instrumented runners could stop the worker before the completed job's status update was sent, making `TestWorker_SuccessfulCompaction` flaky in the arm64 `-race -cover` job. The tests now use bounded, test-specific completion signals, and the successful compaction test validates the complete status update payload before shutdown.


marcsanmi (approved):
LGTM

PR #5606, opened 2026-09-08
feat(otlp): detect profile language from telemetry.sdk.language resource attribute

luweglarz:
## Summary

OTLP profiles should report their language via the `telemetry.sdk.language` resource attribute (per [OTel resource semconv](https://opentelemetry.io/docs/specs/semconv/resource/#telemetry-sdk)). Today that attribute isn't used for language detection, instead, `pprof.GetLanguage` is used, which matches known filename prefixes/suffixes (`.go`, `.py`, `java/`, etc.) in the profile's string table and silently returns `unknown` when none match

## Changes

The idea is to use the predefined OpenTelemetry semantic convention directly, avoiding the string-table lookup and giving an accurate language 

- `pkg/model/labels.go`: add `AttrTelemetrySDKLanguage` constant
- `pkg/ingester/otlp/ingest_handler.go`: extract `telemetry.sdk.language` per `ResourceProfiles`, set on each resulting `ProfileSeries.Language` 
- `pkg/ingester/otlp/ingest_handler_test.go`: new test covering a custom (non-CPU/off-CPU) profile with the attribute set

### DIFF
--- a/pkg/ingester/otlp/ingest_handler.go
+++ b/pkg/ingester/otlp/ingest_handler.go
@@ -269,6 +269,7 @@ func (h *ingestHandler) export(ctx context.Context, er *pprofileotlp.ExportProfi
 
 	for _, rp := range rps {
 		serviceName := getServiceNameFromAttributes(rp.Resource.GetAttributes())
+		language := getLanguageFromAttributes(rp.Resource.GetAttributes())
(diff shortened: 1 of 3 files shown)


simonswine (approved):
Good idea, implementation looks sounds too. Just needs a rebase now.

simonswine (approved):

PR #5623, opened 2026-09-09
docs: Document write-path sampling and stripped-profile limits

knylander-grafana:
This PR documents Pyroscope v2 write-path sampling and splits reference from how-to content:

- Add a standalone v2 architecture page that explains write-path sampling, why sampled-out profiles are dropped, and how `keep_stripped_profiles` and `include_stripped_profiles` work. Includes a write-path decision flowchart.
- Add a `configure-server/configure-sampling.md` how-to page: per-tenant overrides example (usage groups + probability), troubleshooting, and a pointer to Grafana Cloud Adaptive Profiles for managed sampling.
- Document runtime configuration and per-tenant overrides in `about-configurations.md`.
- Add short cross-references from the distributor and query-backend pages.

Following review feedback, the sampling configuration example was moved off the architecture reference into the new how-to page, so the reference page matches the rest of the v2 architecture docs (concept + pointer, with both pages cross-linked).

Follow-up to review comments on #5599 / #5354.

Made with the help of Cursor.

## Test plan
- [ ] Confirm the Write-path sampling page appears under Reference: v2 Architecture and the flowchart renders
- [ ] Confirm the new Configure write-path sampling page appears under Configure the server
- [ ] Confirm distributor and query-backend links resolve to the sampling page
- [ ] Confirm config-reference links to `keep_stripped_profiles` and `include_stripped_profiles` resolve
- [ ] Confirm the sampling page links to the new how-to page, and the how-to links to the runtime overrides section and Adaptive Profiles

PR #5630, opened 2026-09-11
docs: add write-path sampling monitoring guide

knylander-grafana:
Add a monitoring page that shows operators how to confirm write-path sampling is active and measure dropped volume by tenant and usage group. Fills the monitoring gap deferred from #5623.

Cross-links to `configure-sampling.md` and the sampling reference page should be added when #5623 merges.


knylander-grafana:
Folded into #5623 per review feedback — the monitoring page and cross-links are now included in that PR.

PR #5636, opened 2026-09-14
fix(symdb): handle line-less call site leaves

simonswine:
## Summary
- avoid treating unresolved physical leaves as flat call sites
- preserve total attribution for matching caller locations
- add regression coverage for sample and Parquet call-site statistics
- include the related Go PGO line-less callee protection

## Testing
- `go test ./pkg/phlaredb/symdb -run '^TestStackTraceSelection_LineLessLeaf$' -count=1`


marcsanmi (approved):
LGTM

PR #5662, opened 2026-09-19
fix(metastore): Reject compaction jobs for deleted blocks

### LINKED ISSUE #5436: Reject compaction jobs for deleted blocks
In a rare case, either due to configuring a very short retention period or deploying an undersized compaction worker pool, v2 compaction workers could compact blocks that were removed due to retention. This can resurrect data that was deleted.

Add checks before `ReplaceBlocks` that all blocks in `compacted.SourceBlocks` still exist. On mismatch, skip `ReplaceBlocks` and tombstone the output block that was already written. Mark the job as completed.

subhramit:
Closes https://github.com/grafana/pyroscope/issues/5436

Reject compaction jobs for deleted blocks.
Refs. https://github.com/grafana/pyroscope/pull/5427#discussion_r3666530308

### Details

Fixes a metastore compaction edge case where a completed v2 compaction job could resurrect data that had already been removed by retention.

Before applying `ReplaceBlocks`, the metastore now verifies that all blocks in `compacted.SourceBlocks` still exist. If any source block is missing, it rejects the compaction output, skips replacing the source blocks in the index, and tombstones the newly written output block objects instead, while still marking the job completed.

Have run `make fmt check/unstaged-changes`, `make lint`, `make go/test GO_TEST_FLAGS="-race -cover"` and `make go/test` and everything passes.


subhramit:
Adding @aleks-p, @marcsanmi for review.

Could I please get an approval for the CI run?

Also, I am unsure why the summary says my commits are unsigned, as I have signed both of them. 
I am also unsure why there are multiple CLA warnings instead of one.

cc @korniltsev-grafanista, @knylander-grafana

Edit - Apologies, I figured out it requires key-signing and not just signed commits. Configured that and rebased.

subhramit:
@aleks-p could I get an approval for CI run

subhramit:
@simonswine can I get an approval for CI and, if possible, a review?

subhramit:
Hi everyone, CI is green. Could I get a review?

pkg/metastore/compaction_raft_handler.go
```diff
@@ -201,6 +204,29 @@ func (h *CompactionCommandHandler) UpdateCompactionPlan(
 			level.Warn(h.logger).Log("msg", "compacted blocks are missing; skipping", "job", job.State.Name)
 			continue
 		}
+		// The source blocks may have been deleted while the job was in
+		// progress, e.g., by the retention policy. If we replace them, we
+		// resurrect the deleted data. Instead, we reject the job output:
+		// the new blocks are not added to the index, and their objects are
+		// tombstoned. The source blocks that still exist stay in the index
+		// as they are. The job is completed in the schedule either way.
+		if missing := h.index.MissingBlocks(tx, compacted.SourceBlocks); len(missing) > 0 {
```
simonswine: During roll out this could have detrimental effects: while the updated replicas rejects and deletes it, the up-to-date replicas will accept it and delete all the source blocks.

I think this needs a two step activation, once all replicas support it. Not too sure if we have ever done something like that before? @aleks-p

pkg/metastore/compaction_raft_handler.go
```diff
@@ -201,6 +204,29 @@ func (h *CompactionCommandHandler) UpdateCompactionPlan(
...
+				"missing_blocks", len(missing),
+				"source_blocks", len(compacted.SourceBlocks.Blocks),
+			)
+			for _, t := range tombstonesForRejectedBlocks(job.State.Name, compacted.NewBlocks) {
+				if err = h.tombstones.AddTombstones(tx, cmd, t); err != nil {
+					level.Error(h.logger).Log("msg", "failed to add tombstones", "err", err)
+					return nil, err
+				}
+			}
+			continue
```
simonswine: For blocks that are still valid, we ignore their compaction. I think this a fair assumption, as they will likely fall out of retention soon after.

We should probably document it though.

subhramit: I agree and this is intentional. In the mismatch case we reject the job output and leave surviving source blocks untouched. Since this is more of an internal edge case I documented that in the inline comment above this branch.
>// tombstoned. The source blocks that still exist stay in the index
		// as they are.

If you meant document somewhere else, please point me. One place I have in mind is probably `metastore/compaction/README.md`?

subhramit: I put it there in the last commit.

pkg/metastore/compaction_raft_handler.go
subhramit: You are right. I have something in mind, will need more changes but let me try.

subhramit: Done. Could you take a look at https://github.com/grafana/pyroscope/pull/5662/changes/2eb93ac06d7f40611602503c42d778cb01a31a17?
Now we make the leader decide whether compaction output should be rejected and encoding that in the replicated plan, with rejected output tombstoned separately.

